Issue #87, opened 2023-06-15
[CON-44] Parse tracebacks in multiple languages - start with python and javascript? Or


sestinj:
We have some of this in the DLT thing, just need to make more reliable and include other languages

See ValidatePipelineStep/steps line 91. Write a function that takes the output of a command that was run (str) and returns either None if no traceback, or the string of the traceback if there was. Make a (separate) function like this for javascript and python.

<sub>From [SyncLinear.com](https://synclinear.com) | [CON-44](https://linear.app/continue/issue/CON-44/parse-tracebacks-in-multiple-languages-start-with-python-and)</sub>

TyDunn:
Used Continue to generate first attempts at these

PR #99, opened 2023-06-15
parsing python and javascript tracebacks

TyDunn:
#87

### DIFF
--- a/continuedev/src/continuedev/steps/core/core.py
+++ b/continuedev/src/continuedev/steps/core/core.py
@@ -344,6 +344,7 @@ def get_python_traceback(output: str) -> str:
 
 def get_javascript_traceback(output: str) -> str:
     lines = output.splitlines("\n")
-    first_line = lines[0].split(": ")
-    if len(lines) > 0 and len(first_line[0]) > 0 and "at" in lines[1].lstrip():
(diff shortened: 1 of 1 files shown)


continuedev/src/continuedev/steps/core/core.py
```diff
@@ -334,3 +334,16 @@ async def run(self, sdk: ContinueSDK) -> Coroutine[Observation, None, None]:
         self.description = self.prompt
         resp = await sdk.wait_for_user_input()
         return TextObservation(text=resp)
+
+
+def get_python_traceback(output: str) -> str:
```
TyDunn, line 339: @sestinj Since I am just writing these functions and the code that uses them has not been written yet, it's not so clear that I am heading in the right direction. Is there a subset of the output we want to send? Also, where should I put them?

In `ValidatePipelineStep`, it uses `output = "Traceback" + output.split("Traceback")[-1]`, but I'm not sure why and I believe it will be different for JavaScript

sestinj: This is fine for now, will probably work. Only thing is try to be more selective. I could print "I am at U of M" and it would think there was a .js error. What is the most selective you can be without ever missing a traceback?

PR #101, opened 2023-06-16
initial prompt improvement

TyDunn:


TyDunn:
@sestinj I have tested this in the playground, but I have not tested it in Continue yet because I want to get the other PR, which makes a bunch of changes around this in first and then test all of them together

sestinj (changes requested):
The example and requested task are not the same. You don't have the <> tags in the real instructions, but they show up in the example. I suppose the model can understand this, but not ideal

PR #138, opened 2023-06-21
removing captured terminal

### LINKED ISSUE #139: remove captured terminal

TyDunn:
closes #139


extension/src/continueIdeClient.ts
```diff
@@ -326,17 +325,10 @@ class IdeProtocolClient {
...
-      this.continueTerminal = new CapturedTerminal({
-        name: "Continue",
-      });
-    }
-
-    this.continueTerminal.show();
-    return await this.continueTerminal.runCommand(command);
+    const terminal = vscode.window.createTerminal();
+    terminal.show();
+    terminal.sendText(command);
```
sestinj: Only thing to change: if a terminal already exists, use that one:

if vscode.window.terminals.length:
use terminals[0]
else:
create terminal

sestinj (changes requested):
good after this!

PR #299, opened 2023-07-26
Close Client after use

sanders41:
This client only gets used once in this function and is not used again. Putting it in a context manager will close the client after use freeing up it's resources.


sestinj:
Thanks for this! And also for your work on the meilisearch Python clients : )

PR #317, opened 2023-07-28
let ggml server url be customized by config

lun-4:
there may be ggml servers that live outside of the main machine (e.g colab, or some cloud gpu somewhere with a llama-70b + cuBLAS build of llama.cpp). it'd be easier for someone to only have to get `llama-cpp-python[server]` running on that external machine and point their pre-existing continuerun install to that, instead of spinning up a separate continuerun installation on that server.

this pr may be denied, i'm fine with that.

my argument for configurable ggml rather than *just* configurable continuerun endpoint is that the server running ggml is (as i understand) stateless (it only provides model inference). in the case of colab free tier *or* some cloud gpu provider that you ran out of credits for, that server may go down, and your data with it (in my opinion, especially important for future usage of [the suggestions.json file](https://continue.dev/docs/collecting-data))!

inspired by [this reddit comment](https://www.reddit.com/r/LocalLLaMA/comments/15b565t/best_oss_coding_assistant_for_vs_code/jtphvo9/). ;)

### DIFF
--- a/continuedev/src/continuedev/core/config.py
+++ b/continuedev/src/continuedev/core/config.py
@@ -50,6 +50,7 @@ class ContinueConfig(BaseModel):
     on_traceback: Optional[List[OnTracebackSteps]] = []
     system_message: Optional[str] = None
     azure_openai_info: Optional[AzureInfo] = None
+    ggml_server_url: Optional[str] = None
 
     context_providers: List[ContextProvider] = []
 
--- a/continuedev/src/continuedev/core/sdk.py
+++ b/continuedev/src/continuedev/core/sdk.py
@@ -116,7 +116,10 @@ class Models:
 
     @cached_property
     def ggml(self):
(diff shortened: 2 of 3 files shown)


sestinj:
I actually like this a lot! You followed the pattern I had with Azure, but ideally the approach would be different (I've been meaning to change this for Azure as well). I'd like to avoid such arbitrary config properties and instead pass an LLM object for the default_model property (not a str). That said, this does no harm and it'll still be easy to make the ideal change later. If you want to take a stab at passing an LLM that would be great, but if you want to just play with models rn, I'll take a closer look tonight and get this merged.

lun-4:
I did intentionally create `ggml_server_url` based on the existing azure config key, lol!

I do agree with passing an LLM object on the config! I'll close this PR and attempt that

sestinj:
Awesome! Let me know if any questions

PR #318, opened 2023-07-28
make Config receive LLM objects

lun-4:
see background here: https://github.com/continuedev/continue/pull/317#issuecomment-1654904324

this is a draft for feedback, code is at the moment very broken.

the idea is to pass LLM objects into the config, to do that we have to:
 - [x] split `api_key` reading away from class initialization. at the moment i created a start/stop async callback for LLM classes to implement. this lets the concern of which model string goes to which env var be *inside* the respective LLM class
   - this would also let the `GGML` class store its own `AsyncSession` instead of continuously creating one on every action
     - immediate pro: it's what aiohttp wants, see https://docs.aiohttp.org/en/stable/client_quickstart.html: `Don’t create a session per request. Most likely you need a session per application which performs all requests together. [...] A session contains a connection pool inside. Connection reusage and keep-alives (both are on by default) may speed up total performance.`
     - future pro: socks5 proxy (who knows what the future may provide, but that won't be implemented here.)
- [x] pass sdk object to llm classes so they can access secrets
- [x] make api key be fetched declaratively by LLM instead of passing sdk object
- [x] ~~fix plugins making explicit references to specific models like gpt35 or starcoder. reference only the configured model (multiple model support sounds like a good idea, so you could use gpt3.5 for some task, and WizardCoder-GGML for another, but that can be done after this PR)~~
- [x] multi model support (small/medium/large model roles)
- [x] refactor llm attributes left over in `util`-type files and put all into `LLM` class
    - prices, max tokens, chat capabilities, etc
- [x] replace plugins calling hardcoded models (gpt35, starcoder, etc, just review all `models` accesses) to call the respective model role
    - `gpt35` becomes `medium`
    - `starcoder` becomes `medium`
    - `gpt4` becomes `large`
- [x] make `MaybeProxyOpenAI` LLM class, use it in default_config
- [x] turn `GGML`'s `aiohttp.AsyncSession` into an instance attribute
- [x] fix SDK `get_api_secret` not using prompt (maybe by removing it entirely?)
- [x] update documetation
- [x] ???
- [x] profit


continuedev/src/continuedev/libs/llm/ggml.py
```diff
@@ -35,6 +35,7 @@ async def stream_complete(self, prompt, with_history: List[ChatMessage] = None,
         messages = compile_chat_messages(
             self.name, with_history, args["max_tokens"], prompt, functions=args.get("functions", None), system_message=self.system_message)
 
+        # TODO move to single self.session variable (proxy setting etc)
```
sestinj: keeping a single session for each instance of GGML (and other LLM too) is great! Definitely an improvement. Since the `aiohttp.ClientSession` __init__ is synchronous, can happen inside GGML.__init__

continuedev/src/continuedev/libs/llm/anthropic.py
```diff
@@ -9,16 +9,18 @@
...
 
-    def __init__(self, api_key: str, default_model: str, system_message: str = None):
-        self.api_key = api_key
+    def __init__(self, default_model: str, system_message: str = None):
         self.default_model = default_model
         self.system_message = system_message
 
-        self.async_client = AsyncAnthropic(api_key=api_key)
+    async def start(self):
+        self.async_client = AsyncAnthropic(api_key=await self.sdk.get_api_key("ANTHROPIC_API_KEY"))
```
sestinj: The SDK is quite a powerful object, so it's best not to pass it around to too many other classes. 2 options here are:
a) have an "api_key_env_var_name: Union[str, None]" property on the `LLM` class, then the Models class can use that to call `sdk.get_user_secret` and pass the result to `start(api_key)`
b) pass the `get_user_secret` method to the `start` method, so that it has access to only this rather than the entire SDK (I'm leaning toward this)

continuedev/src/continuedev/core/sdk.py
```diff
@@ -252,6 +198,10 @@ async def delete_directory(self, path: str):
         path = await self._ensure_absolute_path(path)
         return await self.run_step(FileSystemEditStep(edit=DeleteDirectory(path=path)))
 
+    async def get_api_key(self, env_var: str) -> str:
+        # TODO support error prompt dynamically set on env_var
+        return await self.ide.getUserSecret(env_var)
+
```
sestinj, line 162: I think it's okay to just use get_user_secret here, though I see why you'd want to abstract over the prompt (is this what the TODO means?). If you do that, it might make sense for `get_api_key` to live inside the Model class.

continuedev/src/continuedev/core/sdk.py
```diff
@@ -174,6 +119,7 @@ async def create(cls, autopilot: Autopilot) -> "ContinueSDK":
             ))
 
         sdk.models = await Models.create(sdk)
+        await sdk.models.start()
```
sestinj: This could live inside of Models.create

continuedev/src/continuedev/core/sdk.py
```diff
@@ -90,54 +80,9 @@ def __load_anthropic_model(self, model: str) -> AnthropicLLM:
...
-        elif model_name == "ggml":
-            return self.ggml
-        else:
-            raise Exception(f"Unknown model {model_name}")
-
     @property
     def default(self):
-        default_model = self.sdk.config.default_model
-        return self.__model_from_name(default_model) if default_model is not None else self.gpt4
+        return sdk.config.llm if sdk.config.llm is not None else ProxyServer(default_model="gpt-4")
```
sestinj: This should respect the logic that says "if there's an openai key, then use OpenAI, otherwise use ProxyServer". I think the best way here is: In Models.create, start the sdk.config.llm, then set the property `self.default` to `sdk.config.llm if sdk.config.llm is not None else (<CHECK FOR API KEY> ... ProxyServer/OpenAI)`, and then there's no need for the @property decorator

sestinj (commented):
Left a few comments about implementation details, but overall looks great. The only serious comment I have is about removing multiple model support, because this is already a pretty key part of the plugins used most frequently in the extension by default (gpt3.5 for summarization, gpt4 for everything more important). My idea on this is something like "model roles."

Since it should be possible to literally use _only_ a local model for everything, the calls to sdk.models.gpt35 can't stay. But unfortunately we also can't postpone the ability to call a faster/cheaper model. The solution seems to be able to "fall back" to the default model, but provide a property for customization. For example, the config could look like this:
```python
config = ContinueConfig(
...
  models: Models(
    default: OpenAI("gpt4",...),
    small: OpenAI("gpt-3.5",...),
    large: None,
    custom: { "my-custom-model-name", MyCustomModel(...), ... }
  )
)
  ```
  
  I like this solution because it:
  1. Maintains strong typing on the sdk.models.
  2. Lets people add whatever models they want without having to modify the Models class
  3. Lets us keep the common functionality of using a smaller model for simpler tasks like summarization
  
Note that doing this would require using an async "start"/"initialize" method on the Models class, instead of a "create" classmethod, but this feels like a fair tradeoff.

**Comment by lun-4** [comment]
> My idea on this is something like "model roles."

I didn't understand the separation between small/medium/large, but "summarizer" is a good role already for something like GPT-3.5. Could you elaborate on what `custom` would mean on that example?

File: continuedev/src/continuedev/libs/llm/ggml.py (hunk 1 above)
Comment (lun-4) [reply continuedev/src/continuedev/libs/llm/ggml.py]: even though creating the ClientSession class is synchronous, [aiohttp will warn against it](https://github.com/aio-libs/aiohttp/pull/1468), in theory we could pass the loop as suggested in the issue, but [loop parameters are deprecated in modern asyncio](https://stackoverflow.com/questions/60312374/what-are-all-these-deprecated-loop-parameters-in-asyncio), so it would still have to fit on an async `start`

File: continuedev/src/continuedev/libs/llm/anthropic.py (hunk 2 above)
Comment (lun-4) [reply continuedev/src/continuedev/libs/llm/anthropic.py]: I'm leaning towards a different version of a), or could be a mixture of a) and b), something like this

I lean towards a) because it lets LLMs declaratively ask for what they want, specifically, rather than giving them access to all secrets through the SDK function

```python
class OpenAI(LLM):
    def wanted_permissions(self):
        # a) return desired privileged information
        return {"secrets": ["OPENAI_API_KEY"]}
        
        # a and b) return desired sdk method
        return {"methods": ["get_user_secret"]}
    
    async def start(self, privileged):
        # a) sdk takes care of giving llm the secret
        privileged.secrets["OPENAI_API_KEY"]
        # b) sdk returns wanted method
        await privileged.sdk.get_user_secret("OPENAI_API_KEY") # special sdk object, other methods raise AttributeError
```

continuedev/src/continuedev/core/sdk.py
lun-4: I noticed the `prompt` parameter was not used on `get_user_secret`, even though we pass it when we did api key verification on `Models`, so I thought I'd set a "default" prompt if the secret is missing: that one being `f'Please add your {env_var} to the .env file'`

continuedev/src/continuedev/core/sdk.py
lun-4: I think we could create a `MaybeProxyOpenAI()` (other name: `AutomaticOpenAIFallback`) LLM class that does the switch based on secret existence, rather than having that logic inside the SDK

continuedev/src/continuedev/libs/llm/ggml.py
sestinj: Good points, putting it in `start` sounds great.

continuedev/src/continuedev/libs/llm/anthropic.py
sestinj: Ok, let's go with a), but I think the complexity of wanted_permissions is unlikely to be needed for now since all LLMs I imagine will be written only need a single API key, so let's just add a `required_api_key: Optional[str] = None` attribute (like wanted_permissions, but not a function) on the LLM class, and then start would have an `api_key: Optional[str] = None` parameter.

continuedev/src/continuedev/core/sdk.py
sestinj: Not a bad idea. If it can be done without too much work that would make sense

sestinj:
> > My idea on this is something like "model roles."
> 
> I didn't understand the separation between small/medium/large, but "summarizer" is a good role already for something like GPT-3.5. Could you elaborate on what `custom` would mean on that example?

Small/medium/large just meaning larger/slower vs. smaller/faster models. Seemed a bit more general than "summarizer", because gpt35 can be used for other tasks as well. Custom should exist because for example the small/medium/large/default might not be enough slots for all the LLMs someone wants, or at least it could be limiting. Or let's say they have their own super specific LLM that is used in a custom Step they've written. Then they'd want to be able to reference it with `sdk.models.custom["my_custom_model_name"]`.

lun-4:
> the small/medium/large/default might not be enough slots for all the LLMs someone wants, or at least it could be limiting. 

I don't understand how would a custom model be used from the perspective of the code that already wants `models.large`, I can understand mapping for a custom Step, but that reads as if it's a separate example than the first sentence. If the intention was just to provide custom Step support, that's fine, and I apologize for misunderstanding.

sestinj:
> > the small/medium/large/default might not be enough slots for all the LLMs someone wants, or at least it could be limiting.
> 
> I don't understand how would a custom model be used from the perspective of the code that already wants `models.large`, I can understand mapping for a custom Step, but that reads as if it's a separate example than the first sentence. If the intention was just to provide custom Step support, that's fine, and I apologize for misunderstanding.

No worries—I think there's a good point here: you can still get custom LLMs that aren't in sdk.models.small/medium/large/default by just importing it yourself into a custom step. So let's remove the `custom` thing in `Models`, but keep small/medium/large:

```python
config = ContinueConfig(
...
  models: Models(
    default=OpenAI("gpt4",...),
    small=OpenAI("gpt-3.5",...),
    large=None,  # This would fall back to the default model (these are all optional kwargs, except default)
  )
)
```

continuedev/src/continuedev/core/sdk.py
lun-4: as `Models` are now instantiated from config due to model roles, I don't think this is needed -- latest draft changes these lines to do async start

lun-4:
I'm taking a longer look into what's going to be needed to make this ready and it seems like it is a very large refactor job... I don't see an avenue where both "LLM objects in config" can come without "very big PR", any ideas? I left some TODOs in the current PR that show points where future work can happen

sestinj:
I'm okay with big PR to be honest. I think the general direction is great, and had expected a bit of cleanup would be needed around the LLM classes. I'll make a few edits myself tomorrow to try and pull this through, then do a bit of testing.

sestinj:
Working on this now. Next will make a commit so that all references like `sdk.models.gpt35` use `sdk.models.default/small` instead

sestinj:
@lun-4 I'm going to stop here to let you take a look, but afterward I'd like to merge main to here and test a bit. Pretty excited about this PR because it sets up to allow easily running llama2 with Continue!

lun-4:
Thank you very much! I had the maybe-proxy already written up on my machine, but completely forgot to commit, oops!

continuedev/src/continuedev/libs/constants/default_config.py.txt
```diff
@@ -31,7 +31,10 @@ class CommitMessageStep(Step):
 
         # Ask gpt-3.5-16k to write a commit message,
         # and set it as the description of this step
-        self.description = await sdk.models.gpt3516k.complete(
+        gpt3516k = OpenAI(model="gpt-3.5-turbo-0613")
```
lun-4: Should we use `sdk.models.default` so that newcomers to the SDK API don't become encouraged to hardcode models on their own steps? I will say this is some "psychological 5D chess" on my part.

continuedev/src/continuedev/plugins/steps/chat.py
```diff
@@ -168,7 +169,10 @@ async def run(self, sdk: ContinueSDK):
             msg_content = ""
             msg_step = None
 
-            async for msg_chunk in sdk.models.gpt350613.stream_chat(await sdk.get_chat_context(), functions=functions):
+            gpt350613 = OpenAI(model="gpt-3.5-turbo-0613")
```
lun-4, line 170: Shouldn't this be `sdk.models.medium` instead of starting gpt350613?

continuedev/src/continuedev/plugins/steps/core/core.py
```diff
@@ -201,8 +200,9 @@ async def get_prompt_parts(self, rif: RangeInFileWithContents, sdk: ContinueSDK,
 
         # If using 3.5 and overflows, upgrade to 3.5.16k
         if model_to_use.name == "gpt-3.5-turbo":
-            if total_tokens > MAX_TOKENS_FOR_MODEL["gpt-3.5-turbo"]:
-                model_to_use = sdk.models.gpt3516k
+            if total_tokens > model_to_use.context_length:
```
lun-4, line 225: This is an interesting question in terms of replacement. On one hand, the idea is to decrease amount of hardcoded models on the steps, so that users can switch to anything. On the other, model upgrades may be wanted based on context size...

I don't know if `sdk.models.large` would be a good replacement here unless we explicitly define "large" as a model that is "actually large in size, expensive" **and** "has a large context size compared to the current models" (that latter part being something that should be documented, in my opinion. it would be a footgun if someone sets their `large` model to something smaller in context to `model_to_use`).

One possible avenue to this is letting the LLM object decide if it should upgrade in the background, letting the user decide if that's behavior they want to have. I can see a future where someone does not want to use any OpenAI models, ever, and they want to upgrade from LLaMA-2 to [LLongMA-13B](https://huggingface.co/conceptofmind/LLongMA-13b) in the cases where the context is too large to warrant an upgrade.

One drawback of that is most likely we'd always blow past the normal context window and use the long model, which for the case of an OpenAI user it'd mean a giant hole in their wallet.

Maybe `model_to_use = await model_to_use.maybe_upgrade()` would be a valid compromise. Let's LLMs decide the upgrade path, while only letting specific steps that desire large context windows signal that to the LLM.

sestinj:
^and thanks for cleaning up those few things. All clear to merge main into this branch?

continuedev/src/continuedev/libs/util/count_tokens.py
```diff
@@ -2,97 +2,93 @@
...
-CHAT_MODELS = {
-    "gpt-3.5-turbo", "gpt-3.5-turbo-16k", "gpt-4", "gpt-3.5-turbo-0613"
-}
 DEFAULT_ARGS = {"max_tokens": DEFAULT_MAX_TOKENS, "temperature": 0.5, "top_p": 1,
                 "frequency_penalty": 0, "presence_penalty": 0}
 
 
-def encoding_for_model(model: str):
-    return tiktoken.encoding_for_model(aliases.get(model, model))
+def encoding_for_model(model_name: str):
```
lun-4, line 21: This might be fine for most models, but should `tiktoken` be used for all? I know that in the case of [LLaMA](https://huggingface.co/docs/transformers/main/model_doc/llama#overview)

> The LLaMA tokenizer is a BPE model based on [sentencepiece](https://github.com/google/sentencepiece). One quirk of sentencepiece is that when decoding a sequence, if the first token is the start of the word (e.g. “Banana”), the tokenizer does not prepend the prefix space to the string.

continuedev/src/continuedev/libs/util/count_tokens.py
```diff
@@ -2,97 +2,93 @@
...
 
 
-def prune_raw_prompt_from_top(model: str, prompt: str, tokens_for_completion: int):
-    max_tokens = MAX_TOKENS_FOR_MODEL.get(
-        model, DEFAULT_MAX_TOKENS) - tokens_for_completion
-    encoding = encoding_for_model(model)
+def count_chat_message_tokens(model_name: str, chat_message: ChatMessage) -> int:
+    # Doing simpler, safer version of what is here:
+    # https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
+    # every message follows <|start|>{role/name}\n{content}<|end|>\n
```
lun-4, line 38: I really am scrutinizing the OpenAI dependency here, huh? :P

This might be fine for most uses (and I didn't really read this code heavily to understand the context), but different models have different chat styles, https://github.com/oobabooga/text-generation-webui/tree/main/characters/instruction-following, we probably don't need to pass this to the LLM class *in this PR*, but might be something to think about in the future

lun-4 (commented):
Needed a bit of time to wrap up the changes, but after these we should be good to go!

continuedev/src/continuedev/libs/constants/default_config.py.txt
sestinj: Good call, this would make Continue appear much more user friendly

continuedev/src/continuedev/plugins/steps/chat.py
sestinj: As far as I know, it's still only the 0613 version that is trained for [function calling](https://openai.com/blog/function-calling-and-other-api-updates), which is needed in this specific step.

continuedev/src/continuedev/plugins/steps/core/core.py
sestinj: I do think this logic could be handled more automatically, but the uses are too sparse right now to justify adding that as a part of each LLM class. I think this is one of those that will become more apparent once it's required a few more times.

continuedev/src/continuedev/libs/util/count_tokens.py
sestinj: Want to be careful about increasing the dependency footprint, and I think the token counts won't differ so much except in specially designed inputs. So it's okay for now....but the way to solve this might be more a question of packaging—perhaps different LLMs/tokenizers can be downloaded as packages on the side?

continuedev/src/continuedev/libs/util/count_tokens.py
sestinj: Yup, and this is already happening in the Anthropic model. Loads more going into the LLM class than I would have expected, but given how (relatively) infrequently a new one should need to be created, this is probably okay. For another PR...

continuedev/src/continuedev/libs/util/count_tokens.py
lun-4: That sounds like a good idea, though I'd also say beyond the scope of this PR :P

sestinj: haha yes, yes it is

sestinj:
I believe we have a PR ready for merge! Feel free to add any other tweaks, then give me a thumbs up and I'll click the button 🚀  : )

docs/docs/customization.md
```diff
@@ -4,11 +4,25 @@ Continue can be deeply customized by editing the `ContinueConfig` object in `~/.
...
+config = ContinueConfig(
+    ...
+    models=Models(
+        default=MaybeProxyOpenAI(model="gpt-4"),
+        medium=MaybeProxyOpenAI(model="gpt-3.5-turbo")
+    )
+)
+```
+
+The `default` model is the one used for most operations, including responding to your messages and editing code. The `medium` model is used for summarization tasks that require less quality. The values of these fields are both of the [`LLM`](https://github.com/continuedev/continue/blob/main/continuedev/src/continuedev/libs/llm/__init__.py) class, which implements methods for retreiving and streaming completions from an LLM.
```
lun-4: Should we document what `small` and `large` mean too? It's fine if it'd be better to wait until we feel there's a need for such roles

sestinj:
woo! 🎉 congrats on a great PR, and thanks for such great attention to detail!

sestinj:
Hi @lun-4, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

lun-4:
I'm not on the discord nor have plans to be in it, what was the statement made there?

PR #333, opened 2023-08-01
load correct llama2 model

mchiang0610:


sestinj:
Closing as it was faster to make change in main (other were needed as well). Thank you for catching this!

PR #336, opened 2023-08-02
Fix: Don't ignore files with numbers in the basename

mdp:
This PR removes the `/\d/` regex test on fsPath, which results in Continue silently ignoring any file you're editing that contains a number in it's basename. It's definitely an edge case, but one that I ran into recently with a Python file.

To be clear, I'm unsure what the original intentions were with regards to 'editorIsTerminal'. It looks like in every case it's being called with 'vscode.window.visibleTextEditors' which in my testing never returns a terminal tab. But I'm new to this codebase, so I could definitely be missing something.

### DIFF
--- a/extension/src/continueIdeClient.ts
+++ b/extension/src/continueIdeClient.ts
@@ -125,7 +125,7 @@ class IdeProtocolClient {
 
     // Setup listeners for any selection changes in open editors
     vscode.window.onDidChangeTextEditorSelection((event) => {
-      if (this.editorIsTerminal(event.textEditor)) {
+      if (!this.editorIsCode(event.textEditor)) {
         return;
       }
       if (this._highlightDebounce) {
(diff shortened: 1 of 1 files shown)


extension/src/continueIdeClient.ts
```diff
@@ -426,33 +426,32 @@ class IdeProtocolClient {
...
   // Respond to request
 
-  private editorIsTerminal(editor: vscode.TextEditor) {
+  private editorIsCode(editor: vscode.TextEditor) {
     return (
-      !!path.basename(editor.document.uri.fsPath).match(/\d/) ||
-      (editor.document.languageId === "plaintext" &&
-        editor.document.getText() === "accessible-buffer-accessible-buffer-")
+      editor.document.languageId === "plaintext" &&
+      editor.document.getText() === "accessible-buffer-accessible-buffer-"
```
sestinj, line 435: Given the name change to the function, its output should be the opposite, so I would add a "!" to make it `!(editor.document.... "buffer-")`. And then the filter functions below should lose the "!".

sestinj:
VS Code does in fact give terminal editors when you have them positioned side-by-side, hence this check. But the digit check should be redundant, so this is a great solution!

I gave some small feedback, but I think this is otherwise completely mergable!

extension/src/continueIdeClient.ts
mdp: Yep, that makes sense. Resolved in latest.

mdp:
> VS Code does in fact give terminal editors when you have them positioned side-by-side, hence this check.

That's what I figured, but I wasn't sure. I couldn't get it to replicate in my vscode (terminal side-by-side on 1.81 on Linux), but I imagine it's probably different for other platforms or versions (or I'm just doing it wrong).

I updated the code based on the feedback and now the naming makes more sense. I also added a comment to clarify the purpose of the check.

Thanks for the quick feedback and thanks for building this product. Continue has been really fun to use, especially on refactoring work.

sestinj:
Nice work 🙌🏻  Glad you've enjoyed Continue. Time to merge!

PR #347, opened 2023-08-06
Add support for Azure, OpenAI, Palm, Anthropic, Cohere, Replicate Models - using litellm

ishaan-jaff:
This PR adds support for models from all the above mentioned providers using https://github.com/BerriAI/litellm

TLDR: continue gets:
- more models out of the box
- instant integration of models by modifying the `model` param (Easy to add models in the future, just change the model param)
- log I/O to Posthog, Sentry, Helicone, Supabase without making any code changes & get API Key management 

- instead of maintaining multiple classes / files for each provider `anthropic.py`, `openai.py` etc, continue can use one file to maintain all model Input/Output. litellm handles translating input/outputs

Here's a sample of how litellm completion is used:

```python
from litellm import completion

## set ENV variables
# ENV variables can be set in .env file, too. Example in .env.example
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion("command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/continuedev/src/continuedev/libs/llm/openai.py
+++ b/continuedev/src/continuedev/libs/llm/openai.py
@@ -2,6 +2,8 @@ from typing import Any, Callable, Coroutine, Dict, Generator, List, Literal, Uni
 
 from pydantic import BaseModel
 import openai
+from litellm import acompletion
+import litellm
 
 from ...core.main import ChatMessage
 from ..util.count_tokens import compile_chat_messages, DEFAULT_ARGS, count_tokens, format_chat_messages, prune_raw_prompt_from_top
@@ -18,6 +20,8 @@ class OpenAIServerInfo(BaseModel):
 CHAT_MODELS = {
     "gpt-3.5-turbo", "gpt-3.5-turbo-16k", "gpt-4", "gpt-3.5-turbo-0613"
 }
+CHAT_MODELS.update(litellm.model_list) # add litellm supported models
+
 MAX_TOKENS_FOR_MODEL = {
     "gpt-3.5-turbo": 4096,
     "gpt-3.5-turbo-0613": 4096,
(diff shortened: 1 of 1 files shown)


ishaan-jaff:
cc @sestinj can you please take a look at this ? 

If the initial commit looks good, happy to docs/testing and address any feedback 😊

sestinj:
@ishaan-jaff Cool stuff! My biggest concerns here are:

a) litellm depends on openai[datalib], which includes numpy and pandas, and would significantly increase our build size
b) making a single class for all of these providers would lead to a slightly less homogeneous experience in the Continue config file. We would still need a different class for connecting to local models or other custom models not supported by litellm, so this would cause most of these commercial providers to be in a single class, while other more specific providers (e.g. Ollama) would have their own unique class. This also means that litellm wouldn't completely remove the need for our LLM class to be implemented in a few separate files.

The solution I'm leaning toward is a "docs integration." We could give an example of how to use litellm [here](https://continue.dev/docs/customization), and allow users to use litellm, but without including it in the builtin library.

Open to your thoughts here, and really do appreciate what you've suggested.

ishaan-jaff:
@sestinj thanks for the feedback! 

> We would still need a different class for connecting to local models or other custom models not supported by litellm

Curious if we get liteLLM to parity with all the models you support currently, would you be open to an integration ? 



> The solution I'm leaning toward is a "docs integration." We could give an example of how to use litellm [here](https://continue.dev/docs/customization), and allow users to use litellm, but without including it in the builtin library.

Would you recommend creating a liteLLM class in continuedev and then adding it to your docs? It looks like we'd need a liteLLM class for a docs integration. Open to suggestions here

sestinj:
@ishaan-jaff As mentioned in email, I think now isn't the right time to integrate with LiteLLM. I'm going to close it to avoid too stale a PR.

For anyone who comes across this, we closed this not because LiteLLM isn't great, but because we made the crazy choice of trying to do on our own many things that LiteLLM does before we knew it existed! LiteLLM has correctly anticipated features that we ended up needing to build even months later. So, _if you are starting your own project using LLMs_, **I recommend using LiteLLM!** It'll save you time early on _and_ down the road.

..and there may come a time where Continue switches to LiteLLM - will open a new PR then!

Issue #346, opened 2023-08-04
Improve environment setup for new contributors


bra1nDump:
Ideally a new contributor would clone the project, open root in vscode, hit F5 and it will start both server / extension and they can immediately set breakpoints in both server / extension and start exploring how to code works.

# Current issues
- There is no single place in the documentation that describes how to run both VSCode extension + server with debugging enabled
- The projects have to be debugged from different VSCode windows
- [Needs a separate issue] Documentation on getting it to work is scattered (especially for the VSCode extension portion across extension/src/README.md and extension/DEV_README.md extension/README.md)

# Suggestions
I am experimenting locally with the suggestions below locally and **will create a PR once I make progress**.
The order is my suggested order of implementation.

## 1. Improve debugging setup for VSCode extension
- Create its own launch.json entry to debug extension so you don't have to be focused on a .ts file or be in the extension root
- Get watch build task to work so on file save the new code gets picked up by the extension host (where the extension being debugged is running)
  - [Requires a reload of the extension host](https://vscode-docs.readthedocs.io/en/stable/extensions/debugging-extensions/#:~:text=%24%7BworkspaceRoot%7D.-,Changing%20and%20Reloading%20your%20Extension,-Since%20the%20TypeScript)

## 2. Improve overall experience for workspace (aka Server + Extension development)
The goal is to allow seamless debugging of all components

### Option 1 - multi root workspace
- [Developer experience] Open root or sub projects to work on either server or extension, they open Run and debug sidebar and run a different configuration depending what they want to work on, the options available will depend on if the top level workspace is opened or an individual project.
- Create nested launch.json configurations in each subproject and reference them in top level
- [Con] Not a popular feature - 7 projects on github use this https://github.com/search?q=.+path%3A*..code-workspace&type=Code&ref=advsearch&l=&l=
- [Pro] Keeps status quo by allowing to open server only or extension only to avoid distractions by opening the root folder

### [My vote] Option 2 - move all launch configurations to root
- [Developer experience] Similar to last option, but always open root of the project to work on either server or extension
- Have 3 launch configurations in the top level - server only, extension only and a both (using launch.json compounds)
- [Pro] More common practice it seems

## 3. Improve dependencies setup
The goal is to have a single vscode task / bash script we can run to install all components needed for development. Motivation for lower priority is this is a one time setup fee where as the above improvements will be useful every time the project is worked on.

- Installing poetry
- Creating the virtual environment
- Setting up python interpreter
- npm install

# References
[...]

PR #352, opened 2023-08-07
Improved 2 click setup experience for new devs

bra1nDump:
Addresses #346

# Adds
- Adds a top level vscode task to install all dependencies (including poetry & creating .venv) in server, extension GUI, extension itself
- Adds top level launch.json entries to start 2 debugging sessions for both server and extension - no need to have 2 windows opened anymore
- Checks in settings.json that sets default python interpreter to the one created by poetry

# Does not address
There are currently typescript errors related to testing which I am fixing on a different branch, I will uncomment tsc build task once I fix those

# Testing
How experience for new devs will look like now:
https://github.com/continuedev/continue/assets/12608159/9f531edc-f485-4a49-a75c-48aa2f1c8157

Tested basic /edit commands still work

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,7 +131,6 @@ dmypy.json
 
 **/node_modules
 **/out
-**/.vscode/settings.json
 notes.txt
 cached_embeddings.pkl
 .ruff_cache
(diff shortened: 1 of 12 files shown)


sestinj:
@bra1nDump This is spotless, nice work! Only improvement I would suggest is instead of changing `npm run esbuild` to use esbuild-base, it might be better to use `esbuild.mjs` as the common denominator. The file includes
```
loader: {
    ".node": "file",
  },
```

which was necessary to build certain packages, and I think we're bordering on the number of flags that a file is cleaner.

Issue #359, opened 2023-08-09
@file-name is not scoped to the current workspace


bra1nDump:
@ file context completion is polluted with entries from old sessions. Happens when switching between workspaces in vscode.

## To Reproduce
1. Open some project A in vscode and open Continue sidebar, type @ - see completions from this project
2. Open another project B and type @ - see completions from both A and B

## Expected behavior
When using continue in workspace A I want to limit the file context to that project

## Root cause
Search index is never cleared between sessions.

### Fresh server per VSCode window [Suggested long term solution]
I would argue starting a fresh continue server for each vscode session, and naturally only keep meilisearch instance in memory without persisting it across sessions.
Starting a separate server for each session seems like it will prevent some hard to diagnose bugs by keeping different workspaces completely isolated.
As continue python package is similar to a language server in a lot of ways, according to chatgpt vscode usually starts a new language server per window - https://chat.openai.com/share/ad1ce898-ed78-4638-bfdf-ed5c757d6e07

Upsides:
- Addresses the bigger problem: Connecting to an existing server with some internal state has additional unpredictability.
- Easier debugging + preventing issues to be materialized later due to cross workspace bug pollination

Downsides (no particular order):
- Startup is slower
- Sharing context between projects is more challenging - I think its uncommon to reference files outside of the project directory for context. One use case I can think of is referencing some of personal prompt collection aka good code snippets that you reuse for all projects. For that use case we can allow you to provide additional context path setting in configuration file. We would include those files for every session.
- Architecture has assumed to this point being able to handle multiple clients - I imagine 
- Perf overhead - least of my concerns as machines get faster
- Sharing log files, config files - I think we should not share a single log file for all sessions as those become not very useful when logs come from different workspaces. Even now it should be fine as config is readonly and logs are append only.
- Meilisearch is accessed directly from GUI, so it would need to know a new port to access it on - I would argue we should proxy this through the server to simplify architecture dependency graph 

### Filter on demand [Suggested workaround]
When returning @file-prefix completions filter out only entries for the workspace where the request originates from

## Before opening an additional project - 4992 files
<img width="517" alt="image" src="https://github.com/continuedev/continue/assets/12608159/dc0bd2f8-60ae-4f17-854e-abdaee325426">

## After opening additional (and after closing it too) - 5024 files
<img width="836" alt="image" src="https://github.com/continuedev/continue/assets/12608159/7b6a8104-9ad7-4417-a000-598619e550af">

sestinj:
[...]

PR #360, opened 2023-08-09
Fixup search cross workspace polution

bra1nDump:
## Fixed an issue with files from other workspaces showing up in @context search

- Adding workspace folder information to react-app that uses
- Adding workspace_dir to the meilisearch documents
- Keeping the original singlton index as that requires a bigger change
- Truncated the context descriptions from absolute paths to only include path relative to workspace directory
- Minor build fix for react-app

Before
image
After
image

## Note:
I have freely changed the store.config model from workspacePath to workspacePaths as it was unused, can't think of a way how it would backfire at the moment with older versions that might have been using it. Might be missing something here

Workaround for #359

## Next steps suggested for @context search:
- Clear out files when workspace is not opened - either way you have to read it again because they might have changed since last read
- Route search through the sever for consistency
- Watch filesystem for changes and update the context items on changes


sestinj:
I like the filtering solution! Will give it a test tomorrow morning. Code looks great

PR #369, opened 2023-08-10
Ci testing

sestinj:
This adds testing to the GitHub Workflow so it's impossible to publish the extension/binaries if the extension is not capable of at least starting the server. More unit tests should be written in the future, but this should at least curb all startup failures.


sestinj:
Waiting until https://www.githubstatus.com/ is resolved : ( before we merge to main, so that the workflow can take effect

PR #396, opened 2023-08-22
feat: :children_crossing: keep file context up to data by listening f…

sestinj:
…or filesystem events


bra1nDump (commented):
Nice! Just a call out tho - vscode based implementation might cause issues with headless mode

sestinj:
Yup, would implement with watchdog for headless mode, should still work

PR #429, opened 2023-08-30
fix: 🐛 typo in core.py

eltociear:
Seperate -> Separate


sestinj:
Absolutely right sir! Thanks for the catch

PR #495, opened 2023-09-20
The shebang must be on the first line. Delete blanks and move comments. See SC1128.

Mearman:
https://github.com/koalaman/shellcheck/wiki/SC1128

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -8,7 +8,7 @@ rm -rf continuedev/.venv
 
 # 2. Create a new virtual environment and activate it
 python3 -m venv env
-source env/bin/activate
+. env/bin/activate
 
 # 3. Install the required packages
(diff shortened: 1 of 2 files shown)


sestinj:
Nice find—if it's safe to assume you found these while trying to build from source, did everything else go smoothly afterward?

Mearman:
@sestinj not got it fully up and running yet.
Are you happy for me to submit some other quality of life stuff in the install scripts etc?

sestinj:
@Mearman absolutely, will never say no to improved quality of life.

Is there anything in particular you're stuck on? I'm guessing you're following this documentation: https://continue.dev/docs/walkthroughs/manually-run-continue

One thing that is maybe a bit hidden is the need to select the "Manually Running Server" checkbox

PR #550, opened 2023-10-13
improving docs seo

TyDunn:
Followed reccs in docusaurus.io/docs/seo


TyDunn:
@sestinj can you review and merge if you are happy with it?

PR #581, opened 2023-10-24
fix: minor issues

LangLangBart:
### description

-  `install_from_source.py` script can't find the `poetry.lock` file
- PR #533 moved files from `continuedev` to `server`
```bash
git log --diff-filter=R --summary --oneline --follow server/poetry.lock
#   279593a8 Strong typing (#533)
#    rename {continuedev => server}/poetry.lock (100%)
```

---

- some typos, `explanation` is rather counterintuitive to write, English language has no repo on
  Github to request a change
	- [Explanation or Explaination | How to spell it? | Spelling ](https://word.tips/spelling/explanation-vs-explaination/)

---

updated `poetry lock --no-update` in the `continue/server` directory

---

- updated `prepackage.js` to throw an error before starting to move files around
- improved error handling in `install_from_source.py`


LangLangBart:
@sestinj 

I have a little issue; maybe you can help.

When the `install_from_source.py` script runs the line `resp = run("cd ../../../server; poetry install; poetry run typegen")`, declaration files are created. This causes the `npm run build` command in the `continue/gui/` directory to fail when it tries to execute the `tsc` command.

https://github.com/continuedev/continue/blob/f3d1ef79be90335c742fd3e505fd1bfe40b9f402/gui/package.json#L7

```mermaid
flowchart TD
    A[install_from_source.py] -->|poetry run typegen| B(prepackage.js)
    B -->|npm run build| C(continue/gui/package.json)
    C -->|tsc| Error
```


How can I solve this?

LangLangBart:
forced-pushed to rebase with `main`

sestinj:
I believe this is the issue: https://github.com/continuedev/continue/blob/f3d1ef79be90335c742fd3e505fd1bfe40b9f402/server/continuedev/core/main.py#L105

I changed it to `Any` instead of `Step` because there was a circular imports dependency, which forced me to use a "fake" Step class. But then this later caused the problem that when a Pydantic class was being initialized, it would make a copy of the real `Step` object that I passed in, causing big issues

But this would be why the "unknown" type error is coming up

LangLangBart:
That is it. I won't be addressing this issue as part of my current pull request.

sestinj:
Sounds good, I'll be making the fix later today. PR is pristine as always

sestinj:
@LangLangBart just to keep you updated, this has been fixed in main. Had to do a workaround for now. Turns out the problem is:
- Pydantic copies models when validating if they don't pass `isinstance`
- Importing a `Step` subclass from `config.py` dynamically with importlib causes `isinstance` not to pass (most likely because the module name for the __bases__ property is different)
- And so to avoid the problems I was seeing with `AnswerQuestionChroma`, I just mapped from the imported class to the original version imported the usual way

PR #584, opened 2023-10-25
Make give_user_input async

masegraye:
The only callsite in gui.py expects this to be async or return a coroutine.

Without this, uses of `sdk.wait_for_user_confirmation` will always throw in the asyncio library.


sestinj:
Thanks! This is absolutely right

PR #592, opened 2023-10-30
Prevent duplicate calls to keyDown events due to IME

benedict-lee:
**Problem**
When typing non-latin characters (such as Korean, Japanese or  Chinese) in the MainInputWindow, the KeyDown event is called duplicately due to IME.  As a result, non-latin character input does not work properly.

- Intellij extention : Korean input is almost impossible.
https://github.com/continuedev/continue/assets/4496185/66b61907-f137-4680-946a-a37ee9f25134
- VS code extention : It is better than IntelliJ, but sometimes letters are entered twice.

**Solution**
Since the moment isComposing is true is the composition phase by the IME, we can use methods to prevent events from occurring at this stage.
```
onKeyDown: (event) => {
    // Prevent duplicate calls to keyDown events due to IME.
    if (isComposing) {
        return;
    }
    // handle event
```

### DIFF
--- a/gui/src/components/ComboBox.tsx
+++ b/gui/src/components/ComboBox.tsx
@@ -1081,6 +1081,11 @@ const ComboBox = React.forwardRef((props: ComboBoxProps, ref) => {
                 setInputFocused(false);
               },
               onKeyDown: (event) => {
+                // Prevent duplicate calls to keyDown events due to IME.
+                if (isComposing) {
+                  return;
+                }
+                
                 dispatch(setBottomMessage(undefined));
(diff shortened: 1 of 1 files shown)


sestinj:
@benedict-lee thanks for making this change! I'm going to wait just a bit so I can merge it in with a big refactor that I have in the works. Shouldn't be later than today, I just wanted to be sure this didn't get lost in a merge conflict

sestinj:
@benedict-lee merging now, and the changes will be included in a published version later today. Thank you for taking the time to do this!

PR #593, opened 2023-10-30
Fix URL for Together.ai models list in models.md

pzaback:


TyDunn (approved):
Thanks for fixing this!

Issue #576, opened 2023-10-23
Dropdown selection obscured by sticky title


LangLangBart:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- Continue: 0.0.413 (2023-10-23T03:23:34.417Z)
```

### Description
In the `ComboBox.tsx` file, the `sticky` title and its description of the dropdown menu result in the currently selected item being partially or fully obscured. 

#### Example 
This issue is particularly evident with items at the top of the dropdown list, as shown in the GIF below. It often occurs when navigating upwards using the arrow keys.

<img src="https://github.com/continuedev/continue/assets/92653266/98beb21e-6d9e-45de-8e45-8688eefbfac0" width="500">

---

### Bisect
This issue was first observed in commit: 8d423fd8

https://github.com/continuedev/continue/blob/8d423fd8d1d5b136e8138a906e8594ab93ec1982/extension/react-app/src/components/ComboBox.tsx?plain=1#L697-L720

---

### fix attempt 
The required solution shall ensure that the list items do not go underneath the sticky title and description.

The `50px` shouldn't be hardcoded, but rather the height of the `div` with the `sticky` title.

```diff
--- a/gui/src/components/ComboBox.tsx
+++ b/gui/src/components/ComboBox.tsx
@@ -1311,6 +1311,7 @@ const ComboBox = React.forwardRef((props: ComboBoxProps, ref) => {
               {nestedContextProvider.description}
             </div>
           )}
+          <div style={{ maxHeight: `${UlMaxHeight - 50}px`, overflow: 'auto' }}>
           {downshiftProps.isOpen &&
             items.map((item, index) => (
               <Li
@@ -1364,6 +1365,7 @@ const ComboBox = React.forwardRef((props: ComboBoxProps, ref) => {
                 )}
               </Li>
             ))}
+          </div>
           {downshiftProps.isOpen && items.length === 0 && (
             <Li
               key="empty-items-li"
```

<img src="https://github.com/continuedev/continue/assets/92653266/d19e6f84-b306-42f0-b35d-f52b11a067ce" width="500">

[...]

PR #605, opened 2023-11-06
fix: dropdown with the height of the sticky div

LangLangBart:
### description
- use the height of the sticky div, rather than the `50px`
	- this PR is based on the commit written by `sestinj` that was lost in some merge
- Ref: #576

- if you increase the font size the `50px` aren't enough
	- the GIF below uses `Font Size: 25`


sestinj:
thanks for cleaning this up : )

PR #618, opened 2023-11-12
fix(vscode): handle null values for edits and page in getSidebarContent

LangLangBart:
### description
- when `edits` or `page` is undefined, the expressions `${edits && }` and `${page && }` will
  evaluate to `undefined` and that gets converted to the string "undefined" in the template literal

- to reproduce the issue, open 'Continue' and scroll down below the LLM selector, optionally open
  the 'Developer Tools'

- issue was introduced with commmit: 4763da10c99b6eda490cdfd510b73e199dc2be42
	- released in **0.5.2 Pre-Release  2023-11-10T20:02:30.66Z ✹ 10/Nov/23**


sestinj:
Thank you again for beating me to this!

PR #619, opened 2023-11-12
feat(gui): add flex-grow to TdDiv in history page

LangLangBart:
### description
The key point here is to ensure that the `TdDiv` is taking up the full width of the parent div so that
clicking anywhere on the history row triggers the click event.


sestinj:
Nice! Looks perfect to merge

PR #629, opened 2023-11-16
initial commit of the OpenAI Agent POC

birdperson1970:
This is still rought and needs a new UI field to allow the user to set their own assitant_id if you comment out:

[  self.assistant_id='thread_lCb2rKUzIOodcA3MZaZu61Bv'](https://github.com/birdperson1970/continue_openai_agent_poc/blob/a42dca2dd82ceda74792abc2fe46e22f402f874e/server/continuedev/libs/llm/openai_agent.py#L67
)

It will create a basic assitant each time which will be really annoying to cleanup....


server/continuedev/libs/llm/openai_agent.py
```diff
@@ -0,0 +1,289 @@
...
+    async def _stream_complete(self, prompt, options):
+         async for chunk in self._stream_chat(
+                [{"role": "user", "content": prompt}], options
+            ):
+                if "content" in chunk:
+                    yield chunk["content"]
+
+
+
+    async def _stream_chat(self, messages: List[ChatMessage], options):
```
sestinj, line 96: This looks like the really important part!

There's a high-level thing that I want to point out right away. It's critical to make sure that every sub-class of LLM has the same interface, and it would probably require changing something here to give it access to the SDK to call tools.

So the first thing I might think is that you could find a way to keep a lot of this logic inside of the OpenAIAgent class and just call the SDK from inside of a step. But the question then becomes: how do we pass the output of the tools back into the assistant?

It might actually be possible to do the following in a Step: [call stream_chat] -> [get response] -> [call tool based on response (in the Step, not OpenAIAgent)] -> [pass results of tool back as the input to another stream_chat call] -> [repeat].

The important thing is to make sure that no "state" is stored in the LLM class, so that it can be called multiple times without worrying about what previous calls did. I think the key to this is creating a new thread for each stream_chat call, and making separate stream_chat calls whenever a tool is invoked

birdperson1970: I understand that the LLMs needs to be reentrant.  OpenAI Agent conversational Threads are designed to capture the entire context with optimisation taking place once the it approaches the 128k context window.  

Would you object to the following changes:
	1. A thread_id UUID field is added to SessionState - Given that the sessionState.title(ide) is editable a new persistent identifier would seem appropriate. This would give a key to the SessionState and server you well in the long term.
	2. Map the SessionState.thread_id to a [OpenAI.thread.id](http://openai.thread.id/) - If you could point me in the right direction on how to do this in a  persistent manner …
	3. Add thread_id to the CompletionOptions for the call to stream_chat.

sestinj:
Going to close this PR for the same reason I mentioned in the other, but once again thank you for taking the time to push Continue forward! We're going to do something very similar to this in typescript really soon, so the idea lives on!

PR #636, opened 2023-11-17
Expand information on model roles, particularly summarize

travellingprog:
This provides more information on the model roles, particularly the `summarize` role. This is important because that role is set by default to OpenAI's free trial, so the user is likely to want more information on it, eventually.


sestinj:
Great work! Appreciate you doing this : )

PR #638, opened 2023-11-20
Update template configuration docs

pzaback:
Attempting to add `config.models[0].template_messages = template_alpaca_messages` or
`config.models[0].prompt_templates["edit"] = "<INSERT_TEMPLATE_HERE>"` to config.py results in `[WARNING] Failed to modify config: 'Models' object is not subscriptable` when saving config.py
This PR modifies those lines in documentation to account for the fact that "models" is an instantiated Models object containing LLM objects


sestinj:
thank you!! this was an important thing for me to overlook

PR #642, opened 2023-11-29
Dockerize the continue server app

djcopley:


sestinj:
@djcopley this is awesome! I think a Docker container would have huge benefit to Continue users and make setup much easier, something that we deeply care about. I'd actually love to hear more about some of the potential troubles that led to creating this, and what you are looking for in a reproducible setup.

To be quite honest about the PR: Many users will want to deploy in ways other than with Ollama, making this a bit of an unfair endorsement. But more importantly, anything we add to this repository requires responsibility, and though it's just one file, a Dockerfile would be an extreme focal point. As much as we might state otherwise, it would be treated as a product, one that comes with troubleshooting, security concerns, and lots of maintenance. For this reason, it's something we want to _really_, _really_ get right, and are actively working toward. Until then, it wouldn't be right to host this.

But what we can do is:
1. Help make setup as easy as possible for you! Docker containers are a great way to do this and if you could use our help getting one going we're more than happy to assist
2. Kickstart a conversation—what are the needs you have when setting up Continue, what guarantees would you want out of a setup process, how might we test those? This would be a great way to help us move toward the version that we _will_ maintain, and you'll probably come out of it having shaped an even better version of Continue : )

sestinj:
Fortunately for Continue users, but unfortunately for this PR, we've moved completely away from using the separate Python server, which means that Continue is a self-contained extension with no need for a separate server. Going to close this for that reason, but seriously thank you again for taking the time with this!

PR #676, opened 2023-12-14
add python linting github workflow

danieleades:
- adds linting with [ruff](https://docs.astral.sh/ruff/) to CI
- addresses *most* of the errors. There's still a couple left to resolve, related to usage of types which don't appear to actually exist


sestinj:
As mentioned in the issue we won't be using python any longer, so I'll probably disable this, but would be good to have around just in case we write more python in the future...so going to merge!

PR #678, opened 2023-12-15
use dependabot to keep github actions up-to-date

danieleades:


sestinj:
love it! may end up adding dependabot hooks for other parts of the codebase as well, I think will be quite helpful

danieleades:
> love it! may end up adding dependabot hooks for other parts of the codebase as well, I think will be quite helpful

yeah i'm a big fan of dependabot (or Renovate). Especially for libraries. For application code the PRs can get pretty noisy since every patch release of a dependency can trigger a PR, so you might have to play around with the config to clamp it down to something you're happy with. Or get your CI tight enough you can automerge dependency updates. Or something.

PR #691, opened 2023-12-17
Move OpenAI "v1" to apiBase

capdevc:
Currently, any usage of the openai provider other than azure automatically adds a "v1" into the API URL. This makes using the openai provider with servers that don't use "v1" impossible. This was specifically motivated by wanting to try out the perplexity.ai API which serves at `https://api.perplexity.ai/chat/completions`

I have moved the "v1" to the default apiBase in the provider. This also aligns with how OpenAI themselves construct URLs in their packages:

https://github.com/openai/openai-python/blob/15d3aba42ed40713ef2d2757462d19846049523a/src/openai/_client.py#L301-L304

```
        if base_url is None:
            base_url = os.environ.get("OPENAI_BASE_URL")
        if base_url is None:
            base_url = f"https://api.openai.com/v1"
```

### DIFF
--- a/core/llm/llms/Mistral.ts
+++ b/core/llm/llms/Mistral.ts
@@ -4,7 +4,7 @@ import OpenAI from "./OpenAI";
 class Mistral extends OpenAI {
   static providerName: ModelProvider = "mistral";
   static defaultOptions: Partial<LLMOptions> = {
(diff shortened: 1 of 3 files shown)


yieme:
Unsure why this was closed. An alternative might be to support baseUrl if the provider is "openai"

PR #703, opened 2023-12-23
fix: add focus to InputToolbar on click

LangLangBart:
### description

When a user interacts with the `MainTextInput` area, there is a non-responsive space between the
`+ Add Context` and `⏎ Enter` buttons. It would be beneficial if clicking this space also focused the
`MainTextInput`, enabling immediate typing.

See GIF below for more clarity


sestinj:
Nice!

PR #716, opened 2023-12-29
fix: editor focus handling on click events

LangLangBart:
### issue
When a user attempts to focus on the text input by clicking within the text area, there is a
portion of dead space that remains unresponsive.

In the image below, the `EditorContent` is highlighted in red and the `InputToolbar` in green. These
are the only areas where a user can click to activate the text input, aside from the two special
buttons that retain their unique click functions in the `InputToolbar`. 

In version 0.7.58 the user could click anywhere.

---

The fix intends to add the click event to the `InputBoxDiv`.


sestinj:
nice!

PR #717, opened 2023-12-30
refactor: Simplify focus command in TipTapEditor

LangLangBart:
### description
I frequently encounter an issue where, when pressing

- <kbd>⌘ Command</kbd> + <kbd>M</kbd> or
- <kbd>⌘ Command</kbd> + <kbd>⇧ Shift</kbd> + <kbd>M</kbd>

with some code as input, the cursor inexplicably lands in the title of the preview code.

This issue occurs consistently when a user clicks on the `History` icon in the top right of the
extension.

Refer to the GIF below for clarity.

---

Interestingly, simply removing one `focus()` method and running the focus again after a brief
timeout seems to resolve the issue. Is there a specific reason the `focus()` was added twice?

### DIFF
--- a/gui/src/components/mainInput/TipTapEditor.tsx
+++ b/gui/src/components/mainInput/TipTapEditor.tsx
@@ -328,10 +328,10 @@ function TipTapEditor(props: TipTapEditorProps) {
                 item,
               },
             })
-            .focus("end")
             .run();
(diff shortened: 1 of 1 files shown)


sestinj:
I think was just a mistake after having tried to solve this exact problem. Thank you for doing this!

PR #728, opened 2024-01-07
Add SAP Generative AI Hub as Model Provider

mbrner:
This PR is to add the [SAP Generative AI Hub](https://blogs.sap.com/2023/12/21/generative-ai-hub-out-now/) as a model provider.

The SAP Generative AI Hub provides a proxy for commercial models such as OpenAI and the ability to deploy your own models. OAuth is always used for authentication, i.e. if you use OpenAI models via the Generative AI Hub, you do not need an API key, but the OAuth token.

In this PR, I have slightly modified the OpenAI class and introduced protected methods for determining the URLs and for generating the request headers. Based on the modified class, I have added a SAPGenAI class that adapts the two methods for the Gen AI Hub.


sestinj:
This looks really good! Thanks for being making the PR so comprehensive. I'd prefer to leave SAP GenAI Hub out from modelData because we want to keep the UI uncluttered—we already leave out a few lesser-used providers, otherwise it might be confusing to new users.

I'm guessing this is just a detail of GenAI Hub, but could you explain briefly what the vllm type (vs. azure or openai) is used for?

Other than these two things, I'll run some tests on my own, and if it seems to be working well on your side also, then this is a PR that I am happy to move forward with.

mbrner:
I will remove the  GenAI Hub from model data:
I also got to reopen a new PR, because I did some commits using the wrong account...
Sorry for that

sestinj:
No worries

PR #699, opened 2023-12-21
Add support for phi2 model via Ollama

apawloski:
This adds support for Microsft's [phi-2](https://ollama.ai/library/phi) model via ollama. Thought this would be a simple first PR. If there are tests or docs that should be updated, I'm happy to do that but could use a pointer as to where.

PR #732, opened 2024-01-08
WIP: Add support for Phi-2 via Ollama

apawloski:
This is a rework of https://github.com/continuedev/continue/pull/699 to use `preview` instead of `main`.

I've followed the steps [here](https://github.com/continuedev/continue/blob/preview/CONTRIBUTING.md#adding-models), though skipping `modelData.ts` changes as noted [here](https://github.com/continuedev/continue/pull/699#issuecomment-1867929206).

As for prompts (in `index.ts`), I'm not sure what your preference is. Was thinking about adding to [this block](https://github.com/continuedev/continue/blob/preview/core/llm/index.ts#L51C1-L58) vs creating new prompt type or re-using one. Thoughts on this?


sestinj:
All looks good, thank you! As for prompt type, I've not actually used phi2 before so would defer to you to verify this, but it looks like the model [expects a prompt template](https://huggingface.co/microsoft/phi-2#qa-format) that is different from any of the others that we have [already implemented](https://github.com/continuedev/continue/blob/main/core/llm/templates/chat.ts). If this is right, then we should add the chat template [here](https://github.com/continuedev/continue/blob/main/core/llm/templates/chat.ts) and edit template [here](https://github.com/continuedev/continue/blob/main/core/llm/templates/edit.ts). Happy to make this change if it's easier, lmk!

apawloski:
I took a stab at adding the chat/edit templates. For Chat, I tried to use a phi-2 specific prompt, whereas for Edit, I reused the Simplified prompt. (In the docs for Phi-2 the code usage expects to be completion of [code described in comment](https://huggingface.co/microsoft/phi-2#code-format), but I think simplified works better with the arbitrary highlighting context. Thoughts?

sestinj:
I think this sounds good as a start, and if response quality ends up saying otherwise after a bit of usage then it will be no problem to make the adjustment. Thanks for such a comprehensive PR!

sestinj:
Hi @apawloski, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

PR #736, opened 2024-01-09
Support SAP Generative AI Hub Models

mbrner:
This PR is to add the [SAP Generative AI Hub](https://blogs.sap.com/2023/12/21/generative-ai-hub-out-now/) as a model provider (reopened from https://github.com/continuedev/continue/pull/728).

The SAP Generative AI Hub provides a proxy for commercial models such as OpenAI and the ability to deploy your own models. OAuth is always used for authentication, i.e. if you use OpenAI models via the Generative AI Hub, you do not need an API key, but the OAuth token.

In this PR, I have slightly modified the OpenAI class and introduced protected methods for determining the URLs and for generating the request headers. Based on the modified class, I have added a SAPGenAI class that adapts the two methods for the Gen AI Hub.

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -23,6 +23,10 @@ declare global {
     apiType?: string;
     region?: string;
     projectId?: string;
+    resourceGroup?: string;
+    authURL?: string;
+    clientID?: string;
+    clientSecret?: string;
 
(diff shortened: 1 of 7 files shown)


mbrner:
Sorry for the mess... Still have to get rid of the wrong username...

PR #737, opened 2024-01-09
Sap generative ai hub

mbrner:
This PR is to add the [SAP Generative AI Hub](https://blogs.sap.com/2023/12/21/generative-ai-hub-out-now/) as a model provider (reopened from https://github.com/continuedev/continue/pull/728).

The SAP Generative AI Hub provides a proxy for commercial models such as OpenAI and the ability to provide your own models. OAuth is always used for authentication, i.e. if you use OpenAI models via the Generative AI Hub, you do not need an API key, but the OAuth token.

In this PR, I have slightly modified the OpenAI class and introduced protected methods for determining the URLs and for generating the request headers. Based on the modified class, I added a SAPGenAI class that adapts the two methods for the Gen AI Hub.

ToDos:

- Check how to adjust config validation, because the GenAI Hub's `azure` `openai/apiType` does not require `engine` + `apiKey`and the `openai` type does not require `apiKey`.

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -23,6 +23,10 @@ declare global {
     apiType?: string;
     region?: string;
     projectId?: string;
+    resourceGroup?: string;
+    authURL?: string;
+    clientID?: string;
+    clientSecret?: string;
 
     _fetch?: (input: any, init?: any) => Promise<any>;
 
@@ -198,6 +202,12 @@ declare global {
     // GCP Options
     region?: string;
     projectId?: string;
+
+    // SAP Gen AI Core options
+    resourceGroup?: string;
+    authURL?: string;
+    clientID?: string;
+    clientSecret?: string;
   }
   type RequireAtLeastOne<T, Keys extends keyof T = keyof T> = Pick<
     T,
@@ -230,6 +240,7 @@ declare global {
   >;
 
   // IDE
+
 
   export interface DiffLine {
     type: "new" | "old" | "same";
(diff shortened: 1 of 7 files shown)


sestinj:
@mbrner This looks really solid. Can we group together those options into a single interface like this?

```typescript
interface SapGenAiOptions {
    // SAP Gen AI Core options
    resourceGroup?: string;
    authURL?: string;
    clientID?: string;
    clientSecret?: string;
}
```

We didn't do this with the Azure OpenAI options but I may go back and do that on my own—it seems like things will start to clutter without this.

Otherwise I'd consider this basically ready as long as it seems to be working on your side

sestinj:
There's a set of clauses for the models array inside of an `allOf` that has good example of how to conditionally require certain properties. In this case after grouping into the interface, you can just require the one.

A sample: https://github.com/continuedev/continue/blob/a3c6d04e6dfa88fa64a0a1b738509d930a9147f3/extensions/vscode/config_schema.json#L276C9-L288C11

sestinj:
Hi @mbrner, wanted to check in on this—happy to make these changes myself if it's easier, just let me know! Otherwise, since the PR is in draft state, I'll wait until I get a green-light from you.

sestinj:
Hey @mbrner, I'm going through older PRs and it looks like this one got stuck in the draft state. Can you let me know whether you're planning on preparing this for merge, or whether it is actually ready in your mind? (other than merge conflicts of course, I could clean those up). Otherwise I'll probably close it out and keep it around for reference

PR #752, opened 2024-01-14
Fix logic error in streamed JSON object parsing

matbee-eth:
buffer becomes any character after `\n`, as anything before `\n` should be considered a valid JSON object


sestinj:
@matbee-eth Will pull into main but I'm going to first publish this just in pre-release.

sestinj:
Thanks for making the change, it works perfectly! And as far as other streaming implementations: Ollama was actually a bit special, so all of the others use a single function, and it doesn't seem to have this problem

Issue #688, opened 2023-12-17
HuggingFace TGI provider missing from menu


erezsharon:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue: 0.6.4 and onward
- Python: 3.11
- IDE: VSCODE 1.85.1
```

### Description

When opening the ""add provider"" menu the option to select HuggingFace TGI is now missing from the menu. This was working fine until version 0.6.4

### To reproduce

Click on the "+" sign and scroll down to the end - no option to select HuggingFace TGI

sestinj:
@erezsharon we've removed this from the visual menu because it's used less frequently than some of the others and tends to be more difficult to set up. But if you type "provider": "huggingface-tgi" in your config.json, the option still exists

PR #754, opened 2024-01-16
Add Hugging Face TGI to ModelProvider UI

simon376:
Add the 🤗Text Generation Inference Model Provider to the User Interface

For the short description I've used the [tagline of its repo](https://github.com/huggingface/text-generation-inference?tab=readme-ov-file#text-generation-inference), for the longer one, [your own documentation](https://github.com/continuedev/deploy-os-code-llm?tab=readme-ov-file#tgi)

### DIFF
--- a/gui/src/util/modelData.ts
+++ b/gui/src/util/modelData.ts
@@ -737,6 +737,27 @@ After it's up and running, you can start using Continue.`,
       },
     ],
   },
+  "huggingface-tgi": {
+    title: "Hugging Face Text Generation Inference",
+    provider: "huggingface-tgi",
+    description:
+      "A Rust, Python and gRPC server for text generation inference. Used in production at HuggingFace to power Hugging Chat, the Inference API and Inference Endpoint.",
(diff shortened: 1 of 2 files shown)


simon376:
seems like this was removed one purpose :/
 #688

sestinj:
@simon376 it was, sorry. If there's a really strong case for this I'm open to hearing it, but the need to keep the UI simple is very top-of-mind for us, especially since the number of providers is growing. The basic heuristic on this is that providers that are made for multiple people and take a while to setup probably won't make the page. Aim of the UI is mostly to make things accessible for beginners

simon376:
> @simon376 it was, sorry. If there's a really strong case for this I'm open to hearing it, but the need to keep the UI simple is very top-of-mind for us, especially since the number of providers is growing. _The basic heuristic on this is that providers that are made for multiple people and take a while to setup probably won't make the page. Aim of the UI is mostly to make things accessible for beginners_

I see, that makes sense 👍

Issue #609, opened 2023-11-07
Query server for model availability of OpenAI-compatible servers


surak:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

OpenAI-compatible inference servers have the `/v1/models` endpoint which lists their models and capabilities. e.g.:

```
curl --header "Authorization: Bearer MY_OPENAI_KEY"   https://api.openai.com/v1/models
```

This returns a JSON list of the models and some of their capabilities. Compatible servers implement the same.

This would make the configuration of such servers much easier. Instead of writing the config.py manually, one can just query the server for the models...

sestinj:
To be sure, the use case here is to populate the dropdown if you're using the OpenAI class with a personal API token?

surak:
That or any OpenAI model server, where models can have completely different names from that.

but yea, the autopopulate part is correct

simon376:
i tried using continue with a small team, and it was horrendous to set up, especially since we had to switch models a few times.
A way to automatically set things up would be great.
e.g. when using TGI there's a /info endpoint that gives model name, context window, and more information. this could be used to configure the prompt template, special tokens, and so on automatically by only having the user select TGI + entering the URL.

If I can help let me know

sestinj:
@simon376 Thanks for the honesty here. Model setup is something we're planning to spend most of next week on, so if you don't mind I'd actually love to hear the entirety of your experience. Is GitHub issues the best way to communicate, Discord, or even a quick call? Whatever is most comfortable. I'll have many questions and can hopefully share some early ideas with you

Our goal is like you said, to get this to a point where you wouldn't need to worry about the config file at all

simon376:
I will message you on discord when I find the time. my choice of words was a bit harsh tbh 😅 maybe I was a bit ill-prepared

sestinj:
No worries! "horrendous" was properly descriptive in my opinion : )

Many of the noted improvements are now available in the pre-release. Still yet to autodetect available models, but I still think this is a good idea

PR #502, opened 2023-09-23
Added a function to docs how to discover all availble models from oll…

meganoob1337:
…ama automatically

This is a function i wrote so the Config gets Prefilled with the Ollama Models that are present in the local installation, 
you don't need to accept the PR, but maybe you can add it somehow if you want.
Here is my config.py:

```python
"""
This is the Continue configuration file.

See https://continue.dev/docs/customization to for documentation of the available options.
"""
from continuedev.src.continuedev.libs.llm.ollama import Ollama

from continuedev.src.continuedev.core.models import Models
from continuedev.src.continuedev.core.config import CustomCommand, SlashCommand, ContinueConfig
from continuedev.src.continuedev.plugins.context_providers.github import GitHubIssuesContextProvider
from continuedev.src.continuedev.libs.llm.maybe_proxy_openai import MaybeProxyOpenAI

from continuedev.src.continuedev.plugins.steps.open_config import OpenConfigStep
from continuedev.src.continuedev.plugins.steps.clear_history import ClearHistoryStep
from continuedev.src.continuedev.plugins.steps.feedback import FeedbackStep
from continuedev.src.continuedev.plugins.steps.comment_code import CommentCodeStep
from continuedev.src.continuedev.plugins.steps.share_session import ShareSessionStep
from continuedev.src.continuedev.plugins.steps.main import EditHighlightedCodeStep
from continuedev.src.continuedev.plugins.steps.cmd import GenerateShellCommandStep
from continuedev.src.continuedev.plugins.context_providers.search import SearchContextProvider
from continuedev.src.continuedev.plugins.context_providers.diff import DiffContextProvider
from continuedev.src.continuedev.plugins.context_providers.url import URLContextProvider
from continuedev.src.continuedev.plugins.context_providers.terminal import TerminalContextProvider
import requests

def get_model_names(url):
    try:
        # Send a GET request to the URL
        response = requests.get(url)

        # Check if the request was successful (status code 200)
        if response.status_code == 200:
            # Parse the JSON data from the response
            data = response.json()

            # Extract the "name" field from each object in the "models" array
            names = [model["name"] for model in data.get("models", [])]

            return names
        else:
            # Print an error message if the request was not successful
            print(f"Failed to fetch data. Status code: {response.status_code}")
            return []
    except Exception as e:
        # Handle any exceptions that may occur during the request
        print(f"An error occurred: {str(e)}")
        return []

# Function to process model names
def process_model_name(name):
    return Ollama(
			context_length=2048,
			model=name,
[...]

PR #755, opened 2024-01-16
Query 🤗TGI /info endpoint to get model name and context window

simon376:
Similar to #609 and #502
The Hugging Face Text Generation Inference Server exposes a `/info` endpoint (see [here](https://huggingface.github.io/text-generation-inference/#/Text%20Generation%20Inference/get_model_info)) which gives us the model name and the context size, allowing for automatic population of these parameters.

code was adapted from the Ollama `/api/show` implementation [here](https://github.com/continuedev/continue/blob/982c0cd15674b415ae24536899b736f471ef9f84/core/llm/llms/Ollama.ts#L20)


sestinj:
love this, nice work!

PR #765, opened 2024-01-20
address some of the ruff warnings

danieleades:
i know the python implementation isn't going to be used in the long-term (i'm assuming that's still the assumption?) but in the meantime, there are errors in `main` that are causing CI to fail.

Consistently failing jobs means CI feedback on PRs is less meaningful.

i've squashed most of the warnings, except for these-

```
continuedev/libs/util/copy_codebase.py:104:18: F821 Undefined name `ManualEditAction`
continuedev/plugins/context_providers/embeddings.py:28:29: F821 Undefined name `ChromaIndexManager`
continuedev/plugins/context_providers/embeddings.py:36:34: F821 Undefined name `ChromaIndexManager`
```

where are these objects defined??

### DIFF
--- a/server/continuedev/core/steps.py
+++ b/server/continuedev/core/steps.py
@@ -13,7 +13,6 @@ from ..libs.util.count_tokens import DEFAULT_MAX_TOKENS
 from ..libs.util.devdata import dev_data_logger
 from ..libs.util.strings import (
     dedent_and_get_common_whitespace,
-    remove_quotes_and_escapes,
 )
 from ..libs.util.telemetry import posthog_logger
 from ..libs.util.templating import render_prompt_template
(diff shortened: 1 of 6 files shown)


sestinj:
It looks like they are defined...nowhere. ManualEditAction is *extremely* old code, and I think `ChromaIndexManager` was a WIP, or just ended up unused.

I've moved the Python server into a [separate repo](https://github.com/continuedev/legacy-python-server) so it might be okay to leave these few missing parts.

sestinj:
I'm happy to merge this, but it would get cleared when preview merges to main. Do you want to make a separate PR on the other repo?

danieleades:
> I'm happy to merge this, but it would get cleared when preview merges to main. Do you want to make a separate PR on the other repo?

can do. might be a few days before i get a chance.

Where's the best place to track progress on the new server implementation?

sestinj:
Hey @danieleades, just cleaning up some older PRs and it looks like this one is probably pretty stuck because of how much we moved around parts of the codebase. Many of the ideas here have lived on however, so I appreciate the time you took here and in the other ones that did end up getting merged. Hope all is well!

PR #766, opened 2024-01-20
apply more aggressive ruff configuration

danieleades:
this PR is an experiment in applying a more aggressive ruff linting configuration


danieleades:
i'll close this. it's clear this changes would need to be in separate PRs, even if they would be considered. but there's not point if the python server is going to be deprecated anyway

PR #769, opened 2024-01-21
Introduce Interactive Azure AD Login Workflows for Azure OpenAI Models

elabbarw:
### Description

This PR enables interactive Azure AD or Azure Entra login workflows for OpenAI models on Azure. Users can control access by assigning AD users and groups to the Azure OpenAI resource using the "Cognative Services OpenAI User" role without exposing any api keys.

### Changes Made

- Added the @azure/identity 4.0.0 module to the package.json of core.
- Added methods to handle interactive login and token expiry within the core openai.ts file.
- Modified the streamchat methods within the core openai.ts file to allow the usage of the AAD token.
- Added the apiType of "azuread" to config_schema.json.

### Benefits

- **Security:** Allows developers to use continue.dev with private models on Azure until token expiry without exposing API keys.
- **Audit Trail:** Provides an Azure and AD audit trail for model access.

### Usage

1. Add AD users/groups to the Azure OpenAI resource using the "Cognative Services OpenAI User" role 
image

2. Add your Azure OpenAI models as you normally would in the plugin's config.json file.
2. Set apiType to "azuread" and leave apiKey empty.
3. Start a session or chat to trigger interactive login.
4. Tokens are saved for subsequent requests.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -209,7 +209,7 @@ jobs:
 
       # 5 Send to Discord Webhook
       - name: Discord Commits
-        uses: Sniddl/discord-commits@v1.6
+        uses: Sniddl/discord-commits@1.7
         with:
           webhook: ${{ secrets.DISCORD_WEBHOOK }}
           template: "avatar-with-link"
--- a/core/llm/llms/OpenAI.ts
+++ b/core/llm/llms/OpenAI.ts
@@ -1,4 +1,5 @@
 import { ChatCompletionCreateParamsBase } from "openai/resources/chat/completions";
+import { InteractiveBrowserCredential } from "@azure/identity";
 import { BaseLLM } from "..";
 import {
   ChatMessage,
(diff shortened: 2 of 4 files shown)


elabbarw:
@sestinj For your kind attention :smile:

elabbarw:
Works ok with VScode but not with IntelliJ... Setting as draft..

sestinj:
@elabbarw Thanks for this thoughtful PR! My first thought is that a pattern is emerging in seeing a few attempts at authentication. There are going to be a ton of different providers, all with their own form of authentication, and as you've seen here with JetBrains not working, the requirement to download an entirely new NPM package, and the complication of extra parameters...it will add up to a lot of complexity for both users and maintainers.

With this in mind, we're hoping to do authentication once and do it really right (possibly a single point of auth for Continue, or at least something that doesn't require multiple different client-side auth providers). My preferred way of moving forward would be to revisit this in the near future when we start to build such "beyond the extension" features. But as we're designing this, it would be extremely helpful to understand what requirements led to this PR. Hopefully we can get all of them resolved in a more well-packaged product soon.

sestinj:
@elabbarw I wouldn't mind at all. And please keep me updated if there's anything else Continue can do to make life easier. In the very near future we'll be tackling this class of problem, and would love to hear more feedback leading up to then. Will be in touch : )

PR #796, opened 2024-01-31
Fix streamDiff tests

my8bit:
There was a use case that where failing because it expects the `old` ( or `-` ) to come only after the `new` (or `+`). 
Code produces the correct diff but is structured in a way that injects the `+` first. In this PR instead of inserting the `+` first, the algorithm stores `-` in the stack and interest before the upcoming `+` this is the way it reprioritizes the `-` changes over `+` in the final diff. 

Example before the change:
```diff
    + D
    + E
    - A
    - B
      C
    + F
    + C
```

now becomes
```diff
    - A
    - B
    + D
    + E
      C
    + F
    + C
```

But exploring the logic around that there are some open questions still for me that I'd like to understand.
Here is some logic, that aims to prevent bugs I assume which I commented on to make a match:

```ts
      if (
        newLine.trimStart() === oldLines[i].trimStart() // &&
        // (permissiveAboutIndentation) || newLine.trim().length > 8) 
      ) {
```

in this case when we compare the ` }` with `  }` we trim the start and get the `}` and `}` but as long as `permissiveAboutIndentation` is `false` and `"}".trim().length` is not longer than 8 it's not matching the closing brackets correctly.

@sestinj could you please have a look at the updated logic, that fixes the test? Is it something that the test intends to cover initially? If yes, then let's think about how we can adjust this logic ` (permissiveAboutIndentation) || newLine.trim().length > 8) ` to not cause the potential regression. Because seems that it was needed to cover some edge cases. If you can give an example of that, it's also possible to cover it here in the test to avoid regression in the future.

### DIFF
--- a/core/diff/streamDiff.ts
+++ b/core/diff/streamDiff.ts
@@ -37,7 +37,7 @@ export async function* streamDiff(
         yield { type: "old", line: mutatedOldLines.shift()! };
       }
       if (postponedLines.length > 0) {
-        for (let i = 0; i < postponedLines.length; i++) {
+        for (let i = 0; i <= postponedLines.length; i++) {
           yield { type: "new", line: postponedLines.shift()! };
         }
       }
--- a/core/diff/streamDiff.ts
+++ b/core/diff/streamDiff.ts
@@ -13,14 +13,15 @@ export async function* streamDiff(
   oldLines: string[],
   newLines: LineStream
 ): AsyncGenerator<DiffLine> {
-  oldLines = [...oldLines]; // be careful
+  const mutatedOldLines = [...oldLines]; // be careful
   let seenIndentationMistake = false;
 
   let newLineResult = await newLines.next();
+  const postponedLines: string[] = [];
   while (oldLines.length > 0 && !newLineResult.done) {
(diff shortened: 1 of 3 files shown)


core/diff/streamDiff.ts
```diff
@@ -13,14 +13,15 @@ export async function* streamDiff(
   oldLines: string[],
   newLines: LineStream
 ): AsyncGenerator<DiffLine> {
-  oldLines = [...oldLines]; // be careful
+  const mutatedOldLines = [...oldLines]; // be careful
```
my8bit, line 16: @sestinj ideally it's a good practice to not reuse the same variable name but to store the cloned array in a separate variable. 

Do you have plans to introduce some project-wide linter, btw?

sestinj: @my8bit any good recommendations? eslint is the obvious one. Seems tools like trunk.io can do everything at once, but haven't fully dug into what we should be using

sestinj:
appreciate this, glad it caught a bug. Want to manually test a bit myself before merging but will try to do this tomorrow

PR #805, opened 2024-02-01
update inline code default font

yuuuxt:


yuuuxt:
@sestinj seems this PR seems got reverted (by mistake?)

sestinj:
Yeah, thanks for catching that. Just got lost while there was lots of movement trying out Shiki -> Monaco -> highlight.js

Back to the way it should be in preview branch now

PR #809, opened 2024-02-02
case insensitive model name when selecting tab autocomplete templete

benedict-lee:
If the model name contains uppercase letters, the model template is not applied correctly. 
Therefore, when selecting a tab autocomplete template, change the model name to lowercase.


sestinj:
nice! thank you

PR #823, opened 2024-02-05
check for edge cases

restlessronin:
The library now checks for cases such as where a source file has no definitions or references (according to the tags query) so there's nothing to rank.

I'm not entirely sure what's the right context to return. For now I've gone with an empty string as the content value.


sestinj:
Empty string is the right thing to do, that way nothing gets displayed to the LLM

Issue #803, opened 2024-02-01
Export functions to add/remove custom models to other VS Code extensions


lupreCSC:
### Validations

- [X] I believe this is a way to improve. ~I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions~
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

We are trying to use `continue` with some locally hosted models were happy to see that there is way to provide a custom model querying code via the `config.ts`. Sadly, this is somewhat limited because it does not allow access to the `vscode` API object, which we would to interact with VSCode's authentication providers. It gets even more difficult if we want to use some features provided by another extension inside the `config.ts` code. A more flexible programmatic way to add models to the `continue` extension would be appreciated.

### Solution

VS Code allows extensions [to export an API object with functions for other extensions](https://code.visualstudio.com/api/references/vscode-api#extensions) to invoke from the `activate` function. It would be great if `continue` could export a function to add a custom model.

Would that be a feature you would be willing to consider?

(Unfortunately unable to join discord).

sestinj:
@lupreCSC thanks for making the detailed PR! I apologize for taking so long to get back with you.

I feel this adds a good amount of complexity, so I'd want to understand the reasons really well. I think I understand the need to use authentication for custom LLMs at a high level, but what is the exact mechanism you'll want to use for this? And when you say "we want to use some features provided by another extension inside the config.ts code", what sorts of features are you interested in using?

Depending on what you're looking for, I think that a possible simpler option would to be to pass the `vscode` object to the `modifyConfig` function

lupreCSC:
Hi @sestinj and please excuse the delay in my response.

Our setting is that we are running a model that is not using the OpenAI API and requires OAuth / OpenID Connect based authentication from our authorization servers. We have now set up a VSCode extension that implements model API requests as well as the authentication and are looking for a way to register that into the model list of the Continue extension.

The following are the problems with `modifyConfig` as we currently see them:

- it forces suppliers of additional models to "outsource" the model implementation to the user in the sense that the set up instruction would always be something like "please copy the following into the `config.ts` in this particular location" - and will likely become especially complex for users that would want to use multiple custom models; by contrast, the ability to register models programmatically would be as simple as installing another extension for the user
[...]

PR #827, opened 2024-02-06
First draft for an API to add models from other VSCode extensions.

lupreCSC:
Proof of concept draft for feature request https://github.com/continuedev/continue/issues/803

Exports functions "addExtensionModel" and "removeExtensionModel" through the VSCode extension API.

`addExtensionModel` accepts a `CustomLLM` object, as well as callbacks to the third-party extension to be notified when the model is added and removed.

Added field `extensionModels` to `ContinueConfig`, which holds the `CustomLLM` model specifications for models added by extensions. This field is not serialized to `config.json`. Config loading routines are adadpted to construct `CustomLLMClass` instances from all models declared in `extensionModels` and add them to `models`.

Also reworked `vscode/src/loadConfig.ts:ConfigHandler` to keep track of extension models on the VSCode extension side and improve configuration update event handling generally.

There seems to have been a bug/leak in earlier code where models removed by the user where removed in the webview / continue server configuration but not in the configuration object held by the VSCode extension: `debugPanels` handler for `deleteModel` would not remove the model from the config object and not trigger a config reload. This should be resolved by the changes.

Not sure if this would be the best way to handle this - it feels a bit hacky and there might be a better way to keep track of the extension models. The whole interaction of extension code with continue server(?)/webview code appears quite complex and I have missed a better way. Please consider this a proof of concept draft - any feedback welcome.


sestinj:
@lupreCSC just wanted to make sure I also referenced other discussion in this PR: https://github.com/continuedev/continue/issues/803#issuecomment-1963329296

Questions there would be good to understand before moving forward.

PR #831, opened 2024-02-07
Resolve Dependency Installation Issue with Yarn on OSX(ARM) for VSC

rootedbox:
Resolve Dependency Installation Issue with Yarn on OSX(ARM) for VSC.. Not sure if this effects other systems.


simonbukin:
I was having the same issue (also on an M1 Macbook), and this fixed it. Thanks!

PR #835, opened 2024-02-08
fix: Promise arrays being returned. Added a new definitions-only context provider. fix: typescript tagging.

restlessronin:
There was still an issue with the provider returning an array of Promise objects. This should now be fixed.

There was another issue with typescript files not returning definition tags. This should now also be fixed.

I added a new provider.

The first provider, a tag-ranking highlighter tries to restrict the highlighted symbols to the "most important" nth percentile. I'm still playing around trying to figure out the scenarios in which this is most useful, and what "most important" means.

In the meantime, this new provider shows only the definition line with the name of classes, functions, methods, etc. It very predictably "summarizes" a file. This should be useful in many places where the entire file is too big to send. So, for example

```python
def greet(name):
  return f"Hello, {name}!"

def calculate_area(length, width):
  return length * width

class Point:
  def __init__(self, x, y):
      self.x = x
      self.y = y

  def distance_to(self, other):
      dx = self.x - other.x
      dy = self.y - other.y
      return (dx**2 + dy**2)**0.5

# Example usage
print(greet("Alice"))
print(calculate_area(5, 3))
point1 = Point(2, 3)
point2 = Point(4, 5)
print(point1.distance_to(point2))
```
is 'outlined' to 
```
test_python_code.py
█def greet(name):
⋮...
█def calculate_area(length, width):
⋮...
█class Point:
█  def __init__(self, x, y):
⋮...
█  def distance_to(self, other):
⋮...
```

### DIFF
--- a/core/context/providers/CodeHighlightsContextProvider.ts
+++ b/core/context/providers/CodeHighlightsContextProvider.ts
@@ -6,7 +6,7 @@ import {
 } from "../..";
 import { getBasename } from "../../util";
 
-const highlighterPromise = import("llm-code-highlighter/dist/index.continue");
+import { getSourceSetHighlights } from "llm-code-highlighter/dist/index.continue";
 
 class CodeHighlightsContextProvider extends BaseContextProvider {
   static description: ContextProviderDescription = {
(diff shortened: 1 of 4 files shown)


sestinj:
Nice! Can't wait to try it

PR #836, opened 2024-02-09
Enhancement/send keyboard shortcuts to gui page

rootedbox:
Currently if someone tries to use a continue keyboard shortcut and they are on the help screen nothing will happen.  This fix intercepts the message switches to the GUI screen and bubbles up the message.

### DIFF
--- a/gui/src/hooks/useNavigationListener.tsx
+++ b/gui/src/hooks/useNavigationListener.tsx
@@ -1,0 +1,42 @@
+import { useEffect } from 'react';
+import { useNavigate, useLocation } from 'react-router-dom';
+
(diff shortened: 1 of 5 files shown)


sestinj:
@rootedbox This is really good, but would leave less room for future error if we are explicit about which messages types are being listened for, instead of defining by negation.

I think the types we want are all found via cmd+F `debugPanelWebview?.` in this file: https://github.com/continuedev/continue/blob/preview/extensions/vscode/src/commands.ts

sestinj:
Looks awesome and works great! Love the packaging in a clean react hook

PR #840, opened 2024-02-10
Use TypeScript Declaration Merging to get rid of some `any`s

nperez0111:
This adds some properties that were being accessed on the window object, onto the window object

I'm not completely certain that I got all of the types 100% correct, but I pulled them from the two places that I see them being set:
 - https://github.com/continuedev/continue/blob/1e75a7e8ff9b438c36c9d783d8c849ec12815d94/extensions/vscode/src/debugPanel.ts#L722-L733
 - https://github.com/continuedev/continue/blob/1e75a7e8ff9b438c36c9d783d8c849ec12815d94/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt#L124-L128

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt
@@ -123,7 +123,7 @@ class ContinuePluginToolWindowFactory : ToolWindowFactory, DumbAware {
                             val jsonData = mutableMapOf(
                                     "type" to "onLoad",
                                     "windowId" to continuePluginService.windowId,
-                                    "workspacePaths" to continuePluginService.worksapcePaths,
+                                    "workspacePaths" to continuePluginService.workspacePaths,
                                     "vscMachineId" to getMachineUniqueID(),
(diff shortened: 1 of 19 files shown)


extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt
```diff
@@ -123,7 +123,7 @@ class ContinuePluginToolWindowFactory : ToolWindowFactory, DumbAware {
                             val jsonData = mutableMapOf(
                                     "type" to "onLoad",
                                     "windowId" to continuePluginService.windowId,
-                                    "workspacePaths" to continuePluginService.worksapcePaths,
+                                    "workspacePaths" to continuePluginService.workspacePaths,
```
nperez0111, line 126: Just caught a typo here and figured it was worth addressing

sestinj:
This is fabulous, and the fact that a typo was caught too...this is exactly why we need strong typing : )

I think it's ready for merge if you can confirm?

nperez0111:
Yes, this is ready for merging @sestinj!

sestinj:
awesome, good work!

PR #841, opened 2024-02-11
Fully typed redux store

nperez0111:
This makes the GUI's redux store be fully typed by:
 - Following Redux Tookits' recommendation of deriving [the root state type from the root reducer](https://redux-toolkit.js.org/usage/usage-with-typescript#getting-the-state-type)
 - Adding `PayloadAction` type to slice payloads (thereby typing their usage)
 - Simplifying some of the reducers in the redux store to be more idiomatic (i.e. using mutation over state spreading)

### DIFF
--- a/gui/src/redux/slices/stateSlice.ts
+++ b/gui/src/redux/slices/stateSlice.ts
@@ -1,7 +1,8 @@
-import { createSlice } from "@reduxjs/toolkit";
+import { PayloadAction, createSlice } from "@reduxjs/toolkit";
 import { JSONContent } from "@tiptap/react";
 import {
   ChatHistory,
+  ChatHistoryItem,
   ChatMessage,
   ContextItemId,
   ContextItemWithId,
@@ -9,7 +10,6 @@ import {
 } from "core";
 import { BrowserSerializedContinueConfig } from "core/config/load";
 import { v4 } from "uuid";
-import { RootStore } from "../store";
 
 const TEST_CONTEXT_ITEMS: ContextItemWithId[] = [
   {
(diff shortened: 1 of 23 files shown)


gui/src/redux/slices/stateSlice.ts
```diff
@@ -281,91 +261,66 @@ export const stateSlice = createSlice({
...
+      state.active = false;
     },
-    streamUpdate: (state, action) => {
-      if (state.history.length > 0) {
+    streamUpdate: (state, action: PayloadAction<string>) => {
+      if (state.history.length) {
         state.history[state.history.length - 1].message.content +=
           action.payload;
       }
     },
```
nperez0111, line 283: This one is a bit strange and has a type error at it's call-site. The type here [is MessageContent](https://github.com/nperez0111/continue/blob/8b11e1009497a50973464b2c0c4353d7fa59bc00/core/index.d.ts#L181) but, it seems that this code only ever assumes streamUpdate to be called with a string. So this change exposed a type incompatibility that I wasn't sure how you guys wanted to resolve. I wrote up how I'd resolve this in my suggestion. But I don't know how you guys were planning on outputting image urls (the text case is straightforward of course.


```suggestion
    streamUpdate: (
      state,
      action: PayloadAction<ChatHistoryItem["message"]["content"]>
    ) => {
      if (state.history.length) {
        const historyItem = state.history[state.history.length - 1];
        if (typeof action.payload === "string") {
          historyItem.message.content += action.payload;
        } else {
          historyItem.message.content += action.payload
            .map((message) => {
              if (message.type === "text") {
                return message.text;
              }
              if (message.type === "imageUrl") {
                return message.imageUrl.url;
              }
              return "";
            })
            .join();
        }
      }
    },
```

sestinj:
Hi @nperez0111, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

nperez0111:

PR #849, opened 2024-02-13
Fit Token Budget

restlessronin:
Added a binary search over percentile of ranked tags to find the maximum tags that fit in the token budget.


sestinj:
very nice

PR #851, opened 2024-02-13
add openai embeddings

bjornjorgensen:
https://platform.openai.com/docs/guides/embeddings/embedding-models

image


bjornjorgensen:
oh.. should I commit this to continuedev:preview insted of main?

PR #861, opened 2024-02-15
add new openapi embeddings

bjornjorgensen:
https://platform.openai.com/docs/guides/embeddings/embedding-models

image


bjornjorgensen:
it might be that we need to update lancedb to have support for those latest models.

https://github.com/lancedb/lancedb/pull/912 

https://github.com/lancedb/lancedb/pull/743

PR #870, opened 2024-02-17
Fix search context provider

Elpaggio:
On Windows VSCode the search context provider implemented during commit 'bb94da0c' seems broke.
You will receive an error no matter what query you enter.

I receive this error message:
image
This is fixed by:
_Set itemType for querySubmenuItem
This is necessary before calling props.command the same way we do it for non-query item types. Otherwise we receive an 'Unknow provider...' error message in the 'getContextItems' case in the 'panel.webview.onDidReceiveMessage, callback._

Fixing the above mentioned error a new error occurs:
image

This is fixed by:
_include @vscode node module folder
we need the @vscode folder containing ripgrep in the distribution._

_Amended path for ripgrep executable
We have to preprend the 'out' folder in order to find the ripgrep executable in the distribution.
Prepending 'out' during development is not necessary. We need a way to distinguish in 'ideProtocol'?_

As mentioned we need a way to distinguish between development and production mode in order to know when to prefix the path to 'ripgrep' with the 'out' folder. Any suggestions?


sestinj:
@Elpaggio thanks for looking into this. The solution looks good. The best way to check whether you are in development mode is to check `context.extensionMode === vscode.ExtensionMode.Development`. `context` comes from the `activate` function in `activate.ts` and can be passed down to wherever it is needed. This is a bit easier if you upgrade to newer versions of the preview branch

PR #874, opened 2024-02-18
Fix markComplete removal

Elpaggio:
Deleted and previously index files were not removed from the tag_catalog database table.

To reproduce just add a file (not ignored by .gitignore or .continueignore) let continue index it and remove it again.

MarkComplete is called for every added, removed or changed item found. Therefore it seems odd (ignoring the performance perspective) that we try to remove all items at once. The other cases also only handle one single item at the time. Nevertheless the "remove" case did not behave as expected and did not remove anything from the tag_catalog.

With the code change we now handle one single item at the time and the item is removed from the tag_catalog as expected.


sestinj:
@Elpaggio thank you once again. This was a really good find

Issue #879, opened 2024-02-19
Cursor Not Always Shown


rootedbox:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:osx
- Continue: v0.8.12 
- IDE: VSCode and Jetbrains
```

### Description

So this is very random.. it will happen both when I'm running continue stand alone or if I make a build and I'm doing debugging. How ever it does tend to happen more when I'm debugging the cursor will not exist and then it won't exist for the whole session.

sestinj:
@rootedbox I've seen this in JetBrains, is it happening in VS Code for you as well?

In JetBrains, I can always get the cursor to come back by clicking the editor and then coming back to the sidebar. Curious if this matches what you've seen

rootedbox:
> @rootedbox I've seen this in JetBrains, is it happening in VS Code for you as well?
> 
> In JetBrains, I can always get the cursor to come back by clicking the editor and then coming back to the sidebar. Curious if this matches what you've seen

Yes that seems about right.. when I first started using continue was on jet brains and I saw the bug the first few times I ran it.. but yes I've switched to VS code lately and when I compile to debug the plugin I see this issue pop up I would say half the time.. that being said I should be ale to fix it if it's 50% of the time.

PR #881, opened 2024-02-20
🐛  Bugfix cursor sometimes not shown initially.

rootedbox:
There seems to be a bug in tip-tap that the cursor doesn't come up initially and waits for a blur event meaning a user won't have a cursor unless some other screen gets focus. this forces the blur on initialization causing the no curser issue to go away


rootedbox:
fixes https://github.com/continuedev/continue/issues/879

sestinj:
@rootedbox I trust this works so going to just merge and then test myself. Thanks for tracking this down, we've been seeing it for quite a while in JetBrains, will make many users happy

PR #955, opened 2024-03-12
Added Locals context provider

commandblock2:
image

Type `@locals` to reference the contents of the local variables with top n level (defaulting to 3) of call stack for that thread. A dropdown will appear, allowing you to select a specific thread to see the local variables in that thread.
```json
{
  "name": "locals",
  "params": {
    "stackDepth": 3
  }
}
```


commandblock2:
Not sure if the prettier config has changed, it used to keep the `,` but not sure why it has changed

sestinj:
@commandblock2 awesome work here!

PR #957, opened 2024-03-13
Refactored web crawling code for document indexing. The updated code …

rootedbox:
…uses a breadth-first search, filters out specific paths, and ensures that websites are not downloaded twice.


sestinj:
@rootedbox everything looked good! Only small change I made was related to the filtering of links based on subpath. I was finding a ton of sites that had most pages filtered out, so made the requirement just to match the hostname instead of also being prefixed by the base URL path

PR #959, opened 2024-03-14
bug fix: Changing the subMenuTitle in the context provider has led to an inablity to add new documentation sources.

rootedbox:
bug fix: Changing the subMenuTitle in the context provider has led to an inablity to add new documentation sources.


sestinj:
Good find!

PR #971, opened 2024-03-15
Fix chat msg formatting in Output for multimodal models

pzaback:
For models that support images, all chat messages showed up as [object Object] in the Output, rather than message text. The completion text for the model response is unaffected. 
```<user>
[object Object]

<assistant>
[object Object]

<user>
[object Object]```

This fix ensures that the text portion shows up correctly in the output. Though it might be nice to have the output include some token to represent any image pieces of a chat dialog, for now they're just stripped from the formatted message just as if the user included an image for a model that doesn't support them.

### DIFF
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -29,6 +29,7 @@ import {
   compileChatMessages,
   countTokens,
   pruneRawPromptFromTop,
+  stripImages,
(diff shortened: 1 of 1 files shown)


sestinj:
@pzaback good fix, important detail there making the copy!

PR #975, opened 2024-03-17
Bump vectordb version

germankrause:
I had some problems with the `@Docs` feature, couldn't add docs nor use preindexed. It was giving an `ERR Table 'docs' was not found: Error: Table 'docs' was not found`. Seems like this was caused by a bug in lance, so I just bumped `vectordb` version. Please add this to the nightly release.


sestinj:
@germankrause good find, thank you!

PR #988, opened 2024-03-19
Add SystemContextProvider to context providers so that users can use …

rootedbox:
# PR Title: Feature: System Context Provider

This PR introduces a new context provider, the `SystemContextProvider`. This provider is designed to reference the operating system and CPU information.

## Impact

With the addition of the `SystemContextProvider`, users can now reference their operating system and CPU information. This will be particularly useful in scenarios where system-specific information is necessary.

### DIFF
--- a/core/context/providers/OSContextProvider.ts
+++ b/core/context/providers/OSContextProvider.ts
@@ -11,8 +11,8 @@ import os from "os";
 
 class OSContextProvider extends BaseContextProvider {
   static description: ContextProviderDescription = {
-    title: "system",
-    displayTitle: "System",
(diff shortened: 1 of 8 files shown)


pcuci:
I totally misread this to mean "system message" context provider, as in "let's change the system message for the whole thread, and discard any prior system message updates"

What do you think about renaming this to `/os` instead?

sestinj:
@rootedbox looks nice! I agree with Paul about naming this "os" to a) avoid slight confusion and b) make it faster to type. Other than that good to merge

rootedbox:
image

PR #1002, opened 2024-03-20
Update message to match revised shortcut

pzaback:
If user attempts to use /edit without any context selected they get a message prompting them to do so. Selection shortcut was recently changed from ctrl/cmd-M to L; this revises the message to match


sestinj:
thank you @pzaback !

PR #1006, opened 2024-03-21
Config: add ui.codeBlockToolbarPosition

germankrause:
- Introduce UI configuration to `~/.continue/config.json`
- Added option to move down the code block toolbar

ui-config

### DIFF
--- a/core/config/default.ts
+++ b/core/config/default.ts
@@ -74,6 +74,9 @@ export const defaultConfig: SerializedContinueConfig = {
     provider: "ollama",
     model: "starcoder-3b",
   },
+  ui: {
+    codeBlockToolbarPosition: "top",
+  }
 };
 
 export const defaultConfigJetBrains: SerializedContinueConfig = {
@@ -137,4 +140,7 @@ export const defaultConfigJetBrains: SerializedContinueConfig = {
       params: {},
     },
   ],
+  ui: {
+    codeBlockToolbarPosition: "top",
+  }
 };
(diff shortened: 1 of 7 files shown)


core/config/default.ts
```diff
@@ -74,6 +74,9 @@ export const defaultConfig: SerializedContinueConfig = {
     provider: "ollama",
     model: "starcoder-3b",
   },
+  ui: {
```
sestinj: Can we leave this out of the default config? I think always better to keep things simple for beginners. Would also mean removing from the slice default value

gui/src/redux/slices/stateSlice.ts
```diff
@@ -117,6 +117,9 @@ const initialState: State = {
       },
     ],
     contextProviders: [],
+    ui: {
```
sestinj: Remove here as mentioned above

sestinj (changes requested):
To be honest when I first saw your comment about moving the buttons to the bottom I was quite unsure...but I really like the design choice you made to create a single object in config! I can see many other possibilities like font size, whether to display certain buttons, etc... Excited to see the community experiment with this!

After removing from default config I'm ready to merge this!

core/config/default.ts
german-swan: Of course, I wasn't sure where else it should be defined besides the types

gui/src/redux/slices/stateSlice.ts
german-swan: Done, also fixed a debug statement

sestinj (approved):

PR #1007, opened 2024-03-21
Fix/autocomplete errors

### LINKED ISSUE #1004: [CON-219] Autocomplete errors are spammny

- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:*
- Continue:*
- IDE: VS Code
```

### Description

If there is an error with tab-autocomplete, it tends to show up on every keystroke, which is annoying. Ideally, Continue will show any given error the first time, and then not display it a second time.

The error in VS Code is displayed [here](https://github.com/continuedev/continue/blob/d431c60c4dd6ca5e7b67b5ffe30cbc2a20b1b60e/extensions/vscode/src/autocomplete/completionProvider.ts#L18). It's possible that errors are escaping to [here](https://github.com/continuedev/continue/blob/d431c60c4dd6ca5e7b67b5ffe30cbc2a20b1b60e/extensions/vscode/src/autocomplete/completionProvider.ts#L113), but the intended usage is that `onError` gets passed to the CompletionProvider [here](https://github.com/continuedev/continue/blob/d431c60c4dd6ca5e7b67b5ffe30cbc2a20b1b60e/extensions/vscode/src/autocomplete/completionProvider.ts#L41) and then the CompletionProvider handles de-duplicating error messages [here](https://github.com/continuedev/continue/blob/d431c60c4dd6ca5e7b67b5ffe30cbc2a20b1b60e/core/autocomplete/completionProvider.ts#L304-L310). In the last couple of days a few instances of errors on every keystroke have shown up, so there must just be a small mistake in how this is setup.

This should be easy to test by purposefully throwing errors in different spots when debugging and making sure they are properly handled.

### To reproduce

_No response_

### Log output

_No response_

<sub>[CON-219](https://linear.app/continue/issue/CON-219/autocomplete-errors-are-spammny)</sub>

Iamshankhadeep:
fixes #1004

### DIFF
--- a/extensions/vscode/src/autocomplete/completionProvider.ts
+++ b/extensions/vscode/src/autocomplete/completionProvider.ts
@@ -20,8 +20,8 @@ export class ContinueCompletionProvider
       if (val === "Documentation") {
         vscode.env.openExternal(
           vscode.Uri.parse(
-            "https://continue.dev/docs/walkthroughs/tab-autocomplete",
-          ),
+            "https://continue.dev/docs/walkthroughs/tab-autocomplete"
+          )
         );
       }
     });
@@ -32,14 +32,14 @@ export class ContinueCompletionProvider
   constructor(
     private readonly configHandler: ConfigHandler,
     private readonly ide: IDE,
-    private readonly tabAutocompleteModel: TabAutocompleteModel,
+    private readonly tabAutocompleteModel: TabAutocompleteModel
   ) {
     this.completionProvider = new CompletionProvider(
       this.configHandler,
       this.ide,
       this.tabAutocompleteModel.get.bind(this.tabAutocompleteModel),
       this.onError.bind(this),
-      getDefinitionsFromLsp,
+      getDefinitionsFromLsp
     );
   }
 
@@ -47,7 +47,7 @@ export class ContinueCompletionProvider
     document: vscode.TextDocument,
     position: vscode.Position,
     context: vscode.InlineCompletionContext,
-    token: vscode.CancellationToken,
+    token: vscode.CancellationToken
     //@ts-ignore
   ): ProviderResult<InlineCompletionItem[] | InlineCompletionList> {
     const enableTabAutocomplete =
@@ -75,7 +75,7 @@ export class ContinueCompletionProvider
       const outcome =
         await this.completionProvider.provideInlineCompletionItems(
           input,
-          signal,
+          signal
         );
 
       if (!outcome || !outcome.completion) {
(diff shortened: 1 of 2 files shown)


extensions/vscode/src/autocomplete/completionProvider.ts
```diff
@@ -100,17 +100,15 @@ export class ContinueCompletionProvider
...
           ),
           {
             title: "Log Autocomplete Outcome",
             command: "continue.logAutocompleteOutcome",
             arguments: [outcome, logRejectionTimeout],
-          },
+          }
         ),
       ];
-    } catch (e: any) {
```
rootedbox, line 112: maybe instead of remove.. post and store the error.. if the next error is the same just ignore.. if it's a new error then post the error

Iamshankhadeep:
``` javascript
Telemetry.capture("autocomplete", {
          accepted: outcome.accepted,
          modelName: outcome.modelName,
          modelProvider: outcome.modelProvider,
          time: outcome.time,
          cacheHit: outcome.cacheHit,
});
```
there is 2 instance where the error can occur, one is the upper one and other one is this
``` javascript
new vscode.InlineCompletionItem(
          outcome.completion,
          new vscode.Range(
            position,
            position.translate(0, outcome.completion.length)
          ),
          {
            title: "Log Autocomplete Outcome",
            command: "continue.logAutocompleteOutcome",
            arguments: [outcome, logRejectionTimeout],
          }
        ),
```        
one thing to note here is do we even care to store this and show it to the user ?  cause posthog capture can fail but showing it too the user seems a bit much and with `vscode.InlineCompletionItem` these too. so I thought of removing it.

sestinj: I'm actually okay with this. I checked the behavior and VS Code will end up logging the error in the console if something happens in this small area. If something goes wrong here, completions will probably never show up and we'll find out from every user all at once / it will be obvious in testing. In that case I might rather it not spam, and I don't necessarily want to duplicate the logic for showing an error only once (just adds a bit of complexity)

sestinj (approved):

sestinj:
Nice work and thanks @rootedbox for looking out!

One final thing before I merge: @Iamshankhadeep would be be able to point this PR to the preview branch instead? Here and in the future this is best, so that we have the chance to test the change in pre-release

PR #1024, opened 2024-03-26
Bedrock chat completion, Anthropic models

tfire:
Didn't quite get _streamComplete functioning yet. This does mean the model has to finish its request before output comes back, which is not ideal. But moreover, Claude3 is not available with the streaming API, at least not yet.

### DIFF
--- a/extensions/intellij/src/main/resources/config_schema.json
+++ b/extensions/intellij/src/main/resources/config_schema.json
@@ -128,6 +128,7 @@
             "openai",
             "free-trial",
             "anthropic",
+            "bedrock",
             "together",
             "ollama",
             "huggingface-tgi",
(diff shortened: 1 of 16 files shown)


extensions/intellij/src/main/resources/config_schema.json
```diff
@@ -529,6 +531,34 @@
...
+        {
+          "if": {
+            "properties": {
+              "provider": {
+                "enum": ["bedrock"]
+              }
+            },
+            "required": ["provider"]
+          },
+          "then": {
```
sestinj, line 543: Could you also add the "region" property as required here, as an enum of "us-east-1", etc...? I think it will be easy for users to forget that that's necessary and could go a long way by giving intellisense

sestinj (commented):
@tfire Nice work! Really glad this was able to work out so cleanly with fetch. Hopefully they release streaming for Claude soon, but in the meantime this is the perfect first step. With the one edit mentioned below I think this is perfect

extensions/intellij/src/main/resources/config_schema.json
tylerfxn: Pushed, the commits got a little jumbled but it's nothing

PR #1052, opened 2024-03-31
Bug Fix: Add ternary operator to prevent nonexistant value error

justinmilner1:
Screenshot from 2024-03-31 16-17-29

[This commit](https://github.com/continuedev/continue/commit/4053f1ad60a61a32bbd3a5543e373f4be4e36508) created a bug where a nonexistant value (next.value.content) is attempted to be read. 

I'm guessing it was added for cases where next.value.content will exist, so adding a ternary operator enables either case.


sestinj:
Looks great, thank you! I made only a slight change so content is allowed to be undefined

PR #1056, opened 2024-04-01
✨Add copy to workspace button for history items

pzaback:
I want to be able to keep a log of my LLM interactions for a given project within the repo for that project and this feels like one decent way to do that (I've also started work on a related feature to enable a config setting that enables automatically saving a copy of a session to the workspace in which it's created). I haven't gotten my dev environment for JetBrains builds up and running yet so I'd appreciate assistance testing that the way I've implemented getting the workspace directory works for that and in general would appreciate any suggestions on how this could be implemented better.

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -22,35 +22,35 @@ declare global {
       ) => void;
     }
   }
-  
+
   export interface ChunkWithoutID {
     content: string;
     startLine: number;
     endLine: number;
     otherMetadata?: { [key: string]: any };
   }
-  
+
   export interface Chunk extends ChunkWithoutID {
     digest: string;
     filepath: string;
     index: number; // Index of the chunk in the document at filepath
   }
-  
+
   export interface IndexingProgressUpdate {
     progress: number;
     desc: string;
   }
-  
+
   export interface LLMReturnValue {
     prompt: string;
     completion: string;
   }
   export interface ILLM extends LLMOptions {
     get providerName(): ModelProvider;
-  
+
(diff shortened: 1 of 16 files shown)


sestinj:
@pzaback Is there any way that we might be able to achieve what you are looking for without adding a directory to the workspace? It would help to know what sort of outcome you'll want from the recorded sessions. Perhaps better filtering capabilities for the `~/.continue/sessions` folder would accomplish the same without adding complexity?

sestinj:
Makes sense! 

Adding a quick comment to follow up on the good convo from OH today: I think starting by adding an "outputDir" option to the "params" of the ShareSlashCommand would be a really good starting solution for this

And if deciding to go further, then refactoring the HistoryManager to not be a global variable might create a quick, minimal code changes way of achieving the initial idea, though possibly this adds some complexity

pzaback:
This may well be another one of those things where you'd have immediately had an idea about the block I ran into or at least more quickly known where to look for the bug, but it was good for me to learn by working through what was causing the issue. I'm about to file a separate PR for this but just wanted to quickly mention here that there was (I believe) a bug with params not being correctly set for slash commands, such that they were undefined within the run function (e.g., `params?.tokenLimit,` in EditSlashCommand was undefined, as were all other such instances). I'll also file a separate PR for the enhancements to /share rather than modify this one.

pzaback:
Implemented similar functionality via modifications to `/share` slash command in https://github.com/continuedev/continue/pull/1128 as discussed above.

PR #1059, opened 2024-04-02
Add abstraction for fetch to easily allow using request options

tobiajung:
The `ModelDescription` property used for the configuration of the llm providers in the `config.json` file uses the `RequestOptions` object which allows to use custom proxy and ssl configurations for llm requests.
Context providers like the Jira context provider do not have such an option. On some cases it might be desired to use different configurations on the components of Continue. For example, one might use a proxy for the Jira context provider while no proxy should be used for the llm. Moreover, there might occur some problems when using the continue plugin on a Jetbrains IDE with SSL requests. Therefore, it should be possible to deactivate the certificate verification or to add one.
This pull requests adds an abstraction for the node-fetch `fetch` function which accepts the `RequestOptions` object declared for the configuration of LLMs. This can be easily implemented for other requests as it is done for the Jira context providers in this pull reqeust.

### DIFF
--- a/core/config/handler.ts
+++ b/core/config/handler.ts
@@ -1,8 +1,3 @@
-import { http, https } from "follow-redirects";
-import * as fs from "fs";
-import { HttpProxyAgent } from "http-proxy-agent";
-import { HttpsProxyAgent } from "https-proxy-agent";
-import fetch from "node-fetch";
 import { ContinueConfig, ContinueRcJson, IDE, ILLM } from "..";
 import { IdeSettings } from "../protocol";
 import { Telemetry } from "../util/posthog";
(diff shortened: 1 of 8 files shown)


sestinj (approved):

sestinj:
@tobiajung This looks good. Have you had the chance to test it with both v2 and v3?

tobiajung:
I only got the chance to test it with v2. Unfortunately, we do not have access to v3.

sestinj:
All good! The changes are pretty straightforward and look correct, so I'm going to merge.

But also tagging @benedict-lee here—I believe you had made the original PR for v3, and just want to notify that this has changed. It will remain in pre-release until we have confidence that things are working.

sestinj:
Hi @tobiajung, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

tobiajung:

Issue #1060, opened 2024-04-03
"Unable to read from the browser's clipboard" when using code-server with Firefox


ading2210:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Debian 11 (server), Debian 12 (client)
- Continue: v0.8.22
- IDE: code-server 1.87.2
- Browser: Firefox 115.9.1esr
```

### Description

When I use Continue with code-server on the Firefox browser, the following error message appears as a notification when I type:

> Unable to read from the browser's clipboard. Please make sure you have granted access for this website to read from the clipboard.

I've tried enabling `dom.events.asyncClipboard.clipboardItem` and `dom.events.asyncClipboard.readText` in `about:config` like suggested in [other forums](https://github.com/microsoft/vscode/issues/164769), but this only results in a "paste" button appearing under the cursor which blocks any typing.

The extension works perfectly on Chromium under the same conditions.

https://github.com/continuedev/continue/assets/71154407/2b44dd17-069b-469b-a814-ce0878431d2f

### To reproduce

1. Install [code-server](https://github.com/coder/code-server) and install the Continue extension
2. Open code-server in Firefox
3. Set up a provider and model
4. Type anywhere in a text editor

### Log output

```
Unable to read from the browser's clipboard. Please make sure you have granted access for this website to read from the clipboard. 2 notificationsAlerts.ts:42:13
    c notificationsAlerts.ts:42
    b notificationsAlerts.ts:28
    y event.ts:1127
    z event.ts:1138
    fire event.ts:1162
    addNotification notifications.ts:228
    notify notificationService.ts:253
    prompt notificationService.ts:331
    readText clipboardService.ts:46
    readText clipboardService.ts:42
```

ading2210:
Looking into the extension's code, it seems like it tries to read the user's clipboard on every keystroke to feed the autocomplete system? This seems like a poor design choice and a possible privacy issue.

ading2210:
Actually, it looks like the clipboard text is never even used? In that case it should be safe to completely remove the code that reads the clipboard.

[...]

PR #1061, opened 2024-04-03
Don't read the user's clipboard for autocomplete

ading2210:
Currently, the Continue VSCode extension reads the user's clipboard on every keystroke to feed the autocomplete system, but it doesn't do anything with the clipboard text. On some browsers, trying to get the clipboard can be problematic, so I believe the best solution would be to simply remove the code which reads the clipboard.


ading2210:
This would fix #1060.

ading2210:
Closed in favor of #1062.

Issue #1037, opened 2024-03-28
"Toggle Full Screen" will cause the Continue panel to malfunction.


danny-su:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 14.4.1
- Continue: v0.9.96
- IDE: VS Code 1.87.2
```

### Description

"Toggle Full Screen" will cause the Continue panel to malfunction.

### To reproduce

Step 1:
<img width="505" alt="image" src="https://github.com/continuedev/continue/assets/12178855/4aedf827-48f6-4862-82b7-9abf673d96b7">
Step 2:
<img width="444" alt="image" src="https://github.com/continuedev/continue/assets/12178855/7d2031d1-8ade-4be6-9a60-695e20ace1bd">
Step 3:
<img width="720" alt="image" src="https://github.com/continuedev/continue/assets/12178855/86f8f8cd-a172-4baa-b240-2c556577d72b">
**PS: The icon buttons will not work for the Continue panel but will work for the newly opened Continue window (opened by "Toggle Full Screen").**

### Log output

```Shell
Nothing.
```

justinmilner1:
I was able to replicate this - I'll start an investigation.

justinmilner1:
Update - I've been debugging this, it seems to be a bigger issue than just using the existing toggleFullScreen functionality with an onDidDispose listener.

RomneyDa:
@justinmilner1 thanks for digging into this!
Looks like addressed on https://github.com/continuedev/continue/pull/1065

PR #1065, opened 2024-04-03
Con-1037: Toggle full screen bug

justinmilner1:
Fix for #1037

When user toggles full screen on, the webviewProtocol.webview is set to the full screen's. 

When user toggles full screen off, this property was not reassigned to the sidebar's webview, so requests would hang.

I've added a listener for full screen panel closure and a function to reset the webview when this happens.

Note: 
If a user has full screen open, the sidebar chat is still unusable - I think enabling both to be functional concurrently would require a larger rework of decoupling webviewProtocol from specific webview instances. 
This might involve keeping a list of active webviews and some method of differentiating them.

### DIFF
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -373,13 +373,12 @@ const commandsMap: (
     const fullScreenTab = getFullScreenTab();
 
     // Check if the active editor is the Continue GUI View
-    if (fullScreenTab && fullScreenTab.isActive) {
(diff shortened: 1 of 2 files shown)


sestinj:
@justinmilner1 Very good work, it was definitely the right move to make the partial fix for now, and as we approach a larger cleaning to more rigorously use something that looks like "services", we can aim for the full fix.

sestinj (approved):

Issue #1036, opened 2024-03-28
The slash commands /http and /issue don't work.


danny-su:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 14.4.1
- Continue: v0.9.96
- IDE: VS Code 1.87.2
```

### Description

Config:
<img width="591" alt="image" src="https://github.com/continuedev/continue/assets/12178855/4b20e2ce-763f-4930-a7a4-95d4903f116f">

Input:
<img width="388" alt="image" src="https://github.com/continuedev/continue/assets/12178855/e3e47e20-65f9-441b-88e0-48387924fb8b">

Unexpected Result:
<img width="375" alt="image" src="https://github.com/continuedev/continue/assets/12178855/90f78d40-6eb8-400a-859d-75b32dfab49d">

<img width="521" alt="image" src="https://github.com/continuedev/continue/assets/12178855/8a7cdc67-acf2-489b-bd17-cebf63506ce9">

### To reproduce

As above.

### Log output

```Shell
Nothing.
```

Issue #1035, opened 2024-03-28
The /so slash command doesn't work.


danny-su:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 14.4.1
- Continue: v0.9.96
- IDE: VS Code 1.87.2
```

### Description

<img width="419" alt="image" src="https://github.com/continuedev/continue/assets/12178855/0521ef75-3e83-46ea-9f63-dc3a4397287d">

<img width="499" alt="image" src="https://github.com/continuedev/continue/assets/12178855/40953e0b-495a-4606-b4de-8368d2df13e8">

### To reproduce

Type /so + anything.

### Log output

```Shell
==========================================================================
==========================================================================
Settings:
contextLength: 4096
model: mistralai/Mixtral-8x7B-Instruct-v0.1
maxTokens: 1024

############################################

<user>
/so How to handle null values in Rust?

<assistant>
```

skicreature:
I'm experiencing the same issue
OS: Windows 11
continue: v0.8.29
IDE: VS code 1.89.1

RomneyDa:
Stack overflow context provider was removed
https://github.com/continuedev/continue/pull/2605

PR #1087, opened 2024-04-07
Complete draftIssue and add slash commands to default config

justinmilner1:
### Resolves:
-  https://github.com/continuedev/continue/issues/1036 

### Related bug: https://github.com/continuedev/continue/issues/1035
- /so is waiting upon re-initialization of external server (I didn't re-add it) 

### I added:
- interfacing with github create issue API
- default config setups for /issue, /http, and /commit (otherwise they don't show up in drop-down menu until user manually adds them to their config) 

### Evidence:
Screenshot from 2024-04-07 13-59-54
---
On gitlab:
Screenshot from 2024-04-07 14-00-12


pzaback:
It looks like this actually overlaps in a way with this PR I submitted to resolve the fact that referencing params wasn't working for any existing slash commands: https://github.com/continuedev/continue/pull/1084
I had also identified as you did here that this.params also works, but given that multiple examples omitted "this", I went searching for a root cause and it seemed like the issue was that params were mapped elsewhere as "options" instead. I don't know if one or the other is "more" correct, but just thought it worth mentioning in case there are some potential foot-guns with either that I'm not aware of.

sestinj:
@justinmilner1 Part of the goal with generating a link was to give users a chance to modify the issue before it originally gets published. When you create via API, I think it just immediately creates the issue, is this how that works? I think it would be nice to have a verification flow where users don't need to fear that the model outputs something scary without a chance to review. Can you think of any kind of in between solution? I don't know if there's an API for generating drafts, or if perhaps we just need to generate a more obvious link (possibly an h1 with a link rather than the small text at the bottom, or just better messaging, like "please click here to create the issue")

justinmilner1:
@pzaback Good catch! Just tested and pushed up your fix to avoid using 'this'.

@sestinj I definitely agree. Feels very wrong to publish the issue without manual editing or confirmation, even if the issue is heavily labeled a draft. ...Unfortunately, github doesn't have a draft option for issues currently. 

I think two changes should be made:

Screenshot from 2024-04-10 10-41-33

1) A 'publish' button should be added into the chat bar. Alternative could be a code lens in the generated issue.

2) The draft should be generated in a new editor, rather than in the sidebar. This way the user can edit the generated issue as they wish. 

These are fairly heavy changes but I think they're necessary for this feature to really be valuable.

Any thoughts?

sestinj:
@justinmilner1 it's not a direct API that they have, but you can create a URL that opens up a draft of an issue when clicked. This is what I was going for with the link at the bottom of the slash command (https://github.com/continuedev/continue/blob/98521a2120fc02be749ba0b87cb7039dcd357568/core/commands/slash/draftIssue.ts#L51-L53)

As an example: https://github.com/continuedev/continue/issues/new?title=Testing&body=HelloWorld

I think you're right though that it would be nice to have more prominent button. I wonder if there's a method of creating buttons in markdown. If this was possible it would be useful in many other slash commands as well. In `StyledMarkdownPreview.tsx` we use rehype/remark and they both have really great plugin ecosystems. It might be that there's an existing plugin solution we could use, and if not it could be possible to write something that would convert something like `<a href="https://github.com/continuedev/continue/issues/new?title=Testing&body=HelloWorld"><button>Draft Issue</button></a>` into an actual button

justinmilner1:
@sestinj Ah I think I understand where my confusion is coming from:

When I click on the link you provided, I am _not_ taken to a draft page - instead I'm directed to the template page. When I use the issue feature on the preview branch, it takes me to 404.
If I try to create this 'draft issue' on a repository that _I own_, I am indeed taken to a 'draft' page.

I added issue templates to my repository, and still am able to reach the draft page - so I think the ability to draft the issue is indeed about repository ownership.

Note: The api is able to create issues both on repos I own and don't own.


So what to do?:
While I wish their was a native gitlab issue draft feature, I do think the 'Publish Draft' button design is sufficient confirmation to create an actual issue, _if_ the issue is labeled 'draft' (in the title/body) _and_ a window is opened with the url for the generated draft - would you agree? 


Is it worth it?:
This is a lot of additions - however, I could see this _interactivity within chat through buttons_ being used for numerous other features. Most user-guided interaction with external APIs will need some form of confirmation/rejection.... if this is something to build we should probably have a longer discussion about this.

sestinj:
@justinmilner1 Oooh this explains everything, I hadn't realized that at all. A bit unfortunate, but nevertheless I think the larger point about user confirmation/input is important. I would say this is interesting enough that we should go ahead and try it. Probably the first attempt won't be the exact thing that lasts forever, but would love to experiment.

The first problem to solve is probably just how to best display a button from the markdown view, and once that is accomplished, rather than making some large refactor so that slash commands can emit buttons in a general way, it might be good to just do something that allows for this to be tested more immediately

justinmilner1:
@sestinj Got it. I'll experiment with how the button could be displayed.

justinmilner1:
@sestinj Looks like it's very easy to _render_ a button in markdown - but it's not straightforward how the button's functionality would be initialized, or how slash commands would listen for button activity, or how we would track issue generations to button activity.

My takeaway is that it's something which could be done - and that's good to know! ...but I think it's better to wait until we have more (quantity and importance) use cases for a change this large - what do you think?

sestinj:
@justinmilner1 I agree with the conclusion. Let's close this PR then, but link it for future reference. I've created an issue (https://github.com/continuedev/continue/issues/1264) that I believe describes where this work will go next, but please add any context I might have left out!

PR #1126, opened 2024-04-15
tsx files are typically typescript files

jaresty:
These files were not showing it by that code search. This way might be why.

## Description

## Checklist

- [ x] The base branch of this PR is `preview`, rather than `main`

### DIFF
--- a/core/util/treeSitter.ts
+++ b/core/util/treeSitter.ts
@@ -34,7 +34,7 @@ export const supportedLanguages: { [key: string]: string } = {
   ts: "typescript",
   mts: "typescript",
   cts: "typescript",
-  tsx: "tsx",
+  tsx: "typescript",
   vue: "vue",
   // The .wasm file being used is faulty, and yaml is split line-by-line anyway for the most part
   // yaml: "yaml",


sestinj:
@jaresty by code search do you mean the @ codebase context provider, cmd+enter, @ files, or something else? The mapping here is for tree-sitter wasm files, and there happens to be a specific one or .tsx files that is separate from typescript

jaresty:
I mean that when I do Code search in my project, I do see symbols but only symbols from files that end in ".ts", so my lsp is working. 

I noticed that in the file I linked, these things are true:

- ts is mapped to typescript
- js is mapped to javascript 
- jsx is mapped to javascript
- tsx is mapped to tsx 

Since one of these look inconsistent, I'm theorizing that it could be a mistake. I'm not sure if it's the whole picture.

jaresty:
I was referring to the "Code" context provider.

jaresty:
I'm not sure how to test. I was hoping that someone here could help me with that.

jaresty:
A method from that file is used in the context provider as well [here](https://github.com/continuedev/continue/blob/af64211954d8d2ad34b1c9c7fc2306f42f98f1e9/core/indexing/CodeSnippetsIndex.ts#L68)

rootedbox:
@jaresty  Just a heads up TSX is another valid grammar within tree-sitter separate from typescript..  https://github.com/tree-sitter/tree-sitter-typescript

jaresty:
Isn't it also true of jsx files that they are a valid grammar too?  Why is the mapping different there?

https://github.com/tree-sitter/tree-sitter-javascript

sestinj:
@jaresty true that .jsx should probably have it's own grammar, I think this is probably just because there wasn't a readily available jsx grammar found and jsx is slightly less common.

Here is a guide to getting set up and running from source: https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md#environment-setup

And it seems like [this function](https://github.com/continuedev/continue/blob/preview/core/indexing/CodeSnippetsIndex.ts#L64-L94) is probably where you'll want to set a conditional breakpoint to find out what's going on

sestinj:
@jaresty I'm going through older PRs so wanted to check in on this. Should I close an let it come up again later, or do you plan to look into this soon?

jaresty:
I'll probably look into this at some point, that I'm not using continue very much these days so I'm not sure when. I don't mind if you want to close this PR.

sestinj:
Sounds good. Would be curious to hear what workflow you're enjoying most these days, but a GitHub PR might not be the best place for that : ) If you're ever around the Discord and want to nudge us in any direction, it would be really nice to hear feedback.

Appreciate you taking the time to look into this!

PR #1128, opened 2024-04-15
Feat/enhanced share

pzaback:
## Description

Added ability to specify an output directory in which to export markdown session file with the "share" slash-command
- Includes option to do pseudo-tilde expansion in case user enters that instead of full path
- Included ability to use the constant `CURRENT_WORKSPACE` in the file path
- default filename includes a datetimestamp to avoid clobbering previously exported sessions

## Checklist

### DIFF
--- a/core/commands/slash/share.ts
+++ b/core/commands/slash/share.ts
@@ -22,8 +22,18 @@ const ShareSlashCommand: SlashCommand = {
     if (outputDir.startsWith("~")) {
       outputDir = outputDir.replace(/~/, homedir);
     }
+    const workspaceDirs = await ide.getWorkspaceDirs();
+    
+    // Although the most common situation is to have one directory open in a
+    // workspace it's also possible to have just a file open without an
+    // associated directory or to use multi-root workspaces in which multiple
+    // folders are included. We default to using the first item in the list, if
+    // it exists.
+    const workspaceDirectory = workspaceDirs?.[0] || "";
+    outputDir = outputDir.replace(/CURRENT_WORKSPACE/, workspaceDirectory);
+    const outPath = path.join(outputDir, `session.md`); //TODO: more flexible naming
 
-    const outPath = path.join(outputDir, `session.md`);
+
     await ide.writeFile(outPath, content);
     await ide.openFile(outPath);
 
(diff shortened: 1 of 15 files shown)


sestinj:
Very nice! Made a couple of comments, but overall design is awesome

core/commands/slash/share.ts
```diff
@@ -1,24 +1,73 @@
...
+  date.setHours(date.getHours() - offsetHours);
+  date.setMinutes(date.getMinutes() - offsetMinutes);
+
+  return date;
+}
+
+function asBasicISOString(date: Date): string {
+  const isoString = date.toISOString();
+
+  return isoString.replace(/[-:]/g, "").replace(/\.\d+Z/, "");
```
rootedbox: can be compacted to isoString.replace(/[-:.\d+Z]/g, "");

pzaback: This appears to replace all the digits, leaving only the letter T for any given ISOString such as "2023-05-18T12:34:56.789Z" but I see where you're going with simplifying to a single replace call and will do something similar. By any chance did you use your new /review slash command to identify/generate this? Regexes seem to me like the perfect sort of thing for LLMs to help with, at least for me given that I understand but am not an expert with them. But in limited testing so far I've found I have to be very careful with how I describe what I want, and even providing input/output examples doesn't guarantee success.

rootedbox: I was using mixtrail in continue and asked `merge the regex in the 2 replaces and then explain yourself isoString.replace(/[-:]/g, "").replace(/\.\d+Z/, "");`

it responded..

Sure, you can chain the two replace() methods together like this:
```
return isoString.replace(/[-:.\d+Z]/g, "");
```

This regular expression (/[-:.\d+Z]/g) matches any of the characters -, :, ., +, or Z that appear in the ISO string, and the g flag at the end of the expression tells it to do this for all occurrences of the matched characters, not just the first one.

This will remove all -, :, ., +, and Z characters from the ISO string, leaving only the digits that represent the date and time.

pzaback:
OK, I've finished making the changes we discussed during OH and rebased on the latest preview commit and as far as I'm concerned this is ready for final review.

Issue #991, opened 2024-03-19
Overwrites Previous Function when use continue in option


Chhunneng:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 14.2.1
- Continue: v0.9.90
- IDE: VSCode 1.87.1
```

### Description

When I use `continue` to edit like `write a docstring for this code` it works. but when finished I try to select the new function and use the same feature to `write a docstring for this code` again, it writes on the old `function` again.

### To reproduce

I try to use continue and it work
<img width="749" alt="Screenshot 2024-03-19 at 17 02 16" src="https://github.com/continuedev/continue/assets/47348453/6e5aae42-5b84-4ab8-afb1-13dd2423b8c7">

<img width="890" alt="Screenshot 2024-03-19 at 17 06 42" src="https://github.com/continuedev/continue/assets/47348453/39eb206b-2a8d-4816-b130-755e35189314">

After that, I try on new function but it back to old function
<img width="815" alt="Screenshot 2024-03-19 at 17 07 31" src="https://github.com/continuedev/continue/assets/47348453/ba4ddda2-d707-481a-9bc8-c1a1165f3ef8">

<img width="855" alt="Screenshot 2024-03-19 at 17 07 59" src="https://github.com/continuedev/continue/assets/47348453/43b5f19e-fb29-44eb-880e-47617fbd9323">

TheLapinMalin:
Bumping this, as I've had this issue as well with Ctrl+I:

> Ctrl+I without selecting code seems to write code in other places (if you've used Ctrl+I once in the file somewhere, then move somewhere else and use Ctrl+I again, it generates code at the first Ctrl+I location again).

justinmilner1:
I was able to replicate this - I'll add that the issue only occurs if you don't accept or reject the suggested diff. 

It _does_ seem to generate a new docstring for the newly highlighted function, but it annotates the previous function. 

1) Generate docstring for function 1
2) Don’t accept or reject
3) generate another docstring for function 2
4) See that newly generated docstring is describing function 2, but is annotated upon function 1

[...]

PR #1137, opened 2024-04-16
Con 991: Diff Manager Improvements

justinmilner1:
# This work has been continued in https://github.com/continuedev/continue/pull/1198

## Description
This MR fixes two main issues:
1) Edits to a file while a vertical diff was open would not be acknowledged by diff manager, and so the accept/reject would be applied at the wrong location.
2) Diff closure was not handled properly when new diffs were created in same file, so new diffs would generate at old diff location

Also, a bug was discovered: Reject/Accept all does not work when there are multiple blocks - instead only the first block receives the apply (I plan on fixing this soon (this weekend))
Update: 4/18 - added accept/reject all blocks on 'accept/reject all'

### Resolves:
https://github.com/continuedev/continue/issues/991

### DIFF
--- a/extensions/vscode/src/diff/verticalPerLine/manager.ts
+++ b/extensions/vscode/src/diff/verticalPerLine/manager.ts
@@ -19,7 +19,9 @@ export class VerticalPerLineDiffManager {
 
   filepathToCodeLens: Map<string, VerticalDiffCodeLens[]> = new Map();
 
-  constructor(private readonly configHandler: ConfigHandler) {}
+  constructor(private readonly configHandler: ConfigHandler) {
+    this.setupDocumentChangeListener();
+  }
 
   createVerticalPerLineDiffHandler(
     filepath: string,
@@ -77,6 +79,38 @@ export class VerticalPerLineDiffManager {
 
   getHandlerForFile(filepath: string) {
     return this.filepathToHandler.get(filepath);
+  }
+
+  //called by constructor
+  //Creates a listener for document changes
+  private setupDocumentChangeListener() {
+    console.log("setting up onDidChangeTextDocument listener")
+    vscode.workspace.onDidChangeTextDocument((event) => {
+      // Check if there is an active handler for the affected file
+      const filepath = event.document.uri.fsPath;
+      const handler = this.getHandlerForFile(filepath);
+      if (handler) {
+        // If there is an active diff for that file, handle the document change
+        this.handleDocumentChange(event, handler);
+      }
+    });
+  }
+
+  private handleDocumentChange(
+    event: vscode.TextDocumentChangeEvent,
+    handler: VerticalPerLineDiffHandler
+  ) {
+    // Loop through each change in the event
+    console.log("Handling doc change")
+    event.contentChanges.forEach((change) => {
+      // Calculate the number of lines added or removed
+      const linesAdded = change.text.split('\n').length - 1;
+      const linesDeleted = change.range.end.line - change.range.start.line;
+      const lineDelta = linesAdded - linesDeleted;
+  
+      // Update the diff handler with the new line delta
+      handler.updateLineDelta(event.document.uri.fsPath, change.range.start.line, lineDelta);
+    });
   }
 
   clearForFilepath(filepath: string | undefined, accept: boolean) {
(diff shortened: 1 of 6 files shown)


extensions/vscode/src/diff/verticalPerLine/manager.ts
```diff
@@ -142,38 +147,50 @@ export class VerticalPerLineDiffManager {
...
       setTimeout(resolve, 200);
     });
+
+    const startLine = editor.selection.start.line;
+    const endLine = editor.selection.end.line;
+
+    //Create new handler
     const diffHandler = this.createVerticalPerLineDiffHandler(
       filepath,
-      existingHandler?.range.start.line ?? startLine,
```
justinmilner1, line 160: I wasn't sure why flexibility to reuse the existing handler was enabled - was the intention at some point to support multiple diffs per file?

extensions/vscode/src/diff/verticalPerLine/manager.ts
```diff
@@ -142,38 +147,42 @@ export class VerticalPerLineDiffManager {
...
+      this.acceptRejectVerticalDiffBlock(false, filepath)
+    }
+
     existingHandler?.clear(false);
+    
     await new Promise((resolve) => {
       setTimeout(resolve, 200);
     });
+
+    const startLine = editor.selection.start.line;
```
justinmilner1: determine start/end _after_ the old block has been deleted

extensions/vscode/src/diff/verticalPerLine/manager.ts
sestinj: I believe this is used for cmd+I to "add further instructions". https://github.com/continuedev/continue/blob/preview/extensions/vscode/src/commands.ts#L238

This might actually complicate this PR a bit. If there is a currently opened diff, then using cmd+I again should prompt the user to add further instructions, which looks like "<prev instructions>, ".

One way of moving forward is to remove the previous diff (as you've implemented here) if there is any text selected, or "re-prompt" if the cursor selection is empty

justinmilner1: I passed the 'previousInput' variable from commands into streamEdit as an optional parameter and used it as a flag.
This variable will have a value if it comes from cmd+I, or will be undefined otherwise.

justinmilner1:
Last commits makes these changes compatible with cmd+I as well as actually reject/accept all blocks in the given file when the user clicks the 'reject/accept all' code lens.

extensions/vscode/src/diff/verticalPerLine/manager.ts
```diff
@@ -4,6 +4,7 @@ import { getMarkdownLanguageTagForFile } from "core/util";
 import { streamDiffLines } from "core/util/verticalEdit";
 import * as vscode from "vscode";
 import { VerticalPerLineDiffHandler } from "./handler";
+import { start } from "repl";
```
sestinj, line 7: Is this just leftover?

sestinj: @justinmilner1 this is last thing, assuming it's safe to delete I can go ahead and do that while I merge

justinmilner1: @sestinj whoops yeah that's not supposed to be in there

extensions/vscode/src/diff/verticalPerLine/manager.ts
```diff
@@ -114,66 +121,86 @@ export class VerticalPerLineDiffManager {
...
+    if (!all) {
+      toHandle = blocks[index];
+      if (!toHandle){
+        return;
+      }
+    } else {
+      toHandle = blocks
+    }
+
+    for (const block of toHandle) {
```
sestinj: @justinmilner1 cd extensions/vscode && npx tsc raises an error here, and it looks like toHandle could be either a list or single value from blocks. I assume the intended logic is to loop through the list if it's a list, otherwise handler.acceptRejectBlock for just blocks[index]?

justinmilner1: @sestinj Sorry that was sloppy.

Pushed up a fix. Tested more thoroughly.

sestinj:
@justinmilner1 I'm doing some tests now locally. First thing that comes up is accept/reject (not all, but the buttons to do it for each block individually) seems like it's regressed a bit. The first thing I saw is that if I were to "Accept" the top block, then the full "Accept All / Reject All / Accept / Reject" codelens would remain in the same position. Whereas in current versions of the extension, you can see it move down to the new top block.

I'm not certain which part of the PR this is coming from, but my guess is that whatever attempts to solve the new problem you found are the cause?

Issue #1041, opened 2024-03-28
Extension API for VS Code: e.g. allow arbitrary requests from other extensions


FunkMonkey:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Continue already provides access to a diverse set of LLM providers and models as well as editor features such as a diff preview for LLM-based code suggestions. It would be nice, if one could reuse this feature set in custom VS Code extensions. Particularly simply sending prompts (with context) and receiving the completions would be a nice basic feature for other extension authors. 

### Solution

Provide an API object via the extension's `activate` function, as also proposed in #803. The initial focus should be on crafting a fundamental API, with a primary aim of determining which aspects of continue's functionality are most beneficial to expose for other extension authors. 

I guess we could just brainstorm on this here in this issue.

PR #1144, opened 2024-04-17
Extension API for VS Code v1

FunkMonkey:
## Description

This is a PR for #1041. It is as simple as possible, basically exporting the `VsCodeExtension` instance to be usable by other extensions. 

This could for example be used to directly communicate with the models (e.g. via `model.streamChat` etc.) or to communicate with the chat webview, for example:
```ts
continueInstance.webviewProtocol.request("userInput", {
  input: "Please do X for me!",
});
```

I believe it is currently ok to export the `VsCodeExtension` instance, despite giving full access to internals of `continue`, in order to figure out what features are actually useful for other extension authors. Later on, the `instance` property should be deprecated in favor of a better designed external API. 

Maybe the author of #803 could make use of this simple API, too, instead of adding the proposed functionality to `continue` itself.

For a better Typescript experience, the types of the API should also be exposed as an npm package.

## Checklist

### DIFF
--- a/extensions/vscode/src/activation/activate.ts
+++ b/extensions/vscode/src/activation/activate.ts
@@ -40,7 +40,7 @@ function showRefactorMigrationMessage(
   }
 }
 
-export async function activateExtension(context: vscode.ExtensionContext) {
+export async function activateExtension(context: vscode.ExtensionContext): Promise<VsCodeExtension> {
   // Add necessary files
   getTsConfigPath();
 
@@ -67,4 +67,6 @@ export async function activateExtension(context: vscode.ExtensionContext) {
       extensionVersion: getExtensionVersion(),
     });
   }
+
+  return vscodeExtension;
 }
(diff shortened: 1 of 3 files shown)


sestinj:
@FunkMonkey I'm going through all of the many PRs that have ended up on the back burner, and I realize now that I completely missed this. The principle I think we're coming to is that it's not best to accept "input" from other extensions. What I mean by this is that you could do something like registering a custom "slash command" or "context provider" (like [here](https://github.com/continuedev/continue/pull/1288)), but we don't want to allow arbitrary usage of Continue resources without the user triggering them. Especially when something like OAuth is in place to allow users to access resources from an account, this could pose a potential security risk.

I'm going to close this PR because of this, but I would be curious to hear in the GH issue whether you think that there might be another way of accomplishing what you're looking to do. Hope to hear from you, and hopefully we can make something work!

PR #1151, opened 2024-04-18
🐛 set supportsCompletions based on useLegacyCompletionsEndpoint model setting

### LINKED ISSUE #1132: Continue: Write comments for this code produces math questions

- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 14.4
- Continue: v0.9.112 (pre-release)
- IDE: VSCode 1.88.0
```

### Description

When selecting code and asking continue.dev to comment on it, it will produce math equations instead

Using OpenAI configured like this:

```json
    {
      "title": "Mistral",
      "provider": "openai",
      "model": "mistral-7b-instruct",
      "apiKey": "your-api-key",
      "apiBase": "https://your-base-url.com/llm/v1"
    }
```

Using Mistal with Ollama the extension produces the expected result, but it seems using the OpenAI provider only empty strings get send, which will lead Mistral to dream about math.

```sh
async_llm_engine.py:436] Received request cmpl-6def2b5f4c4d4d61952dbf6a82262f4d: prompt: '<s>',
```

### To reproduce

1. Select this function and Press `CMD + P`
```python
   def get_attribute(attributes: list[dict[str, str]], key: str) -> Optional[str]:
      for attribute in attributes:
          if attribute.get("key") == key:
              return attribute.get("value")
      return None
 ```

2. Write `Continue: Write comments for this code`
3. See the result below

image

### Log output

```Shell
async_llm_engine.py:436] Received request cmpl-6def2b5f4c4d4d61952dbf6a82262f4d: prompt: '<s>',
```

bufferoverflow:
Closes #1132

## Description

For self-hosted OpenAI compatible endpoints `/v1/completions`  is used instead of `/v1/chat/completions`. With this PR the setting `useLegacyCompletionsEndpoint` is considered as well so people can control the endpoint to be used.

## Checklist

### DIFF
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -52,7 +52,8 @@ export abstract class BaseLLM implements ILLM {
     if (this.providerName === "openai") {
       if (
         this.apiBase?.includes("api.groq.com") ||
-        this.apiBase?.includes(":1337")
+        this.apiBase?.includes(":1337") ||
+        this._llmOptions.useLegacyCompletionsEndpoint?.valueOf() === false
       ) {
         // Jan + Groq don't support completions : (
         return false;
--- a/docs/docs/reference/Model Providers/openai.md
+++ b/docs/docs/reference/Model Providers/openai.md
@@ -52,4 +52,10 @@ If you are [using an OpenAI compatible server / API](../../model-setup/select-pr
 }
 ```
 
+To force usage of `chat/completions` instead of `completions` endpoint you can set
+
+```json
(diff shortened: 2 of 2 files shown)


sestinj:
@bufferoverflow I'm going to make the one change that we need to default to setting this value to `undefined` at the top of the `OpenAI` class, but otherwise this is a nice clean solution

PR #1159, opened 2024-04-21
✨ add Cohere as Reranker

### LINKED ISSUE #1122: Support Cohere as Reranker
### Validations

### Problem

[Codebase retrieval](https://continue.dev/docs/walkthroughs/codebase-embeddings) could use Cohere's [Rerank](https://docs.cohere.com/docs/reranking) models

### Solution

Add supporting code under [core/context/rerankers](https://github.com/continuedev/continue/tree/d40b4c5d4bf593bc81f0b909ccae9837e8785bfb/core/context/rerankers)

maxbrunet:
## Description

Closes #1122

## Checklist

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -646,7 +646,7 @@ declare global {
     embed(chunks: string[]): Promise<number[][]>;
   }
   
-  export type RerankerName = "voyage" | "llm" | "free-trial";
+  export type RerankerName = "cohere" | "voyage" | "llm" | "free-trial";
   
   export interface RerankerDescription {
     name: RerankerName;
--- a/core/context/rerankers/cohere.ts
+++ b/core/context/rerankers/cohere.ts
@@ -1,0 +1,47 @@
+import fetch from "node-fetch";
+import { Chunk, Reranker } from "../..";
+
+export class CohereReranker implements Reranker {
+  name = "cohere";
+
+  static defaultOptions = {
+    apiBase: "https://api.cohere.ai/v1/",
+    model: "rerank-english-v3.0",
+  };
+
+  constructor(
+    private readonly params: {
+      apiBase?: string;
+      apiKey: string;
+      model?: string;
+    },
+  ) {}
+
+  async rerank(query: string, chunks: Chunk[]): Promise<number[]> {
+    let apiBase = this.params.apiBase ?? CohereReranker.defaultOptions.apiBase
+    if (!apiBase.endsWith("/")) {
+      apiBase += "/";
+    }
+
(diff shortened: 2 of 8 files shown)


sestinj:
@maxbrunet flawless, thank you for the PR!

Related to this, I'm curious whether you're aware of any good re-ranking benchmarks outside of those from MTEB, particularly if they are specific to code?

sestinj (approved):

maxbrunet:
Thank you @sestinj, would you have an idea when we can expect these features to be released?

> Related to this, I'm curious whether you're aware of any good re-ranking benchmarks outside of those from MTEB, particularly if they are specific to code?

I asked [Coral, our showcase chatbot](https://cohere.com/chat), and it literally pointed me to one of your blog posts, so I might not bring much news to you unfortunately: https://blog.continue.dev/an-introduction-to-code-llm-benchmarks-for-software-engineers/

By asking around, Human Eval, MBPP, DS1000 were mentioned to me, and these 2 projects not in the blog post:

* https://github.com/github/CodeSearchNet
* https://github.com/Jun-jie-Huang/CoCLR

sestinj:
@maxbrunet I just now am pushing a pre-release version (0.9.119) including all of the updates.

Hilarious that it brought back our post... The links you shared are new to me though, thank you!

sestinj:
Sorry misspoke on releasing yesterday. Working on a fix that had to happen before, then ideally today will be out

maxbrunet:
No worries, would it possible to have matching Git tags on the commits used to build them too? I had not realized there were versions published without Git tag

Also what about JetBrains?

sestinj:
@maxbrunet JetBrains doesn't yet have support for codebase retrieval, but this isn't far off, just a small refactor I want to get in beforehand.

It's true that pre-releases don't get tagged or added to GitHub releases. The plan is to only tag main releases because pre-release happens so much more frequently (unless there's a really strong reason why this should be considered)

maxbrunet:
> JetBrains doesn't yet have support for codebase retrieval

Ok, that's fine, no hurry, beyond cobase retrieval, I was also wondering if Cohere as model provider was released there

> The plan is to only tag main releases because pre-release happens so much more frequently (unless there's a really strong reason why this should be considered)

Git tags are cheap and make it easy to identify if a commit is part of a release, e.g.

image

It also makes it straightforward to bisect between two releases if something breaks

sestinj:
That's a fair enough point about bisecting. Seems reasonable enough for us to start tagging pre-releases then

I'm working on getting JetBrains 0.0.46 out the door right now, but the store will probably accept it on Monday morning. This will include the Cohere model provider

PR #1168, opened 2024-04-22
new docs updates

TyDunn:
## Description

Updating docs and READMEs

## Checklist

- [x] update docs
- [x] update main README
- [x] update VS Code README
- [x] updated JetBrains README


sestinj (commented):
Need to sleep and no time to check these things before then, but:
- need to be sure we've double-checked link references. If pages are deleted, we should add a redirect from the old one (at the bottom of docusaurus.js, there's an example there)
- was it purposeful to remove the discord icons and stuff on the readme?

sestinj (approved):

PR #1174, opened 2024-04-23
📝 rename googlepalmapi.md to googlegeminiapi.md

pzaback:
## Description

Renamed doc file and all references to it from Palm to Gemini

## Checklist


docs/docs/model-setup/select-provider.md
```diff
@@ -57,6 +57,6 @@ You can use commercial LLMs via APIs using:
 - [OpenAI API](../reference/Model%20Providers/openai.md)
 - [Azure OpenAI Service](../reference/Model%20Providers/openai.md) (OpenAI compatible API)
 - [Anthrophic API](../reference/Model%20Providers/anthropicllm.md)
-- [Google Gemini API](../reference/Model%20Providers/googlepalmapi.md)
+- [Google Gemini API](../reference/Model%20Providers/googlegeminiapi.md)
```
sestinj: Only thing to add to this PR is we need to make sure that the docs page here exists. In `docs/docs/reference/Model Providers` can you change `googlepalmapi.md` to this new name? (actually I might choose just the name "gemini" while we're changing things, but having the page exist is the important part : ))

sestinj (changes requested):

docs/docs/model-setup/select-provider.md
pzaback: Ah geez, silly omission. I had actually renamed the file first through GH web interface but then couldn't figure out how to quickly update links in different files in the same commit so I just started over in local repo but forgot that repo-wide find/replace wouldn't include updating the filename itself. 😆

sestinj: All good! And thanks for the rename to geminiapi. If you don't mind...I'm going to attempt to wait just a bit on merging because we deploy docs from the main branch, and I'd like to only release these links into the wild once the page they point to actually exists. So that should be this Friday that I plan for another main release

pzaback: No rush from my end whatsoever. Aside from this being a minor thing I happened to stumble across  after seeing that the corresponding provider module was recently deleted, in general I have no desire to pressure merging of my PRs.

PR #1177, opened 2024-04-24
First Lens.. Mouse over TODO's

rootedbox:
## Description

Added the ability to mouse over and react to TODO's

Screen Shot 2024-04-23 at 21 13 07 PM

### DIFF
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -138,14 +138,14 @@ const commandsMap: (
   sidebar: ContinueGUIWebviewViewProvider,
   configHandler: ConfigHandler,
   diffManager: DiffManager,
-  verticalDiffManager: VerticalPerLineDiffManager,
+  verticalDiffManager: VerticalPerLineDiffManager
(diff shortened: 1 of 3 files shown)


rootedbox:
@sestinj is the auto linter removing the last comma?

sestinj:
@rootedbox Yes most likely Prettier is doing that? But Peter Zaback added `"trailingComma": "all"` to the .prettierrc in his recent PR, which will help standardize

sestinj:
My thoughts after pulling and playing around with this a bit...I like that it's not a codelens because in a way that makes it less invasive. The one downside is that this feels like a bit of a hidden feature that would be extremely rare to discover. More importantly though, I am not sure how frequently this will be able to accomplish a TODO, which in practice is often a multi-file edit, or something difficult enough to accomplish that somebody decided to postpone the work. That is mostly feedback on the code insertion feature. As for chat, I can see more success.

Also: I'm curious what the limits of this hover view are. If there's any way to display something like a button, an image, or just larger text, this might be an extremely valuable UI tool

PR #1182, opened 2024-04-24
Respect Retry-After header when available from 429 responses

nejch:
## Description

Usually a 429 response code comes with a `Retry-After` header, providing the number of seconds to wait until the next request. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429. This PR attempts to respect that value when provided.

Hopefully this makes requests more resilient. In our case we have custom rate limits in place and not sure the exponential backoff is that accurate.

* 77061aa98e9f438dd9386f7605531380302775a1 removes dead code that has been replaced by the `withExponentialBackoff` util.
* 4dfa6a9cfcc66bf4a9a8514c0cfddc9538f72e4f adds a bit of logic to `withExponentialBackoff`  to respect the standard [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header supplied in  

⚒️ with ❤️ by [Siemens](https://opensource.siemens.com/)

## Checklist


sestinj:
@nejch This is great! Thanks for a clean PR : )

A new VS Code pre-release will be available later today with these changes

sestinj (approved):

sestinj:
Hi @nejch, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

PR #1197, opened 2024-04-27
Fix routing of acceptall rejectall commands

kiates:
## Description

Accept/Reject All was inadvertantly being routed to the commands to perform just Accept/Reject.  This fixes that routing and fixes the issue where Accept/Reject All only accpts/rejects a single diff hunk.

## Checklist

- [X] The base branch of this PR is `preview`, rather than `main`


sestinj (approved):
Nice fix, thank you!

PR #1204, opened 2024-04-29
Fix issue with @codebase provider when n becomes odd due to a divide …

kiates:
…by 2 during the full text search portion of the query.

## Description

If nFinal is odd in the @codebase provider, the divide by 2 that happens results in a value of n that is a fractional value.  When this is passed to the sqllite database query it throws a datatype mismatch error.  I picked ceil over floor to ensure that the pathological case of nFinal = 1 results in 1 rather than 0.

## Checklist

- [X] The base branch of this PR is `preview`, rather than `main`

### DIFF
--- a/core/indexing/FullTextSearch.ts
+++ b/core/indexing/FullTextSearch.ts
@@ -121,7 +121,7 @@ export class FullTextSearchCodebaseIndex implements CodebaseIndex {
     let results = await db.all(query, [
       ...tagStrings,
       ...(filterPaths || []),
-      n,
+      Math.ceil(n),
     ]);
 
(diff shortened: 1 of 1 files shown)


sestinj:
Hi @kiates, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

kiates:

PR #1217, opened 2024-05-02
Fix issue with the Docs reranking prompt

JoseVega:
I fixed a bug where you were sending the query variable (which holds the base URL of the doc) to the rerank method, and it made no sense to rerank the chunks based on a URL. So I changed it to extras.fullInput because it should rerank based on the user input, which should provide better results.

## Description

## Checklist


sestinj:
@JoseVega nice catch!

sestinj (approved):

PR #1231, opened 2024-05-04
Active file default context

sestinj:
Use the active file as context by default, unless users explicitly says not to


chenyuxi2002:
How do I change the configuration to explicitly not do this?

Issue #1195, opened 2024-04-26
[CON-232] Bring custom docs to the top of the dropdown


sestinj:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

In the @ docs context provider, only a few suggestions are shown in the dropdown, and ideally these would be ones that have been added manually

<sub>[CON-232](https://linear.app/continue/issue/CON-232/bring-custom-docs-to-the-top-of-the-dropdown)</sub>

PR #1239, opened 2024-05-06
CON-232 bring custom docs to top, alphabetize doc results, make scrol…

justinmilner1:
## Description

Resolves: https://github.com/continuedev/continue/issues/1195

I've made three changes:
- Ensure that custom docs are sorted to the top of the suggestions list (I think they should be even without the sorting regardless though? I couldn't reproduce the custom docs not showing up at the top)
- Alphabetize suggestion list
- Make suggestion list scrollable (not just for docs, but also for folders, and others which make use of MentionList)

Note: Potential downside of making scrollable is that the 'add docs' button isn't always visible - If you think it should be, I can pin that to the bottom (as it's own diff) of the suggestions list so that it's always visible regardless of the scrolling.

Two custom docs show at the top
Screenshot from 2024-05-05 21-11-19

Scroll to the bottom and see add docs
Screenshot from 2024-05-05 21-11-45

### DIFF
--- a/core/context/providers/DocsContextProvider.ts
+++ b/core/context/providers/DocsContextProvider.ts
@@ -24,7 +24,7 @@ class DocsContextProvider extends BaseContextProvider {
     extras: ContextProviderExtras,
   ): Promise<ContextItem[]> {
     const { retrieveDocs } = await import("../../indexing/docs/db");
-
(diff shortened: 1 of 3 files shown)


sestinj (approved):
All looks good!

PR #1255, opened 2024-05-09
Add --no-dependencies to vsce package

djgould:
## Description
`vsce pack` has weird behavior where it tries to package all your dependencies from your node_modules itself. This can cause issues depending on the package manager configuration, and prevents modern package managers from being used.

since we are already running esbuild in the prepublish step we can just skip the dependency packing from vsce with the `--no-dependencies` flag

see: microsoft/vscode-vsce#421 (comment)

## Checklist


sestinj (approved):
this was a great find, thank you!

PR #1256, opened 2024-05-09
Add github action to install dependencies

djgould:
## Description
I noticed some issues running the install script again, so figured this could be useful. At the very least for me to check if it happens in a clean env.

I only see one other github action in the project so let me know if there isn't any interest in adding more.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,15 @@
+name: Install Dependencies
+
+on: [push]
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+      - name: Run install script
+        run: bash install-dependencies.sh


djgould:
Ah just realized I need approval for the workflow to run. You can see the current install failure in my fork. happens locally as well.

https://github.com/djgould/continue/actions/runs/9021010813/job/24787377254

sestinj:
Hm yeah the approval is something we'll keep around for security reasons. Given this, I'm not sure it's super helpful to have the action here in the repo. Thoughts on a script instead to "remove-all-dependencies"? I don't disagree that the fresh start would be nice, and maybe we can just make it easy to do so locally. The other option is a local clone

sestinj:
As for the issue, it seems like a matter of explicitly declaring the dependency? My plan was to move to pnpm anyway, so in the process of doing that I'm sure we'll solve this

djgould:
> Given this, I'm not sure it's super helpful to have the action here in the repo

cool i'll close this then

> Thoughts on a script instead to "remove-all-dependencies"

That would be great!

>As for the issue, it seems like a matter of explicitly declaring the dependency? My plan was to move to pnpm anyway, so in the process of doing that I'm sure we'll solve this

Yeah I think it was just a missing dependency. Not sure how it slipped in. +1 for switching to pnpm

sestinj:
totally! commit incoming for the switch to pnpm

PR #1265, opened 2024-05-11
📦 switch to pnpm

sestinj:
Use pnpm rather than yarn. Possible with --no-package option for vsce


refactorthis:
Why go back to npm after this?

PR #1319, opened 2024-05-20
fix docs social card rendering

TyDunn:
## Description

- uses absolute instead of relative path


TyDunn:
@sestinj can you merge?

sestinj (approved):

PR #1324, opened 2024-05-20
add toString

justinmilner1:
## Description

One of my recent MRs caused this bug if the user has titles that are composed solely of numbers:
Error loading submenu items from docs: TypeError: a.title.localeCompare is not a function

Fix: Convert the titles to strings.


sestinj (approved):
good catch!

Issue #1156, opened 2024-04-20
[CON-228] "url" context provider


sestinj:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

The docs context provider will crawl an entire site and find multiple pages, but oftentimes you only want literally one page

<sub>[CON-228](https://linear.app/continue/issue/CON-228/url-context-provider)</sub>

oldluke92:
is the HTTP context provider what you are looking for ?
https://github.com/continuedev/continue/blob/main/core/context/providers/HttpContextProvider.ts

justinmilner1:
@sestinj I think single-page-crawl should probably just be an option to set from within the AddDocsDialog? (rather than a whole new context provider)

I also see the case for a slider or input box to specify the max number of pages to crawl. Doesn't look like there is a max pages limit currently - I tried with a Wikipedia page and indeed that crawled indefinitely.

What do you think?

sestinj:
@justinmilner1 I think this would be nice. Max depth seems like the right thing, otherwise it's hard to prioritize one page vs. the other when they are same depth.

Let's add this option to `SiteIndexingConfig` and then instead of doing the silly thing that I did and pass a freely defined object over the webviewProtocol for index/addDocs (https://github.com/continuedev/continue/blob/preview/core/protocol.ts#L65-L66), let's pass a `SiteIndexingConfig`. Probably just use the URL input by the user as both starting and root URL so the UI remains simple.

PR #1343, opened 2024-05-22
Add Doc Indexing Max Depth

justinmilner1:
## Description

Continue's add docs feature currently crawls pages until it runs out of links - this pr introduces a 'max depth' parameter which intends to prevent cases where indexing continues excessively.

Originated from [this](https://github.com/continuedev/continue/issues/1156) thread:
Screenshot from 2024-05-28 12-49-42

Screenshot from 2024-05-28 12-50-09

### Changes:
- New input box in the addDocs diagog
- Backend changes to indexing
   - stop crawling at maxDepth
   - catch errors where reasonable (ex: bad urls)
   - condense freely-defined object into SiteIndexingConfig object
   - make github page crawl use 'default' branch
   
### Testing:
Cases below passed

**UI:**
- try entering non-numerical
- try negative input
- try special characters

**Backend:**
- use default input 
     - confirm expected depth
     - confirm docs are indexed
- use custom input
    - confirm expected depth
    - confirm docs are indexed
- Use with wikipedia, maxDepth 1
    - should not index 'forever'


sestinj:
Is it purposeful to remove the special crawler for GitHub pages? I haven't visited this in a while, so it very well could have looked better to remove that.

justinmilner1:
@sestinj Ah, purposeful initially, but I intended to move it back before the pr. 

The special crawler is failing (socket hanging up, urls not being found). I haven't looked that deep into it yet - will do this and let you know.

justinmilner1:
@sestinj Looks like the github crawler was only searching for a 'main' branch, so if 'main' didn't exist then it would fail. 
I've added a step to check for the default branch name, then use that.

sestinj (approved):

sestinj:
Great work @justinmilner1!

PR #1345, opened 2024-05-22
Add unit test option to Continue

julianShi:
## Description

## Checklist

- [ ] The base branch of this PR is `preview`, rather than `main`

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -519,6 +519,7 @@ type ModelProvider =
   | "deepinfra"
   | "flowise"
   | "groq"
+  | "zhiman"
   | "custom";
 
 export type ModelName =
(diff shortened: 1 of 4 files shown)


sestinj:
@julianShi this is a very cool idea! I think that quite frequently unit tests will be written in other files, so we won't want to use the inline edit feature for this, but we could instead send a prompt to the chat window by using something like this:

https://github.com/continuedev/continue/blob/preview/extensions/vscode/src/commands.ts#L387-L392

Would you also mind changing the base branch of this PR to preview rather than main? We merge there first for testing purposes

julianShi:
> @julianShi this is a very cool idea! I think that quite frequently unit tests will be written in other files, so we won't want to use the inline edit feature for this, but we could instead send a prompt to the chat window by using something like this:
> 
> https://github.com/continuedev/continue/blob/preview/extensions/vscode/src/commands.ts#L387-L392
> 
> Would you also mind changing the base branch of this PR to preview rather than main? We merge there first for testing purposes

Thank you so much for the prompt suggestion!! ^_^

PR #1352, opened 2024-05-23
Feat/gemini flash

pzaback:
## Description

Added references to Gemini Flash in a few more locations where 1.5 Pro was listed but Flash wasn't. Also updates documentation and removes a duplicate reference to "gemini".

## Checklist

## Additional comments
- It seems reasonable to me to add Flash to the free trial similar to how Haiku and GPT 3.5 are provided to enable comparison of the lightweight version of those models, but didn't make such changes because anything relating to the trial feels like maybe it should be handled by Nate/Ty.
- In `Gemini.ts (15-18)` the default model is still specified as "gemini-pro". It seems perhaps this should be updated to 1.5, but I don't know for certain what differences may cause unexpected problems so I left it alone for now.
```ts
static defaultOptions: Partial<LLMOptions> = {
    model: "gemini-pro",
    apiBase: "https://generativelanguage.googleapis.com/v1beta/",
  };
```

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -563,6 +563,7 @@ declare global {
     // Gemini
     | "gemini-pro"
     | "gemini-1.5-pro-latest"
+    | "gemini-1.5-flash-latest"
     // Mistral
     | "mistral-tiny"
     | "mistral-small"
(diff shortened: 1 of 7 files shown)


sestinj (approved):

sestinj:
This is great, thank you @pzaback! Agreed that it would make sense to have this in the trial as well

PR #1366, opened 2024-05-26
Make arrows scroll in doc dropdown menu

justinmilner1:
## Description
Doc dropdown menu was made scrollable recently - This pr adds scrolling via arrow keys.

I tested that the correct docs are being selected.

https://github.com/continuedev/continue/assets/42585006/f82ef25c-445d-4463-b23b-a68f822ecce7

### DIFF
--- a/gui/src/components/mainInput/MentionList.tsx
+++ b/gui/src/components/mainInput/MentionList.tsx
@@ -231,13 +231,48 @@ const MentionList = forwardRef((props: MentionListProps, ref) => {
     }
   };
 
-  const upHandler = () => {
-    setSelectedIndex((selectedIndex + allItems.length - 1) % allItems.length);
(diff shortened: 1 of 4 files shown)


gui/src/components/mainInput/MentionList.tsx
```diff
@@ -231,13 +231,45 @@ const MentionList = forwardRef((props: MentionListProps, ref) => {
     }
   };
 
-  const upHandler = () => {
-    setSelectedIndex((selectedIndex + allItems.length - 1) % allItems.length);
```
justinmilner1, line 235: necessary to change because modulo will cause looping

sestinj:
@justinmilner1 finally tested this and it really feels like a big quality of life improvement!

PR #1368, opened 2024-05-26
Fix sidebar indexing status timing-related bugs

justinmilner1:
## Description

Bugs targeted:
- When sidebar is opened until indexing completes, it opens as grey, yet indexing has actually occurred. Re-indexing button is not interactive in this case, so its a dead end for the user.
- When sidebar opened before backend is initialized, it loads in as disabled regardless of disabled/enabled settings, this is confusing for the user.

Changes:
- Rename the light blue state from 'starting up' to 'loading'. Purpose is to indicate that continue is loading, indexing has no status yet (has not started, may not start if disabled)
- Make light blue state default when app is loaded
- Provide a mechanism for loading a non-default state in cases where indexing has completed before the sidebar has been opened (otherwise, the default state will not be updated)
- Re-add yield failed state (it was removed at some point)
- Discovered a flaw in table creation logic (table creation is attempted despite table existing). Made a quick fix. Looking further into the logic is on to-do list

Testing:
- Open in each state before indexing complete
- Open in each state after indexing complete

### DIFF
--- a/binary/build.js
+++ b/binary/build.js
@@ -51,7 +51,7 @@ async function installNodeModuleInTempDirAndCopyToCurrent(packageName, toCopy) {
   console.log(`Copying ${packageName} to ${toCopy}`);
   // This is a way to install only one package without npm trying to install all the dependencies
   // Create a temporary directory for installing the package
-  const adjustedName = packageName.replace(/^@/, "").replace("/", "-");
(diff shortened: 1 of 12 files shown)


sestinj:
Nicely done! I like the UI choice of calling it "loading" in the beginning

sestinj (approved):

PR #1407, opened 2024-06-01
Full project file path in context

tijszwinkels:
## Description

For many types of projects (like index.ts in react/next.js) there's many files with the same file name that live in a different path in the project. Since continue was previously only giving the filename without the path to the LLM, the LLM wouldn't always understand correctly which file is including which.

This MR gives the full file path from the workspace root to the LLM, for:
- @Open Files
- @Files (named files added by hand to the context)
- The 'use active file' (alt-return) context

For example:

Before:
```
"'''test.js
<...>
'''

'''helloNested.py
<...>
'''

What files do you see? Open Files "
```

After:
```
"'''manual-testing-sandbox/test.js
<...>
'''

'''manual-testing-sandbox/nested-folder/helloNested.py
<...>
'''

What files do you see? Open Files "
```

## Checklist

- [🦄] The base branch of this PR is `preview`, rather than `main`

### DIFF
--- a/core/context/providers/FileTreeContextProvider.ts
+++ b/core/context/providers/FileTreeContextProvider.ts
@@ -4,20 +4,12 @@ import {
   ContextProviderExtras,
 } from "../../index.js";
 import { BaseContextProvider } from "../index.js";
+import { splitPath } from "../../util/index.js";
 
 interface Directory {
   name: string;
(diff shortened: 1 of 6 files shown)


sestinj (approved):
I like this, good work!

PR #1441, opened 2024-06-07
Enable Gemini system message use

pzaback:
Older versions of the Google API apparently didn't 
support passing a "system" parameter so a shim 
was used here to convert such a parameter into the
first user message instead. This is no longer
necessary as of the v1beta API, so this change
makes sure the shim is only applied if the user
has specified API version v1. I should also note
that there may still be a bug with this method if
using v1; the reason I found this is I was
experiencing a bug where the message ultimately
being sent after applying the shim was [Object]
[object] or something like that. I fixed it this
way because the system message shouldn't be
removed with v1beta (and presumably later
versions)

## Description

## Checklist

### DIFF
--- a/core/llm/llms/Gemini.ts
+++ b/core/llm/llms/Gemini.ts
@@ -47,7 +47,15 @@ class Gemini extends BaseLLM {
     messages: ChatMessage[],
     options: CompletionOptions,
   ): AsyncGenerator<ChatMessage> {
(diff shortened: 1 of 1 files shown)


sestinj (approved):

sestinj:
nice find! i'd completely forgotten about this little detail : )

Issue #1428, opened 2024-06-06
Offline & install by .vsix file, had error about sqlite3


lcwxz1989:
### Before submitting your bug report

- [ ] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: ubuntu 22.04.4 LTS
- Continue: Continue.continue-0.9.149@linux-arm64.vsix
- IDE: code_1.90.0-1717531825_amd64.deb
```

### Description

Install continue extension by .vsix in offline network. Get error below, I opened developer tools and find error
continue-xxxx/out/build/Release/node_sqlite3.node: cannot open shared object file: No such file or directory.

fry69:
Try again with 0.8.40 or pre-release version 0.9.152 (both just released), according to the commit log there might be a SQLite build fix in it.

priyashpatil:
Might be related issue facing M2:

```
(node:29658) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
[info] Downloading node-sqlite3
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  963k  100  963k    0     0   560k      0  0:00:01  0:00:01 --:--:-- 2439k
x build/Release/node_sqlite3.node
node:fs:3015
  binding.copyFile(
          ^

Error: ENOENT: no such file or directory, copyfile 'bin/darwin-arm64/node_sqlite3.node' -> 'build/node_sqlite3.node'
    at Object.copyFileSync (node:fs:3015:11)
    at /Users/priyash/projects/continue/binary/build.js:226:10
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'copyfile',
  path: 'bin/darwin-arm64/node_sqlite3.node',
  dest: 'build/node_sqlite3.node'
}

Node.js v22.2.0

[...]

PR #1454, opened 2024-06-09
fix: change broken path for node sqlite

anmol242:
## Description

- While installing dependencies, VSCode says `ENOENT: no such file or directory, copyfile 'bin/darwin-x64/node_sqlite3.node' -> 'build/node_sqlite3.node`, this happens because build folder is not inside `binary`, it's inside `targetDir` path.

Fixes [#1428](https://github.com/continuedev/continue/issues/1428)

## Checklist


anmol242:
Not needed, I was installing this on `main`, apologies. Seems like issue is fixed in `preview` branch 😄

PR #1473, opened 2024-06-12
Multi-Build/Package VS Code extensions on single computer.

maxxrdrgz:
## Description

In collaboration with nates prepackage-cross-platform.js script, I've written the rest of the code that will fetch the necessary binaries to build for other platforms decoupling the build process from needing other hardware to build on.

## Checklist

- [ ] Review my code.
- [ ] Test that all builds created are functional (I could only test the builds on what I have locally available) npm run package-all
- [ ] Ensure single packaging still works by running npm run package

### DIFF
--- a/binary/build.js
+++ b/binary/build.js
@@ -224,12 +224,17 @@ async function installNodeModuleInTempDirAndCopyToCurrent(packageName, toCopy) {
     );
 
     // Copy to build directory for testing
-    const [platform, arch] = target.split("-");
-    if (platform === currentPlatform && arch === currentArch) {
-      fs.copyFileSync(
-        `${targetDir}/node_sqlite3.node`,
-        `build/node_sqlite3.node`,
-      );
+    try {
+      const [platform, arch] = target.split("-");
+      if (platform === currentPlatform && arch === currentArch) {
+        fs.copyFileSync(
+          `${targetDir}/node_sqlite3.node`,
+          `build/node_sqlite3.node`,
+        );
+      } 
+    } catch (error) {
+      console.log("[warn] Could not copy node_sqlite to build");
+      console.log(error)
     }
 
     fs.unlinkSync(`${targetDir}/build.tar.gz`);
--- a/extensions/vscode/scripts/prepackage-cross-platform.js
+++ b/extensions/vscode/scripts/prepackage-cross-platform.js
@@ -117,15 +117,6 @@ async function package(target, os, arch, exe) {
     lancePackageToInstall,
     "@lancedb",
   );
-
-  const esbuildPackageToInstall = {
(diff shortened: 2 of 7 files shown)


binary/build.js
```diff
@@ -224,12 +224,17 @@ async function installNodeModuleInTempDirAndCopyToCurrent(packageName, toCopy) {
...
+      if (platform === currentPlatform && arch === currentArch) {
+        fs.copyFileSync(
+          `${targetDir}/node_sqlite3.node`,
+          `build/node_sqlite3.node`,
+        );
+      } 
+    } catch (error) {
+      console.log("[warn] Could not copy node_sqlite to build");
+      console.log(error)
     }
```
maxxrdrgz, line 238: try catch is needed here, else I get an error on darwin-arm64 when installing dependencies for the first time

sestinj: @maxxrdrgz is this happening because the build directory doesn't exist, or is there some other error? We might just need to add a line to remove/re-add the build dir at the very top of build.js, right under the "// Clean slate" comment

extensions/vscode/scripts/prepackage-cross-platform.js
```diff
@@ -111,20 +114,22 @@ function isWin() {
...
   // *** esbuild ***
-  await installNodeModuleInTempDirAndCopyToCurrent(
-    "esbuild@0.17.19",
-    "@esbuild",
-  );
+  // await installNodeModuleInTempDirAndCopyToCurrent(
+  //   "esbuild@0.17.19",
+  //   "@esbuild",
+  // );
+  await downloadEsbuildBinary(target);
```
sestinj, line 125: Did you replace this because the other was broken? As much as possible I'd like to download everything from NPM rather than from raw URLs.

This sounds odd, but the reason is that oftentimes people want to build, but have a locked down network with only access to a mirror of NPM.

sestinj (commented):
Everything looks great! I left a couple of comments, just let me know what your thoughts are, but it sounds like this is working, and it looks quite good to me

binary/build.js
maxxrdrgz: looks like someone fixed this d124aa0. Although I'm getting a different error when installing dependencies for the first time on darwin-arm64 which I can make a ticket for.

extensions/vscode/scripts/prepackage-cross-platform.js
sestinj: @maxxrdrgz I think this is good to merge, but if you get the chance, I linked to this question in Discord and would be curious to know if you'd run into a bug with the NPM way of doing it

sestinj (approved):

PR #1475, opened 2024-06-12
Jupiter is not the 9th planet

pzaback:
If the reference to "after Mars," is
desired for some reason, punctuation should
probably be added for clarity. E.g., "planet,
after Mars,", or "planet, following Mars,". Note
also that because this is a FIM example and "the
5th" is part of the sentence to be filled, it
shouldn't be included as part of the completion.

## Description

Fixed a FIM example

## Checklist


fry69:
:heart: the title

sestinj:
@pzaback Commendable work 🫡

sestinj (approved):

PR #1493, opened 2024-06-15
Skip embedding files that do not chunk properly

johnbwang:
## Description

Skip sending a file to the embedding model if any of its chunks are empty. Empty chunks are usually result of chunking a binary blob.

A better solution would be to detect a binary file and skip it or perhaps have an allowlist of extensions, but this is a reasonable work around.

## Checklist

- [x ] The base branch of this PR is `preview`, rather than `main`

### DIFF
--- a/core/indexing/LanceDbIndex.ts
+++ b/core/indexing/LanceDbIndex.ts
@@ -79,13 +79,24 @@ export class LanceDbIndex implements CodebaseIndex {
       const content = contents[i];
       const chunks: Chunk[] = [];
 
+      let hasEmptyChunks = false;
+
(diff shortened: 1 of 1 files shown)


sestinj (approved):

sestinj:
@johnbwang nice work, I think this is going to be the solution to a lot of somewhat opaque problems that people were having. Thanks for the contribution!!

On the note about file extensions, we have a list of disallowed file extensions and folder names [here](https://github.com/continuedev/continue/blob/nate/test-test-test/core/indexing/ignore.ts#L3), so if you have a binary file format in mind that you don't see there, feel free to add it

PR #1494, opened 2024-06-16
CON-240: Updated FileContextProvider, FolderContextProvider and useSubmenuContextProvider to increment display of path parts to ensure unique displays when duplicates exist in the items list

### LINKED ISSUE #1347: [CON-240] Ambiguous @ file entries in the dropdown when they have same folder/filename
### Validations

### Problem

In cases where there are two distinct folders named "util", both of which have a file named "index.ts", there is no way to distinguish in the dropdown which of the two I am selecting:

image

### Solution

We should deduplicate the results by showing their parent paths until any duplicate names are no longer ambiguous. This logic probably belongs here: [https://github.com/continuedev/continue/blob/preview/gui/src/components/mainInput/getSuggestion.ts#L99-L107](https://github.com/continuedev/continue/blob/preview/gui/src/components/mainInput/getSuggestion.ts#L99-L107)

The `getBasename` function may be useful

<sub>From [SyncLinear.com](https://synclinear.com) | [CON-240](https://linear.app/continue/issue/CON-240/ambiguous-file-entries-in-the-dropdown-when-they-have-same)</sub>

scottschluer:
## Description

Fixes #1347: in cases where there are two distinct folders named "util", both of which have a file named "index.ts", there is no way to distinguish in the dropdown which of the two I am selecting.

This PR resolves the issue by grouping file paths by their last two parts and de-duplicating them by incrementing path parts by 1 until it is no longer a duplicate of the others in the group.

## Checklist

- [x ] The base branch of this PR is `preview`, rather than `main`

### DIFF
--- a/core/context/providers/FileContextProvider.ts
+++ b/core/context/providers/FileContextProvider.ts
@@ -5,7 +5,7 @@ import {
   ContextSubmenuItem,
   LoadSubmenuItemsArgs,
 } from "../../index.js";
-import { getBasename, getLastNPathParts } from "../../util/index.js";
+import { getBasename, groupByLastNPathParts, getUniqueFilePath } from "../../util/index.js";
 import { BaseContextProvider } from "../index.js";
 
 const MAX_SUBMENU_ITEMS = 10_000;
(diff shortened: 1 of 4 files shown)


scottschluer:
A couple of notes:

1. Lengthy duplicate folder paths may require expanding the width of the chat window to see the full path.
2. After experimenting with @folder (which had the same issue and is addressed in this PR), I feel like a better user experience might be to ONLY display the path. Right now, because it only displays the name of the folder (but is still sorted by the unseen hierarchy), the sorting makes no sense and you have to hover over it to see the full path.

sestinj (approved):

sestinj:
@scottschluer solid work! You make a good point that maybe we should disambiguate without needing to hover over the item. I'll see how this feels and maybe that will be follow-up work

sestinj:
Hi @scottschluer, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

PR #1498, opened 2024-06-16
fix(core): use `TextDecoderStream` for stream response

### LINKED ISSUE #1492: Non-ASCII response from Gemini is partially broken in VScode extension

- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11 23H2
- Continue: v0.8.40
- IDE: VSCode 1.91.0-insider
- Model: gemini-1.5-pro-latest
```

### Description

# A description of the bug

I'm encountering an issue where responses from Gemini with non-ASCII characters are garbled. This doesn't seem to happen with responses from Groq.

# What you expected to happen

Non-ASCII character responses should be displayed correctly, without any garbled characters or replacement characters (like "�").

# What actually happened

Currently, non-ASCII characters are being replaced with the replacement character "�". This happens consistently.

For example, the following response is affected:

> zh-cn.ts���ァイルの一部ですね。 これは中国語の簡体字で������れ��コードで、音声操作に関するUIのテキストのようです。
> 
> 日本語のメッセージにする場合、どのような文脈で表示されるかを考慮する必要があります。 例えば、以下のように状況を想定して、より自然で適切な日本語���を検討できます。

# Screenshots or videos

screenshot

# Possible solutions

I suspect this is because the buffer is being treated as a string, rather than an ArrayBuffer. Since Gemini responses may contain incomplete Unicode bytes, using a string buffer could be causing the corruption.

solution 1: change `buffer` from string to ArrayBuffer in `streamChatGemini` function

https://github.com/continuedev/continue/blob/aa18568c6096f8f2df7075ff0f1f711f6f75ed01/core/llm/llms/Gemini.ts#L105

solution 2: use [`TextDecoderStream`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoderStream) instead of `TextDecoder` in `streamResponse` function.

https://github.com/continuedev/continue/blob/aa18568c6096f8f2df7075ff0f1f711f6f75ed01/core/llm/stream.ts#L12-L18

```ts
yield* response.body.pipeThrough(new TextDecoderStream());
```

### To reproduce

Ask some questions in the chat panel in Japanese.

### Log output

```Shell
No output during chat response.
```

reosablo:
Closes #1492

## Description

This PR changes the implementation of `streamResponse` function to use `TextDecoderStream` rather than to use `TextDecoder` for each chunk from streaming response.

The current implementation sometimes break the response non-ASCII characters and this PR will fix it. see #1492 for more detail.

## Checklist

### DIFF
--- a/core/llm/stream.ts
+++ b/core/llm/stream.ts
@@ -9,13 +9,9 @@ export async function* streamResponse(
     throw new Error("No response body returned.");
   }
 
-  const stream = response.body as any;
+  const stream = (ReadableStream as any).from(response.body);
 
-  const decoder = new TextDecoder("utf-8");
-
-  for await (const chunk of stream) {
-    yield decoder.decode(chunk, { stream: true, });
-  }
+  yield* stream.pipeThrough(new TextDecoderStream("utf-8"));
 }
 
 function parseDataLine(line: string): any {


reosablo:
Note:

```ts
const stream = (ReadableStream as any).from(response.body);

yield* stream.pipeThrough(new TextDecoderStream("utf-8"));
```

- `response` is not an instance of `globalThis.Response` and `response.body` is not `ReadableStream`. So we need to convert `response.body` into `ReadableStream` to use `TextDecoderStream`.
- TypeScript doesn't recognize that `ReadableStream` has `from` static method so `as any` is required.

sestinj (approved):

sestinj:
Thanks for the contribution! This is quite important : )

vvvlad:
I believe this causes the following error:
ReadableStream.from is not a function

reosablo:
> I believe this causes the following error:
> ReadableStream.from is not a function

I found that `ReadableStream.from()` seems to be only supported in the latest VSCode.

[MDN](https://developer.mozilla.org/docs/Web/API/ReadableStream/from_static) says `ReadableStream.from()` is supported Node.js v20.6.0+ and the VSCode built-in Node.js version seems to be like the bellow.

| VSCode version | Electron version | Node.js version |
|-|-|-|
| [May 2024 (version 1.90)](https://code.visualstudio.com/updates/v1_90) | 29 | 20.9.0 |
| [April 2024 (version 1.89)](https://code.visualstudio.com/updates/v1_89) | 28 | 18.18.2 (`ReadableStream.from()` is NOT available) |
| [March 2024 (version 1.88)](https://code.visualstudio.com/updates/v1_88) | 28 | 18.18.2 (`ReadableStream.from()` is NOT available) |

reosablo:
The bellow code can be a remedy of the error if you want to support older VSCode versions.

```ts
// This triple slash directive doesn't seem to be recognized by TypeScript on compile time so we can't use this instead of `as any`.
// /// <reference lib="DOM.AsyncIterable" />

export async function* streamResponse(
  response: Response,
): AsyncGenerator<string, void, undefined> {
  if (response.status !== 200) {
    throw new Error(await response.text());
  }

  if (!response.body) {
    throw new Error("No response body returned.");
  }

  // Use ReadableSteam constructor instead of Readable.from() function
  const body = response.body;
  const stream = new ReadableStream<Uint8Array>({
    async start(controller) {
      for await (const chunk of body as any) {
        controller.enqueue(chunk);
      }
      controller.close();
    },
  }) as any;

  yield* stream.pipeThrough(new TextDecoderStream("utf-8"));
}
```

PR #1505, opened 2024-06-17
quick fix: ocktokit bug in crawl.ts

nmcostello:
## Description

According to [here](https://github.com/octokit/octokit.js?tab=readme-ov-file#octokitrest-endpoint-methods) and [here](https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/main/src/generated/method-types.ts#L9243), the correct method is `octokit.rest.repos` NOT `octokit.repos`. 

## Checklist


nmcostello:
closed: am idiot

PR #1509, opened 2024-06-18
Project file path in LLM context for highlighted code

tijszwinkels:
## Description

Follow-up to #1407
Adds the project file-path to the LLM for selected/highlighted code, rather than just the filename.

Allow the LLM to understand the selected code in context of @filetree, or to allow the LLM to distinguish between multiple files with the same name (but in a different path).

Screenshot 2024-06-18 at 16 04 21

Notice how the editor just shows the filename, but the LLM sees the project path.

 
## Checklist


sestinj (approved):
Very nice!

PR #1511, opened 2024-06-18
Feat/gemini error handling

pzaback:
## Description

This PR checks that the return has the correct format before attempting to parse it and prints a more helpful error message to the debug console if not. This is an incomplete first step to handling instances where the Gemini streaming API returns something other than the desired response text. In particular, in testing with certain parameters I was getting a partial response and then the Continue chat panel popped up an error message that wasn't very helpful and the partial response disappeared. Troubleshooting revealed it to be that the API was returning FinishReason: Recitation (described in frustrating lack of detail here: https://ai.google.dev/api/rest/v1beta/Candidate , and a potentially broader Gemini issue here: https://issuetracker.google.com/issues/331677495 ), which in turn caused Continue to error out when the expected content text field was missing. A more complete solution would be to ensure coverage of the various ways the API might respond and how Continue should handle each, but for now this is at least a step in that direction.

## Checklist

### DIFF
--- a/core/llm/llms/Gemini.ts
+++ b/core/llm/llms/Gemini.ts
@@ -181,23 +181,30 @@ class Gemini extends BaseLLM {
           data = JSON.parse(part);
         } catch (e) {
           foundIncomplete = true;
-          continue;
+          continue; // yo!
         }
         if (data.error) {
           throw new Error(data.error.message);
         }
-
-        // Incrementally stream the content to make it smoother
-        const content = data.candidates[0].content.parts[0].text;
-        const words = content.split(/(\s+)/);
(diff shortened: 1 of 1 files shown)


core/llm/llms/Gemini.ts
```diff
@@ -181,23 +181,28 @@ class Gemini extends BaseLLM {
           data = JSON.parse(part);
         } catch (e) {
           foundIncomplete = true;
-          continue;
+          continue; // yo!
```
sestinj, line 184: nice

sestinj (approved):

sestinj:
This is great @pzaback ! Thanks for cleaning up after Gemini : )

PR #1512, opened 2024-06-19
Bugfix for /comment slash command which does nothing.

maxxrdrgz:
## Description

- the default behavior of /comment doesn't do anything unless the user adds additional prompt. It'll essentially perform the same as an /edit with no additional instruction.
- edit.ts was ignoring the default input prompt from comment.ts

## Checklist

- [ ] Review code


sestinj (approved):

sestinj:
Thanks for the fix! It passes my review, is that what the checklist item is for? If so I'll go ahead and merge

maxxrdrgz:
@sestinj yep, should be good to go!

PR #1520, opened 2024-06-19
Add stable versions of latest gemini models

pzaback:
## Description

Enables users to specify the latest _stable_ version of gemini 1.5 Pro and Flash. 
See [Gemini API Docs](https://ai.google.dev/gemini-api/docs/models/gemini#model-versions)

## Checklist

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -562,7 +562,9 @@ declare global {
     // Gemini
     | "gemini-pro"
     | "gemini-1.5-pro-latest"
+    | "gemini-1.5-pro"
     | "gemini-1.5-flash-latest"
+    | "gemini-1.5-flash"
     // Mistral
     | "mistral-tiny"
     | "mistral-small"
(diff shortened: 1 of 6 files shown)


Patrick-Erichsen:
@pzaback Great catch! 

Looks you got all of the typings updated. 👍 

My only question for @sestinj is whether we should update the versions in the model selection page to default to these stable versions rather than the latest.

https://github.com/continuedev/continue/blob/main/gui/src/util/modelData.ts#L466-L490

Relevant docs from Gemini:
> **Latest**: Points to the cutting-edge version of the model for a specified generation and variation. The underlying model is updated regularly and might be a preview version. Only exploratory testing apps and prototypes should use this alias.

Although it does look like we're using `latest` for a few other models, e.g. [Codestral](https://github.com/continuedev/continue/blob/main/gui/src/util/modelData.ts#L381-L392).

sestinj:
I tend to prefer latest on the assumption that most users won't think to upgrade/won't always be aware that a newer version has been released. This and I haven't encountered too many actual instabilities

sestinj (approved):

PR #1522, opened 2024-06-20
Multi messages

rootedbox:
## Description

Allows both vscode and non-vscode plugins to receive IPC via tcp / stdio / VSCIpc.

I needed this to integrate my thirdparty app with continue.


sestinj:
@rootedbox I think we're going to end up not wanting to merge this, but I remember it seemed like you might have had an alternative workaround. How do you want to move forward with this?

rootedbox:
@sestinj understood I've closed this out.. I don't think there is a great way for me to interop at this time.. in my app I'll probably just check for a continue config and ask the user if they want to configure my app from the continue config.

PR #1534, opened 2024-06-20
fix(extensions): schema for db ctx provider

Patrick-Erichsen:
## Description

Fixes an invalid schema for the database context provider. Previously it wasn't checking for a `params` property, just a top level `connections` property.

See the relevant docs here: https://docs.continue.dev/customization/context-providers#database-tables

## Checklist


sestinj (approved):

sestinj:
Good catch

PR #1593, opened 2024-06-27
Add IndexingQueue to handle indexing tasks sequentially

priyashpatil:
## Description

- Introduce IndexingQueue class
- Update core.ts to use IndexingQueue

## Checklist


sestinj:
I think we're going to want to do something slightly different here. The biggest concern for me is that the docs indexing won't happen until main indexing has completed, which might force the user to wait some time.

The alternative that I think would be ideal is to keep a "stack" of currently running jobs. They would all run in parallel, but all updates would be attached to a certain "job id" and the last created job id would always be the one displayed in the loading bar, until it was done, at which point the job beneath it, if still progressing, would take over. On hover, it could display the full stack of current jobs.

I think this would require a) updating the IndexingProgressUpdate type to include a "jobId", b) setting that "jobId" to a random UUID at the beginning of any job and making sure that the same ID is used throughout the job, c) writing the basic "stack" logic in the IndexingProgressBar component

I'm going to close this, but the thought process here was really valuable, so thank you for making the draft

Issue #1422, opened 2024-06-05
Continue interfering with `"terminal.integrated.allowChords"` 


y8:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS Sonoma 14.5 (23F79)
- Continue: v0.8.39
- IDE: VSCode 1.89.1 (Universal)
```

### Description

When Continue is enabled disabling chords in integrated terminal using `"terminal.integrated.allowChords": false` settings is not working which breaks `cmd + K` as terminal clear shortcut

### To reproduce

1. Set `"terminal.integrated.allowChords": false` in user preferences
2. Open terminal, press `Cmd\Ctrl + K`
3. Terminal output should be cleared
4. Enable Continue extension
5. Press `Cmd\Ctrl + K`
6. VSCode enters chord mode and waits for next key

### Log output

```Shell
There is no output in continue log
```

sestinj:
I don't think that Continue should be doing anything to interfere with cmd+K (we use cmd+L and cmd+I). Are you certain that this is happening only with the Continue extension installed?

zbarbuto:
I can confirm I was also having this issue only after installing Continue.

As a workaround I added an extra terminal clear keybind in my user keybindings (for the same key combination) which seemed to fix it:

```
  {
    "key": "cmd+k",
    "command": "workbench.action.terminal.clear",
    "when": "terminalFocus"
  },
```

spew:
This is fixed by https://github.com/continuedev/continue/pull/1671

PR #1671, opened 2024-07-05
Fix an issue where CMD+K does not clear the terminal when the terminal has focus

spew:
## Description

On MacOS, ⌘+K is bound, by default, to Terminal:Clear. Without this change ⌘+K does not clear the terminal but instead iniates a chord sequence and waits for the next stroke of the chord.

## Checklist


sestinj (approved):
very good find!

zbarbuto:
I believe this closes #1422

PR #1672, opened 2024-07-05
Change treeSitter to cache the Language objects it loads from wasm

spew:
## Description

Without this change, for a repository with 600 typescript files, the indexer would fail to finish correctly and there would be many of the following errors in the webview console log:

'Unable to load language for file ${path} RuntimeError: table index is out of bounds'

The following bash will create a repo that reproduces the problem:

```lang=bash
current_path="."

for ((i=1; i<=20; i++)); do
    new_folder="folder-$i"
    mkdir -p "$current_path/$new_folder"
    current_path="$current_path/$new_folder"

    for ((a=1; a<=30; a++)); do
      head -c 10000 /dev/urandom | base64 > "$current_path/file-$a.ts"
    done
done
```

## Checklist

### DIFF
--- a/core/util/treeSitter.ts
+++ b/core/util/treeSitter.ts
@@ -93,6 +93,11 @@ export async function getParserForFile(filepath: string) {
   }
 }
 
+// Loading the wasm files to create a Language object is an expensive operation and with
+// sufficient number of files can result in errors, instead keep a map of language name
+// to Language object
+const nameToLanguage = new Map<string, Language>();
+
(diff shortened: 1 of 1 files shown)


sestinj (approved):
Nice!

sestinj:
I can see why the tree-sitter parser wouldn't be happy with /dev/random output, curious if it was a reasonably formatted Typescript file though that initially caused you to find this error?

spew:
It was a user report of a large repository with many typescript files.

PR #1674, opened 2024-07-05
feat: add Quick Actions CodeLens feature

Patrick-Erichsen:
## Description

Adds a new "Quick Actions" affordance. Quick Actions are VSCode Code Lens that allow users to perform common actions on functions and classes, such as explaining how the code behaves or generating a docstring.

## Checklist

## Screenshots

quick-actions-demo-small

## Testing

- Try out both default actions
- Disable the feature in settings, confirm the code lens is removed
- Reenable it, confirm is shows back up
- Try adding a custom action, e.g. 
```json
"experimental": {
    "quickActions": [
      {
        "title": "Detailed explanation",
        "prompt": "Explain the following code in detail, inlcuding all methods and properties.",
        "sendToChat": true
      }
    ]
  }
```
- Confirm the custom command works


extensions/vscode/src/lang-server/codeLens/providers/QuickActionsCodeLensProvider.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    return [explain, comment];
+  }
+
+  async provideCodeLenses(
+    document: vscode.TextDocument,
+  ): Promise<vscode.CodeLens[]> {
+    const editor = vscode.window.activeTextEditor;
+
+    if (!editor) {
```
sestinj, line 109: if (editor.document.uri.fsPath.endsWith("continue_tutorial.py")) {
  return [];
}

because we already have a lot of codelens buttons in that file

extensions/vscode/package.json
```diff
@@ -79,10 +79,15 @@
           "default": true,
           "description": "Show inline suggestion to use the Continue keyboard shortcuts (e.g. \"Cmd/Ctrl L to select code, Cmd/Ctrl I to edit\")."
         },
+        "continue.enableQuickActions": {
+          "type": "boolean",
+          "default": false,
```
sestinj: set to true

docs/docs/walkthroughs/quick-actions.md
```diff
@@ -0,0 +1,92 @@
+---
```
sestinj, line 1: Add full front-matter to improve search like in other files, e.g. codebase-embeddings.md, which has this:
```
---
title: Codebase Retrieval
description: Talk to your codebase
keywords: [talk, embeddings, reranker, codebase, experimental]
---
```

docs/docs/walkthroughs/quick-actions.md
```diff
@@ -0,0 +1,92 @@
...
+toc_max_heading_level: 5
+---
+
+# Quick Actions (experimental)
```
sestinj:
```
# Quick Actions (experimental, VS Code only)
```

extensions/vscode/src/commands.ts
```diff
@@ -281,12 +348,14 @@ const commandsMap: (
         "Write comments for this code. Do not change anything about the code itself.",
       );
     },
-    "continue.writeDocstringForCode": async () => {
+    "continue.writeDocstringForCode": async (range?: vscode.Range) => {
```
sestinj: Adding the range parameter seems to have broken the right click context menu action (https://github.com/continuedev/continue/blob/Patrick-Erichsen-pe/codelens/extensions/vscode/package.json#L215-L221)

Solution is probably to keep them as separate commands

extensions/vscode/src/lang-server/codeLens/providers/QuickActionsCodeLensProvider.ts
```diff
@@ -0,0 +1,142 @@
...
+   * - Classes
+   *
+   * Used to filter symbols when providing Quick Actions.
+   */
+  static quickActionSymbolKinds = [
+    vscode.SymbolKind.Function,
+    vscode.SymbolKind.Class,
+  ];
+
+  customQuickActionsConfig?: ExperimentalConfig["quickActions"];
```
sestinj: Later on we'll move this out of ExperimentalConfig, and then these types will need to be changed to either QuickActionsConfig[] or SerializedContinueConfig["quickActions"]

docs/docs/walkthroughs/quick-actions.md
```diff
@@ -0,0 +1,92 @@
...
+
+Quick Actions automate repetitive tasks and streamline your development workflow. Configure custom actions to execute complex operations with a single click.
+
+![Quick actions example](/img/quick-actions-demo.gif)
+
+## How It Works
+
+:::info[Note]
+Quick Actions are currently only available in VSCode.
```
sestinj: ...for languages for which you have installed the LSP extension (e.g. I don't have rust analyzer installed, so they don't show up for rust functions/classes)

sestinj (commented):
Other than the context menu docstring action being broken these are just edge cases and nitpicks

Future thoughts:
- I think these buttons will tempt users to speed-run the docstrings for an entire file and click the buttons all at once. Currently we only support a single diff stream at one time in a file. It’s far out of scope for this PR though
- I think I'm pretty likely personally to want to configure per language, but we can wait to at least get first word back from VS Code

extensions/vscode/src/commands.ts
```diff
@@ -244,6 +268,49 @@ const commandsMap: (
...
+    },
+    "continue.customQuickActionStreamInlineEdit": async (
+      prompt: string,
+      range: vscode.Range,
+    ) => {
+      Telemetry.capture("quickAction", {
+        type: "customQuickActionStreamInlineEdit",
+      });
+
+      streamInlineEdit("docstring", prompt, true, range);
```
sestinj: If we don't give the user the option to change it in config, then I think `onlyOneInsertion` should default to false.

Otherwise anything like "optimize", "fix", "improve variable names", "add comments", "make more concise", etc... won't work

Patrick-Erichsen: Makes sense, agreed.

extensions/vscode/src/commands.ts
Patrick-Erichsen: Interesting, so what's happening here is that because `writeDocstringForCode` is being invoked from a context menu, it is implicitly receiving a URI for the file that the context menu was invoked on as the first param of the callback. 

This issue mentions this behavior: https://github.com/microsoft/vscode/issues/25716

Good to know.

PR #1677, opened 2024-07-06
Nate/better retrieval

sestinj:


core/context/retrieval/pipelines/RerankerRetrievalPipeline.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (!this.options.reranker) {
+      throw new Error("No reranker provided");
+    }
+
+    let scores: number[] = await this.options.reranker.rerank(input, chunks);
+
+    // Filter out low-scoring results
+    let results = chunks;
+    // let results = chunks.filter(
+    //   (_, i) => scores[i] >= RETRIEVAL_PARAMS.rerankThreshold,
```
AnnoyingTechnology, line 37: This has been commented out, is there a reason for that ? 
I see the threshold is set to 0.3, which seems to high. 
Setting it to 0.1 and de-commenting this section would at least remove chucks that are totally irrelevant. 

Having the `rerankThreshold` configurable in the config.conf's reranker section would also be better than a hard-coded value.

Issue #1693, opened 2024-07-09
Typo in doc for section 'Codebase retrieval/Voyage AI'


Laxenade:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Hi, while reading the doc, there seems to be a typo in [this part](https://docs.continue.dev/walkthroughs/codebase-embeddings#voyage-ai). Should the provider be `voyage` instead of `openai`?

### Solution

Should be a quick change in these two places?

https://github.com/continuedev/continue/blob/c93f530abd0d64300c4d45a8ad8da9ee996f03f8/docs/docs/setup/configuration.md?plain=1#L74

https://github.com/continuedev/continue/blob/c93f530abd0d64300c4d45a8ad8da9ee996f03f8/docs/docs/walkthroughs/codebase-embeddings.md?plain=1#L117

sestinj:
Hi! This isn't a typo—it is called "openai" because Voyage uses the OpenAI API format, and we switch the API Base. I can see how this would be confusing though, and perhaps we'll make a dedicated "voyage" provider

PR #1694, opened 2024-07-09
docs: Fix typos related to voyage AI

Laxenade:
## Description

Fixing a few typos on Voyage AI being used with the wrong provider. I couldn't find any usage in code where the provider is used/validated when the model "voyage-code-2" is being configured. Feel free to close this issue if you don't think it's important enough.

This addresses: https://github.com/continuedev/continue/issues/1693

## Checklist

## Screenshots

N/A

## Testing

N/A

### DIFF
--- a/docs/docs/setup/configuration.md
+++ b/docs/docs/setup/configuration.md
@@ -71,7 +71,7 @@ This setup uses Claude 3.5 Sonnet for chatting, Codestral for autocomplete, and 
     "apiKey": "[CODESTRAL_API_KEY]"
   },
   "embeddingsProvider": {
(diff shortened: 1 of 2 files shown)


sestinj:
@Laxenade thanks for helping out here! These are actually correct—we use "provider" to indicate the format of the API call, which in Voyage's case is the same as OpenAI, hence the "openai" provider with custom "apiBase". I don't think you're wrong though that this is confusing. What would probably be a good outcome here would be making a dedicated Voyage AI embeddings provider.

Closing this for now, but if you would have interest in contributing this, let me know!

PR #1698, opened 2024-07-09
Doug/review merged

dschonholtz:
I am closing this down. 
I may be pulling some pieces of it into separate PRs. And I may re-open it when the PR is ready. Lot's of work required first.

### What this does:
1. Modifies the existing review pipeline such that for any review, it doesn't just review the diff, but also reviews the references that diff touches.

### Currently in progress:
1. Updating references to be code blocks around the reference rather than the reference itself. 
    - Three potential approaches. 1. Get the surrounding function if applicable, 2. Get the surrounding lines of code with some new? utility function that pads a code range func padCodeRange(range, before: int, after: int) and returns the codeRangeWithContents. 3. Do both, so if there is no surrounding function we still pad the code block enough that LLM can understand the context the reference is used in.
    - Reasoning: I am not sure that the references that are currently fetched are actually useful. This is because we only get the one line that reference is used. We may want to modify this to get a code block around the modified reference, or to get that files diff if applicable? Or both? I think the initial naive approach here of getting a few lines before and maybe 10 lines afterwards is likely to be a good starting point.
    - This now appears to work, but I have only really tested it the one time. 
    - For some reason I have enough internet to edit this, but not to git push :( 
    - This is done pending further testing.
2. Structured output from LLM and parsing of that structured output
    - The StructuredLlmReviewOutput we previously defined isn't quite sufficient.
    - status should have an added error field, we want the user to have a different icon for an error talking to the LLM, that we currently handle with the red x, and a code review which says that something can be improved. This means we need a warning icon or something similar to differentiate.
    - This struct also won't work because each review item in a file should have it's own category and comments. The whole review for a file will have many individual critiques, each with their own category.
    - Testing this. I think I have import problems, and I need to go back to basics on making sure I understand how the IDE works 
    - Ok it is time for a break... But I am increasingly convinced that what I am currently doing with respect to importing extension specific logic into core is absolutely horrifying. I am realizing I do not fully understand how the IDE abstraction layer works and I need to move how I request symbol definitions around. I am confused because I am now unsure how this ever worked.
 
New:
enum reviewCategories {
yada | nada
}

interface ReviewPart {
    comment: string,
    category: reviewCategory
}

interface StructuredLlmReviewOutput {
    filepath: string,
    status: "good" | "bad" | "error",
    reviewParts: reviewPart[]
}

Original:
interface StructuredLlmReviewOutput {
  filepath: string;
  status: "good" | "bad";
  categories: "aoifoja" | "iaojsfio";
  comments: string;
}

### What still needs to be done:
1. General clean up. This has a lot of debugging logic in it, and some logic for handling some local environment issues that I have now holistically fixed. That env specific logic can now be removed.
4. Do the holistic UI/UX update
5. Review where the code lives. I'm guessing I have some code that ideally lives in a different config file.

## Interfaces:

interface StructuredLlmReviewOutput {
  filepath: string;
  status: "good" | "bad";
  categories: "aoifoja" | "iaojsfio";
  comments: string;
}

export default function claudePrompt(
  filepath: string,
  rifs: RangeInFileWithContents[], # dependencies
  diff: string, # original diff for this file.
): Promise<StructuredLlmReviewOutput> {

# Draft Notes

These notes are largely for myself as I kept hitting random problems and I didn't want to lose the thread on any of them. Feel free to take a look bu

## Random dev problems...
1. Free trial stuff with continue continues to crop up in weird and annoying ways. So I spent the time to switch everything over to API keys and use the recommended set up for the best experience. ✅
5. After pasting in the anthropic key, when I try to use it I just get a 400 error. So I am currently stuck with 4o? I copy and pasted from the example in docs and from my api key. This is a new api key on a new org and I haven't tested it yet. Might be user error. ✅
    - This was a payment config thing in the anthropic console

### LanceDB not working making indexing impossible in dev

No indexing, no code review so this is important.

6. Source error: Error: Cannot find module '@lancedb/vectordb-darwin-arm64'... This was after running the install all dependencies check... I assume this is just a dependency mismatch.  ✅
7. Node seems to think that I am on x86 arch? Which may explain the above error and some of the previous errors I was seeing. It won't let me install lance for arm. ✅

What is going on here?

the prepackage.js script runs. It then seems to think I'm on an x86 machine. Therefore, when it attempts to install packages, if I have manually installed arm packages to override it's logic it throws a fit, otherwise it installs accordingly on x86. When I run uname or other arch detection cli commands they return as I would expect. However, LLMs seem to think that I must be running node under rosetta 2 somewhere and/or have another version of node running. During the script I do see a different node version get run. I haven't tracked down where that version is coming from, it isn't managed by nvm, but I'm guessing that version of node is run as x86 resulting in all of these issues. 

8. When trying to debug lanceDB, now when I try to install all dependencies I am given a variety of errors seemingly unrelated to any of the code I have touched.

Specifically:

```
> build
> tsc && vite build

../core/core.ts:332:37 - error TS2339: Property 'content' does not exist on type 'ChatMessage | PromptLog'.
  Property 'content' does not exist on type 'PromptLog'.

332         yield { content: next.value.content };
                                        ~~~~~~~

../extensions/vscode/src/stubs/remoteConfig.ts:77:14 - error TS2339: Property 'canParse' does not exist on type '{ new (url: string | URL, base?: string | URL): URL; prototype: URL; createObjectURL(obj: Blob | MediaSource): string; revokeObjectURL(url: string): void; }'.

77     if (!URL.canParse(this.remoteConfigServerUrl)) {
                ~~~~~~~~

../extensions/vscode/src/stubs/remoteConfig.ts:94:21 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(intervalId: string | number | Timeout): void', gave the following error.
    Argument of type 'Timer' is not assignable to parameter of type 'string | number | Timeout'.
      Property '[Symbol.dispose]' is missing in type 'Timer' but required in type 'Timeout'.
  Overload 2 of 2, '(id: number): void', gave the following error.
    Argument of type 'Timer' is not assignable to parameter of type 'number'.

94       clearInterval(this.syncInterval);
                       ~~~~~~~~~~~~~~~~~

  node_modules/@types/node/timers.d.ts:130:17
    130                 [Symbol.dispose](): void;
                        ~~~~~~~~~~~~~~~~
    '[Symbol.dispose]' is declared here.


Found 3 errors in 2 files.

Errors  Files
     1  ../core/core.ts:332
     2  ../extensions/vscode/src/stubs/remoteConfig.ts:77

```

These seem to be legitimate tsc errors? I'm not sure why they are only cropping up now.
I fixed these tsc errors. I am not sure why they are broken. As they seem to be breaking for the first time now. But the errors seemed to make sense so I updated them. After fixing them I was back to the x86 errors ✅

13. I know the codebase is indexed now... But I still can't seem to get the bot to search in prod. When I chat with codebase it just uses the current active file. I'll tried turning off that experimental feature and seeing if that impacts it, but it does not, it just doesn't have any context at all.

I rewrote the below bullet as it was barely english and written in a vaguely sleep deprived haze the first time around.

14. To attempt to be thorough and reset potentially erroneously installed x86 binaries I ran the scripts/uninstall.js script. But now when I attempt to run the install all dependencies task it complains about not knowing about various packages it should  know about. (vscode is an example.) ~~I think this is where I turn in. It's getting late. If I'm feeling it maybe tomorrow I'll attempt to get these packages installed again and find that node install that appears to love running as x86.~~ Nope, I found the fix immediately after typing that.

This is a bit of a journey, the below block fixes all of the arm vs x86 problems and I also outline how I fixed the problem of dependencies not being found after running install after running uninstall.

AHAH!  /usr/local/bin/node was x86! 
In theory, this should not matter. Nvm is configured in .nvmrc, right? But when you run the install-all-dependencies.sh script, it is running as a shell script so it doesn't have my nvm configuration sourced from my .zshrc file. So when running that script, it used my system node, the x86 version resulting in pulling down x86 node binaries wherever applicable. It may make sense to modify the install dependency script to explicitly reference a node version from nvm as on my machine even though the node version used is now arm, it still isn't the node process that is used elsewhere in the project. This may still create subtle issues downstream.

To fix this, I backed that binary up just in case, and then simlinked /usr/local/bin/node to another node version I had on my machine that was arm64. 

Once that was done, I could go back to the previous bug above where vscode and many other packages were not being installed. This was simple when I looked through the install-all-dependencies.sh script. The GUI package was relying on vscode and some other packages that were only installed in the later step extensions/vscode. To fix this, I ran the parts of the script by hand and out of order so I installed and linked the vscode packages and then went back to the GUI packages. This fixed the problem.

I do not know the full dependency chain there, but there is a chance you may want to run the vscode install before GUI more generally in that script? I haven't dug further into it. ✅

15. Various hot reloading problems
    - These are problems for a later date. But I get told it can't find bootstrap if I just try to hot reload, I also don't quite understand which things hot reload vs not when you click the restart button in the running icon for vscode. For now I am just killing all of the processes and re-running all of them with the debug vscode button

## Random unrelated thoughts....
1. This is not how I would normally structure a pull request description...
3. I did not know about the cmd+shift+R and wouldn't have if you guys hadn't mentioned it today. That is REALLY nice. I wonder if there is a way to provide a more obvious user affordance and/or detect errors as they occur or to give a button the user can click that says cmd+shift+R
5. I have been poking at this for a long time at this point, and I still have barely spent any real effort on the core feature. It is all tsc, anthropic, x86 bugs just so I can use the thing reliably enough so I can actually start debugging. I would hope in practice after a couple days all of this will be ironed out.
6. Wow, the plane constantly getting and dropping wifi is not a great workflow when trying to debug an LLM workflow

### Notes

These were some notes before I updated this PR. They are now fairly out of date.

Ok this code was pretty useless. For some reason the context switching with pair programming always has crippled my brain and I maybe should have thought about how to handle that beforehand. I took a couple hours to have dinner and reset. Regardless of outcome, I think it is probably worth making this code actually useful.

I also really should have gotten up on a whiteboard and sketched out my plan before I started trying to write code and how that interacts with the existing code and existing review process. I kinda tried to define what tools were useful, but didn't look at the inputs and outputs to the current review process which would have grounded all of my logic. 

A couple notes...

This util function is not useful. The real input should be the filepath, and the current diff, and the output should the dependency code blocks for that diff. Then for each diff we will pull in the code blocks and add those to our file to be reviewed.

Instead of attempting to test the code in the activate function... I should just update the review code so when we trigger a review we call this new util function. We can ignore the result for right now as long as I can set a breakpoint and debug it. 

This also points out a disconnect with what I was building and what was actually needed.

Currently, the review pipeline takes in a diff, reviews that diff, end of story.

What we need to do, is take in the diff, extract the line numbers in that file from the diff.
Then we need to get that rangeInFile, and then pass that to the getDefinitionsFromFileRange util function (new).
This function also is too complicated right now... It can be much simpler. 

For each of those definitions we need to pass them to the prompt builder.
Then we can get into UI/UX, LLM interactions and more...

## Checklist

## Screenshots

## Testing

### DIFF
--- a/core/review/parseDiff.ts
+++ b/core/review/parseDiff.ts
@@ -32,37 +32,57 @@ export function getChangedFiles(diff: string): string[] {
 }
 
 export async function extractUniqueReferences(
-  diff: string,
+  fileDiff: string,
+  filepath: string,
 ): Promise<RangeInFileWithContents[]> {
+  console.log("Extracting unique references...");
   const uniqueReferences = new Map<string, RangeInFileWithContents>();
   const ideUtils = new VsCodeIdeUtils();
-  const diffPerFile = getDiffPerFile(diff);
+  // TODO - Everything up to this point is working. This function should probably not take in a diff, 
+  // but take in the file and line info we calculated before calling this function
+  // We already have calculated the diff per file stuff we don't need to do it again here.
+  // const diffPerFile = getDiffPerFile(diff); 
+  // console.log("Diff per file:", diffPerFile);
 
-  for (const [filepath, fileDiff] of Object.entries(diffPerFile)) {
+  // for (const [filepath, diff] of Object.entries(diffPerFile)) {
+    console.log('Diff output for filepath:', filepath);
+    console.log('File diff:', fileDiff);
     const range = getDiffRange(fileDiff);
-    if (!range) continue;
+    console.log('Range:', range);
 
-    const rif: RangeInFile = { filepath, range };
-
+  // Handle the case where range is null
+   if (range === null) {
+    console.log('No valid range found in the diff. Skipping reference extraction.');
+    return []; // Return an empty array as no references can be extracted
+  }
+    
     for (let line = range.start.line; line <= range.end.line; line++) {
-      const definitions = await getDefinitionsForLine(filepath, line);
+      const references = await getReferencesForLine(filepath, line);
+      console.log(`References for line ${line}:`, references);
+      console.log('Type of references:', typeof references);
+      console.log('Is references an array?', Array.isArray(references));
+    
 
-      for (const definition of definitions) {
-        const { filepath: defFilepath, range: defRange } = definition;
-        const content = await ideUtils.readRangeInFile(defFilepath, defRange);
+      for (let i = 0; i < references.length; i++) {
+        const reference = references[i];
+        const { filepath: refFilepath, range: refRange } = reference;
+        const content = await ideUtils.readRangeInFile(refFilepath, refRange);
+        if (!content) continue; // Skip if no content could be read
+        console.log('Reference:', { filepath: refFilepath, range: refRange, content });
 
-        const key = `${defFilepath}:${defRange.start.line}:${defRange.start.character}-${defRange.end.line}:${defRange.end.character}`;
+        // Create a unique key for each reference
+        const key = `${refFilepath}:${refRange.start.line}:${refRange.start.character}-${refRange.end.line}:${refRange.end.character}`;
         if (!uniqueReferences.has(key)) {
           uniqueReferences.set(key, {
-            filepath: defFilepath,
-            range: defRange,
+            filepath: refFilepath,
+            range: refRange,
             contents: content,
           });
         }
       }
     }
-  }
 
+  // Convert the Map to an array and return
   return Array.from(uniqueReferences.values());
 }
 
@@ -92,15 +112,15 @@ function getDiffRange(fileDiff: string): Range | null {
   };
 }
 
-async function getDefinitionsForLine(
+async function getReferencesForLine(
   filepath: string,
   line: number,
 ): Promise<RangeInFile[]> {
-  const gotoInput = {
+  const referenceInput = {
     uri: filepath,
     line: line,
     character: 0,
-    name: "vscode.executeDefinitionProvider" as const,
+    name: "vscode.executeReferenceProvider" as const,
   };
-  return await executeGotoProvider(gotoInput);
+  return await executeGotoProvider(referenceInput);
 }
--- a/core/review/parseDiff.ts
+++ b/core/review/parseDiff.ts
@@ -38,49 +38,48 @@ export async function extractUniqueReferences(
   console.log("Extracting unique references...");
   const uniqueReferences = new Map<string, RangeInFileWithContents>();
   const ideUtils = new VsCodeIdeUtils();
-  // TODO - Everything up to this point is working. This function should probably not take in a diff, 
-  // but take in the file and line info we calculated before calling this function
-  // We already have calculated the diff per file stuff we don't need to do it again here.
-  // const diffPerFile = getDiffPerFile(diff); 
-  // console.log("Diff per file:", diffPerFile);
 
-  // for (const [filepath, diff] of Object.entries(diffPerFile)) {
-    console.log('Diff output for filepath:', filepath);
-    console.log('File diff:', fileDiff);
-    const range = getDiffRange(fileDiff);
-    console.log('Range:', range);
+  console.log("Diff output for filepath:", filepath);
+  console.log("File diff:", fileDiff);
+  const range = getDiffRange(fileDiff);
+  console.log("Range:", range);
 
   // Handle the case where range is null
-   if (range === null) {
-    console.log('No valid range found in the diff. Skipping reference extraction.');
+  if (range === null) {
+    console.log(
+      "No valid range found in the diff. Skipping reference extraction.",
+    );
     return []; // Return an empty array as no references can be extracted
   }
-    
-    for (let line = range.start.line; line <= range.end.line; line++) {
-      const references = await getReferencesForLine(filepath, line);
-      console.log(`References for line ${line}:`, references);
-      console.log('Type of references:', typeof references);
-      console.log('Is references an array?', Array.isArray(references));
-    
 
-      for (let i = 0; i < references.length; i++) {
-        const reference = references[i];
-        const { filepath: refFilepath, range: refRange } = reference;
-        const content = await ideUtils.readRangeInFile(refFilepath, refRange);
-        if (!content) continue; // Skip if no content could be read
-        console.log('Reference:', { filepath: refFilepath, range: refRange, content });
+  for (let line = range.start.line; line <= range.end.line; line++) {
+    const references = await getReferencesForLine(filepath, line);
+    console.log(`References for line ${line}:`, references);
+    console.log("Type of references:", typeof references);
+    console.log("Is references an array?", Array.isArray(references));
 
-        // Create a unique key for each reference
-        const key = `${refFilepath}:${refRange.start.line}:${refRange.start.character}-${refRange.end.line}:${refRange.end.character}`;
-        if (!uniqueReferences.has(key)) {
-          uniqueReferences.set(key, {
-            filepath: refFilepath,
-            range: refRange,
-            contents: content,
-          });
-        }
+    for (let i = 0; i < references.length; i++) {
+      const reference = references[i];
+      const { filepath: refFilepath, range: refRange } = reference;
+      const content = await ideUtils.readRangeInFile(refFilepath, refRange);
+      if (!content) continue; // Skip if no content could be read
+      console.log("Reference:", {
+        filepath: refFilepath,
+        range: refRange,
+        content,
+      });
+
+      // Create a unique key for each reference
+      const key = `${refFilepath}:${refRange.start.line}:${refRange.start.character}-${refRange.end.line}:${refRange.end.character}`;
+      if (!uniqueReferences.has(key)) {
+        uniqueReferences.set(key, {
+          filepath: refFilepath,
+          range: refRange,
+          contents: content,
+        });
       }
     }
+  }
 
   // Convert the Map to an array and return
   return Array.from(uniqueReferences.values());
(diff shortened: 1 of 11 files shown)


core/review/parseDiff.ts
```diff
@@ -1,3 +1,9 @@
+import { executeGotoProvider } from "../../extensions/vscode/src/autocomplete/lsp.js";
+import { VsCodeIdeUtils } from "../../extensions/vscode/src/util/ideUtils.js";
+import { RangeInFile, Range } from "../index.js";
```
dschonholtz, line 3: The linter gets angry about the file extensions here.
I need to do some learning on this to see how this should be managed given this are actually .ts files. and the index file is a index.d.ts file.
Given that elsewhere in the code the linter is ignored, that is probably what I should do here.

core/review/parseDiff.ts
```diff
@@ -24,3 +30,77 @@ export function getChangedFiles(diff: string): string[] {
...
+
+  if (startLine === null || endLine === null) return null;
+
+  return {
+    start: { line: startLine, character: 0 },
+    end: { line: endLine, character: Number.MAX_SAFE_INTEGER },
+  };
+}
+
+async function getDefinitionsForLine(
```
dschonholtz: This needs to be references.

scripts/util/index.js
```diff
@@ -37,6 +39,23 @@ function autodetectPlatformAndArch() {
     s390x: "x64",
     x64: "x64",
   }[process.arch];
+
+  // Additional check for ARM-based Macs
+  if (platform === "darwin" && process.arch === "x64") {
+    try {
+      const { execSync } = require("child_process");
+      const cpuType = execSync("sysctl -n machdep.cpu.brand_string").toString();
+      if (cpuType.includes("Apple M")) {
```
dschonholtz, line 48: This code probably shouldn't be checked in.
I think I have some subtle upstream thing where esbuilt is set up to point to x86 somewhere and that is causing havoc downstream. 

As after I did this, I ultimately ended up with errors about esbuild and arch as well as not finding some of the binaries for the arch. 

I think once I can track that down everything will just work

core/review/prompts.ts
```diff
@@ -10,5 +10,14 @@ export const reviewSystemMessage = [
...
+export const reviewPrompt = `
+Please review the following code changes in the file {{filepath}}:
+
+{{diff}}
+
+Relevant definitions used by this code:
+
+{{definitions}}
+
+Please provide a concise code review, focusing on potential issues, improvements, and best practices.
```
dschonholtz: We'd have to update this to give us some structured data. Maybe system prompt update?

scripts/util/index.js
dschonholtz: This code should definitely be removed. x86 stuff is captured in dev problems. This falls under code clean up

core/review/parseDiff.ts
```diff
@@ -24,3 +30,148 @@ export function getChangedFiles(diff: string): string[] {
...
+    character: 0,
+    name: "vscode.executeReferenceProvider" as const,
+  };
+  return await executeGotoProvider(referenceInput);
+}
+
+export async function padReferences(
+  references: RangeInFile[],
+  linesBefore: number = 5,
+  linesAfter: number = 15,
```
dschonholtz, line 130: I think I need to restructure this as it may be confusing to the reader.
We only use these padding numbers as a fallback if we don't find a parent function or class when doing the symbol mapping. At least should have a doc string or comment here.

core/review/parseDiff.ts
```diff
@@ -24,3 +30,148 @@ export function getChangedFiles(diff: string): string[] {
...
+): Promise<RangeInFileWithContents[]> {
+  const paddedReferences: RangeInFileWithContents[] = [];
+  const ideUtils = new VsCodeIdeUtils();
```
dschonholtz, line 133: I think it would take some non-trivial effort to both understand how this should work with respect to finding parent symbols like this, why it did briefly work, and how it should work.

core/review/prompts.ts
```diff
@@ -1,14 +1,70 @@
...
+The review status should be 'good' only if the response is 'LGTM', and 'bad' if there is any actionable feedback.
+
+Available review categories:
+{{#each categories}}
+- {{this}}
+{{/each}}
+
+Example responses:
+
+1. For a good review (LGTM):
```
dschonholtz, line 20: These prompts are untested as of now. We can roll back the commits if we don't want to use them. 
But this should work for getting the data structure we chatted about.

core/review/review.ts
```diff
@@ -5,17 +5,47 @@ import { CodeReviewOptions, IDE, ILLM } from "..";
...
+  performance = "⚡ Performance",
+  security = "🔒 Security",
+  maintainability = "🔧 Maintainability",
+  errorHandling = "🚨 Error Handling",
+  duplication = "🐑 Duplication",
+  naming = "🏷️ Naming",
+  architecture = "🏗️ Architecture",
+  optimization = "🚀 Optimization",
+  readability = "📖 Readability",
+  typeSafety = "🛡️ Type Safety",
```
dschonholtz, line 48: This is probably too many and the category thing might be hard to display well and might be more noise than signal. We'll see though

core/review/review.ts
```diff
@@ -136,49 +168,149 @@ export class CodeReview {
...
+    const finalSnippets = fillPromptWithSnippets(
+      deduplicatedSnippets,
+      maxSnippetTokens,
+      this.llm.model,
+    );
+
+    // Prepare the definitions content
+    const systemPrompt = this.systemPrompt();
+    const prompt = this.reviewPrompt(filepath, finalSnippets, diff);
+    console.log("Final Prompt: ", prompt);
```
dschonholtz, line 233: This appeared to have symbol based results in it.
I need to test what happens when we don't have top level symbols to make sure the naive padding works fairly well.

scripts/install-dependencies.sh
```diff
@@ -5,6 +5,32 @@
...
 set -e
 
+# Function to log the current Node.js and npm versions
+log_versions() {
+    echo "Using Node.js version: $(node -v)"
+    echo "Using npm version: $(npm -v)"
+}
+
+# Try to load nvm
+export NVM_DIR="$HOME/.nvm"
```
dschonholtz, line 15: I'm not married to this, but it would have made my life much easier if it had attempted to install and use the nvmrc instead of using my global node. 
I also think the logs would be nice here. 
It also will fall back to whatever the system node is if it doesn't have luck with nvm.

core/review/parseDiff.ts
```diff
@@ -1,3 +1,9 @@
+import { executeGotoProvider } from "../../extensions/vscode/src/autocomplete/lsp.js";
+import { VsCodeIdeUtils } from "../../extensions/vscode/src/util/ideUtils.js";
```
dschonholtz, line 2: Pretty sure this is turning into a weird circular import problem and this is a bit of an antipattern compared to how the rest of the repo works. 
It will probably take me a minute to understand how this works and what the proper pattern here is. I could be wrong, because it appeared to be working at one point. 
It would just need some time.

gui/src/pages/review.tsx
```diff
@@ -57,17 +66,37 @@ function FileResult(props: FileHeaderProps) {
...
+                  ]
+                }
+              </ReviewPartHeader>
+              <StyledMarkdownPreview
+                showCodeBorder={true}
+                source={part.comment}
+              />
+            </ReviewSection>
+          ))}
+          <ReviewSection>
```
dschonholtz, line 92: This is definitely going to be ugly and it will need some iteration

dschonholtz:
I am closing this down.
I may be pulling some pieces of it into separate PRs. And I may re-open it when the PR is ready. Lot's of work required first.

Issue #1701, opened 2024-07-10
Text Embeddings Interface (TEI) as a Reranker


LLukas22:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

TEI provides a [`/rerank`](https://huggingface.github.io/text-embeddings-inference/#/Text%20Embeddings%20Inference/rerank) endpoint which could be implemented as one of the rerankers, since there are currently a lot of available open source models that perform quite well such as e.g. [jinaai/jina-reranker-v2-base-multilingual](https://huggingface.co/jinaai/jina-reranker-v2-base-multilingual). 

### Solution

Add a simple wrapper around the defined `/rerank` endpoint.

sestinj:
Definitely a good idea, thanks for pointing this out!

RomneyDa:
This has been added!

PR #1711, opened 2024-07-11
Add `HuggingFaceTEIReranker`

LLukas22:
## Description

See #1701

Adds HugginfaceTEI as a reranker

## Checklist

## Testing

1. Create a TEI reranker server with a reranking model, e.g. [BAAI/bge-reranker-base](https://huggingface.co/BAAI/bge-reranker-base)
2. Add the reranker to the `config.json`, e.g.:
```json
{
  "reranker": {
    "name": "huggingface-tei",
    "params": {
        "apiBase": "http://localhost:8080",
        "truncate": true,
        "truncation_direction": "Right"
    }
  },
}
```

### DIFF
--- a/core/context/rerankers/index.ts
+++ b/core/context/rerankers/index.ts
@@ -3,10 +3,12 @@ import { CohereReranker } from "./cohere.js";
 import { FreeTrialReranker } from "./freeTrial.js";
 import { LLMReranker } from "./llm.js";
 import { VoyageReranker } from "./voyage.js";
+import {HuggingFaceTEIReranker} from "./tei.js"
 
 export const AllRerankers: { [key in RerankerName]: any } = {
   cohere: CohereReranker,
   llm: LLMReranker,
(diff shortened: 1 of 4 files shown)


sestinj:
@LLukas22 this is awesome, thank you! Would you be able to update extensions/vscode/config_schema.json as well so that users get intellisense on the file when editing it?

LLukas22:
Yeah i missed that one, sorry i'll add it tomorrow 😅 

I also noticed, that it's possible to get an empty list of chunks which will result in an error being thrown as TEI expects at least 1 entry in it's payload. But i guess that it something that should be handled in the RerankerPipeline as the reranking step can simply be skipped if we don't have any chunks. 🤔

LLukas22:
Should be good to go 👍

sestinj (approved):

sestinj:
Awesome, thanks!

PR #1746, opened 2024-07-16
Configurable `maxChunkSize` for embedding providers

LLukas22:
## Description

Add `maxChunkSize` to `EmbeddingsProvider` and makes it configurable via `config.json`.
An embedding providers `id` will now also include the `maxChunkSize` to avoid collisions between indices with different chunck lengths if it is not the default. 

## Checklist

## Testing

Set `maxChunkSize` for an embedding provider that does support longer sequences, e.g. : 
```json
"embeddingsProvider": {
    "provider": "huggingface-tei",
    "model": "jina-embeddings-v2-base-code",
    "apiBase": "[URL]",
    "maxChunkSize": 1024
  },
```
Then index a new project.

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -743,6 +743,7 @@ export interface EmbedOptions {
   apiType?: string;
   apiVersion?: string;
   requestOptions?: RequestOptions;
+  maxChunkSize?: number;
 }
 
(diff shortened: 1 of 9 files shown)


Patrick-Erichsen:
Great work, thanks for the contribution!

PR #1785, opened 2024-07-20
Index on file change

joshvera:
## Description

Indexes files on save. Fixes an issue where Continue wasn't reindexing the file and reloading the submenu items on save.

Now you can load a code context with @Class, change the class, then load it again with @Class and the submenu item should be loaded with the latest results from the context provider.

## Checklist

## Testing

1. Open a file with Continue.
2. Add a `console.log("Test")` to the file.
3. Ask Continue how many Test logs are in the file. Should return something like 1.
4. Add `console.log("Test2")` to the file.
5. Ask Continue how many Test logs are in the file. Should return something like 2.


core/protocol/core.ts
```diff
@@ -65,7 +65,7 @@ export type ToCoreFromIdeOrWebviewProtocol = {
     },
     ContextItemWithId[],
   ];
-  "context/loadSubmenuItems": [{ title: string }, ContextSubmenuItem[]];
+  "context/loadSubmenuItems": [{ title: string }, Promise<ContextSubmenuItem[]>];
```
sestinj: I think this can stay the same as everything here is implicitly wrapped in a promise. If I'm overlooking something, let me know

sestinj (commented):
All looks great! Only possible change is I think the Promise<> can be removed

joshvera:
Still having issues with this after installing the extension. I'm going to put it in Draft mode and test it some more.

joshvera:
Works pretty well now!

Patrick-Erichsen:
Great work here @joshvera ! Very useful improvement.

PR #1815, opened 2024-07-23
Add CodeGeeX4 code completion support

ekatiyar:
## Description
Handles #1742

I added support for autocomplete using using the new CodeGeeX4 model.
Following their FIM guide, this change also includes support for context through code snippets.

## Checklist

- [x] The base branch of this PR is `dev`, rather than `main`
- [x] The relevant docs, if any, have been updated or created

## Screenshots

[ "For visual changes, include screenshots." ]

## Testing

Modify config.json with the following:
```json
  "tabAutocompleteModel": {
    "title": "Tab Autocomplete Model",
    "provider": "ollama",
    "model": "codegeex4:{quant}"
  },
```

I'm running the quant 9b-all-q6_K, but should work similarly with other quantizations.

Issue #1742, opened 2024-07-15
Add CodeGeeX4 code completion support


milkowski:
### Validations

- [ ] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

It is new quite universal code LLM with code completion which has prompt template not supported by Continue.

### Solution

The exact FIM template is described here:
https://github.com/THUDM/CodeGeeX4/blob/main/guides/Infilling_guideline.md

michael-wettach:
Actually I tested it today and was able to use codegeex for code completion with Continue and got quite nice results. There may be issues that I have not yet seen, of course. 

One weird thing I noticed: after accepting a suggestion with tab, another suggestion appears *in the same code line* and even without a space after the last generated word. However, I was able to reproduce that behaviour with codellama, so it does not seem to be related to codegeex alone.

PR #1827, opened 2024-07-24
Add Missing Tokens for CodeGeeX4 AutoComplete

ekatiyar:
## Description

Follow up on #1815, Handles #1742

Previous attempts at using the normal pipe char with CodeGeeX FIM tokens resulted in empty completion output. However, this was due to `<|user|>` and `<|assistant|>` tokens being necessary for completion. The earlier working change omitted these tokens, and its functionality through using the deepseek pipe char was merely coincidental. This change now properly follows the CodeGeeX infill guides and should match what the model expects.

In addition to `<|user|>` and `<|assistant|>`, the `<|endoftext|>` token was added as a stop token.

## Checklist

## Screenshots

## Testing

Same as #1815

### DIFF
--- a/core/autocomplete/templates.ts
+++ b/core/autocomplete/templates.ts
@@ -168,20 +168,20 @@ const codegeexFimTemplate: AutocompleteTemplate = {
       ...snippets.map((snippet) => snippet.filepath),
       filepath,
     ]);
-    const baseTemplate = `###PATH:${relativePaths[relativePaths.length - 1]}\n###LANGUAGE:\n###MODE:\n<｜code_suffix｜>${suffix}<｜code_prefix｜>${prefix}<｜code_middle｜>`;
+    const baseTemplate = `###PATH:${relativePaths[relativePaths.length - 1]}\n###LANGUAGE:\n###MODE:BLOCK\n<|code_suffix|>${suffix}<|code_prefix|>${prefix}<|code_middle|>`;
     if (snippets.length == 0)
     {
-      return baseTemplate;
+      return `<|user|>\n${baseTemplate}<|assistant|>\n`;
     }
     const references =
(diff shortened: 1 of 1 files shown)


ekatiyar:
Shoutout to @TheLapinMalin for helping to get this fixed and working correctly!

core/autocomplete/templates.ts
```diff
@@ -168,20 +168,20 @@ const codegeexFimTemplate: AutocompleteTemplate = {
       ...snippets.map((snippet) => snippet.filepath),
       filepath,
     ]);
-    const baseTemplate = `###PATH:${relativePaths[relativePaths.length - 1]}\n###LANGUAGE:\n###MODE:\n<｜code_suffix｜>${suffix}<｜code_prefix｜>${prefix}<｜code_middle｜>`;
+    const baseTemplate = `###PATH:${relativePaths[relativePaths.length - 1]}\n###LANGUAGE:\n###MODE:\n<|code_suffix|>${suffix}<|code_prefix|>${prefix}<|code_middle|>`;
```
TheLapinMalin: I think you can now pass the LANGUAGE name as per this commit: https://github.com/continuedev/continue/commit/8d83472792bf41ac7207f8a09514c8d3e82d5657

And we might want to set MODE to "BLOCK", since Continue can handle multi-line completions?

ekatiyar: The AutocompleteTemplate interface currently doesn't provide `language` as a variable available to pre-defined FIM templates that use its template function, but agree that it would be nice to add; however, adding that functionality is outside the scope of this change. @sestinj is that something you guys plan to add?

Good point on setting the mode to `BLOCK`. The CodeGeeX docs mention that this is the default so there should be no change in behavior, but always good to be explicit!

PR #1857, opened 2024-07-28
Disable autocomplete in config.json

FozzieHi:
## Description

Currently, if you edit the `config.json` in an editor with autocomplete enabled, apiKeys may be included in the prompt to the autocomplete provider. Since it is fairly common to edit the `config.json` in editor, this PR replaces configured API keys with the string "SECRET".

### Example
Before:
```
{
  "models": [
    {
      "title": "Claude 3.5 Sonnet",
      "provider": "anthropic",
      "model": "claude-3-5-sonnet-20240620",
      "apiKey": "[ANTHROPIC_API_KEY]"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Codestral",
    "provider": "mistral",
    "model": "codestral-latest",
    "apiKey": "[CODESTRAL_API_KEY]"
  },
  "embeddingsProvider": {
    "provider": "openai",
    "model": "voyage-code-2",
    "apiBase": "https://api.voyageai.com/v1/",
    "apiKey": "[VOYAGE_AI_API_KEY]"
  },
  "reranker": {
    "name": "voyage",
    "params": {
      "apiKey": "[VOYAGE_AI_API_KEY]"
    }
  }
}
```

After:
```
{
  "models": [
    {
      "title": "Claude 3.5 Sonnet",
      "provider": "anthropic",
      "model": "claude-3-5-sonnet-20240620",
      "apiKey": "SECRET"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Codestral",
    "provider": "mistral",
    "model": "codestral-latest",
    "apiKey": "SECRET"
  },
  "embeddingsProvider": {
    "provider": "openai",
    "model": "voyage-code-2",
    "apiBase": "https://api.voyageai.com/v1/",
    "apiKey": "SECRET"
  },
  "reranker": {
    "name": "voyage",
    "params": {
      "apiKey": "SECRET"
    }
  }
}
```

As this is my first contribution, feel free to let me know if any improvements could be made. For example, is this the correct place for the replacement to be made? Is the ConfigHandler the correct file for the `apiKeys()` function? Do we need to traverse like this or could we just get them from the known typed values? Could we just work on the serialised config instead?

I'm also happy to add any tests if needed.

## Checklist

### DIFF
--- a/core/autocomplete/completionProvider.ts
+++ b/core/autocomplete/completionProvider.ts
@@ -52,7 +52,7 @@ import { getTemplateForModel } from "./templates.js";
 import { GeneratorReuseManager } from "./util.js";
 // @prettier-ignore
 import Handlebars from "handlebars";
-import { getContinueGlobalPath } from "../util/paths.js";
+import { getConfigJsonPath } from "../util/paths.js";
 
 export interface AutocompleteInput {
   completionId: string;
@@ -249,6 +249,11 @@ export class CompletionProvider {
         ...DEFAULT_AUTOCOMPLETE_OPTS,
         ...config.tabAutocompleteOptions,
       };
+
+      // Check if we're in the continue config.json file
+      if (path.relative(getConfigJsonPath(), input.filepath) === "") {
+        return undefined;
+      }
 
       // Check whether autocomplete is disabled for this file
       if (options.disableInFiles) {
@@ -535,16 +540,6 @@ export class CompletionProvider {
     if (manuallyPassPrefix) {
       prefix = manuallyPassPrefix;
       suffix = "";
-    }
-
-    const relativePath = path.relative(getContinueGlobalPath(), filepath);
-    // Check if we're in the continue directory
-    if (!relativePath.startsWith("..") && !path.isAbsolute(relativePath)) {
-      const apiKeys = await this.configHandler.apiKeys();
-      apiKeys.forEach(key => {
-        prefix = prefix.replace(key, "SECRET");
-        suffix = suffix.replace(key, "SECRET");
-      });
     }
 
     // Template prompt
--- a/core/autocomplete/completionProvider.ts
+++ b/core/autocomplete/completionProvider.ts
@@ -52,6 +52,7 @@ import { getTemplateForModel } from "./templates.js";
 import { GeneratorReuseManager } from "./util.js";
 // @prettier-ignore
 import Handlebars from "handlebars";
+import { getContinueGlobalPath } from "../util/paths.js";
 
 export interface AutocompleteInput {
   completionId: string;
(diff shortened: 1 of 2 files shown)


sestinj:
@FozzieHi I like the idea here, it's definitely pretty important to avoid API keys. I think it's going to be really difficult to handle all possible cases because sometimes there could also be an API key in a context provider's params. Maybe we ought to just ignore config.json for autocomplete altogether. Would you be interested in making this change?

FozzieHi:
@sestinj There's been a few messages on the Discord on this. I can see the pros and cons of both approaches.

By searching for the API keys directly, you might not necessarily catch every single case but you'll catch the majority of them. You'll also catch cases where the keys are stored outside of the `config.json`, for example in a variable in the `config.ts`. By disabling autocomplete in the `config.json`, you'll catch all cases of API keys in the JSON, but then you'll lose the usefulness of the autocomplete in that file. You also won't catch the API keys in other files like the `config.ts`, unless you disable autocomplete in all files in the `.continue` directory, and then you're reducing the usefulness of autocomplete even more.

So instead of force disabling it in the `.configure` directory,  should we make disabling it in the `.continue` directory a config option which is disabled by default, and document the risk of enabling it in the docs? We could either set a default value for `disableInFiles`, or create a new config option. The risk with setting a default value in `disableInFiles` is that users may inadvertently turn it back on again by configuring their own files and not including the continue directory in the array. Although we could document that as well.

Patrick-Erichsen:
Hi @FozzieHi , thanks for taking the initiative on this!

You made some good points on the pros/cons of both approaches but I would probably lean towards @sestinj 's solution as well - just disabling autocomplete in the Continue folder.

The more robust/long-term solution here is likely to identify any API key in any prompt and strip it out. 

In the short run, I lean towards just disabling in the Continue folder for a few reasons:
- Not much value in autocomplete in a config file
- Easy to implement/maintain

FozzieHi:
@Patrick-Erichsen Great, that's what I was leaning to as well. Have you thought about what I said above about how we'd want to implement this? Either by putting it in as a default entry in the `disableInFiles` option, or creating a new boolean option like `disableInContinueDirectory`?

sestinj:
I think completions should be forcefully turned for config.json, but not for config.ts or other files in the .continue directory. While there's a chance that a few people might want to alter this behavior, I think the urgency wouldn't rise to the level where it would be worth the additional config complexity + docs + risk of reverting the change like you said

FozzieHi:
@sestinj Done, I couldn't find any tests for the `disableInFiles` functionality, so do you think we need one for this?

sestinj (approved):

sestinj:
no need to test, this is perfect. Thanks for your contribution!

PR #1873, opened 2024-07-30
Change the onDidSaveTextDocument handler to only reindex the single saved file

spew:
## Description

This will improve scale as before the indexing process was restarted on every file save. There is a bit of code duplication between this new function and the existing refresh function. It will be solved in the next commit which builds on top of this one.

## Checklist


Patrick-Erichsen:
Excellent work per usual 👌

PR #1876, opened 2024-07-30
Make indexing functional in large workspaces by chunking segments of the workspace instead of working on the entire workspace at once

spew:
## Description

Indexing now works in chunks instead of operating on the entire repo at once. Note: to keep the progress bar working, we do an initial walk of the repository to count the files. This walk is fast in small repositories and can take upwards of 30 seconds in a huge, 100K+ file repository. This is an acceptable price to pay since it would take O(days) to index 100K files.

## Checklist

### DIFF
--- a/core/indexing/CodebaseIndexer.ts
+++ b/core/indexing/CodebaseIndexer.ts
@@ -7,7 +7,7 @@ import { LanceDbIndex } from "./LanceDbIndex.js";
 import { ChunkCodebaseIndex } from "./chunk/ChunkCodebaseIndex.js";
 import { getComputeDeleteAddRemove } from "./refreshIndex.js";
 import { CodebaseIndex, IndexResultType } from "./types.js";
-import { walkDir } from "./walkDir.js";
+import { walkDirAsync } from "./walkDir.js";
 
 export class PauseToken {
   constructor(private _paused: boolean) {}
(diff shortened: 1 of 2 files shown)


sestinj:
This is fantastic. The refactor looks clean to me, but I'm starting to feel that this reaches a level of complexity + criticality that some testing would be beneficial. Even something as simple as making a mock `CodebaseIndex` that lists the name of the files indexed in a temp file so we can validate that what we expect to get indexed gets indexed, and that the sequence of yielded updates is sensical. Any interest in trying this?

spew:
I started down the path of adding a test for CodebaseIndex but the issue is that ConfigHandler is incompatible with jest (doesn't compile for a variety of reasons).

sestinj:
ah you're right and I remember this. Will look into it today to see if we can't get the tsconfig correct

sestinj:
WIP here to convert to ESM so that these tests can run: https://github.com/continuedev/continue/tree/nate/esm. Will lyk when this progresses, and will wait to merge until either that happens or we realize it won't be done in 2 days

sestinj:
@spew this was merged—we've un-skipped your previous test (it's now failing, but for ostensibly non-tsconfig-related reasons). Do you want to wrap up the test and then we'll merge?

spew:
Hi @sestinj I don't think I had previously created a test. I did see there is a test at `test/indexing/CodebaseIndexer.skip.ts`. I tried enabling it but saw errors related to the environment:

```
 FAIL  test/indexing/CodebaseIndexer.test.ts
  ● Test suite failed to run

    vectordb: failed to load native library.
      You may need to run `npm install @lancedb/vectordb-linux-x64-gnu`.

      If that does not work, please file a bug report at https://github.com/lancedb/lancedb/issues
          
      Source error: Error: Cannot find module '@lancedb/vectordb-linux-x64-gnu' from 'node_modules/vectordb/native.js'

      at Object.<anonymous> (node_modules/vectordb/native.js:26:11)
      at Object.<anonymous> (node_modules/vectordb/src/query.ts:20:25)
      at Object.<anonymous> (node_modules/vectordb/src/remote/index.ts:31:1)
      at Object.<anonymous> (node_modules/vectordb/src/index.ts:23:1)

```

sestinj:
Yeah no worries—I've been able to test this enough locally to feel comfortable merging, which I'll do later today. My mistake assuming this was your test

spew:
I do think a test would be great and make development a lot faster but it does help me if we are able to merge this now without it.

spew:
This may not merge without conflict due to the change I made last week to fix .gitignore pattern matching in walkdir. Will update the commit with a rebase.

spew:
It actually rebased cleanly but I pushed the updated branch anyway.

sestinj (approved):

sestinj:
Going to merge. We'll work on testing in short order, and keep a close eye on indexing in the meantime.

PR #1877, opened 2024-07-31
Convert core/ to NodeNext

sestinj:


jueinin:
I suspect this is related to the PR. Since vscode extension version 0.9.192, the models defined in my config.ts are no longer displayed. @sestinj

sestinj:
@jueinin I just tested and reproduced this. Not yet sure what caused it, but I don't believe it was this, because system message is still loaded successfully. More details here: https://github.com/continuedev/continue/issues/1905

PR #1880, opened 2024-07-31
Migrate walkDir from using Minimatch to ignore.Ignore for matching .gitignore patterns

spew:
## Description

The ignore library attempts to pattern match exactly as .gitignore does and is better for this purpose than Minimatch.

## Checklist


RichardoC:
Will this also honour all .gitignore files it fines as it walks the folder?

PR #1850, opened 2024-07-27
[WIP] Concurrent Indexing UI

priyashpatil:
## Description
- [x] Add id support on IndexingProgressUpdate
- [x] Enable publishing docs indexing updates
- [x] fixes docs force indexing ran before indexing 
- [x] add concurrent docs indexing
- [x] fixes style props passing through StyledTooltip
- [x] Update gui for concurrent Indexing Progress Updates
- [ ] Fix UI states for showing errors coming from concurrent operations

## Checklist

- [x] The base branch of this PR is `dev`, rather than `main`
- [x] The relevant docs, if any, have been updated or created

## Screenshots
https://github.com/user-attachments/assets/95e00834-eef7-49e1-a8a5-f2ce89baaeb1

PR #1885, opened 2024-08-01
Enable SQLite WAL (Write-Ahead Logging) mode for improved concurrency

priyashpatil:
## Description

Enabled Write-Ahead Logging (WAL) mode for the SQLite database. This change is intended to enhance both the performance and stability of database operations, particularly when handling concurrent transactions and large volumes of data.

WAL is very standard across many well-known databases eg. MySQL, PostgreSQL, and MongoDB. The WAL mode in SQLite is kinda similar and **is not enabled by default**. More about it here: https://www.sqlite.org/wal.html

This also resolves the concurrency issues pointed out in and related to #1850 . **However, it does not improve the performance of indexing.** The indexing might be actually slower as mentioned in the above WAL references. For large codebase the performance will be noticeably slow by minutes For example the below:

**Codebase** - (Django Framework Repo) 
**Without WAL** - 13:48 Minutes
**With WAL** - 16:48 Minutes

**Note: The current change only enables WAL while creating SQLite files for the first time. A fresh run of continue (delete the index folder to test it out). The migration for the existing SQLite file is to be implemented (in Next PR).** 

## Checklist

### DIFF
--- a/core/indexing/CodeSnippetsIndex.ts
+++ b/core/indexing/CodeSnippetsIndex.ts
@@ -27,6 +27,8 @@ export class CodeSnippetsCodebaseIndex implements CodebaseIndex {
   constructor(private readonly ide: IDE) {}
 
   private static async _createTables(db: DatabaseConnection) {
+    await db.exec("PRAGMA journal_mode=WAL;");
+
     await db.exec(`CREATE TABLE IF NOT EXISTS code_snippets (
         id INTEGER PRIMARY KEY,
         path TEXT NOT NULL,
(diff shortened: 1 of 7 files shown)


sestinj:
Good find—looks good to merge, but I want this one to incubate in pre-release a bit longer as it touches sensitive parts of the code. Will merge after we've pushed current dev into main later today

priyashpatil:
> e migration for the existing SQLite file is to b

Works 👍🏻

sestinj (approved):

PR #1943, opened 2024-08-06
Convert ChunkCodebaseIndexer's sqlite inserts into a single bulk insert to improve performance

spew:
## Description

This change greatly improves the performance when inserting chunks. There is almost no time spent in this indexer after this change. Once a similar change is made for the other Indexers the time spent in indexing will be greatly improved.

## Checklist


spew:
the `tsc` check appears to be failing due to an older node.js version in the check environment.

digitalneedstech:
Looks like this commit is causing an issue when running install all dependencies task in vs code.

spew:
@digitalneedstech thanks for reporting, I've opened up this PR here to fix it: https://github.com/continuedev/continue/pull/1955/files

PR #1944, opened 2024-08-06
Change the LanceDBIndex to compute its new cache rows in parallel and then do a bulk insert

spew:
## Description

This greatly improves the performance of indexing. Combined with the change for bulk inserts in ChunkCodebaseIndexer, indexing is ~4x as fast. However, it will even be much faster when the other indexers are converted to use bulk insert as well (sending those Prs out soon).

In addition, this indexer is **not using the computed chunks in the chunk table**. It would be faster to reference the already computed chunks, however, for the purposes of this PR we are not changing that behavior.

## Checklist

### DIFF
--- a/core/indexing/LanceDbIndex.ts
+++ b/core/indexing/LanceDbIndex.ts
@@ -19,6 +19,7 @@ import {
   PathAndCacheKey,
   RefreshIndexResults,
 } from "./types.js";
+import { RunResult } from "sqlite3";
 
 // LanceDB  converts to lowercase, so names must all be lowercase
 interface LanceDbRow {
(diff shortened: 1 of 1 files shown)


spew:
the `tsc` check appears to be failing due to an older node.js version in the check environment.

Patrick-Erichsen:
@spew - it looks like our JS target doesn't actually support  `cause` . 

[See this comment](https://github.com/microsoft/TypeScript/issues/49535#issuecomment-1154793325): "The constructor overload with cause is only supported with ES2022 or later."

Our config:
```json
"target": "ES2020",
"lib": ["ES2021", ...],
```

The following config resolved the type errors for me. Going to push a commit to update to it.
```json
"target": "ES2022",
"lib": ["ES2022", ...],
```

PR #1946, opened 2024-08-06
Enable async encoding for the GPTEncoder

spew:
## Description
This will improve the performance of indexing when using a GPT model. Also, improved the Async interface to also have an async decode method.

Note: I was only able to test this method with the test I added as I do not have access to a GPT model.

## Checklist

### DIFF
--- a/core/llm/asyncEncoder.ts
+++ b/core/llm/asyncEncoder.ts
@@ -5,22 +5,23 @@ import * as path from "path";
 
 export interface AsyncEncoder {
   encode(text: string): Promise<number[]>;
-  decode(tokens: number[]): string;
+  decode(tokens: number[]): Promise<string>;
+  close(): Promise<void>;
 }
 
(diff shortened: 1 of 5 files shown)


core/llm/asyncEncoder.ts
```diff
@@ -31,17 +32,30 @@ export class LlamaAsyncEncoder implements AsyncEncoder {
...
+  public async close(): Promise<void> {
+    await this.workerPool.terminate();
   }
+}
 
-  decode(tokens: number[]): string {
-    return this.tiktokenEncoding.decode(tokens);
+function workerCodeFilePath(workerFileName: string): string {
+  if (process.env.NODE_ENV === "test") {
+    // `cross-env` seems to make it so __dirname is the root of the project and not the directory containing this file
```
Patrick-Erichsen, line 57: @spew - I believe that config is actually coming from here: https://github.com/continuedev/continue/blob/main/core/jest.config.js#L17

PR #1960, opened 2024-08-07
Fix a merge mistake that results in a full reindex on file save

spew:
In https://github.com/continuedev/continue/pull/1873, the file save handler was changed to only reindex the single saved file. However, in the merge commit d0f3209dff29002e6dd6fa1cc47b5769d5c75106 the change was partially reverted and now the entire index is being refreshed on single file save.

## Checklist


Patrick-Erichsen:
Good catch 👍

PR #1976, opened 2024-08-09
Remove the PRAGMA setting of journal_mode which attempted to revert to the default journal_mode

spew:
## Description

The intention of the PRAGMA was to revert the databse back from WAL, however, this pragma is creating problems when multiple instances of continue are open

## Checklist


Patrick-Erichsen:
Good catch 👍

Issue #1991, opened 2024-08-12
Add a way to allow users to reset indexes on failure


priyashpatil:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Recently, starting to get more sqlite errors on updates. It would be useful if there's indexing error related to sqlite to give an option to "rebuild indexing database".

CXwudi:
I need this as well. It is very frustrating that the indexing kept failing even when I click the red icon to rebuild

fry69:
Hope is on the way (already merged), praise @Patrick-Erichsen  -> https://github.com/continuedev/continue/pull/1993

Patrick-Erichsen:
@fry69 thanks for the follow up here, forgot to link in the PR.

PR #1993, opened 2024-08-13
feat: create "rebuild index" dialog

Patrick-Erichsen:
## Description

This PR allows users to clear their codebase index either through a new command, `Continue: Rebuild codebase index`, or through a dialog that can be activated when clicking on an indexing error status. This dialog will only be available when we've identified the error is from a pre-defined list of indexing errors that we believe warrant a rebuild.

```ts
  // Note that we exclude certain Sqlite errors that we do not want to clear the indexes on,
  // e.g. a `SQLITE_BUSY` error.
  errorsRegexesToClearIndexesOn = [
    /Invalid argument error: Values length (d+) is less than the length ((d+)) multiplied by the value size (d+)/,
    /SQLITE_CONSTRAINT/,
    /SQLITE_ERROR/,
    /SQLITE_CORRUPT/,
    /SQLITE_IOERR/,
    /SQLITE_FULL/,
  ];
```

## Checklist

## Screenshots
Screenshot 2024-08-12 at 5 58 48 PM

## Testing

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -638,8 +638,13 @@ export class Core {
       const rows = await DevDataSqliteDb.getTokensPerModel();
       return rows;
     });
-    on("index/forceReIndex", async (msg) => {
-      const dirs = msg.data ? [msg.data] : await this.ide.getWorkspaceDirs();
+    on("index/forceReIndex", async ({ data }) => {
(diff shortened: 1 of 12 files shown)


priyashpatil:
Closes #1991

PR #2051, opened 2024-08-20
Fix passing codebaseContext parameter and InputToolbar flickering

anrgct:
I discovered two bugs in the VSCode extension. First, the nFinal value set in config.json for the codebase contextProvider wasn't taking effect. After investigating, I found that load.ts wasn't passing the parameter, so I modified it to include the parameter. Second, clicking buttons in input toolbars that weren't the last one wasn't working. I tracked this down to a rapid switching of the hidden state, so I changed it to use a debounced hidden state. I hope my code changes can be merged to fix these two bugs.

### DIFF
--- a/core/config/load.ts
+++ b/core/config/load.ts
@@ -355,9 +355,11 @@ async function intermediateToFinalConfig(
 
   // These context providers are always included, regardless of what, if anything,
   // the user has configured in config.json
+
+  const codebaseContextParams = ((config.contextProviders || []).filter(isContextProviderWithParams).find(cp => cp.name === "codebase") as ContextProviderWithParams | undefined)?.params || {};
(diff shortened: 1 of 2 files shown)


sestinj (approved):
Nice work! Thanks for taking the time to fix these things

sestinj:
Hi @anrgct, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

PR #2054, opened 2024-08-20
Add initial support to NVIDIA NIMs model provider

adawalli:
## Description

Added initial support for NVIDIA NIMs (https://build.nvidia.com)

## Checklist

### DIFF
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -68,6 +68,7 @@ export abstract class BaseLLM implements ILLM {
         this.apiBase?.includes("api.groq.com") ||
         this.apiBase?.includes("api.mistral.ai") ||
         this.apiBase?.includes(":1337") ||
+        this.apiBase?.includes("integrate.api.nvidia.com") ||
         this._llmOptions.useLegacyCompletionsEndpoint?.valueOf() === false
       ) {
         // Jan + Groq + Mistral don't support completions : (
--- a/core/llm/llms/Nvidia.ts
+++ b/core/llm/llms/Nvidia.ts
@@ -1,0 +1,14 @@
+import { LLMOptions, ModelProvider } from "../../index.js";
+import OpenAI from "./OpenAI.js";
+
+class Nvidia extends OpenAI {
+  static providerName: ModelProvider = "nvidia";
+  static defaultOptions: Partial<LLMOptions> = {
+    apiBase: "https://integrate.api.nvidia.com/v1/",
+    model: "meta/llama-3.1-405b-instruct",
+    useLegacyCompletionsEndpoint: false,
+  };
+  maxStopWords: number | undefined = 4;
+}
+
+export default Nvidia;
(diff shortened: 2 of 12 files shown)


core/llm/llms/Nvidia.ts
```diff
@@ -0,0 +1,14 @@
+import { LLMOptions, ModelProvider } from "../../index.js";
+import OpenAI from "./OpenAI.js";
+
+class Nvidia extends OpenAI {
```
agirault, line 4: Should other stuff from openai be overriden? Such as `NON_CHAT_MODELS` and `CHAT_ONLY_MODELS` used in [`_streamChat`](https://github.com/continuedev/continue/blob/main/core/llm/llms/OpenAI.ts#L196)

adawalli: Mostly I am emulating the recent add to OpenRouter, which is similar to NVIDIA's implementation. I do think we will need to tweak this list over time though

core/llm/index.ts
```diff
@@ -68,6 +68,7 @@ export abstract class BaseLLM implements ILLM {
         this.apiBase?.includes("api.groq.com") ||
         this.apiBase?.includes("api.mistral.ai") ||
         this.apiBase?.includes(":1337") ||
+        this.apiBase?.includes("integrate.api.nvidia.com") ||
```
agirault, line 71: am I understanding right that this means NIMs don't support completion when using the openai provider? But it does work fine (or should) otherwise with the nvidia provider?

If yes, seems worthy of a comment, the comment below this just make it sounds like these providers just don't support completion at all.

core/llm/llms/Nvidia.ts
```diff
@@ -0,0 +1,14 @@
...
+  static providerName: ModelProvider = "nvidia";
+  static defaultOptions: Partial<LLMOptions> = {
+    apiBase: "https://integrate.api.nvidia.com/v1/",
+    model: "meta/llama-3.1-405b-instruct",
+    useLegacyCompletionsEndpoint: false,
+  };
+  maxStopWords: number | undefined = 4;
```
agirault, line 11: This was the main issue with issue the `openai` provider class with NIMs, right? Can you comment why?

adawalli: NVIDIA NIMS api are currently limited to only 4 stops - we're going to see if that can be adjusted - but for now it's a max of 4

core/llm/index.ts
adawalli: I was initially trying to block the completions endpoint due to the issues I'm running into (this would only have done it on the `openai` type, not the new `nvidia` type) - I will end up removing this if I can figure out why completions aren't working that great

core/llm/llms/Nvidia.ts
agirault: Right, I meant that's worthy of a comment in code. Though as we've seen during testing, that's only for starcoder NIM

adawalli:
@sestinj - I still have some more things to work through, I'm going to close this request for now

PR #2057, opened 2024-08-21
Reorder stableCodeFimTemplate.stop order

adawalli:
Some providers have a limited number of STOPs that can be accepted (e.g., NVIDIA NIMs, groq, etc). This means that the ordering of the first 4 default STOPS is important. Starcoder has an interesting issue where it emits <file_sep> as part of it's completion and having this supplied as part of the 4 first stops helps to provide better completions.

## Description

Reordered key 3 and 4 (0 indexed) so that `<file_sep>` makes it's way into even the most restricted of `STOP` lists

## Checklist

## Screenshots

## Testing


core/autocomplete/templates.ts
```diff
@@ -33,8 +33,8 @@ const stableCodeFimTemplate: AutocompleteTemplate = {
       "<fim_prefix>",
       "<fim_suffix>",
       "<fim_middle>",
-      "<|endoftext|>",
       "<file_sep>",
+      "<|endoftext|>",
```
agirault, line 37: You're updating stablecode which maybe is fine, but did you not mean to update starcoder? https://github.com/continuedev/continue/pull/2057/files#diff-369bc8b51fad8f522e230fb546b9c8ee7fd69650421509eadaf6877320082fd0R137-R138

adawalli: I don't know the history here, but the starcoder template is not currently used - `stable` is used

https://github.com/continuedev/continue/blob/1f20433f36e50bbb344d0de656c7d0b98666a682/core/autocomplete/templates.ts#L314-L330

adawalli: that being said, I should probably update the starcoder template as well in case that line gets uncommented?

agirault: oh, uh. Feedback from @sestinj needed then.

adawalli: I went ahead and updated the `starcoder2FimTemplate` as well to be safe

sestinj: yeah starcoder uses the same template as stablecode so we just have named is stablecode. The changes here look good

sestinj (approved):

core/autocomplete/templates.ts
agirault: ty @sestinj. What is `starcoder2FimTemplate` used for then?

PR #2068, opened 2024-08-22
added terminal context provider.

uxroc:
## Description

- Added Intellij terminal context provider as requested in https://github.com/continuedev/contribution-ideas/issues/23
 
- Supports tabs, splits, will always capture the most recent used session that is still valid. (As Intellij does not have direct API to do this, I have added listener and kept track of active terminal sessions in the TerminalActivityTrackingService.)

- One thing different from VSCode is that it captures everything from a session rather than the last command and output.

## Checklist

## Screenshots

Screenshot 2024-08-21 at 11 47 23 PM

## Testing

Please feel free to play around on terminal sessions, and use @Terminal to see what the content looks like


sestinj:
@uxroc very well done! Couldn't find anything wrong after playing around myself. One small change that I think would be valuable, and please let me know if you'd be interested in trying to do this: I think we should make an attempt to only get the last command, in order to standardize behavior across IDEs. We have a heuristic for doing that manually in VS Code, and the same thing would work here: https://github.com/uxroc/continue/blob/terminal-context-provider/extensions/vscode/src/util/ideUtils.ts#L363

uxroc:
> @uxroc very well done! Couldn't find anything wrong after playing around myself. One small change that I think would be valuable, and please let me know if you'd be interested in trying to do this: I think we should make an attempt to only get the last command, in order to standardize behavior across IDEs. We have a heuristic for doing that manually in VS Code, and the same thing would work here: https://github.com/uxroc/continue/blob/terminal-context-provider/extensions/vscode/src/util/ideUtils.ts#L363

@sestinj Sure thing! I was looking for Intellij plugin APIs to get the last command but could not find anything. And I will definitely apply the manual extraction method above. Thanks for the suggestion!

uxroc:
- added heuristics to get the last command and results. 
- the method always assumes the last line in terminal is a "clean" terminal prompt which will be used to identify the start point of the last command. 
- The fallback will include the whole history.

sestinj:
Great work @uxroc and thank you again! Merging now and will get this into the next release : )

PR #2097, opened 2024-08-25
fallback for non-sse response

raja-jamwal:
## Description

- Solution for https://github.com/nomic-ai/gpt4all/issues/2174, gpt4all doesn't support SSE, this is a fallback in case the response is non-SSE.

## Checklist

## continue.dev loading from gpt4all

Screenshot 2024-08-25 at 12 33 55 PM

## gpt4all logs

Screenshot 2024-08-25 at 12 39 57 PM

## Testing
- start gpt4all with the server mode.
- in continue.dev use it as an OpenAI compatible provider.
- ask something, it should respond.

### DIFF
--- a/core/llm/llms/OpenAI.ts
+++ b/core/llm/llms/OpenAI.ts
@@ -235,6 +235,10 @@ class OpenAI extends BaseLLM {
       if (value.choices?.[0]?.delta?.content) {
         yield value.choices[0].delta;
       }
+
+      if (value.choices?.[0]?.message?.content) {
+        yield value.choices[0].message;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


sestinj:
@raja-jamwal thanks for taking the time to look into this!

I am slightly concerned about the complexity that this adds, just to account for a special case with GPT4All. My preference would be to make an entirely new provider for GPT4All because it isn't in the true sense "OpenAI-compatible".

This would mean adding a new file in the same folder as OpenAI.ts, which can implement similar code to OpenAI, or subclass it where appropriate. Would you be interested in trying this?

raja-jamwal:
Yes, that makes more sense. I'll make those changes.

sestinj:
Hi @raja-jamwal, just wanted to check in, have you had a chance to take a look at this?

sestinj:
I'm going to close this for now, but let me know if you have the chance to take another look and we can re-open!

PR #2101, opened 2024-08-25
Add rails context provider

rootedbox:
## Description

added a rails context provider so that if you have a question about your rails project simply adding the provider gives the ai full scope into the important parts of a rails app

Gemfile - libraries and rails version used
.ruby-version - version of ruby being used
schema.rb - the schema of the DB which tells active record properties of models
file structure.. so that the ai knows how you are modeling your data and what type of service structure you are using

## Checklist

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -447,6 +447,7 @@ declare global {
     | "database"
     | "code"
     | "docs"
+    | "rails"
     | "gitlab-mr"
     | "os";
   
--- a/core/context/providers/FileTreeContextProvider.ts
+++ b/core/context/providers/FileTreeContextProvider.ts
@@ -27,6 +27,43 @@ function formatFileTree(tree: Directory, indentation = ""): string {
   return result;
 }
 
+export async function fileTrees(extras: ContextProviderExtras): Promise<string[]> {
+  const workspaceDirs = await extras.ide.getWorkspaceDirs();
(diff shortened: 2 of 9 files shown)


Patrick-Erichsen:
Hey @rootedbox, neat idea! Given the limited subset of users developing rails apps though, I think this could be better suited to a[ .prompt file](https://docs.continue.dev/features/prompt-files). Would you be open to moving the contribution to https://github.com/continuedev/prompt-file-examples? 

Looking at the code I believe everything you're doing can be done in a .prompt file.

Eventually we plan to flesh this out much more so it would be cool to eventually have something like a "Rails" category we could put this under. For now just a top level file in that repo would be great though.

rootedbox:
Yah I didn't know about prompt_files.. I can move it there. one question about the docs..

```
{{{ ./path/to/file.js }}} - Any file can be directly referenced
```

is ./ the root of the project open?

sestinj:
Yup, that is how it is intended to work. You should also be able to do path/to/file.js

rootedbox:
Closing and converting to prompt.

Patrick-Erichsen:
@rootedbox if you do convert to a prompt we would definitely be interested in merging it to https://github.com/continuedev/prompt-file-examples. I think this is a great example for framework specific prompting.

Issue #1396, opened 2024-05-30
[IntelliJ Plugin] Unauthorised when using Mistral Codestral (despite proper config which work for VSCode)


tajchert:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 14.5
- Continue: 0.0.49
- IDE: IntelliJ IDEA 2024.2 EAP (Community Edition) also Android Studio Jellyfish | 2023.3.1 Patch 1
- Model: Mistral Codestral
```

### Description

Continue.dev plugin doesn't seems to correctly work with Mistral Codestral, same model config works perfectly for Continue.dev plugin for VSCode but results in `HTTP 401 Unauthorized  from https://codestral.mistral.ai/v1/chat/completions`

Continue plugin config I'm using is:
```
    {
      "title": "Codestral - Mistral",
      "provider": "mistral",
      "model": "codestral-latest",
      "apiKey": "[API_KEY]"
    },
```
Where API_KEY is replaced with proper key. Seems like plugin for IntelliJ doesn't work proper (yet) with Mistral Codestral.

### To reproduce

1. Configure Mistral Codestral according to https://docs.continue.dev/walkthroughs/set-up-codestral
2. Open IntelliJ or Android Studio
3. Use latest Continue.dev plugin (0.0.49) for IntelliJ
4. Use it ex. CMD+J with highlighted code, add message 
Result is : `HTTP 401 Unauthorized  from https://codestral.mistral.ai/v1/chat/completions` even as same steps but in VSCode with same config works fine.

### Log output

```Shell
[2024-05-30T14:02:03] Error: HTTP 401 Unauthorized from https://api.mistral.ai/v1/chat/completions

{
  "message":"Unauthorized",
  "request_id":"b123a93ebdd331585f834d5832bd087f"
}

Code: undefined
Error number: undefined
Syscall: undefined
Type: undefined

Error: HTTP 401 Unauthorized from https://api.mistral.ai/v1/chat/completions

{
  "message":"Unauthorized",
  "request_id":"b123a93ebdd331585f834d5832bd087f"
}
[...]

PR #2120, opened 2024-08-28
autocomplete: prevent unauthorized error per #1396

ChrisChiasson:
## Description

Updating the documentation to reflect current error messages returned when a real Mistral autocomplete API key gets defaulted to the wrong API base because API base is not supplied.

## Checklist

- [ X ] The base branch of this PR is `dev`, rather than `main`
- [ X ] The relevant docs, if any, have been updated or created

## Screenshots

Not applicable

## Testing

Without this change, autocomplete with the paid version of Mistral doesn't work (which is what that part of the docs is trying to show how to use). With this change, it does work.

### DIFF
--- a/docs/docs/features/tab-autocomplete.md
+++ b/docs/docs/features/tab-autocomplete.md
@@ -12,6 +12,7 @@ If you want to have the best autocomplete experience, we recommend using Codestr
     "title": "Codestral",
     "provider": "mistral",
     "model": "codestral-latest",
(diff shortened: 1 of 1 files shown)


sestinj:
@ChrisChiasson Thanks for helping out here! Definitely can be improved. These docs are intended to help people obtain a Codestral API key, rather than a Mistral API key (at least while they continue to be free). I think if anything it would be much better to update the language beforehand and use the specific codestral link: https://console.mistral.ai/codestral

PR #2123, opened 2024-08-28
chore: Update embeddings provider to voyageai

asimchoudhary:
## Description

Changed the embeddings provider from OpenAI to VoyageAI in the `codebase-embeddings.md`

## Checklist

## Screenshots

## Testing


sestinj:
@asimchoudhary thanks for looking into this! The "openai" label actually is correct—we don't have a dedicated Voyage provider, which is why the URL has to be set.

So I'm going to close this PR, but I think that we should probably add a dedicated provider

Issue #2162, opened 2024-09-02
Indexing Silently Failed after updated to ≥0.8.48


YouJiacheng:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows
- Continue: 0.8.49
- IDE: VS Code
- Model:
- config.json:
```

### Description

Before I update to 0.8.49, I can get the error message.

![2cdc3e3ae1fb2f0d6da3326a92c7a77](https://github.com/user-attachments/assets/65b356a4-aa73-49fb-bab0-af46a17f3286)

After I update to 0.8.49, I can only get "Indexing complete" with 1% progress.
![image](https://github.com/user-attachments/assets/ed56cdd2-dbbf-4d1d-bad1-8df137a7b4b5)

Details of the indexing error: The indexing error was caused by `maxBatchSize`, the API provider I used only supports `maxBatchSize=32`. Changing the following 128 to 32 can resolve this error. Maybe we can add a config option for `maxBatchSize`, but that will be a separate PR: #2163 .
![image](https://github.com/user-attachments/assets/fea86c55-a770-4b8e-9875-d18b026e4734)

BTW, `maxChunkSize` and its default value should be documented. Initially I suppose the error was caused by chunk size exceeding limit. I searched in the source code and found this option. I changed it to 8192 and problem resolved, but I suddenly found that the default value is 500. Then I realized it should be a `maxBatchSize` problem instead of `maxChunkSize`.

PR #2163, opened 2024-09-02
Add `maxBatchSize` option for EmbeddingsProvider.

YouJiacheng:
## Description

Add `maxBatchSize` option for EmbeddingsProvider.

Minor changes:
+ fix a bug in getBatchedChunks: `if (chunks.length > this.maxBatchSize) return [chunks]` causes chunks will never be spilt into multiple batches. Originally only Gemini uses it, and its `maxBatchSize` is 2048, so no problem in the most of cases.      
+ consolidate duplicated code implementing `getBatchedChunks` logic in` CohereEmbeddingsProvider`, `FreeTrialEmbeddingsProvider`, `HuggingFaceTEIEmbeddingsProvider` and `OpenAIEmbeddingsProvider`. `TransformersJsEmbeddingsProvider` has identical logic, but its property is called `maxGroupSize`.

## Checklist

## Screenshots

## Testing


Patrick-Erichsen:
Thanks for looking into this @YouJiacheng ! I had noticed that logic error in `getBatchedChunks` yesterday. Nice cleanup to consolidate the logic for the other providers as well.

Getting an error in `core/indexing/embeddings/HuggingFaceTEIEmbeddingsProvider.ts` after this update however: https://github.com/continuedev/continue/actions/runs/10672135505/job/29624337029?pr=2163#step:12:6

@sestinj @YouJiacheng - thoughts on defining a `DEFAULT_MAX_BATCH_SIZE` in the base class? I'm not sure if it's better to define an arbitrarily small max batch, or go with something like `Number.MAX_SAFE_INTEGER`.

Either way I would +1 the [comment here](https://github.com/continuedev/continue/issues/2162#issue-2501372689) that we should add it to the docs. 

@YouJiacheng , mind updating this PR with that documentation once we come to a conclusion?

YouJiacheng:
> Getting an error in core/indexing/embeddings/HuggingFaceTEIEmbeddingsProvider.ts after this update however

>  indexing/embeddings/HuggingFaceTEIEmbeddingsProvider.ts(12,3): error TS2610: 'maxBatchSize' is defined as an accessor in class 'BaseEmbeddingsProvider', but is overridden here in 'HuggingFaceTEIEmbeddingsProvider' as an instance property.

I realized there will be such an error after I opened this PR several ours. I only implemented a getter for `maxBatchSize` in `BaseEmbeddingsProvider`, while `HuggingFaceTEIEmbeddingsProvider` will modify it in its custom handling logic.

It was the first time I wrote TypeScript and I forgot to read `CONTRIBUTING.md` so I didn't test it locally. Is it okay to exercise CI or I should test it locally?

Do you think we should add a `_maxBatchSize` property (not sure if it can be private) and implement a setter in the base class? That's not very elegant. Or we can override the getter in `HuggingFaceTEIEmbeddingsProvider` and add a private property `_maxBatchSize`.

YouJiacheng:
I got `tsc core` passed locally. Not sure about whether we should further consolidate the logic: now most `embed` are similar.
image
image
image

sestinj (approved):
I would be fine to merge this now, but I also support further consolidating the logic

sestinj:
I'm going to merge this now and if we want to move the logic all to the base class we can do that in a separate PR

sestinj (approved):

YouJiacheng:
Oh I forgot to update `config_schema.json`
https://github.com/continuedev/continue/blob/455ca6bde068f3ebbfff225b5ed428b2da16d1a7/extensions/vscode/config_schema.json#L2110-L2116

PR #2212, opened 2024-09-09
fix!: change mistral api base

aretrace:
## Description

Change the Mistrial API base to the official endpoint, replacing old and nonworking codestral endpoint.

## Checklist


sestinj:
@aretrace as far as I am aware Codestral is still available. Are you having trouble connecting?

aretrace:
@sestinj, yes for `tabAutocompleteModel`, I changed `apiBase` per https://github.com/continuedev/continue/issues/1396#issuecomment-2142603194 and it worked (this is still using the Codestral model).
I don't think `"https://codestral.mistral.ai/v1/"` is available (I get a 404, not a 401 or 200).

sestinj:
@aretrace I think the problem here comes from Mistral having separate API keys for Codestral and Mistral, where one key won't work with the other's API base. Until this changes, we wish to keep Codestral as the default because it is free for individuals to use

aretrace:
@aretrace I see, would some disambiguation in the docs be welcomed regarding this?

sestinj:
Definitely! We'd done so here (https://continue.dev/walkthroughs/set-up-codestral) but are also working on a docs rewrite where we'll make it even more clear (also in product onboarding)

PR #2251, opened 2024-09-13
Use XDG directory on Linux.

### LINKED ISSUE #558: Migrate `~/.continue/config.py` to XDG_CONFIG_HOME or `~/.config/continue/config.py`
### Validations

- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

I see the default of `~/.continue/config.py` is used to store the configuration settings, but a lot of tools have started moving their configs to follow the XDG Base configuration: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

From that spec, config files usually go in `~/.config/<app>/<whatever>` as a convention, which keeps a user's `$HOME` directory clean.

### Solution

_No response_

alula:
## Description

Fixes #558.

I personally think that polluting the home directory with dotfiles is a pretty lazy and bad practice.

This PR currently only addresses said problem on Linux, by making Continue [follow the XDG Base Directory specification](https://specifications.freedesktop.org/basedir-spec/latest/). 

This change is backwards compatible and only affects new setups, `~/.continue` is still used if it already exists.

I didn't do anything about other platforms yet, Windows and macOS continues to use `~/.continue` just like before.

## Checklist

### DIFF
--- a/core/util/paths.ts
+++ b/core/util/paths.ts
@@ -9,8 +9,37 @@ import { IdeType, SerializedContinueConfig } from "../index.js";
 
 dotenv.config();
 
-const CONTINUE_GLOBAL_DIR =
-  process.env.CONTINUE_GLOBAL_DIR ?? path.join(os.homedir(), ".continue");
+function getContinueGlobalDir(): string {
+  // Use user-configured directory if set.
+  const envGlobalDir = process.env.CONTINUE_GLOBAL_DIR;
+  if (envGlobalDir) {
+    return envGlobalDir;
+  }
+
+  const defaultDotDir = path.join(os.homedir(), ".continue");
+
+  // Backwards compatibility: use ~/.continue if it already exists
+  if (fs.existsSync(defaultDotDir)) {
+    return defaultDotDir;
+  }
+
+  // on Linux, prefer XDG directories by default
+  // https://specifications.freedesktop.org/basedir-spec/latest/index.html
+  const xdgDataDir = process.env.XDG_DATA_HOME;
+  if (xdgDataDir) {
+    return path.join(xdgDataDir, "continue");
+  }
+
+  // on Windows, prefer ~\AppData\Local
+  const localAppDataDir = process.env.LOCALAPPDATA;
+  if (localAppDataDir ) {
+    return path.join(localAppDataDir , "Continue");
+  }
+  // Use ~/.continue everywhere else.
+  return defaultDotDir;
+}
+
+const CONTINUE_GLOBAL_DIR = getContinueGlobalDir();
 
 export function getChromiumPath(): string {
   return path.join(getContinueUtilsPath(), ".chromium-browser-snapshots");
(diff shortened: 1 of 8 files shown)


core/util/paths.ts
```diff
@@ -9,8 +9,32 @@ import { IdeType, SerializedContinueConfig } from "../index.js";
...
+    return defaultDotDir;
+  }
+
+  // on Linux, prefer XDG directories by default
+  // https://specifications.freedesktop.org/basedir-spec/latest/index.html
+  const xdgDataDir = process.env.XDG_DATA_DIR;
+  if (xdgDataDir) {
+    return path.join(xdgDataDir, "continue");
+  }
+
```
rimmer, line 32: Would you mind adding same for Windows?
```suggestion

  // on Windows, prefer ~\AppData\Local
  const localAppDataDir = process.env.LOCALAPPDATA;
  if (localAppDataDir ) {
    return path.join(localAppDataDir , "Continue");
  }
```

rimmer:
On Windows, `LOCALAPPDATA` can be used to retrieve ~\AppData\Local, where app data is stored. Perhaps, u'd like to add it in your PR, otherwise I can file a PR once this is accepted. 

I hate seeing dotfiles in my homefolder too

ShayBox:
Agreed, especially because it's constantly conflicting with `.config`'s autocomplete. This is so infuriating.

ChrisLane:
@alula your current code doesn't work per the spec.
You're checking for `XDG_DATA_DIR` when you should be checking `XDG_DATA_HOME`

alula:
> @alula your current code doesn't work per the spec. You're checking for `XDG_DATA_DIR` when you should be checking `XDG_DATA_HOME`

Good catch, I have changed it from XDG_**CONFIG**_DIR last minute before commiting 😅

ChrisLane:
Additionally, an improvement to this would be to place config files in `XDG_CONFIG_HOME`. Also, documentation should be updated to avoid people being confused when they can't find ~/.continue while having these environment variables set.

ChrisLane:
Opening the config file via the plugin in Intellij seems to create/open files in `~/.continue/`

alula:
I've updated the IntelliJ plugin and other places too, now my question: how the former `~.continue/` directory present in maaaany places in the documentation be called from now?

alula:
I skipped adding relevant logic to the Rust code as it seems to be broken and not used yet.

sestinj:
Hey all, apologies for the late review here. After thinking about it for a bit, we're going to stick with ~/.continue, the primary reason being that we're often fighting not software bugs, but difficulties in making sure to guide users down the right path when they are customizing Continue. Moving the relevant files to a path that will vary per user is a real opportunity for confusion.

I appreciate the work that went into this though, so if there were interest in doing something like some easter egg flag that would have the directory stored instead at the XDG path that might be fine. Yes, it would still be complex, but I figure for the handful of users that care deeply about this that might be okay

alula:
This PR contains some fixes that make CONTINUE_GLOBAL_DIR work correctly and removes few hardcoded paths. Should I open another one without the change that makes default directory different?

Issue #2056, opened 2024-08-21
[CON-261] Only first module detected in JetBrains editors


sestinj:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:*
- Continue:*
- IDE:JetBrains
```

### Description

When there are multiple modules, or microservices in a workspace in a JetBrains IDE, Continue is only picking up on one of them. This is coming from the array that we return for getWorkspaceDirs, but the root is probably our use of project.basePath here: https://github.com/continuedev/continue/blob/dev/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/activities/ContinuePluginStartupActivity.kt#L120

<sub>[CON-261](https://linear.app/continue/issue/CON-261/only-first-module-detected-in-jetbrains-editors)</sub>

mphilippnv:
Can confirm this is a problem in Pycharm 2024.2.1 with Continue 0.0.64. I had posted in Discord about it but figured I'd post here just for visibility. Thanks!

PR #2259, opened 2024-09-13
support indexing of multiple modules in jetbrains ides

dschling:
## Description
Fixes issue #2056 

workspaceDirectories function updated to return all _top-level_ modules from the project.

## Checklist

- [ x ] The base branch of this PR is `dev`, rather than `main`
- [ x ] The relevant docs, if any, have been updated or created

## Testing

From an existing jetbrains project, import an additional module by using File -> New -> Module from Existing Sources... Reload the Continue plugin window and reindex. Indexes should now be created for both modules.


sestinj (approved):
This is great! Thanks for taking the time to dig into it

PR #2263, opened 2024-09-15
Greptile context provider

justinmilner1:
## Description

https://github.com/user-attachments/assets/2709c546-2b6f-46f8-bfce-bb0793e55bcf

Added an initial version of a Greptile context provider. 

Future versions: Design was limited by lack of an endpoint to get what indices the user has available to them - this cp would be more useful as a dropdown menu context provider.

The current design is to automatically search using the git url correlating to the currently active vscode project. User must have repo indexed prior to using this cp.

The user enters their query into the text box popup. Greptile returns a list of files for context, continue uses this context (Greptile does not provide the response, just a list of files).


core/context/providers/GreptileContextProvider.ts
```diff
@@ -0,0 +1,186 @@
...
+      description: "Insert query to Greptile",
+      type: "query",
+    };
+  
+    async getContextItems(
+      query: string,
+      extras: ContextProviderExtras,
+    ): Promise<ContextItem[]> {
+      const greptileToken = this.getGreptileToken();
+      const githubToken = process.env.GITHUB_TOKEN;
```
sestinj: can this be changed to `this.options` and we can add `githubToken` as one of the expected `params` in config.json for this context provider?

sestinj (changes requested):

sestinj:
Sorry it took a second to get around to a full review, but @justinmilner1 this is awesome! Just a small comment about how to get the github token (probably this schema should also be in config_schema.json), but otherwise I'm more than happy to merge this!

sestinj (approved):

PR #2291, opened 2024-09-18
fix-aws-expired-profile

chezsmithy:
## Description

Update the fromIni method used to retrieve AWS credentials so that it ignores credential file caching.

Without this setting the credentials are read a single time when the extension is started and if the user refreshes the credentials manually via SSO login for example, the extension won't see the refreshed credentials and will continue to throw security token errors. Currently the only way to fix this is to restart VSCode completely.

## Checklist

## Screenshots

## Testing

I have manually tested Imported, Base and Embedded models with AWS Bedrock.
I have expired the AWS credentials and observed the token errors.
I have logged back into AWS and noted the credentials are refreshed and the extension stars working correctly again.

### DIFF
--- a/core/indexing/embeddings/BedrockEmbeddingsProvider.ts
+++ b/core/indexing/embeddings/BedrockEmbeddingsProvider.ts
@@ -78,7 +78,8 @@ class BedrockEmbeddingsProvider extends BaseEmbeddingsProvider {
     try {
       return await
       fromIni({
-        profile: this.profile
+        profile: this.profile,
+        ignoreCache: true
       })();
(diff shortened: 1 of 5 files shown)


chezsmithy:
I looked for some better ways to only read from the non-cached copy by detecting errors and by attempting to compare the expiration results from fromIni. Unfortunately, fromIni doesn't return the expiration timestamp, and catching the errors in the client call requiring a large re-setup of the whole client seemed a bit overkill vs just reading each time.

sestinj (approved):
nice!

PR #2351, opened 2024-09-23
Fix 2349

steverhoades:
## Description
The instructions use `ndims` which does a model name check, the shift to use Vector[1536] fixes the issue and allows for the tutorial to continue.

Source of the dimension
https://docs.voyageai.com/docs/embeddings

## Checklist


sestinj (approved):
Thanks for taking the time to do this!

PR #2372, opened 2024-09-24
Cache turn-by-turn conversation in Anthropic Claude

tijszwinkels:
## Description

This implements turn-by-turn conversation caching for Claude, in addition to the pre-existing system message caching.
See: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching
 
Perhaps we should enable this by default?

## Checklist

## Testing
- Add `"cacheConversation": true` to the anthropic config. Like:
```
{
      "title": "Claude 3.5 Sonnet",
      "provider": "anthropic",
      "model": "claude-3-5-sonnet-20240620",
      "apiKey": "<YOUR_API_KEY>",
      "cacheSystemMessage": true,
      "cacheConversation": true
    }
```

- uncomment the following line in Anthropic.ts: `if (value.type == "message_start") console.log(value);`
- Select the claude model, add at least 1000 tokens context
- When running in the debugger, see in the debug console things like:
```
{input_tokens: 4, cache_creation_input_tokens: 1212, cache_read_input_tokens: 0, output_tokens: 2}
```

then the next request:

```
{input_tokens: 4, cache_creation_input_tokens: 224, cache_read_input_tokens: 1212, output_tokens: 1}
```

Hopefully Anthropic will soon put caching statistics in their api console as well.

### DIFF
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -121,6 +121,7 @@ export abstract class BaseLLM implements ILLM {
   watsonxFullUrl?: string;
 
   cacheSystemMessage?: boolean;
+  cacheConversation?: boolean;
 
   private _llmOptions: LLMOptions;
 
@@ -194,6 +195,7 @@ export abstract class BaseLLM implements ILLM {
     this.watsonxFullUrl = options.watsonxFullUrl;
 
     this.cacheSystemMessage = options.cacheSystemMessage;
+    this.cacheConversation = options.cacheConversation;
 
     if (this.apiBase && !this.apiBase.endsWith("/")) {
       this.apiBase = `${this.apiBase}/`;
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -14,6 +14,7 @@ import {
   PromptTemplate,
   RequestOptions,
   TemplateType,
+  CacheBehavior
 } from "../index.js";
 import { logDevData } from "../util/devdata.js";
 import { DevDataSqliteDb } from "../util/devdataSqlite.js";
@@ -102,6 +103,7 @@ export abstract class BaseLLM implements ILLM {
   llmRequestHook?: (model: string, prompt: string) => any;
   apiKey?: string;
   apiBase?: string;
+  cacheBehavior?: CacheBehavior;
   capabilities?: ModelCapability;
 
   engine?: string;
@@ -119,9 +121,6 @@ export abstract class BaseLLM implements ILLM {
   watsonxStopToken?: string;
   watsonxApiVersion?: string;
   watsonxFullUrl?: string;
-
-  cacheSystemMessage?: boolean;
-  cacheConversation?: boolean;
 
   private _llmOptions: LLMOptions;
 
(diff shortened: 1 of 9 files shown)


core/llm/index.ts
```diff
@@ -194,6 +195,7 @@ export abstract class BaseLLM implements ILLM {
     this.watsonxFullUrl = options.watsonxFullUrl;
 
     this.cacheSystemMessage = options.cacheSystemMessage;
+    this.cacheConversation = options.cacheConversation;
```
sestinj: Can we make a single object called "cacheBehavior"? I think this will make it cleaner in the future if we want to offer further control

core/llm/llms/Anthropic.ts
```diff
@@ -35,17 +35,40 @@ class Anthropic extends BaseLLM {
...
+    const lastTwoUserMsgIndices = filteredmessages 
+      .map((msg, index) => (msg.role === "user" ? index : -1))
+      .filter((index) => index !== -1).slice(-2);
+
+    const messages = filteredmessages.map((message, filteredMsgIdx) => {
+        // Add cache_control parameter to the last two user messages
+        // The second-to-last because it retrieves potentially already cached contents,
+        // The last one because we want it cached for later retrieval.
+        // See: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching
+        const addCaching = !!this.cacheConversation && lastTwoUserMsgIndices.includes(filteredMsgIdx);
```
sestinj: Isn't previous user message already cached from the previous request?

sestinj:
Love this! Yeah I think we'll want to make this default after a very short period of sanity check testing

core/llm/index.ts
tijszwinkels: @sestinj Like this?

core/llm/llms/Anthropic.ts
tijszwinkels: See: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching

_Cache hits require 100% identical prompt segments, including all text and images up to and including the block marked with cache control. The same block must be marked with cache_control during cache reads and creation._

So it needs the marker both to retrieve the previous cache (so in the exact same place as the previous request), as well as to mark up to when the request should be cached now. (creation)

This is also how it's done in the 'Continuing a multi-turn conversation' example on their page.

tijszwinkels:
@sestinj Ready for re-review!

core/llm/llms/Anthropic.ts
sestinj: Thanks for clarifying! This is perfect then

core/llm/index.ts
sestinj: yup, exactly, this is great

sestinj (approved):
excellent work!

tijszwinkels:

chezsmithy:
@sestinj looks like we might have gone a bit backwards with our recent prompt caching decision to move to using promptCaching as the key under default completion options. Thoughts?

sestinj:
In what way specifically are you thinking?

chezsmithy:
@sestinj referencing this decision to standardize on a single promptCaching setting. https://github.com/continuedev/continue/pull/5371#issuecomment-2852535503

PR #2374, opened 2024-09-24
fix: modify docs Next Page link

DongjaJ:
## Description

When I was on the "getting-started/install", I found the next page link at the bottom was "getting-started/install".

The same phenomenon occurred in "chat/how-to-use-it", "autocomplete/how-to-use-it", "edit/how-to-use-it", and "actions/how-to-use-it".

It has been modified appropriately.

## Checklist

## Screenshots

#### as-is
continue docs bug

#### to-be
continue after

## Testing


sestinj:
@DongjaJ thanks for your attention to detail here! I'm wondering if there's a cleaner way we can solve this. It seems like there is something fundamentally wrong causing it to link to itself, perhaps the page index in the frontmatter?

DongjaJ:
@sestinj  

I think the 'link' attribute was misused inside siderbar.js.

Let's take "getting-started" as an example.

Docusaurus reads sidebar settings and identifies the order and hierarchy between documents.
If a category has a link property, the document it points to is considered an index document for that category.

Typically, Docusaurus generates "Next" and "Previous" links in the order listed in the sidebar.

And when you give a link attribute to a category in Docusaurus, it treats the document as if it were a virtual "index.md " of that category.

As a result, Docusaur tries to find the next page of "getting-started/install" on "getting-started/index.md " but points to itself because "getting-started/install" is the current page.

After the link is removed, the "install" page is simply treated as an item in the category.
"Next page" will now correctly point to the next item in the sidebar ("getting-started/overview").

sestinj:
Thanks @DongjaJ, the new change looks great to me. Appreciate you taking a second look!

PR #2389, opened 2024-09-25
Installation 092424

shermanhuman:
## Description

- Added VS Code and Jetbrains logos, following their respective use guidelines found here:

    https://code.visualstudio.com/brand

    https://www.jetbrains.com/company/brand/#company-brand-logos-and-icons-jetbrains-logos

- Minor copy edits
- Style changes: spacing of ordered lists accomplished via css instead of adding spaces
- Some formatting changes to try to improve visual appeal

## Checklist

## Screenshots

image

image

## Testing

I checked out the rest of the doc site just to make sure the changes to unordered lists and list items didn't have any side effects and didn't see any.

I've only tested on a 3440x1440 monitor on Chrome on Windows 11.

To test:

- `cd docs` in the project root
- `npm start`
- navigate to 'Installation' and make sure it looks kosher.


shermanhuman:
Is the formatting too much?  I generally agree that less is more, but it does feel more readable to me compared to the original:

image

I think grouping the subject 'Continue extension page' is good, and makes it more readable.  

I also think splitting up the expository from the instruction is good, but maybe italics isn't the right tool.  It feels like I'm de-emphasizing the important part.  Maybe it's just too much italics.

Issue #2313, opened 2024-09-19
@open Context Provider doesn't work


Snowman-25:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11 23H2 (Build 22631.4169)
- Continue:   v0.8.52
- IDE:        VSCode 1.92.1 (user setup)
- Model:      deepseek-coder-v2:latest
- config.json:
  
{
  "models": [
    {
      "title": "Ollama",
      "provider": "ollama",
      "model": "AUTODETECT"
    }
  ],
  "customCommands": [
    {
      "name": "test",
      "prompt": "{{{ input }}}\n\nWrite a comprehensive set of unit tests for the selected code. It should setup, run tests that check for correctness including important edge cases, and teardown. Ensure that the tests are complete and sophisticated. Give the tests just as chat output, don't edit any file.",
      "description": "Write unit tests for highlighted code"
    },
    {
      "name": "desc",
      "prompt": "{{{ input }}}\n\nWrite a comprehensive comment for each block of the selected code. It should describe what it does and show possible caveats.",
      "description": "Comment the highlighted code"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Starcoder 3b",
    "provider": "ollama",
    "model": "starcoder2"
  },
  "allowAnonymousTelemetry": false,
  "embeddingsProvider": {
    "provider": "transformers.js"
  },
  "contextProviders": [
    {
      "name": "open",
      "params": {
        "onlyPinned": false
      }
     }
  ],
  "experimental": {
    "readResponseTTS": true
  },
  "ui": {
    "showChatScrollbar": true
  },
  "docs": []
}
```

### Description

@open doesn't send the correct context or sometimes ANY content to the LLM

The Errors about config.json not found in the log output are instances where Continue tried to use config.json as the only context to my prompt, despite having multiple files open and config.json not being the active Tab.
Also, the file is existant in that location.
Note, that while I'm running VSCode on my Win11 Computer, my workspace resides on a Unix-Server using the Remote-SSH extension.

### To reproduce

1. Add @open to your config.json
[...]

PR #2424, opened 2024-09-29
Fixes issue #2313 - Convert WSL2 file paths to Windows file paths

steverhoades:
## Description

When using VSCode on Windows where projects are hosted in a linux WSL2 environment the Open Files Context Provider fails. The reason is that VSCode does not know how to interpret the filepaths.

This PR adds a conversion step to all file paths if the given path is coming from a WSL2 environment.

## Checklist

## Testing

Added a unit test for the VsCodeIde.ts class covering these specific changes.


extensions/vscode/src/VsCodeIde.ts
```diff
@@ -431,7 +431,42 @@ class VsCodeIde implements IDE {
...
+    return convertedPaths;
+  }
+
+  /**
+   * Tests the remote name of the vscode.env.remoteName to see if it starts with 'wsl'
+   *
+   * @returns boolean
+   */
+  private isWsl(): boolean {
+    return !vscode.env.remoteName?.toLowerCase().startsWith('wsl');
```
sestinj: Should this be `!!`?

extensions/vscode/src/VsCodeIde.ts
```diff
@@ -431,7 +431,42 @@ class VsCodeIde implements IDE {
...
+
+  private async convertWslPathsIfNeeded(paths: string[]): Promise<string[]> {
+    if (!this.isWsl()) {
+      return paths;
+    }
+
+    const convertedPaths = await Promise.all(paths.map(async (path) => {
+      if (path.startsWith('/mnt/')) {
+        try {
+          const [stdout, stderr] = await this.subprocess(`wslpath -w "${path}"`);
```
sestinj: I'm not overly familiar with `wslpath`—can it be very safely assumed that this command will exist whenever `isWsl` is true, or is it expectation that this `catch` statement will be necessary frequently?

sestinj (commented):
A couple of comments, but I'm a fan of this PR! A larger change we need to make is passing around full URIs at all times rather than just absolute paths, but this is a good stop gap

extensions/vscode/src/VsCodeIde.ts
steverhoades: It's a valid question. I believe it's widely support since 2018 across the WSL 2 environments.
See: https://devblogs.microsoft.com/commandline/windows10v1803/

I would anticipate this being available on most (if not all?) those environments utilizing WSL 2.

extensions/vscode/src/VsCodeIde.ts
steverhoades: It should I fixed my oops.

steverhoades:
@sestinj is there something that is blocking this from being merged, anything I can address? It'd be nice to get that sweet open file functionality in my instance of VSCode without running my local copy.

steverhoades:
Just a heads up, I think i've discovered a much simpler approach to solve this issue without having to run the wslpath process. I'll be pushing the update here shortly.

steverhoades:
@sestinj I've push my change to remove the process call to wslpath, i've tested this locally and it works. I suspect this could also fix the remote-file issue I read about earlier however I have not had a chance to test that.

steverhoades:
Potentially fixes #2451.

sestinj:
@steverhoades this is excellent! Thanks for finding a cleaner way, and sorry for delay. If you could do me a quick favor and remove the package-lock/package.json changes (they look unnecessary now) I'll go ahead and merge!

steverhoades:
done, there was one small version change that was picked up, let me know if you want that reverted as well - happy to do it.

sestinj (approved):

sestinj:
Beautiful, thank you!

PR #2439, opened 2024-10-02
Support remote access toolwindow

956237586:
For intellij idea 
1.Some headless env use jcef runtime need extra effort, or impossiable.
2.For some remote env screen size too small to use continue.

So i changed tool window factory logic, use embedded ktor websever for remote access. When splitMode enabled jcef will not load instead of ktor server, it can be access from remote or out of jetbrains ide.

## Description
1.support force config osr enabled when using jcef
2.support ktor embedded websocket and webjar server for headless or remote enc, also support port mapping config

## Checklist

## Screenshots
image

## Testing
1. change `config.json` for enabled splitMode
```
  "splitMode": true,
  "serverHost": "127.0.0.1",
  "serverPort": "2333",
  "serverToken": "123456"
  "serverExternalHostPort":"reverse-proxy.xxx:port"//optional
```
2. continue gui canbe access from 127.0.0.1:2333 out of ide env

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -15,6 +15,8 @@ declare global {
     };
     colorThemeName?: string;
     workspacePaths?: string[];
+    project:string,
+    embeddedSocketInfo?: IEmbeddedSocketManager;
     postIntellijMessage?: (
       messageType: string,
       data: any,
@@ -22,7 +24,17 @@ declare global {
     ) => void;
   }
 }
-
+export interface IEmbeddedSocketManager {
+  socket: WebSocket,
+  authed: boolean;
+  isConnected: boolean;
+  error?: Error;
+  internal?: {
+    timeout?: NodeJS.Timeout;
+  };
+  project:string?,
+  serverToken:string
+}
 export interface ChunkWithoutID {
   content: string;
   startLine: number;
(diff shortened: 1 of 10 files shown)


sestinj:
@956237586 Can you help me understand how prevalent this problem might be? I'm wondering what versions of JetBrains or which OS, or other factors might contribute. But I do think overall it is probably unlikely that we would merge this, given that the maintenance cost here appears quite high.

One alternate solution is I might be happy to allow an environment variable so that you could then run the sidebar server on your own (rather than setting up ktor within the extension)

956237586:
thanks for you reply

i can't use ssh protocol based remote dev solotion because firewall and other security problem. so i found [Projector](https://jetbrains.github.io/projector-client/mkdocs/latest/) it can use websocket render remote intellij ide. but jcef app depend on jdk native render function, modify Projector server support jcef render too difficult for me when official support is stopped, and message transfer between ide and webview can't found perfect solotion for all scene at now

the key feature i needed is reuse message handler and reply logic, but it current coupled in ContinueBrowser, so i simply split message handler and ui render logic based on current branch code structure in this PR, and embedded ktor server just because ktor already in dependency list

if you accept split message handler rlated logic from render, maybe i will change implement solution and can reduce too much intrusion for current code structure

sestinj:
@956237586 I think I'm going to have to close this as it feels much too likely that it will cause a lot of complexity and difficulty to maintain. If you wanted to attempt a new PR with a much more minimal solution (such as ability to specify the [defaultGuiUrl](https://github.com/continuedev/continue/blob/05cba74b183693c5aaec0b689b96ee453065036b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt#L44) ), then I might be willing to take a second look

PR #2458, opened 2024-10-03
feat: make deletion line highlight wider

Patrick-Erichsen:
## Description

This is the main fix: https://github.com/continuedev/continue/compare/nate/jetbrains-inline-diff-red...pe/jb-highlights?expand=1#diff-17cb756c324178a0fc001dd39fd6d58f172f6180a0f19a8c89c0fbbfbe0d045cR93

## Checklist

## Screenshots

Screenshot 2024-10-03 at 3 50 30 PM

## Testing

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/DiffStreamHandler.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/DiffStreamHandler.kt
@@ -9,6 +9,7 @@ import com.intellij.openapi.components.ServiceManager
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.colors.EditorFontType
 import com.intellij.openapi.editor.colors.TextAttributesKey
+import com.intellij.openapi.editor.markup.EffectType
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
This is close but not perfect. A few shortcomings:
- Unwanted right + left margin
- New code is getting a gutter highlight from VCS styling

I think it might be a better approach to flip the insertion/deletion logic, e.g. we use a `addLineHighlighter` for deletions without actually deleting the code, and then we use the `editorComponentInlaysManager` for insertions. This would at least get around the gutter highlight issue, and be more of a true "proposed" diff since we won't actually be modifying any source code, just applying highlights.

sestinj:
I think the biggest problem is still the width, not that I wouldn't merge this, but that flipping the green/red wouldn't be drastically better

so yeah same conclusion

I think this is what you were getting at with the link, but I don't know if it brought me to the exact lines. What are the biggest changes here that weren't already in the original? trying to parse out formatting stuff

sestinj:
Ok I see it's largely the thing in InlaysManager line 93?

Thompson1985 (approved):

Thompson1985 (approved):

PR #2468, opened 2024-10-04
feat: Automatic Chat Titles

mkummer225:
## Description

This PR adds automatic title generation for chat sessions. Currently it runs after the user submits their first chat (there's an opportunity for more robust logic added here). We piggy back off the user's currently selected model while limiting the maxTokens (response tokens) to 24; the [baked-in prompt](https://github.com/continuedev/continue/pull/2468/files#diff-b5b664dda0f9bdd6a2075c79fcf5e2e382fabba4f0f9c76173a267459223697aR23) asks the model to reply with 4-12 words.

Currently this feature is enabled using the experimental config parameter: `getChatTitles: boolean;`

There appears to have been a prior effort to add this functionality via the `disableSessionTitles` config param (enabled by default then the user could opt-out). When this feature is ready to come out of experimental we should clean up this prior param (assuming I understood it's purpose correctly)

## Checklist

## Screenshots

No visual changes were made; the titles are shown the same as before

image

image

## Testing

Enable this feature by setting `experimental.getChatTitles` to `true`


mkummer225:
While building this feature I spent a decent amount of time thinking about ways to improve search & organization of past chats – it might be cool to have a set of tags that the LLM also generates when generating the title. Then in the history tab the user is able to click-select multiple tags while filtering.

core/core.ts
```diff
@@ -401,6 +402,11 @@ export class Core {
       if (config.experimental?.readResponseTTS && "completion" in next.value) {
         void TTS.read(next.value?.completion);
       }
+      
+      // TODO: better detection of when we should prompt for a title; currently only doing it on the user's first chat
+      if (config.experimental?.getChatTitles && "completion" in next.value && msg.data.messages.filter(m => m.role === "user").length === 1) {
```
sestinj: Initiation of the request for the description should happen from the GUI. We may wish to use llmStreamChat for various reasons other than literally a chat session message, and wouldn't want this spending tokens in those cases.

This warrants a new core protocol message type, which can then be called directly from the GUI

core/util/chatDescriber.ts
```diff
@@ -0,0 +1,65 @@
...
+        {
+          role: "user",
+          content: ChatDescriber.prompt + message,
+        },
+      ],
+      completionOptions,
+    );
+
+    // Set the title
+    ChatDescriber.messenger?.send("setCurrentChatTitle", titleResponse.content.toString());
```
sestinj: i'd recommend trimming the response and removing quotes. I believe we already have utilities for stripping quotes from last time in core/util somewhere

sestinj (changes requested):
Excellent work overall! The two comments are fairly crucial imo, but happy to merge after that

core/core.ts
mkummer225: Yeah that definitely makes sense!

mkummer225: Updated execution to be housed within useHistory's `saveSession` method and removed `setCurrentChatTitle` in favor of directly returning the string value when invoking `chatDescriber/describe`

core/util/chatDescriber.ts
mkummer225: updated

sestinj:
great work!

sestinj:
Hi @mkummer225, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

PR #2472, opened 2024-10-05
add watsonx embeddings provider + add granite chat template + add wat…

mq200:
…sonx models

## Description

- add watsonx embeddings provider
- intelliJ @folder: enable selection of nested folders
- add granite chat template
- add all granite code models
- fix missing watsonx icon in intelliJ plugin
- fix @codebase behavior not respecting configured nFinal/nRetrieve (n being used as num files instead of num chunks for recently edited files)
- align watsonx config options to default options and add apikey URL

## Checklist


Thompson1985 (approved):

core/index.d.ts
```diff
@@ -830,6 +832,13 @@ export interface EmbedOptions {
 
   // AWS and GCP Options
   region?: string;
+
+  // Watsonx Options
+  watsonxUrl?: string;
```
sestinj: Can these be turned into a single object? (watsonXConfig)

core/llm/index.ts
```diff
@@ -177,7 +175,7 @@ export abstract class BaseLLM implements ILLM {
       autodetectTemplateFunction(
         options.model,
         this.providerName,
-        options.template,
+        templateType || options.template,
```
sestinj: Let's make this a separate PR—I'd like to be able to separately review so that this doesn't hold up the rest of the code (it's a particularly touchy part of the codebase)

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IdeProtocolClient.kt
```diff
@@ -641,9 +641,14 @@ class IdeProtocolClient(
                     "listFolders" -> {
                         val workspacePath = workspacePath ?: return@launch
                         val workspaceDir = File(workspacePath)
-                        val folders =
-                            workspaceDir.listFiles { file -> file.isDirectory }?.map { file -> file.absolutePath }
-                                ?: emptyList()
+                        val folders = mutableListOf<String>()
```
sestinj, line 575: Ideally this would be a separate PR as well, but I'm okay with this one being here, as it looks much safer than the template change I commented on above

sestinj (commented):
thanks for all of this!

core/llm/index.ts
mq200: I removed this from PR for now- watsonx template names have to be specified in model configs

core/index.d.ts
mq200: Updated the model config to align fields with default options- now the only field specific to watsonx is the deployment ID (as well as projectId for EmbedOptions)

PR #2479, opened 2024-10-06
Fixes issue 2313 - Fix Open Files Context Provider for WSL and SSH

steverhoades:
## Description

Remote SSH and WSL files are represented with a uri scheme of `vscode-remote`. There is a check in the `openFiles` method of the ideUtils class that checks to see if the current uri is of a schema `file`, if it's not it get's filtered out.

This breaks the OpenFilesContextProvider for the `vscode-remote` schemes.

## Checklist

## Screenshots

Before fix
https://github.com/user-attachments/assets/9c13b347-3440-4d88-a6f2-869254173189

After Fix
https://github.com/user-attachments/assets/4dbfead7-ca5c-4c4c-8a24-5512b41cfbc6

### DIFF
--- a/extensions/vscode/src/util/ideUtils.ts
+++ b/extensions/vscode/src/util/ideUtils.ts
@@ -233,7 +233,7 @@ export class VsCodeIdeUtils {
   // In some cases vscode.window.visibleTextEditors can return non-code editors
   // e.g. terminal editors in side-by-side mode
   private documentIsCode(uri: vscode.Uri) {
-    return uri.scheme === "file";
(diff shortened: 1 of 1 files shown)


steverhoades:
@sestinj apologies for another PR on this issue. This PR reverts my prior change and includes the necessary change to fix the nature of the issue. I'm not entirely sure all the ways that the `documentIsCode` function is being utilized but I suspect that this will fix lingering issues in those other methods as well.

sestinj:
This looks great, thanks!

sestinj (approved):

PR #2493, opened 2024-10-08
Full screen UI improvements

tomasz-stefaniak:
## Description

Added history sidebar on large screens.

## Checklist

## Screenshots

https://github.com/user-attachments/assets/b8fadffc-a987-4063-9d93-6e2a59a38fae

## Testing

### DIFF
--- a/gui/src/context/LastSessionContext.tsx
+++ b/gui/src/context/LastSessionContext.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, useState, useEffect } from "react";
+import React, { createContext, useState } from "react";
 
 export const LastSessionContext = createContext<{
   lastSessionId: string | undefined;
@@ -12,13 +12,6 @@ export const LastSessionProvider: React.FC<{
   children: React.ReactNode;
 }> = ({ children }) => {
   const [lastSessionId, setLastSessionId] = useState<string | undefined>();
-
-  useEffect(() => {
-    const storedId = localStorage.getItem("lastSessionId");
-    if (storedId) {
-      setLastSessionId(storedId);
-    }
-  }, []);
 
   return (
     <LastSessionContext.Provider value={{ lastSessionId, setLastSessionId }}>
--- a/gui/src/context/LastSessionContext.tsx
+++ b/gui/src/context/LastSessionContext.tsx
@@ -1,0 +1,30 @@
+import React, { createContext, useState, useEffect } from "react";
+
+export const LastSessionContext = createContext<{
+  lastSessionId: string | undefined;
+  setLastSessionId: (id: string | undefined) => void;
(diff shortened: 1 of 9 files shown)


gui/src/context/LastSessionContext.tsx
```diff
@@ -0,0 +1,30 @@
...
+  setLastSessionId: () => {},
+});
+
+export const LastSessionProvider: React.FC<{
+  children: React.ReactNode;
+}> = ({ children }) => {
+  const [lastSessionId, setLastSessionId] = useState<string | undefined>();
+
+  useEffect(() => {
+    const storedId = localStorage.getItem("lastSessionId");
```
Patrick-Erichsen: We have a file with typings for all of our localStorage accessors: https://github.com/continuedev/continue/blob/main/gui/src/util/localStorage.ts

So I think we should add this there and update to use the `getLocalStorage` fn.

gui/tailwind.config.cjs
```diff
@@ -10,12 +10,15 @@ module.exports = {
   theme: {
     // Note that these breakpoints are primarily optimized for the input toolbar
     screens: {
-      xxs: "170px", // Smallest width for Primary Sidebar in VS Code
+      "2xs": "170px", // Smallest width for Primary Sidebar in VS Code
```
Patrick-Erichsen, line 13: Thanks for making this update! 😁

gui/src/context/LastSessionContext.tsx
tomasz-stefaniak: It's already there 😄 I'm just reusing an already existing value 
image

tomasz-stefaniak: This code actually wasn't needed. Removed it 👌

TyDunn (changes requested):
@tomasz-stefaniak Looks like a good next iteration! My two thoughts:
1. I suspect we'll want to increase the breakpoint a little, so that the history section is only added if the sidebar is expanded a bit wider than the current breakpoint
2. I think we should work on making the history page look / feel much nicer after we get this merged

tomasz-stefaniak:
> @tomasz-stefaniak Looks like a good next iteration! My two thoughts:
> 
> 1. I suspect we'll want to increase the breakpoint a little, so that the history section is only added if the sidebar is expanded a bit wider than the current breakpoint
> 2. I think we should work on making the history page look / feel much nicer after we get this merged

1. I updated it to 1180px from 1140px. I think it's the right balance - any wider and it wouldn't open at all unless the left bar is removed.
2. I agree - I wanted to get this PR out of the way quickly but there's plenty of room for improvement on the history page UI

sestinj (approved):

PR #2515, opened 2024-10-13
Improve Cmd + I menu

tomasz-stefaniak:
## Description

1. User highlights code
2. Presses cmd+I
3. Enters message
4. Presses enter
5. [NEW] Immediately show a new quick pick
    1. Accept all
    2. Reject all
    3. Text input to try again - can reuse “handleQuickPickChange”
    4. Close menu (if the list is too long we need to remove one option)
6. *diff begins streaming

## Checklist

## Screenshots

image

## Testing


extensions/vscode/src/quickEdit/QuickEditQuickPick.ts
```diff
@@ -253,9 +324,11 @@ export class QuickEdit {
 
     const fileName = vscode.workspace.asRelativePath(uri, true);
 
-    const { start, end } = !!this.range
-      ? this.range
-      : this.editorWhenOpened.selection;
+    if (!this.range) {
```
sestinj, line 312: probably worth just talking about whenever we next meet, but making a note for myself: what situation is this happening in?

extensions/vscode/src/quickEdit/QuickEditQuickPick.ts
```diff
@@ -326,31 +391,22 @@ export class QuickEdit {
...
-      alwaysShow: true,
-    };
+  async _getInitialQuickPickVal(): Promise<{
+    label: QuickEditInitialItemLabels | undefined;
+    value: string | undefined;
+  }> {
+    const modelTitle = await this.getCurModelTitle();
 
-    const noResultsItem: vscode.QuickPickItem = { label: "No results found" };
+    if (!modelTitle) {
```
sestinj, line 387: same here (I'm assuming it's just added error handling we were missing before, and nothing new)

sestinj (commented):
This works super well!! I was basically unable to break anything. Things that I feel would be important to add after having played around with this:
- we should leave the previous input in the input box when the accept/reject dialog shows up. In JetBrains what we do is `\`${previousInput}, \`` so that it sort of prompts the user to append instructions to the previous. I think right now typing new instructions just completely starts over
- I think that if the user 1) cmd+I and generates code 2) closes the dialog 3) presses cmd+I again while the new code blocks are still not accepted/rejected, then we should bring back up the accept/reject dialog

sestinj:
And glad to see the code in a cleaner state than before as well : )

PR #2534, opened 2024-10-15
Validate config.json

tomasz-stefaniak:
## Description

Validation of config.json

## Checklist

## Screenshots

Screenshot 2024-10-15 at 10 15 53 PM

image

## Testing


core/config/validation.ts
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Validates a SerializedContinueConfig object to ensure all properties are correctly formed.
+ * @param config The configuration object to validate.
+ * @returns An array of error messages. If the array is empty, the config is valid.
+ */
+export function validateConfig(config: SerializedContinueConfig) {
+  const errors: ValidationErrorMessage[] = [];
+
+  // Validate models
+  if (!Array.isArray(config.models) || config.models.length === 0) {
```
sestinj, line 17: Don't worry about changing it in this PR, but just making the note here that we can switch to using zod later (definition is in packages/config-types/src/index.ts)

gui/src/components/Footer.tsx
```diff
@@ -1,32 +1,38 @@
...
+    (store: RootState) => store.state.configError,
+  );
 
   function onClickMore() {
-    if (location.pathname === "/more") {
-      navigate("/");
-    } else {
-      navigate("/more");
-    }
+    navigate(pathname === ROUTES.MORE ? "/" : ROUTES.MORE);
```
sestinj, line 31: Nice

sestinj (commented):
I noticed a couple of things trying this out:
1) the warning icon turns out not to be super obvious. I think it is done perfectly for cases of non-fatal warnings, but if we have any errors that entirely break the config, then we might want to show it in a dialog or something. None of the current things seem necessarily critical though, so this can come in later work
2) it seems like the validation errors completely stop the config from loading, and I'm not actually sure that we want this. It's still possible to use the model, but it won't reflect the changes that the user thought they made in config.json. Especially when we want to return things that are just optional suggestions, we shouldn't stop the config from loading. This might mean we have to explicitly return rather than throwing and catching

core/config/validation.ts
tomasz-stefaniak: I tested it - it's very easy to switch to zod, it took about 5 minutes of work, but it's a bit too eager and fails even valid configs. We'd need to look at the zod schema and see if it's correct.

sestinj:
Sorry to be back and forth again on this—part of it is time zones—but I think we still need to both:
1) Give an extremely obvious/intrusive error dialog when a validation error is stopping the config from loading
and
2) Not stop the config from trying to load when we find validation errors

At least the latter is necessary when the validation errors are only "suggestion" status and not "critical". We should probably expand the `ValidationErrorMessage` type now to be an object with status and message to make sure we're doing this right

Thompson1985 (approved):

PR #2542, opened 2024-10-16
Better history UI

tomasz-stefaniak:
## Description

UI improvements

## Checklist

## Screenshots

image

image

HOVER STATE:
image

## Testing


TyDunn (changes requested):
This is looking really good! Fits with the rest of Continue much better. Here are some other thoughts:
- I still don't understand the "Show All" button. I also always forget that it only shows sessions from my current workspace by default. When I open Continue without a workspace, it then gives me the confusing "No past sessions found" even when I just had some sessions. I need to click "Show All" to reveal them
- I think we should make the "Open New Window" icon smaller to better fit with the other icons in the top right
- I think we should only have one way to go full screen
-

Patrick-Erichsen:
Agreed, looks really solid! I wouldn't be opposed to just making the history "Show all" by default. Seems sensible.

sestinj:
Screenshot 2024-10-16 at 12 57 10 PM

My feedback is just to add a faint (0.5px, light gray) line to separate the history from the full screen chat (like in above scenario)

tomasz-stefaniak:
> This is looking really good! Fits with the rest of Continue much better. Here are some other thoughts:
> 

> * I think we should make the "Open New Window" icon smaller to better fit with the other icons in the top right
> * I think we should only have one way to go full screen


> * I still don't understand the "Show All" button. I also always forget that it only shows sessions from my current workspace by default. When I open Continue without a workspace, it then gives me the confusing "No past sessions found" even when I just had some sessions. I need to click "Show All" to reveal them

> Agreed, looks really solid! I wouldn't be opposed to just making the history "Show all" by default. Seems sensible.

I agree, I also find that confusing. I didn't realize that's what the behavior was and it just seemed kind of random. I'd also rather see all by default, and maybe in the future we can add filters?

> I think we should make the "Open New Window" icon smaller to better fit with the other icons in the top right

I'll look into it, although as far as I know we don't set the sizes of these icons, they're all VSCode defaults. I might be wrong though.

> I think we should only have one way to go full screen

I think that would make sense, I just didn't want to gut existing features. Do you think we should get rid of this full screen view then?
image

> My feedback is just to add a faint (0.5px, light gray) line to separate the history from the full screen chat (like in above scenario)

I'll look into it. What theme is this so that I can debug locally? There is a gray line but it's not visible on this theme so I might have accidentally used the wrong variable.

TyDunn:
Yes, let's go with all by default for now and add more filters later

If we don't set size, then I think the padding should be increased around the icon if possible

Yes, I think we should pick one and roll with it

sestinj:
Given we're picking one full screen option, don't worry about the icon sizing (not worth us making a custom image, which is what would be required), just use the one we've previously been using

Probably use the `lightGray` variable. It's exactly in the middle of grayscale so should show up on any background (# 888)

tomasz-stefaniak:
Summary of the changes:

- Fixed the border but used a different color. lightGray was very hard on a black background
- Retained only one full screen mode, using the old icon
- Removed "*" / clear all from search to display all workspaces results by default. We can restore some of these filters later, under a dedicated menu

Thompson1985 (approved):

Issue #2519, opened 2024-10-14
java.lang.IndexOutOfBoundsException: Wrong line: -1. Available lines count: 1


danny-su:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 15.0.1
- Continue version: 0.0.75
- IDE version: IntelliJ IDEA 2024.2.3 (Ultimate Edition)
- Model:
- config.json:
```

### Description

java.lang.IndexOutOfBoundsException: Wrong line: -1. Available lines count: 1
	at com.intellij.openapi.editor.impl.LineSet.checkLineIndex(LineSet.java:197)
	at com.intellij.openapi.editor.impl.LineSet.getLineEnd(LineSet.java:191)
	at com.intellij.openapi.editor.impl.DocumentImpl.getLineEndOffset(DocumentImpl.java:1029)
	at com.github.continuedev.continueintellijextension.listeners.ContinuePluginSelectionListener.getSelectionInfo(ContinuePluginSelectionListener.kt:86)
	at com.github.continuedev.continueintellijextension.listeners.ContinuePluginSelectionListener.access$getSelectionInfo(ContinuePluginSelectionListener.kt:20)
	at com.github.continuedev.continueintellijextension.listeners.ContinuePluginSelectionListener$updateTooltip$1.invoke$lambda$0(ContinuePluginSelectionListener.kt:68)
	at com.intellij.util.concurrency.ContextRunnable.run(ContextRunnable.java:27)
	at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:229)
	at com.intellij.openapi.application.TransactionGuardImpl.access$100(TransactionGuardImpl.java:22)
	at com.intellij.openapi.application.TransactionGuardImpl$1.run(TransactionGuardImpl.java:191)
	at com.intellij.openapi.application.impl.AnyThreadWriteThreadingSupport.runIntendedWriteActionOnCurrentThread$lambda$1(AnyThreadWriteThreadingSupport.kt:184)
[...]

PR #2543, opened 2024-10-16
fix java.lang.IndexOutOfBoundsException: Wrong line: -1. Available li…

malaki12003:
…nes count: 1 #2519

## Description

When the code checks if the selection is a full-line selection (isFullLineSelection), it doesn't account for the case where the selection is on the first line of the code. It attempts to check the previous line, even though there is no code before the first line.

## Checklist

## Testing

In IntelliJ IDEA, when you select only part of the first line, the issue no longer occurs.


sestinj (approved):
very nice, thanks for the fix!

PR #2544, opened 2024-10-16
added cerebras provider

varadanvk:
## Description

Added [Cerebras Systems](cerebras.ai) (Fast inference) as a provider.  Cerebras is a GPU manufacturer that focuses on large GPUs. They have Llama 3.1 8b running at 2k tokens/second. 

Currently they're in beta for their cloud API, so only some people can get access to an API key, but I think it would still be a good idea to have this available as it's super fast. Right now, they only support Llama 3.1 8B and 70B.

## Checklist

- [x] The relevant docs, if any, have been updated or created (added Cerebras to provider docs as per the contributing guide)

## Screenshots

Not much to show, 

image

It works!

## Testing

Add a new model and try to use the cerebras api key. They have limited access right now so only some people can access. 

## Possible Issues
I noticed that the PNG image I added for the Cerebras logo didn't show once or twice in the "add provider" page. Not sure if this is a problem or not


varadanvk:
someone else already did this nvm

PR #2548, opened 2024-10-17
Prettier + Tailwind

tomasz-stefaniak:
## Description

Added prettier tailwind plugin to automatically order prettier class names in a sensible way.

IMPORTANT: Please check that I didn't break the Windows installation script, I can't test it locally.

## Checklist

## Screenshots

## Testing

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "tabWidth": 2,
   "useTabs": false,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "plugins": ["prettier-plugin-tailwindcss"]
 }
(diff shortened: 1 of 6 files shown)


Patrick-Erichsen:
This will be great! 

I think after we merge this we should follow up with a second PR to apply this rule to all of the files.

tomasz-stefaniak:
> This will be great!
> 
> I think after we merge we should follow up with a second PR to apply this rule to all of the files.
> 
> I'm uncertain if adding a top level package.json is the appropriate way to go about adding this however. Given that the gui folder is the only place we use tailwind, could we just install the plugin to that folder?

We could do that but then we'd need a separate `.prettierrc` for the gui. I figured we might need some root-level dependencies and commands at some point anyways. For example, I'd like to add a `tsc:check:all` command at the root level, which checks for type errors in all packages at once, etc.

sestinj:
It's true, there's a PR open for npm workspaces support that I'm planning to merge soon so I think this is okay

sestinj (approved):

Thompson1985 (approved):

Issue #2495, opened 2024-10-08
Handle 5xx errors more gracefully in Chat


Patrick-Erichsen:
### Before submitting your bug report

- [ ] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Description

When we receive a 5xx error from a provider while streaming Chat responses, there's a few ways we could improve the experience:

- Make it clearer that the error is on the provider, not Continue, and recommend retrying the prompt
- Don't clear out the chat history (e.g. 90% of the message was streamed, last 10% hits a 5xx - current behavior is to clear the 90% that has already streamed)

We've had multiple users report 529 errors from Anthropic, so this issue has a relatively wide impact.

---

Original user feedback: https://discord.com/channels/1108621136150929458/1292863736100687923/1292863736100687923

> Yeah. I personally don't mind the current notification or going into the logs, but it was entirely unclear that the overloaded issue was just the API failing. 

> However, my biggest suggestion would be in how the chat feature handles this type of error. The most frustrating part of this is how the completion completely disappears from the chat history. This is especially annoying when the completion is like 90% done, and I don't really even need the part that failed. 

> My ideal situation would be this:
If a completion fails, Continue shows the error modal like it is, but with a more clear message indicating that Anthropic's API is what failed. Continue keeps the portion of the completion that was successful in the window until I click a "retry" or "continue anyways" button. 

> I am not really sensitive to pricing and I am not sure if Anthropic charges you for tokens in a failed completion, but I feel like some Continue customers could be irritated that they lost tokens they paid for if Continue wipes the completion on error. 

---

Implementation

I think a good UX here could be to put the error notification directly into the Chat rather than a popup.

drcrallen:
I'm in agreement with the discord thread. This is pretty terrible experience with the Continue VS Code plugin. The current behavior drops any partial response AND your chat input. So if you are typing a complex query, and half way through the response you get an error like
```
Malformed JSON sent from server: {"type":"error","error":{"details":null,"type":"overloaded_error","message":"Overloaded"} }
```
[...]

PR #2558, opened 2024-10-18
Handle 5xx errors more gracefully in Chat #2495 #2557

malaki12003:
## Description

Make it clear that the error is on the provider’s side, not Continue’s, and ensure the chat history is preserved.

image

## Checklist

### DIFF
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -596,8 +596,9 @@ export abstract class BaseLLM implements ILLM {
 
     let completion = "";
 
-    try {
-      if (this.templateMessages) {
+   
+    if (this.templateMessages) {
+      try {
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen (changes requested):
Hey @malaki12003 , thanks for the contribution here! 

A few points of feedback
- Instead of looking at the string message of the error which can be vary, we should look at the status code
- We don't want to yield the error text back into the chat. The original thought here was just to give a clearer description in the error toast to show that the issue isn't with Continue but instead with the upstream provider

sestinj:
I'm seconding Patrick's comments here. I think it's probably good to start fresh here, so I'm going to close this particular PR. If we could still use the Toast, but just display some more useful message that would help a lot

PR #2574, opened 2024-10-19
fix: history bug by writing to history after chat completion.

kyleli626:
## Description

Fixed bug where last message of history could be lost when switching between chats. 

Step to replicate:
1. create a new chat 
2. send a message
3. go to another chat thru history
4. return to the new chat thru history
5. type another
6. go to history and click on the same new chat

bug - https://www.loom.com/share/e912c87c229d48e59f36d9af26133949
fixed - https://www.loom.com/share/dfda1e458f204b72985c9962af65d86e

## Checklist

## Testing
tried again. 

Using effect might be against best practices! Please let me know what you think is the best way to do this. Calling saveSession directly didn't work in the finally clause - it wouldn't register the new message or dispatched state changes. 

We might want to remove the pattern of calling saveSession to spawn a new session as well. This happens in a few other spots, and is the reason why clicking the "history" button and reselecting the same chat discards the last message. I added an optional parameter here, but it makes sense to just completely replace all other saveSessions with just like a newSession button, now that the latest message is automatically saved.

### DIFF
--- a/extensions/vscode/package-lock.json
+++ b/extensions/vscode/package-lock.json
@@ -116,6 +116,7 @@
         "dotenv": "^16.4.5",
         "fastest-levenshtein": "^1.0.16",
         "follow-redirects": "^1.15.5",
+        "google-auth-library": "^9.14.2",
         "handlebars": "^4.7.8",
         "http-proxy-agent": "^7.0.1",
         "https-proxy-agent": "^7.0.3",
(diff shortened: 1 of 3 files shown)


Patrick-Erichsen:
Thanks for the writeup on `saveSession` - I'd agree with your logic that it's not the best way to spawn a new session. For now though this is a good improvement 👍  Much appreciated!

Issue #2538, opened 2024-10-15
'inferenceConfig.stopSequences' failed to satisfy constraint: Member must have length less than or equal to 4


hongbo-miao:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

- OS: macOS 15.0 (24A335)
- Continue version: 0.0.75
- IDE version: IntelliJ IDEA 2024.1.4 & VSCode 1.94.1
- Model: Claude 3.5 Sonnet through provider AWS Bedrock
- config.json:
  
```json
{
  "models": [
    {
      "title": "claude-3-5-sonnet",
      "provider": "openai",
      "apiBase": "https://litellm.example.com",
      "apiKey": "xxx",
      "model": "claude-3-5-sonnet"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Tab Autocomplete",
    "provider": "openai",
    "apiBase": "https://litellm.example.com",
    "apiKey": "xxx",
    "model": "claude-3-5-sonnet",
    "maxStopWords": 4
  },
  "tabAutocompleteOptions": {
    "debounceDelay": 500,
    "maxPromptTokens": 1500
  },
  "slashCommands": [
    {
      "name": "edit",
      "description": "Edit highlighted code"
    },
    {
      "name": "comment",
      "description": "Write comments for the highlighted code"
    },
    {
      "name": "share",
      "description": "Export the current chat session to markdown"
    },
    {
      "name": "cmd",
      "description": "Generate a shell command"
    }
  ],
  "customCommands": [
    {
      "name": "test",
      "prompt": "{{{ input }}}\n\nWrite a comprehensive set of unit tests for the selected code. It should setup, run tests that check for correctness including important edge cases, and teardown. Ensure that the tests are complete and sophisticated. Give the tests just as chat output, don't edit any file.",
      "description": "Write unit tests for highlighted code"
    }
  ],
  "contextProviders": [
    {
      "name": "diff",
      "params": {}
    },
    {
      "name": "open",
      "params": {}
    },
    {
      "name": "terminal",
      "params": {}
    }
  ],
  "embeddingsProvider": {
[...]

PR #2606, opened 2024-10-21
Fix Max Stop Words not being set for openai and watsonx models

Lash-L:
## Description

A model i'm using with continue requires under 6 stop words. I tried setting maxStopWords, realized it did not work and looked on github for issues. I found #2538. 

After doing some debugging, I realized that maxStopWords was set during the super constructor, but blank in the child classes constructor.

This was because of the class level setting of maxStopWords = undefined. Removing that, it correctly accesses this.maxStopWords during _convertArgs()

## Checklist

## Screenshots

## Testing

- Use a openAi model
- Set maxStopWords
- Verify stopWords is under the limit specified


sestinj (approved):
thanks!

PR #2607, opened 2024-10-21
Respect stream argument for openAI

Lash-L:
## Description

stream is set inside of convertArgs based off of the config or defaulted to true. By overriding it here, you make it impossible to set it as anything different inside the config.

## Checklist

## Screenshots

## Testing

### DIFF
--- a/core/llm/llms/OpenAI.ts
+++ b/core/llm/llms/OpenAI.ts
@@ -204,7 +204,6 @@ class OpenAI extends BaseLLM {
       headers: this._getHeaders(),
       body: JSON.stringify({
         ...args,
-        stream: true,
       }),
     });
 


sestinj:
@Lash-L Are you using

```
"completionOptions": {
        "stream": true
      }
  ```

to set stream: false?


I think this is partially a problem of ill-defined logic, but rather than forcing all of the instances of _streamComplete to also support non-streaming, it would be better to call _streamComplete vs. _complete depending on whether `stream` is set to true or false

Lash-L:
> @Lash-L Are you using
> 
> 
> 
> ```
> 
> "completionOptions": {
> 
>         "stream": true
> 
>       }
> 
>   ```
> 
> 
> 
> to set stream: false?
> 
> 
> 
> 
> 
> I think this is partially a problem of ill-defined logic, but rather than forcing all of the instances of _streamComplete to also support non-streaming, it would be better to call _streamComplete vs. _complete depending on whether `stream` is set to true or false

Yeah, that's what I am currently doing - your approach seems much better. I'll see if I can change it

Lash-L:
So the problem with calling stream complete vs complete is that complete calls _streamChat, which in turn calls _legacystreamComplete.

sestinj:
I see what you mean. Seems what we should do here is write _complete so that it uses its own logic instead of _streamChat, and generally untangle the logic where everything points toward _streamChat

RomneyDa:
https://github.com/continuedev/continue/pull/2700
did this for ollama

sestinj:
Hey @Lash-L, I'm going to close this for now since the scope has changed a bit—I think the right thing to do here is untangle the _streamChat and _streamComplete functions first

Lash-L:
> Hey @Lash-L, I'm going to close this for now since the scope has changed a bit—I think the right thing to do here is untangle the _streamChat and _streamComplete functions first

Sounds good, thank you!

PR #2608, opened 2024-10-21
Editor pass on getting-started/install

shermanhuman:
## Description

1. Removed lines in between list items, a markdown list shouldn't have lines in between items  
2. Added padding to list items in custom.css (_which is where I think we are supposed to place any changes to the standard docusaurus theme_)
3. Used the `<kbd>` tag instead of code ticks to indicate keys - it was a little unclear if '+' was a key be pressed
4. Created custom style for kbd for both light and dark themes 

## Checklist

## Screenshots

image

image

## Testing

Run a local docusaurus instance and test locally.  =)

### DIFF
--- a/docs/docs/getting-started/install.md
+++ b/docs/docs/getting-started/install.md
@@ -18,7 +18,7 @@ If you have any problems, see the [troubleshooting guide](troubleshooting.md) or
 
 ## JetBrains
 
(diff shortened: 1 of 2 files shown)


sestinj:
@shermanhuman this looks really nice!! I hadn't even considered that it was possible. Thanks for the awesome PR : )

PR #2618, opened 2024-10-22
Add Not Diamond Provider

dragos199993:
## Description
Hi there! I am a big fan of Continue and would love to integrate Not Diamond into it. In case you’re unfamiliar with Not Diamond, it automatically routes your prompts to the AI model best-suited to respond to that prompt. By using Not Diamond as an AI “meta-model,” you can achieve state-of-the-art performance that beats every foundation model on every major benchmark.*

## Screenshots

image
image

## Testing

Simply add Not Diamond api key and the preferred providers api keys and it will work out of the box.

We didn't tried to battle how Continue is built, we only added few inputs to match the Not Diamond requirement.

**If you’d like to learn more about our routing, check out [[this example walkthrough video with Tomas](https://x.com/tomas_hk/status/1818296673749328199)](https://x.com/tomas_hk/status/1818296673749328199), our CEO. You can also* 

- *see how we work by using [[our Chat app](https://chat.notdiamond.ai/)](https://chat.notdiamond.ai/),*
- *learn more at [[our website](https://www.notdiamond.ai/)](https://www.notdiamond.ai), or*
- *[[read our docs](https://notdiamond.readme.io/)](https://notdiamond.readme.io/).*


sestinj:
@dragos199993 very excited to see this! A couple of quick observations:
- It looks like there is an extraneous .idea file or two in the PR
- Is Not Diamond OpenAI-compatible? (or at least close to this except having an array as the models parameter and a small handful of others). It would be quite nice if we could use the existing logic from the OpenAI class rather than importing the separate notdiamond-continue package (which appears to have a lot of dependencies itself)

The second point would be the biggest for us—ideally we can support NotDiamond requiring no dependencies and just using fetch, and something close to OpenAI-compatibility would probably make that easiest

dragos199993:
> @dragos199993 very excited to see this! A couple of quick observations:
> 
> * It looks like there is an extraneous .idea file or two in the PR
> * Is Not Diamond OpenAI-compatible? (or at least close to this except having an array as the models parameter and a small handful of others). It would be quite nice if we could use the existing logic from the OpenAI class rather than importing the separate notdiamond-continue package (which appears to have a lot of dependencies itself)
> 
> The second point would be the biggest for us—ideally we can support NotDiamond requiring no dependencies and just using fetch, and something close to OpenAI-compatibility would probably make that easiest

Likewise, @sestinj 🚀 !

I addressed the first point and fixed the conflict (the version in package.json).

Unfortunately on the second point it might be a bit more complicated and the results might be more verbose 😢
We are not fully compatible with OpenAI and we use inside `notdiamond-continue` a few langchain packages to handle the streaming for multiple providers and models.

If we do a fetch on `model-select` we only get the name of the provider + model, but that's all 😢. I could try to reduce the amount of dependencies from `notdiamond-continue` because (as the same suggests) it's specifically for the continue integration. I can make it as slim as possible, only langchain stuff to be there.

Or if you recommend a different approach, I'll gladly address it 🙏

sestinj:
I really do appreciate the dedication to make a package just for Continue. Unfortunately with langchain included at all, we won't be able to merge. It's 3MB, which is a non-trivial amount when we have a limit for the JetBrains extension store.

I do feel that it is going to be necessary for Not Diamond to become OpenAI-compatible anyway if other tools like Continue are to integrate easily. Whatever work is being done on the client should probably be done on the server. That said, I do admit that the extent to which this feels obvious to me might also be the extent to which I am missing something about how Not Diamond is able to provide privacy-preserving routing. Is there a more fundamental limitation?

acompa:
> I really do appreciate the dedication to make a package just for Continue. Unfortunately with langchain included at all, we won't be able to merge. It's 3MB, which is a non-trivial amount when we have a limit for the JetBrains extension store.
> 
> I do feel that it is going to be necessary for Not Diamond to become OpenAI-compatible anyway if other tools like Continue are to integrate easily. Whatever work is being done on the client should probably be done on the server. That said, I do admit that the extent to which this feels obvious to me might also be the extent to which I am missing something about how Not Diamond is able to provide privacy-preserving routing. Is there a more fundamental limitation?

Hi there @sestinj. You're correct about our privacy-focused approach to routing, and how we avoid offering proxied (server-side) invocation as a result of that.

Would it suffice for us to skip the Langchain dependency, instead relying on Continue's already-existing dependencies on OpenAI and Anthropic TS client libraries? With this approach we could still submit requests client-side via those libs, without introducing a large binary dependency which violates size limits for JetBrains extensions.

sestinj:
@acompa Thanks a lot for the clarification, yes! This would be completely sufficient for me to merge

Issue #2437, opened 2024-10-01
Disable extension notifications / popups


evertjr:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue: v0.9.213 (pre-release)
- IDE: vscode
- Model:
- config.json:
```

### Description

For many reasons this extensions triggers popups in vscode and it does not respect our notifications preference so they keep appearing without consent.

This is especially distracting for notifications such as this:

```
Warning: qwen2.5-coder:1.5b-instruct-q8_0 is not trained for tab-autocomplete, and will result in low-quality suggestions. See the docs to learn more about why: https://docs.continue.dev/features/tab-autocomplete#i-want-better-completions-should-i-use-gpt-4
```

it's okay to show this warning once but not every single time, specially since this is a false positive, the model works fine.

### To reproduce

1. check "turn off all notifications from Continue..."
2. still get notifications from continue

Patrick-Erichsen:
Hi @evertjr , thanks for the heads up here. Could you share where you have configured notifications to be disabled? I'm not seeing much in the VS Code docs around how to check this setting so that we can then respect it.

evertjr:
e believe it's from the native Notification Center, this notification shows up even if I enable do not disturb.

<img width="620" alt="image" src="https://github.com/user-attachments/assets/13b0c533-7b32-420b-9761-de82246850a9">

RomneyDa:
Hi @evertjr, 

Would agree that it should not be throwing an error; we can change this to a warning message, since error messages will be shown regardless of the info/warning settings to make sure real errors aren't silent. Any other specific errors notifications that have been showing incessantly?

Notifications playground: clone and debug this repo: [Sample notifications extension](https://github.com/microsoft/vscode-extension-samples/tree/main/notifications-sample)

RomneyDa:
note to self, error messages are shown by:
- uncaught error in core, ide, or gui
- ide.showToast("error" | "warning" | "info", ...) from core
- 
Quick scan through the ~200 Error throws and show, all look like valid error events except this one and one other. Will open PR

[...]

PR #2619, opened 2024-10-22
Replace tab-autocomplete untrained model error with warning

RomneyDa:
## Description

Untrained autocomplete model was shown as error. 
Disabling extension notifications didn't prevent this.
Changing it to warning (showToast("warning", ...))

## Checklist

- [x ] The base branch of this PR is `dev`, rather than `main`
- [ x] The relevant docs, if any, have been updated or created

See [issue 2437](https://github.com/continuedev/continue/issues/2437)

### DIFF
--- a/core/autocomplete/completionProvider.ts
+++ b/core/autocomplete/completionProvider.ts
@@ -480,9 +480,8 @@ export class CompletionProvider {
       !llm.model.toLowerCase().includes("codestral")
     ) {
       shownGptClaudeWarning = true;
-      throw new Error(
-        `Warning: ${llm.model} is not trained for tab-autocomplete, and will result in low-quality suggestions. See the docs to learn more about why: https://docs.continue.dev/features/tab-autocomplete#i-want-better-completions-should-i-use-gpt-4`,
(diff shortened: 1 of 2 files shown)


sestinj (commented):
Looks great—it just occurred to me that we might be able to do something even nicer than this: Tomasz recently added the ability to emit validation errors while we are loading config.json, and then display them in the GUI. Going forward, we're hoping that this can be a single place for us to point users in the right direction when either their config.json is entirely broken, or they've just made less than ideal decisions. This would fit very nicely.

Want to add a check and `errors.push` a non-fatal warning [here](https://github.com/continuedev/continue/blob/nate/autocomplete-tests/core/config/validation.ts#L14)?

RomneyDa:
@sestinj updated

Could also remove the warning popup and _only_ show it in config, with the reasoning that it's a duplicate message and seems to be the kind of message that would make people disable continue notifications anyways.

sestinj:
Yeah that's a good call

RomneyDa:
toast removed

PR #2633, opened 2024-10-23
Fix false positives when checking if a file is in a workspace

kylewlacy:
## Description

Depending on the workspace layout, I would sometimes get this error toast from Continue:

VS Code error toast with the following message: path should be a `path.relative()`d string, but got "..\foo-bar\test3.js"

## Checklist

## Screenshots

N/A

## Testing

This fixes a bug when a directory in a workspace has a path that overlaps with another directory in a workspace. In this case, `path.relative()` would return a path containing `..` components. The fix was to tweak the logic to detect if a path is contained within another path to make it more robust ([this Stack Overflow answer](https://stackoverflow.com/a/45242825) gives some helpful context, although I went with a different solution because I believe that answer _also_ has a bug if one of the paths has a name starting with `..`, such as `..foo`)

Anyway, here's how I reproduced this issue:

2. Create an empty directory called `foo`
3. Create another directory next to it called `foo-bar`
4. Open `foo` in VS Code
5. In the Command Palette, run "Workspaces: Add Folder to Workspace...", then add `foo-bar` to the workspace
6. Create a file under `foo-bar`, then edit it with inline completions enabled

On both the latest release (v0.8.54) and pre-release (v0.9.217) versions of the Continue VS Code extension, this reliably triggers an error for me. With this PR, the error goes away and completions work as expected.

### DIFF
--- a/core/autocomplete/completionProvider.ts
+++ b/core/autocomplete/completionProvider.ts
@@ -268,7 +268,10 @@ export class CompletionProvider {
         const workspaceDirs = await this.ide.getWorkspaceDirs();
         let filepath = input.filepath;
         for (const workspaceDir of workspaceDirs) {
-          if (filepath.startsWith(workspaceDir)) {
+          const relativePath = path.relative(workspaceDir, filepath);
+          const relativePathBase = relativePath.split(path.sep).at(0);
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen:
Great catch here @kylewlacy ! Appreciate the detailed writeup as well.

PR #2655, opened 2024-10-24
Added support for MSSQL to the DatabaseContextProvider

DifferentialityDevelopment:
## Description

I've added support for MSSQL to the DatabaseContextProvider, honestly just a few config tweaks were required.
The package dbinfoz that DatabaseContextProvider relies on has already been updated to support MSSQL.

## Checklist

## Screenshots

## Testing

Install SQL Server
Ensure local connectivity to SQL Server
Update continue.dev config for @database context provider and add mssql connection paramaters.
Try out @database with the mssql server

### DIFF
--- a/docs/static/schemas/config.json
+++ b/docs/static/schemas/config.json
@@ -1737,6 +1737,7 @@
                 "highlights",
                 "outline",
                 "postgres",
+                "mssql",
                 "code",
                 "currentFile",
                 "url",
@@ -1986,7 +1987,7 @@
                         "connection_type": {
                           "type": "string",
                           "description": "The type of database (e.g., 'postgres', 'mysql')",
-                          "enum": ["postgres", "mysql", "sqlite"]
+                          "enum": ["postgres", "mysql", "sqlite", "mssql"]
                         },
                         "connection": {
                           "type": "object",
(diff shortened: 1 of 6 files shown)


DifferentialityDevelopment:
I closed my previous PR and instead made the changes on the dev branch.

docs/static/schemas/config.json
```diff
@@ -1737,6 +1737,7 @@
                 "highlights",
                 "outline",
                 "postgres",
+                "mssql",
```
Patrick-Erichsen:
```suggestion
```
What this would be doing is implying that there is a new context provider called "mssql" - however, what we're actually doing is adding a `connection_type` to the existing `database` context provider, so this line is not needed.

Patrick-Erichsen (changes requested):

docs/static/schemas/config.json
Patrick-Erichsen: I can see the source of your confusion here though since we have a `postgres` provider in addition to the `database` provider.

I'm not actually sure why that's the case. Ideally we just have the single `database` provider.

@rootedbox @sestinj - any context here? Would it be possible to remove `postgres` and consolidate on a single `database` provider using "dbzinfo"?

rootedbox: @Patrick-Erichsen we might want to remove the postgres provider as the database provider can do the same thing.. and more. What do you think?

Which I think partially led to this confusion.

sestinj: Yes, I am generally in agreement. I believe they were contributed at a similar point in time. There are certainly people using the postgres provider so we'll probably want to either throw in a validation warning to help people migrate (in the new validation.ts)

sestinj (changes requested):
It is my understanding from the thread above that we can leave this PR to just being the upgrade to dbinfoz?

DifferentialityDevelopment:
> It is my understanding from the thread above that we can leave this PR to just being the upgrade to dbinfoz?

Not entirely correct, the arguments for mssql is slightly different than for postgres, namely where postgres you pass 'host' argument, in mssql you pass 'server' argument instead, so a few of my changes were to add the necessary validation for it.

sestinj (approved):

DifferentialityDevelopment:
Hey there!

I've got the latest version installed, also tried release version but for some reason still shows "Unsupported database type" when trying to link up my local mssql db in my continue.dev config file.
Not sure how to go about fixing it, from what I can tell the dbinfoz package is indeed correct version in continue.dev's package.json ie 0.14.0.

Anyone have any idea how to fix it?

DifferentialityDevelopment:
image
I opened up the installed extension folder and searched for dbinfo in the compiled code but cannot find the mssql adapter there.
But it does seem to be using dbinfoz package 0.14.0 so not sure what's going on.

DifferentialityDevelopment:
Perhaps the old code is being cached somehow during the workflow?
https://github.com/continuedev/continue/blob/main/.github/workflows/main.yaml

DifferentialityDevelopment:
Perhaps the old code is being cached somehow during the workflow?
https://github.com/continuedev/continue/blob/main/.github/workflows/main.yaml
@Patrick-Erichsen @rootedbox

DifferentialityDevelopment:
This has recently stopped working for whatever reason, when I find some time I'll try and figure out why it stopped working, but any help would be appreciated.

PR #2667, opened 2024-10-25
fix config json schema validation

RomneyDa:
only uses config_schema.json validation for config.json files in the .continue folder
adds separate check/plugin for continue_rc_schema.json json schema validation in intellij
and copying of continue_rc_schema.json to intellij resources

Working in VsCode
NOT WORKING IN INTELLIJ - NEED TO FIGURE OUT WHY NOT GETTING SUGGESTIONS IN INTELLIJ


RomneyDa:
seems like suggestions in intellij will show, you just have to manually enable the json suggestions in the setttings

sestinj:
@RomneyDa merge on your own once file is renamed

PR #2700, opened 2024-10-27
add missing Ollama support for stream etc.

RomneyDa:
## Description
- Ollama provider typing/methods made more readable/sensible
- stream and options.minP handling added

## Checklist

### DIFF
--- a/core/llm/llms/Ollama.ts
+++ b/core/llm/llms/Ollama.ts
@@ -8,6 +8,59 @@ import { stripImages } from "../images.js";
 import { BaseLLM } from "../index.js";
 import { streamResponse } from "../stream.js";
 
+interface OllamaChatMessage extends ChatMessage {
+  images?: string[];
+}
+
+// See https://github.com/ollama/ollama/blob/main/docs/modelfile.md for details on each parameter
+interface ModelFileParams {
+  mirostat?: number;
+  mirostat_eta?: number;
+  mirostat_tau?: number;
+  num_ctx?: number;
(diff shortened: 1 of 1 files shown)


RomneyDa:
wonderings:
- frequencyPenalty or presencePenalty correlate to repeatPenalty?
- For `api/generate` Not supported yet?: `context`, `images`, `system`, `template`
- template is not included in `BaseCompletionOptions` and never passed in actual api call but `template` and `promptTemplate` are in `config_schema`?

sestinj:
repeatPenalty should equal frequencyPenalty

It's okay not to allow images for /api/generate—we'd only use images in chat (at least for as far as I can see into future)

context I believe is just another way of passing in the context of the current chat session, which we already do by sending chat messages

system you are right to call out (though generally /api/generate should actually be incredibly rare to use over chat so this is all fairly low priority

`template` and `promptTemplates` both correspond to general properties of the `BaseLLM` class: https://github.com/continuedev/continue/blob/b9d3ef4a77041711e3445f0f75647b3eda8da54c/core/llm/index.ts#L141-L142 and https://github.com/continuedev/continue/blob/b9d3ef4a77041711e3445f0f75647b3eda8da54c/core/llm/index.ts#L171-L174

`template` being the name of a template that we have pre-made, like "alpaca" or something. "promptTemplates" being customizable way to change the Edit template, though both of these are fairly legacy now that a) every provider handlers templating for us (almost) and b) people probably shouldn't really be able to come up with a better Edit prompt than we can

Issue #2688, opened 2024-10-25
LRU Cache requests Sqlite LIKE or GLOB pattern too complex


RomneyDa:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
OS: Windows_NT x64 10.0.22631 (Windows 11 Home)
Continue version: I think v0.8.55, I got this from checking the extension page. Not sure if it s showing currently instaleld or latest version
IDE version: 1.94.2 (user setup)
Model: starcoder:3b
config.json:
  "tabAutocompleteModel": [
    {
    "title": "Free Trial",
    "provider": "free-trial",
    "model": "codestral-latest"
    },
    {
      "title": "Starcoder 3B Q8",
      "provider": "lmstudio",
      "model": "brittlewis12/starcoder2-3b.Q8_0.gguf",
      "contextLength": 16384,
      "completionOptions": {
        "maxTokens": 1024
      }
    },
    {
      "title": "DeepSeek Coder V2 16b Q8",
      "provider": "lmstudio",
      "model": "bartowski/DeepSeek-Coder-V2-Lite-Instruct-GGUF/DeepSeek-Coder-V2-Lite-Instruct-Q8_0.gguf",
      "contextLength": 16384,
      "completionOptions": {
        "maxTokens": 1024
      }
    }
  ],
  "tabAutocompleteOptions": {
    "maxPromptTokens": 16384
  },
```

### Description

Per @la la land on discord

>All of a sudden I keep getting: SQLITE_ERROR: LIKE or GLOB pattern too complex
>SQLITE_MAX_LIKE_PATTERN_LENGTH can't be updated at runtime

seems to be happening in Autocomplete LRU cache 
https://www.sqlite.org/limits.html#max_like_pattern_length
default length is 50000

SQLITE_MAX_LIKE_PATTERN_LENGTH can't be updated at runtime, so for now you'd have to recompile the db with this setting adjusted to be higher
#define SQLITE_MAX_LIKE_PATTERN_LENGTH 1000000

[...]

PR #2703, opened 2024-10-27
Ensure don't go over sqlite max like pattern length

RomneyDa:
## Description

Truncate LIKE comparison pattern in autocomplete LRU cache

## Checklist

See discussion in 
https://github.com/continuedev/continue/issues/2688

### DIFF
--- a/core/autocomplete/cache.ts
+++ b/core/autocomplete/cache.ts
@@ -4,11 +4,19 @@ import sqlite3 from "sqlite3";
 import { DatabaseConnection } from "../indexing/refreshIndex.js";
 import { getTabAutocompleteCacheSqlitePath } from "../util/paths.js";
 
+const SQLITE_MAX_LIKE_PATTERN_LENGTH = 50000;
+
+function truncateToBytes(input: string, maxBytes: number): string {
+  let bytes = 0;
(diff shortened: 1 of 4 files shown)


RomneyDa:
This is one way but maybe better, reevaluate how the prefix is constructed and why it would get so long:

https://github.com/continuedev/continue/blob/334705c0a3c4fd081c22faf13f9f5478a57600d4/core/autocomplete/completionProvider.ts#L536-L601

sestinj:
@RomneyDa this is a good backup to have so thanks for doing it. And yes we should also re-evaluate, though I'd asked the person whether changing to smaller contextLength/maxPrefixTokens solves it, as this is a likely explanation.

One thing to change on this PR before we merge: the LIKE pattern (at least the autocomplete one) should be truncated from the front first. Right now we'll leave out chars at the end, but if it's too long we should be leaving out tokens at the start

RomneyDa:
@sestinj updated

sestinj:
@RomneyDa thanks this is great. I made a small change myself because it was super minimal and biased to just getting this merged, but I threw in a few (Continue-generated) tests to show examples of what was intended here

PR #2710, opened 2024-10-28
Add Nebius AI Studio LLM and embedding providers

vadjs:
## Description

Added Nebius AI Studio LLM and embedding provider.

## Checklist

## Screenshots

No visual changes

## Testing

It's not a feature, it's a new provider. The testing is simple: User Nebius provider with their API key. It just works.


core/llm/llms/Nebius.ts
RomneyDa, line 1: Hi @vadjs, nice addition, looks like Nebius is OpenAI compatible, could you switch this out for a simpler class that extends OpenAI? 

Kind of like:
https://github.com/continuedev/continue/blob/dev/core/llm/llms/Kindo.ts

or did I miss a reason for extending BaseLLM?

RomneyDa (changes requested):

sestinj (changes requested):
Thanks for the PR! I'd like to second the comment from Dallin: since it appears Nebius is OpenAI-compatible, I would prefer that this is just a subclass of the OpenAI class. I also noticed that the default base URL for the embeddings provider was a Voyage AI URL, I think this needs to be cleaned up

core/llm/llms/Nebius.ts
vadjs: Hi @RomneyDa, thank you for your advise!
I've changed `Nebius` provider accordingly. Now it extends `OpenAI`.

vadjs:
Hi @sestinj,
Thank you for the good point. I've changed apiBase URL in the `NebiusEmbeddingsProvider` to the correct one.
Also extended `Nebius` LLM provider from `OpenAI` class as proposed before.

sestinj (approved):

sestinj:
Awesome, thank you!

PR #2722, opened 2024-10-28
Dallin/lint modules

RomneyDa:
cache core modules
run vscode lint command in checks


RomneyDa:
tested by adding dallin/lint-modules to the `on` in yaml temporarily

Issue #2734, opened 2024-10-30
Recently Edited files cache doesn't check ignore patterns


RomneyDa:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
would affect all versions
```

### Description

Recently edited env files, etc. are getting injected into the context

### To reproduce

recently open/edited e.g. `.env` files aren't checked for gitignore/continueignore patterns

PRANJALRANA11:
@sestinj could you assign this to me also

sestinj:
For sure! Recommend using our built-in ignore.ts functionality to implement. Also we'll probably want to make this happen before end of tomorrow

PRANJALRANA11:
@RomneyDa could you tell where it is getting injected ! like when we do @ there it shows up the file or after editing it's automatically gets injected

RomneyDa:
@PRANJALRANA11 
injected here
https://github.com/continuedev/continue/blob/dev/core/core.ts

check if file should be ignored on adding to recently edited cache, current file, files, etc.
see https://github.com/continuedev/continue/blob/dev/core/indexing/walkDir.ts

Don't worry about active/current file injection yet, will worry about that later

RomneyDa:
@PRANJALRANA11 let me know if you start on this in the next bit, if not can knock out in the morning, high priority

PRANJALRANA11:
@RomneyDa yep doing it rit now

RomneyDa:
awesome! sorry for delayed response

PRANJALRANA11:
@RomneyDa yep done

RomneyDa:
partially fixed by https://github.com/continuedev/continue/pull/2751
leaving open for more thorough fix

RomneyDa:
Fixed by #4531

PR #2751, opened 2024-10-31
add check for ignore for recently edited file

PRANJALRANA11:
fix for #2734 

added check for ignore for recently edited file

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -704,15 +704,40 @@ export class Core {
       const url = await getAuthUrlForTokenPage();
       return { url };
     });
-
     on("didChangeActiveTextEditor", async ({ data: { filepath } }) => {
-      const dirs = await this.ide.getWorkspaceDirs();
-      const paths = await walkDir(dirs[0], this.ide);
-      if (paths.includes(filepath)) {
+      let dirs = await this.ide.getWorkspaceDirs();
+      let memoizedFunction = await this.memoizedFunction(dirs);
+      if (memoizedFunction.includes(filepath)) {
+        console.log("yep running");
         recentlyEditedFilesCache.set(filepath, filepath);
       }
     });
   }
+
+  private memoize<T extends (...args: any[]) => Promise<any>>(
+    fn: T,
+  ): (...args: Parameters<T>) => Promise<ReturnType<T>> {
+    const cache = new Map<string, Promise<ReturnType<T>>>();
+
+    return async function (...args: Parameters<T>): Promise<ReturnType<T>> {
+      const key = JSON.stringify(args);
+      if (cache.has(key)) {
+        return cache.get(key) as Promise<ReturnType<T>>;
+      }
+      const result = await fn(...args);
+      cache.set(key, Promise.resolve(result));
+      return result;
+    };
+  }
+
+  private async checkIgnoredFiles(dirs: string[]): Promise<string[]> {
+    console.log("called");
+    const paths = await walkDir(dirs[0], this.ide);
+    return paths;
+  }
+
+  // Memoize the async checkIgnoredFiles function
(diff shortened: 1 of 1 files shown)


RomneyDa (approved):
@PRANJALRANA11 nice quick fix!

RomneyDa:
@sestinj good point, would be bad for large repos
@PRANJALRANA11 can you check ignore for just the default ones and I'll open a separate branch for the walking?

https://github.com/continuedev/continue/blob/dev/core/indexing/walkDir.ts

```
ignore().add(defaultIgnoreDir).add(defaultIgnoreFile).add(globalIgnoreFile)
ignore.ignores(path)
```

RomneyDa (changes requested):
see comments above

PRANJALRANA11:
> @PRANJALRANA11 Thanks for taking the time to do this! This is going to be the perfect band-aid for small codebases, but I'm trying to think ahead to what happens when there is a huge codebase: with this implementation, we would walk the entire directory every time the user changes files, which is very expensive.
> 
> I'm thinking we'll need to instead have a way of storing the .gitignore and .continueignore locations, or instead just walking up the file tree from a given location to find them. This is a bit more involved, but in the meantime what you could do is use the default ignore patterns in `ignore.ts` (which include .env, etc.) here and it would solve this problem for 90% of users

If we try to do some sort of memorization here it then ! I think the no. Of times walkdir run will greatly reduced, it will only run one time for a codebase and then we can just loop around the results , this will also expensive but less

PRANJALRANA11:
@sestinj @RomneyDa  I have tried some sort of memoization func and now the walk dir only runs when workspace path changes
else it gives out the memoized results lmk if that works

RomneyDa (changes requested):
@PRANJALRANA11 
Interesting approach! this might speed up the the walking, but the cache isn't updated and to add that I'd probably move the cache off of core, like the recently edited files cache. To handle the updates properly would need to add additional handling for onsave, etc. An ignored files cache is probably on the right track for the future

For now to handle 90% of users, could you update to just do the default file ignores? Something like

```
ignore().add(defaultIgnoreFile)
ignore.ignores(path)
```

RomneyDa:
Maybe add a TODO comment about adding an ignored files cache that is cleared/filtered on ignore file updates, about speed concerns, etc.

PRANJALRANA11:
ok sure

RomneyDa:
@PRANJALRANA11 apologies for the stop n go on this one. Bumping priority on this one, do you think you'll have a chance to knock out just the defaultIgnoreFile ignores? If not I can snag in a few hours. Thanks!

PRANJALRANA11:
@RomneyDa hey sorry man I'm in IST timezone so it was late night yesterday

PRANJALRANA11:
Will resolve it today no worries

RomneyDa:
Got it, no worries thanks!

RomneyDa:
@sestinj tested, works. Benchmarked the entire event from start to cached at 4 ms, should be fine

RomneyDa (approved):

sestinj (approved):

RomneyDa:
@PRANJALRANA11 Thank you!!

sestinj:
Hi @PRANJALRANA11, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

PR #2768, opened 2024-11-01
Refactor autocomplete

sestinj:


core/autocomplete/filtering/streamTransforms/StreamTransformPipeline.ts
```diff
@@ -0,0 +1,86 @@
...
+    //   lang.topLevelKeywords,
+    //   fullStop,
+    // );
+    for (const lineFilter of helper.lang.lineFilters ?? []) {
+      lineGenerator = lineFilter({ lines: lineGenerator, fullStop });
+    }
+
+    lineGenerator = stopAtSimilarLine(lineGenerator, lineBelowCursor, fullStop);
+
+    lineGenerator = showWhateverWeHaveAtXMs(lineGenerator, 250);
```
tomasz-stefaniak: This value is hardcoded, I believe it should be taken from the options defined in `parameters.ts` instead.

core/autocomplete/postprocessing/index.ts
tomasz-stefaniak, line 1: I like what you did with `AutocompleteLanguageInfo.lineFilters`. What do you think about doing something similar with `postprocessCompletion` and various models? At the moment it has model-specific logic for `qwen` and `codestral`.

core/autocomplete/CompletionProvider.ts
```diff
@@ -0,0 +1,305 @@
...
+      outcome.completion,
+      outcome.filepath,
+    );
+  }
+
+  public markDisplayed(completionId: string, outcome: AutocompleteOutcome) {
+    this.loggingService.markDisplayed(completionId, outcome);
+  }
+
+  private async _getAutocompleteOptions() {
```
tomasz-stefaniak, line 125: Let's decide on a convention for private method names. At the moment we use a mix of leading underscore and no leading underscore. IMO it's better to have no underscore since TypeScript provides the `private` keyword but I'm open to anything that's consistent.

sestinj: We'll do no underscore from now on

sestinj:
Note to self: address loader bug before merging

core/autocomplete/postprocessing/index.ts
sestinj: Yeah this makes sense to me

PR #2776, opened 2024-11-01
Gitignore jetbrains config schema files

RomneyDa:
Gitignore copied config_schema files from jetbrains resources
These files are added/updated on package and on install-all-dependencies


sestinj (approved):
👌🏼

PR #2781, opened 2024-11-01
Fix activate vscode extension

RomneyDa:
Originally thought it was due to a missing dependsOn, but on second take looks like it was caused by an insufficient endsPattern (something @sestinj @Patrick-Erichsen) mentioned last week but I didn't get at the time

How to know?
vscode-extension:build task alone generates extension.js fine
but Launch "Launch extension" attempts to launch extension host before extension.js has been updated.

Solution:
try a few problem matchers, this one works

note, root issue seems to be that the problem matcher used to be ">" but then an npm script `esbuild-base` was added that caused an addition premature > to be output

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -5,11 +5,11 @@
     {
       "label": "vscode-extension:build",
       "dependsOn": [
-        // To detect compile errors and output extensions.js
+        // To detect compile errors
         "vscode-extension:tsc",
         // To build the React app that is used in the extension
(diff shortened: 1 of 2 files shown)


RomneyDa:
@sestinj @Patrick-Erichsen can you verify that endsPattern works on your computers?
`node scripts/esbuild.js --sourcemap --watch`

Wondering if this script outputs are actually due to some global setting that could make it not work for some users

PR #2784, opened 2024-11-02
copy prompt logs button

RomneyDa:
Adds a small button for people to be able to copy logs for the current chat history item

image

Copies context, prompts and completions, llm info, etc. to the clipboard


RomneyDa:
`clicked_copy_chat_history_logs` posthog event

Issue #2739, opened 2024-10-30
.continueignore should override .gitignore


RomneyDa:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Should be able to index ignored files
would also allow be a path to allow indexing external folders that people have been requesting
https://discord.com/channels/1108621136150929458/1301030122824138805

RomneyDa:
@PRANJALRANA11 assigned, thanks. Continue's take on the issue:
>The key is to apply the patterns from .gitignore first, followed by patterns from .continueignore

So approach might be to look through the `walkdir` and make sure that both global and all local continueignores are applied _after_ any gitignores

PRANJALRANA11:
@RomneyDa can't get it could you tell what's the problem and what could be the possible solution for that

PRANJALRANA11:
@RomneyDa yep done

k0d3r1s:
anything new here? i would like this to be implemented.  

my use case is pretty simple - i want continue to index vendor directories in php project otherwise it is generating obviously wrong suggestions

RomneyDa:
Fixed by https://github.com/continuedev/continue/pull/4531

PR #2802, opened 2024-11-04
chore:  apply gitignore before continueignore in walkdir

PRANJALRANA11:
## Description

fixes : #2739


sestinj:
I really appreciate the PR, but I've been thinking about it a bit and since the behavior here is so minor, and this adds a decent amount of complexity to the code, I'm unfortunately going to close it for now. The biggest thing is that I think walkDir is a very core function, which could cause major problems if any bug is introduced, and the more variables/things to keep track of we add, the easier that will be to do

PR #2839, opened 2024-11-07
Fix config schema for tab autocomplete models

amytimed:
Old schema used a key which isn't used by Continue, and thus you couldn't specify tab autocomplete model by following schema. This updates it to how it is in the the real

### DIFF
--- a/extensions/vscode/config_schema.json
+++ b/extensions/vscode/config_schema.json
@@ -2684,25 +2684,13 @@
             }
           ]
         },
-        "tabAutocompleteModel": {
-          "title": "Tab Autocomplete Model",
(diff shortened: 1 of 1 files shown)


RomneyDa:
Hi, @amytimed, thanks for the PR!

Mentioned in discord
>Hi @Amy, in config.json it uses tabAutocompleteModel, which can be set to one model or an array of models. Then it converts from intermediate to final config here, and the s is added for the final type
https://github.com/continuedev/continue/blob/dev/core/config/load.ts
Adding the s fixed it because now it's not finding your tabAutocomplete model so it's defaulting to starcoder2:3b, which is not causing the memory issue
It's a bit confusing but the root issue seems to be that deepseek-coder:1.3b-base is somehow requesting way too much memory, or perhaps we're not sending some param that deepseekcoder needs for memory management. Might be related to this
https://github.com/continuedev/continue/pull/2793

sestinj (approved):

sestinj:
As mentioned above, "tabAutocompleteModels" actually isn't intended to be supported, so going to close. Thank you though for taking the time to open a PR!

Issue #2798, opened 2024-11-03
Context provider "Active file" (ex Alt+Enter) doesn't work for unsaved files


Astlaan:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows
- Continue version: v0.8.55
- IDE version: VS Code 1.95.1
```

### Description

ContextProvider for active file (ex. when submitting prompt about the file in Chat with Alt+Enter) doesn't work for unsaved files. This happens despite the UI saying that it has been passed as context, however it is seen as an Object content, not visible by the model.

This is the behaviour I get:

![image](https://github.com/user-attachments/assets/d75dc0bf-5fbc-4cfe-949f-6cfa3e272243)

I haven't tested if this is also the case for the Quick Edits (Ctrl+I) since my Ctrl+I isn't working due to another bug (#2346 , https://github.com/continuedev/continue/issues/2346#issuecomment-2439655702 ).

### To reproduce

1. Open a new file, write some stuff, but don't save to disk
2. Open Chat, make a query about the file, add the active file as context with Alt+Enter

sestinj:
Thanks for bringing this up @Astlaan. My guess is that the solution to https://github.com/continuedev/continue/issues/1226 will be the same as for this, and it's something we plan to fix later this week

RomneyDa:
Looks like it does check open text document case here
https://github.com/continuedev/continue/blob/885e7c138d725dfed88d9165df99f1f692c49fb8/extensions/vscode/src/VsCodeIde.ts#L450-L457

Wondering if path issue, will debug

RomneyDa:
just for clarity it does work for unsaved files that already exist, just not for brand new unsaved files like in @Astlaan's case

RomneyDa:
Fixed by https://github.com/continuedev/continue/pull/2841

Astlaan:
> just for clarity it does work for unsaved files that already exist, just not for brand new unsaved files like in @Astlaan's case

Yes, sorry, could have been clearer

Issue #1226, opened 2024-05-03
[CON-234] Autocomplete only generates '}' in unsaved files


sestinj:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:*
- Continue:0.9.124
- IDE:vscode
```

### Description

Completions only show '}', regardless of what is in the file

This is local workspace
There is no output for "Continue - LLM Prompts/Completions" when the file is not saved. If I save the file, I see prompts and autocompletes

Probably an issue reading unsaved files (in ideUtils.ts readFile)

### To reproduce

This error is reproducible when I create a new file (selected language as python from status bar) and this file is not yet saved. 

1. Make new file
2. Type some stuff
3. The first completion might be fine, but subsequent are only '}'

<sub>[CON-234](https://linear.app/continue/issue/CON-234/autocomplete-only-generates-in-unsaved-files)</sub>

harshiniKumar:
I am facing the same issue.

fry69:
> I am facing the same issue.

I cannot confirm this issue (it is an old issue from May 3). I tried the steps mentioned to reproduce:

- fresh unsaved file (turned off autosave etc)
- started typing in that file
- autocomplete worked as expected, just pressed `tab` a few times to generate following code

<img width="508" alt="autocomplete" src="https://github.com/user-attachments/assets/85a6e5b1-b256-430e-a122-d39f6bc12918">

VSCode "About":
```
Version: 1.92.1
Commit: eaa41d57266683296de7d118f574d0c2652e1fc4
Date: 2024-08-07T20:16:39.455Z
Electron: 30.1.2
ElectronBuildId: 9870757
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.5.0
```

Pre-Release Continue (0.9.195) with Codestral via Mistral API

harshiniKumar:
It still happens. The only difference I see between us is OS and the Ollama local instance instead of API. 

[...]

PR #2841, opened 2024-11-07
Get current file untitled support

RomneyDa:
## Description

Current pattern is getCurrentFile -> readCurrentFile

The problem is if it's untitled there's no way to deduce that reliably from a single path

Also in every single use case, getCurrentFile read the file contents

So I just made getCurrent file return the file contents

Adjusted some types here and there

`isUntitled` is unused at the moment

## Checklist

Fixes https://github.com/continuedev/continue/issues/2798
https://github.com/continuedev/continue/issues/1226 seems related but as far as I can tell it's already been fixed


extensions/vscode/src/extension/VsCodeExtension.ts
```diff
@@ -278,7 +278,7 @@ export class VsCodeExtension {
         // Reindex the workspaces
         this.core.invoke("index/forceReIndex", undefined);
       } else {
-        // Reindex the file
+        // Reindex the filex
```
Patrick-Erichsen:
```suggestion
        // Reindex the file
```

sestinj (approved):
Nice application of the augmenting data types to solve problems principle

PR #2844, opened 2024-11-08
Set up root path context tests, add tree sitter definitions for typescript return types

tomasz-stefaniak:
## Description

Set up root path context tests, add tree sitter definitions for typescript return types.

## Checklist

- [ ] ~The base branch of this PR is `dev`, rather than `main`~

## Screenshots

## Testing


core/autocomplete/context/root-path-context/RootPathContextService.ts
```diff
@@ -77,28 +66,37 @@ export class RootPathContextService {
...
+        line: endPosition.row,
+        character: endPosition.column,
+      },
+    });
+    const newSnippets = await Promise.all(
+      definitions.map(async (def) => ({
+        ...def,
+        contents: await this.ide.readRangeInFile(def.filepath, def.range),
+      })),
+    );
```
tomasz-stefaniak, line 97: I'm still on the fence about making the tests more comprehensive by only mocking `gotoDefinition` instead of `getSnippets`. The tradeoff is that we'd have to manually update the ranges each time the test files change.

core/autocomplete/context/root-path-context/test/RootPathContextService.test.ts
```diff
@@ -0,0 +1,23 @@
+import { testRootPathContext } from "./testUtils";
+
+describe("RootPathContextService", () => {
+  describe("TypeScript should return expected snippets when editing inside a:", () => {
+    test("function", async () => {
+      await testRootPathContext(
+        "typescript",
+        "file1.ts",
+        { start: { line: 3, character: 2 }, end: { line: 3, character: 24 } },
+        ["EXPECTED SNIPPET 1", "EXPECTED SNIPPET 2"],
```
tomasz-stefaniak: The effect of mocking `getSnippets` is that we are essentially testing that the right number of snippets is returned in the right order but we don't check the actual content of the snippets.

sestinj: Yeah I think this is problematic as it exists now. What we're trying to test is that the tree-sitter queries are able to find the correct range to go to definition on. We need to test not just the number of times that getSnippets or gotoDefinition is called, but the range that it is called with.

sestinj (approved):

PR #2848, opened 2024-11-08
Scroll chat to bottom on session change and load

RomneyDa:
## Description

Currently when a chat loads it stays at the top, which is pretty much never the desired behavior, you'd want to start where you left off.

Snap scroll isn't bad but there's a few reasons I did the slighly delayed smooth scroll
- In the case of a long chat message, etc. it's nice to have a brief reminder that there _was_ chat history and where it started
- Often snap on load doesn't work because some components take time to render and there's some timing issues, more complicated to check that all children are loaded than to just delay a bit

## Checklist

## Screenshots

https://www.loom.com/share/d83e76e997104015a8eb07aa0581ed05


sestinj (approved):
👌🏼

PR #2854, opened 2024-11-09
Dallin/search in chat

RomneyDa:
## Description

Added custom search widget that highlights words and code lines

## Checklist

https://www.loom.com/share/6f18951ce95449579d088f10959e2afb

Notes on design decisions:
- It's implemented as a hook that takes a container ref and returns a find widget, search results, and highlight overlay elements. This is because the highlights _must_ be within the searched container, but the widget must be outside of it
- When chat starts generating it hides the search. It already disables while generating for performance, but it's also a feature to make sure any edge cases around cancellation, error, etc. don't cause weird search results
- Results and navigation arrows are hidden on smaller screens
- Query input and window resize handling are debounced for performance
- VS code theme highlight colors are used
- For styled markdown containers that might be wider than the screen, e.g. with horizontal scrollbars, it highlights the entire line, since it is complicated to adjust highlight overlay positions in this case

There is an edge case where expanding code blocks still messes up highlight positions
looking into it


sestinj (approved):
@RomneyDa this is so well done

I'm just assuming getting the overlay to not go full line width in code blocks was sufficiently tricky, and it's fine enough to leave for now imo

only other thing is I think we should probably make sure to get the autoformatter stuff straight before merging this one. I saved in the FindWidget file and it totally changed everything. Can you double check your prettier setup? I looked through all of the available options and I'm pretty confident that we've made everything either explicit or a default in the config file. If nothing obvious may can look closer monday

RomneyDa:
Nothing obvious on the prettier, let's look monday

PR #2880, opened 2024-11-12
Render Markdown Codeblocks that contain code

RomneyDa:
## Description

Fixes markdown codeblock rendering when they contain code

## Approach
Detect the beginning of markdown code blocks and replace  their ```s with ~~~s

People will be copying the code block with no backticks in the vast majority of cases anyways, not the entire generated output

## Notes
Benchmarked at sub-millisecond or around 1ms for longer outputs

Doesn't work if language isn't specified in codeblocks, but usually it is
Doesn't work if tildes or more than 3 backticks are used in top-level markdown block, which is an extreme edge case

I wrote a version that does achieve that and included commented out, didn't use for performance/complexity concerns. It essentially predicts if it's another start/end codeblock based on if there's a new line after the ticks or not, but it doesn't work well mid-generation

## Checklist

image


sestinj (approved):

snwfdhmp:
What were the motivations for this ? I'm working on a fork and considering removing this to normalize behavior between all languages.
How would this impact the extension ?
What are the problems you wanted to fix here ?

snwfdhmp:
I've had a response from @RomneyDa 

Putting it here in case anyone needs it in the future :

image

Issue #2896, opened 2024-11-13
[JetBrain IDE] `continue-binary` process does not properly exit when IDE project is closed


jnhyperion:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:Mac OS
- Continue version:0.0.80
- IDE version:PyCharm CE 2024.2.4
- Model:
- config.json:
```

### Description

Start Idea project the plugin will spawn a new `continue-binary` process, but when user close current project window (not quit the entire IDE), the `continue-binary` process does not properly exit.

When user open / close multi projects, continue plugin will keep spawning new process.

PR #2902, opened 2024-11-13
[JetBrain IDE] Fix `continue-binary` process does not properly exit

jnhyperion:
Fix: https://github.com/continuedev/continue/issues/2896

ContinuePluginStartupActivity's life cycle is IDE instance level, its `dispose` will not be called when a project is closed.


sestinj (approved):
nice fix, thank you for taking the time to do this!

PR #2912, opened 2024-11-13
Fix an edge case when stop token is not filtered out in completion response.

Pyroboomka:
## Description
We recently started using Qwen2.5-Coder-7B-Instruct for code completion.
In most cases the response from LLM is quite long, and selected/trimmed completion won't contain "<im_end|>" token.
But in some cases, when completion is short, or LLM does not add extra data (such as examples, comments, etc..) which are usually filtered out, plugin would not filter out <|im_end|"> at the end of the completion.
For example:
For code
```
const notificationSlice = createSlice({
})
```
the suggestion from LLM would be something like
```
const notificationSlice({
...lots of react-redux code...
})
// Export actions
const {...} = notificationSlice
....
export default notificationSlice<|im_end|>
```
and the completion returned to IDE would end up as
```
const notificationSlice({
...lots of react-redux code...
})
```

But for code
```
const log = (x) => {

}
```

the suggestion completion from LLM and returned to IDE would be the same
`console.log(x);<|im_end|>`

After investigation, I found out it occurs when remaining buffer here (https://github.com/continuedev/continue/blob/main/core/autocomplete/filtering/streamTransforms/charStream.ts#L110)
contains stop token, it's not getting filtered out. So, the naive fix was added.

## Testing
Added a test that covers that case. Problem is I can't figure out how to run the suite. Maybe will try that tomorrow.

Thank you for your work, extensions is amazing.

### DIFF
--- a/core/autocomplete/filtering/streamTransforms/charStream.test.ts
+++ b/core/autocomplete/filtering/streamTransforms/charStream.test.ts
@@ -157,4 +157,20 @@ describe("stopAtStopTokens", () => {
 
     expect(output.join("")).toBe("Hello world!");
   });
+
+  it("should handle stop token when remaining buffer is smaller than maximum stop token length", async () => {
+    const mockStream = createMockStream(["Hello world!STOP"]);
+    const stopTokens: string[] = [
+      "STOP",
+      "STOP_TOKEN_THAT_IS_LARGER_THAN_BUFFER",
+    ];
+    const result = stopAtStopTokens(mockStream, stopTokens);
+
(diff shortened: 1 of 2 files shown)


sestinj (approved):
nice catch, and thanks for adding the test!

jtc1246:
But it seems that this problem still occurs, #2953. I installed continue.dev today, and config and situation is same to that issue.

d4rkc0de:
The problem seems to be solved with the latest pre-release v0.9.237

Issue #2838, opened 2024-11-07
Chat/Completion not working with ollama and local models


sbouchet:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Fedora 40
- Continue version: v0.9.225 (pre-release)
- IDE version: VSCode 1.95.1
- Model: granite-8b
- config.json:
  
{
    "models": [
        {
            "model": "granite-code:8b",
            "provider": "ollama",
            "contextLength": 128000,
            "completionOptions": {
                "maxTokens": 4000,
                "temperature": 0.1,
                "topP": 0.9,
                "topK": 40,
                "presencePenalty": 0,
                "frequencyPenalty": 0.1
            },
            "systemMessage": "You are Granite Code, an AI language model developed by IBM. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior. You always respond to greetings (for example, hi, hello, g'day, morning, afternoon, evening, night, what's up, nice to meet you, sup, etc) with \"Hello! I am Granite Code, created by IBM. How can I help you today?\". Please do not say anything else and do not start a conversation.",
            "title": "granite-code:8b"
        }
    ],
    "tabAutocompleteModel": {
        "model": "granite-code:8b",
        "provider": "ollama",
        "contextLength": 128000,
        "completionOptions": {
            "maxTokens": 4000,
            "temperature": 0.1,
            "topP": 0.9,
            "topK": 40,
            "presencePenalty": 0,
            "frequencyPenalty": 0.1
        },
        "systemMessage": "You are Granite Code, an AI language model developed by IBM. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior. You always respond to greetings (for example, hi, hello, g'day, morning, afternoon, evening, night, what's up, nice to meet you, sup, etc) with \"Hello! I am Granite Code, created by IBM. How can I help you today?\". Please do not say anything else and do not start a conversation.",
        "title": "granite-code:8b"
    },
    "customCommands": [
        {
            "name": "test",
[...]

PR #2935, opened 2024-11-14
Cancel stale completion requests as you type

fbricon:
## Description

On VS Code, while editing a file, on every keystroke, Continue sends an http request to the provider. When running against ollama 0.4.x, it quickly becomes unresponsive (see https://github.com/continuedev/continue/issues/2838#issuecomment-2474492124). VS Code provides a cancellation token but it's unused when it comes to querying servers.

This PR ensures the AbortSignal bound to VS Code's cancellation token is used in all fetch queries. In effect, this means that on each character you type actually cancel the previous completion requests to the servers.

I'm not too sure how to handle requests not bound to a CancellationToken so I just created new AbortController().signal everywhere. This leads to more changes than I'd like but I think it makes the API more consistent across the board. But feel free to disagree/point me to better approaches.

~~This is still WIP as I need to fix an issue where all aborted signal spew stacktraces in the console.~~ fixed

While the implementation might be debatable, I'm convinced the overall strategy is sound (cancel stale requests as you type). It feels the overall type completion is snappier (using ollama), but I'm sure other providers would also benefit from it. 

Looking for feedback @sestinj

## Checklist

- [x] No doc change needed

## Screenshots

- [x] No UI changes

## Testing
- install ollama 0.4.1 and pull granite-code:3b
- configure granite-code:3b for autocompletion
>     "tabAutocompleteModel": {
>       "model": "granite-code:3b",
>         "provider": "ollama",
>         "contextLength": 128000,
>         "completionOptions": {
>             "maxTokens": 4000,
>             "temperature": 0.1,
>             "topP": 0.9,
>             "topK": 40,
>             "presencePenalty": 0,
>             "frequencyPenalty": 0.1
>         },
>         "title": "granite-code:3b"
>     },

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,7 +142,7 @@ codeql
 .tiktoken_cache
 
 
-# IntelliJ Plugin 
+# IntelliJ Plugin
 **/**/.gradle
 **/**/.qodana
 **/**/build
@@ -169,3 +169,5 @@ extensions/intellij/.idea/**
 
 extensions/intellij/bin
 extensions/.continue-debug/
+
+*.vsix
--- a/core/autocomplete/CompletionProvider.ts
+++ b/core/autocomplete/CompletionProvider.ts
@@ -29,6 +29,7 @@ const autocompleteCache = AutocompleteLruCache.get();
 const ERRORS_TO_IGNORE = [
   // From Ollama
   "unexpected server status",
+  "operation was aborted"
 ];
 
 export type GetLspDefinitionsFunction = (
--- a/core/autocomplete/generation/CompletionStreamer.ts
+++ b/core/autocomplete/generation/CompletionStreamer.ts
@@ -25,10 +25,10 @@ export class CompletionStreamer {
     // Try to reuse pending requests if what the user typed matches start of completion
     const generator = this.generatorReuseManager.getGenerator(
       prefix,
-      () =>
+      (abortSignal: AbortSignal) =>
         llm.supportsFim()
-          ? llm.streamFim(prefix, suffix, completionOptions)
-          : llm.streamComplete(prompt, {
+          ? llm.streamFim(prefix, suffix, abortSignal, completionOptions)
+          : llm.streamComplete(prompt, abortSignal, {
               ...completionOptions,
               raw: true,
             }),
--- a/core/autocomplete/generation/GeneratorReuseManager.test.ts
+++ b/core/autocomplete/generation/GeneratorReuseManager.test.ts
@@ -4,7 +4,7 @@ import { GeneratorReuseManager } from "./GeneratorReuseManager";
 function createMockGenerator(
   data: string[],
   delay: number = 0,
-): () => AsyncGenerator<string> {
+): (abortSignal: AbortSignal) => AsyncGenerator<string> {
   const mockGenerator = async function* () {
     for (const chunk of data) {
       yield chunk;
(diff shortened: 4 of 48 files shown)


fbricon:
The console is no longer polluted with aborted error stacktraces

sestinj:
@fbricon thanks for investigating this! I definitely agree that we are making too many requests right now. This particular approach is going to stop us from reusing requests in later completion attempts though. The code for this is a bit convoluted at the moment, but to give an idea of what I mean: say I start typing `con`, and the LLM begins to respond, and by the time I type `s`, the LLM has already streamed up to `console.`. Even though the CancellationToken will be fired, we wouldn't want to stop this request, because it is still correct, so we let it keep going in the background, and next time a completion request happens, if it still matches the stream, we keep the stream.

I kind of feel like this is tricky enough to get right that I would hesitate to make large changes until we have some basic tests for `CompletionStreamer.ts` in place

fbricon:
I understand what you're trying to do but Ollama 0.4 is currently unusable in the current situation which is an absolute blocker for us. So if you have ideas/pointers to limit the scope of the changes, I'll be happy to look into it

sestinj:
@fbricon it's not the scope of the changes that was problematic (I completely agree that this is a major problem and worth a large diff to fix it). The problem was that this would regress other parts of our autocomplete, causing many more requests to be sent.

I looked more deeply into this today, and
1. I wrote tests for the services that I am trying not to break while solving the Ollama problem (this PR: )
2. I think I've figured out exactly where we need to add the cancellation (here: https://github.com/continuedev/continue/blob/5b85cad7a9e292ca4060d8188ec60f7d4060de40/core/autocomplete/generation/GeneratorReuseManager.ts)

deboer-tim:
Thanks for looking into this @sestinj. I understand why you wouldn't want to cancel an older request that's in-flight and might still be valid after additional typing, but I wanted to echo Fred's comments that if you're using a local model that's anywhere near your machine's capabilities then Continue is unusable with Ollama 0.4. Multiple requests at a time are enough to push it from sluggish to thrashing badly and unresponsive, and it's much better with this PR.

I'm wondering if this is a case where behaviour should be different between local and remote models since you'd typically optimize differently? If not, then maybe it's worth leaving the first completion going as long as it _still matches current typing_ but cancel any intermediate ones? i.e. if I type `contr` rapidly then at most I'd have the initial completion on 'c' and the last one running, but not one from the o/n/t, and the first would be cancelled too if the first four letters it returned are 'cons'.

fbricon:
@sestinj I changed the code so now the AbortController is managed by GeneratorReuseManager.getGenerator. Calling cancel() on the ListenableGenerator now aborts the signal, that is propagated to all LLM queries. So reusing the same prefix doesn't cancel the current completion request

sestinj (approved):

sestinj:
Looks great, thank you @fbricon and @deboer-tim for pushing this through! I'm going to publish a new pre-release right away

rodrigocarvalheira:
Any idea when these changes on the cancel will be available in intellij? currently it's on v0.0.81 but in tags there's already a v0.0.82 but my intellij doesn't find it... thanks in advance... this problem with autocomplete makes the usage very not optimal... I'm looking forward to this changes...
Thanks!

Patrick-Erichsen:
@rodrigocarvalheira we're working on getting JetBrains into a more stable state before our next release, see here: https://github.com/continuedev/continue/issues/2956

rodrigocarvalheira:
thanks. hopefully it will be released soon... makes it a bit difficult to use like this... but great great work guys

PR #2937, opened 2024-11-14
warn users when maxTokens too close to contextLength

sestinj:


guerrerocarlos:
Anyone facing this error, the fix is to add the `maxTokens` inside the `completionOptions` something like this to your `config.json` like this

```json
{
  "models": [
    {
      "title": "LM Studio",
      "model": "",
      "contextLength": 8000,
      "completionOptions": {
        "maxTokens": 3000
      },
      "provider": "lmstudio",
      "max_tokens": 3000,
    }
  ]
}
```

PR #2948, opened 2024-11-15
fix: Add ContinuePluginDisposable to avoid memory leaks in Jetbrains

lkk214:
## Description

Some components do not have any parent disposable, leading to potential memory leaks,
It should be registered to a parent disposable with the same life cycle as the plugin.


sestinj (approved):
Thanks for doing this!

PR #2950, opened 2024-11-15
fix: The selected item in inline chat is always the first one

lkk214:
## Description

When opening inline chat, the first model in the drop-down list is always displayed, 
and the user has to re-select the desired model each time.


sestinj (approved):
🙌

PR #2963, opened 2024-11-16
ListenableGenerator.test.ts

sestinj:


sestinj:
Fails this one test: https://github.com/continuedev/continue/blob/49a9fcd1e609d72d8196129c2090c978fd021e72/core/autocomplete/filtering/test/testCases.ts#L112-L139

I think that the behavior of this test is wrong, and it should just be deleted

Patrick-Erichsen (approved):

PR #2972, opened 2024-11-18
Docs Indexing UI

RomneyDa:
## Description

A few main things here:
- A more stateful approach to indexing progress for docs
- Several DocsService bug fixes and (a bit too much reinventing the wheel) refactoring
- GUI elements for viewing and managing indexing status

## Checklist

## Screenshots

Docs flow ish
image

Docs indexing statuses in "More" page
image

Indexing status peeks in chat
image

image

New Add docs form flow that shows status after adding
https://www.loom.com/share/5dba457250dc4b5ab64ddaf1b853fa1f


Patrick-Erichsen (dismissed):
This is awesome, ton of nice polish, e.g. clicking on the progress from chat linking to docs, clicking on title linking to the site, the favicon, the new success dialog, the progress text updates - feels super polished 👌 

A few nitpicks/ideas:

- I have a hunch that for long running docs indexing, or indexing multiple, the footer on the homepage will become distracting, and therefore we should allow users to dismiss it. I think this could be a followup task though, not a big deal rn.
- Grabbing lastUpdated from the SQL table could be nice info for the "More" page view of docs
- When reopening the dialog after indexing a doc, it opened to the "Docs added" success page, it should probably default to going back to the "Add a new doc" page.
- The title element for codebase indexing ins't clickable, but this one is, so I think we need to make that clear through the UI. Probably either an underline by default, or a `arrow-top-right-on-square` from hero icons... but the "click to re-index" is already underlined so maybe +1 to an icon
- I think it's more typical to have the "close" button here be on the left and either be an outlined or ghost button. Also a bit smaller in general, I don't think they need to take up the whole width (especially on wider viewports)
Screenshot 2024-11-18 at 12 29 27 PM

RomneyDa:
> * I have a hunch that for long running docs indexing, or indexing multiple, the footer on the homepage will become distracting, and therefore we should allow users to dismiss it. I think this could be a followup task though, not a big deal rn.

Now hideable, will only show up again if all indexing stopped for that type
image

> * When reopening the dialog after indexing a doc, it opened to the "Docs added" success page, it should probably default to going back to the "Add a new doc" page.
Fixed!

> * The title element for codebase indexing ins't clickable, but this one is, so I think we need to make that clear through the UI. Probably either an underline by default, or a `arrow-top-right-on-square` from hero icons... but the "click to re-index" is already underlined so maybe +1 to an icon
Added
image

> * I think it's more typical to have the "close" button here be on the left and either be an outlined or ghost button. Also a bit smaller in general, I don't think they need to take up the whole width (especially on wider viewports)
Had on left as more of a "go to the next thing" placement, so updated text to "Done" and made the add more button secondary
image

> * Grabbing lastUpdated from the SQL table could be nice info for the "More" page view of docs
will look into this

PR #2974, opened 2024-11-18
support `Qwen/Qwen2.5-Coder-32B-Instruct` model provided by siliconflow

AnoyiX:
## Description

support some models provided by siliconflow, like 
- `Qwen/Qwen2.5-Coder-32B-Instruct`
- `Qwen/Qwen2.5-Coder-7B-Instruct`

## Testing

custom config:

```
{
  "models": [
    {
      "title": "Qwen2.5 Coder",
      "provider": "openai",
      "model": "Qwen/Qwen2.5-Coder-32B-Instruct",
      "apiKey": "************************************",
      "apiBase": "https://api.siliconflow.cn/v1/"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Qwen2.5 Coder",
    "provider": "openai",
    "model": "Pro/Qwen/Qwen2.5-Coder-7B-Instruct",
    "apiKey": "************************************",
    "apiBase": "https://api.siliconflow.cn/v1/"
  },
  ...
}
```

AI Edit Example:
image

Autocomplete Example:
image


core/llm/llms/OpenAI.ts
```diff
@@ -288,7 +288,8 @@ class OpenAI extends BaseLLM {
     suffix: string,
     options: CompletionOptions,
   ): AsyncGenerator<string> {
-    const endpoint = new URL("fim/completions", this.apiBase);
+    const url = this.apiBase?.includes("api.siliconflow.cn") ? "completions" : "fim/completions"
```
sestinj: @AnoyiX Can we make a dedicated `LLM` class instead of putting the logic in `OpenAI.ts`? A good example would be in `Deepseek.ts`, where we subclass `OpenAI` and adjust the FIM logic

sestinj (changes requested):

core/llm/llms/OpenAI.ts
AnoyiX: It might be better to allow custom `endpoint`, there are so many providers, it’s hard to support them all, what do you think?

sestinj: I think that's fine as long as they have the exact same expected request body. I would just make it a function like `getFimEndpoint` so that it can be modified by subclasses, and then make a subclass of the endpoint. I'm not worried about having a large number of subclasses

AnoyiX: Sounds good. It's convenient for subclasses to implement `getFimEndpoint`, this way we can avoid duplicating a lot of code in `*_streamFim`

AnoyiX: you can review the latest code ~ 😊

image

core/llm/llms/OpenAI.ts
```diff
@@ -297,7 +297,7 @@ class OpenAI extends BaseLLM {
     signal: AbortSignal,
     options: CompletionOptions,
   ): AsyncGenerator<string> {
-    const endpoint = new URL("fim/completions", this.apiBase);
+    const endpoint = this.apiBase?.includes("completions") ? new URL(this.apiBase) : new URL("fim/completions", this.apiBase);
```
sestinj: Everything else looks perfect, but can this be changed back to what it was before? @AnoyiX

AnoyiX: ok, I'll handle it now.

sestinj (approved):

PR #2975, opened 2024-11-18
Clickable chat symbols

RomneyDa:
## Description

Classes and functions are retrieved and cached per file during context item retrieval. Markdown components can then check for matching symbols based on nearby context items. Only checks for full or `startsWith` matches within code elements. Will hallucinate occasionally but pretty reliable.

https://www.loom.com/share/14827247cd8b48ab9473171df9df08f7


RomneyDa:
found one bug on this where react remark is using outdated state values grabbed during options initialization only, which causes it to only inject symbols into <code> elements if you reload the window, switch sessions, etc. Also affects file symbol links. Looking into this.

RomneyDa:
Bug fixed, ready for review

Patrick-Erichsen:
@RomneyDa also patched merge conflicts here partially related to my recent merge to main, take a second peek at my resolution though.

RomneyDa:
@Patrick-Erichsen thanks for the feedback, fixed the light/dark issues and switched to a simple opacity change on hovering
I tried the codebracket icon, I also think it's a bit noisy for now but there's something there

RomneyDa:
@sestinj 
- fixed bug with current file
- played around and settled on adding ring on hover, which I think looks really clean, found border a bit busy
- changed so uses global symbols. Assumption is duplicate symbol names are somewhat edge cases. Future could move to force e.g. session only

RomneyDa:
ready for review

sestinj (approved):

sestinj (approved):

PR #2978, opened 2024-11-18
Remove Footer

RomneyDa:
## Description

- Removes footer under main chat input, replaces with built-in webview title 
- Replaces profile selection with dialog
- VS Code only (could just add footer back in jetbrains)


Patrick-Erichsen (commented):
Looks great! Big fan of not having the footer. Personal opinion but I feel like the ordering from right to left would be clearer like this: Close, Config, More, Profile

Seems like the commands for switching between profiles isn't setup? 

Screenshot 2024-11-18 at 11 50 09 AM

RomneyDa:
@Patrick-Erichsen 
> Need to discuss approach to teams dropdown and config profile selection
Yes we should talk about this

Currently we can show or hide known submenu items based on context, but can't dynamically load items into the submenu.

We could could achieve the current 3 config options
- Local profile
- Personal workspace
- Team profile

Approach:
- if signed in, show "Account" icon with submenu options
   - "Use local config" shows if context "continue.usingLocalConfig" is false. Clicking triggers "continue.useLocalConfig" command, which sets "continue.usingLocalConfig" to true, "continue.usingTeamConfig" to false, and "continue.usingPersonalConfig" to false
   - same approach as ^ for "Use personal workspace" and "Use Teams workspace" submenu options
   - "Sign out" option triggers sign out command
- if not signed in, show "Account" icon that triggers sign in command

But future might be more profiles so not sustainable
@sestinj

sestinj:
@RomneyDa I think we should probably just have the account button open up a dialog, otherwise as you said it will be pretty limiting.

For now though (because I don't think we can conditionally display one of the toolbar icons, and we want it hidden for anyone who hasn't explicitly enabled the beta), I think we should leave the login button and dropdown we had before in the footer, so that all Continue for Teams users have the footer. When we exit beta, we'll bring it back up to the header

RomneyDa:
Ready for vscode review
especially wanting feedback on account dialog design, it's not spectacular but works
once GUI is good can check on jetbrains

RomneyDa:
could also re-insert footer for jetbrains for now, or maybe adding the jetbrains icons wouldn't be too bad

sestinj (approved):

PR #2988, opened 2024-11-19
Set custom User-Agent header

pczekaj:
## Description

Replaced the default 'node-fetch' User-Agent with 'continue.dev' to better identify requests

## Checklist

## Screenshots

## Testing


sestinj:
Hi @pczekaj, I'd prefer not to have this by default, but for users that want to use it the `requestOptions.headers` option in config.json will allow requests to be identified

Issue #2918, opened 2024-11-14
Github Issues Context Provider , the Get Issue Code maybe Outdataed


csh0101:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:
   macos
- Continue version:
   v0.9.2
- IDE version:

- Model:
- config.json:
```

### Description

the domain github.com/api/v3 maybe de Outdated ..
you can ask github.com/api/v3 in  browser , it is 404 not found..

### To reproduce

1. set config.json Github Issues Context Provider ..
2.find a repo issue 
3.enter 

### Log output

```Shell
Error getting context items from issue: HttpError: Not Found
```

tomasz-stefaniak:
Thanks for bringing this up, I opened a PR: https://github.com/continuedev/continue/pull/2998

PR #2998, opened 2024-11-19
fix: remove unnecessary baseUrl

tomasz-stefaniak:
## Description

This PR fixes the issue described here: https://github.com/continuedev/continue/issues/2918

NOTE: I removed the `domain` field because it's undocumented and we didn't even use it in `loadSubmenuItems`, so it seemed unnecessary. Let me know if it should be restored - in that case we should document it properly and ensure it works with `loadSubmenuItems`.

## Checklist

## Screenshots

## Testing


sestinj (approved):

sestinj:
yeah sort of, it's for people using GH Enterprise Server. Made a change so it will worth in both cases

PR #3013, opened 2024-11-20
Flow AbortSignal from IDE

X9VoiD:
## Description

#2935 had some `AbortController`s with no way to call `abort()` on it. This PR fixes that and only adds AbortSignal where it makes sense. For future proofing, AbortSignal is also passed down to `core/llm/stream.ts` functions.

The abort error that shows up in debug console is also removed.

## Testing

- Tested with `ollama serve`. No requests are left open. Cached completions still work. No regressions are detected so far with manual testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,7 +142,7 @@ codeql
 .tiktoken_cache
 
 
-# IntelliJ Plugin
+# IntelliJ Plugin 
 **/**/.gradle
 **/**/.qodana
 **/**/build
@@ -169,5 +169,3 @@ extensions/intellij/.idea/**
 
 extensions/intellij/bin
 extensions/.continue-debug/
-
-*.vsix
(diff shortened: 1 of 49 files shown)


fbricon:
Binding the http requests to the IDE cancellation token was the 1st approach I used in #2935. But as @sestinj requested, and as https://github.com/continuedev/continue/pull/2935#issuecomment-2488927707 explains, the abort signal has to be controlled by the generator, so you can still reuse pending streaming if the prefix is compatible with previous completion requests, as to minimize requests. This is irrelevant to response caching, that you observed.

X9VoiD:
Right! I apologize I missed that comment. Will fix in a moment.

X9VoiD:
Done. I think `AbortSignal` is outside the scope of `ListenableGenerator`'s tee-ing function and should be managed by the GeneratorReuseManager instead. It makes more sense this way for me.

sestinj:
Because of the other PR we had to similarly address this, I'm going to close this PR. We seem to have abort signals passed through everywhere, and it's actually purposeful that they are different ones from VSCode's token passed through the `InlineCompletionProvider`

I think just an awkward bit of timing with a few people jumping to solve the problem all at once (which I appreciate!), but if I'm missing something here let me know

Issue #1458, opened 2024-06-10
[CON-246] fix syntax highlighting for strings in sidebar code blocks in dark mode


sestinj:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:*
- Continue:*
- IDE: VS Code
```

### Description

![image](https://github.com/continuedev/continue/assets/33237525/3b0a298d-2ec8-4314-816e-7e25da898469)

In many dark themes (e.g. One Dark Pro), Continue uses a color for strings in the side bar that is very close to the background color. We need to select a better color [here](https://github.com/continuedev/continue/blob/preview/gui/src/hooks/useVscTheme.ts#L5-L55)

<sub>[CON-246](https://linear.app/continue/issue/CON-246/fix-syntax-highlighting-for-strings-in-sidebar-code-blocks-in-dark)</sub>

Patrick-Erichsen:
@menjilx I got started on this today but I was unable to reproduce the issue with `Dainty - One Dark Pro`. Could you specify the theme you're using?

dreambit:
Same issue, theme: Learn with Sumit - Vintage

dreambit:
<img width="255" alt="image" src="https://github.com/continuedev/continue/assets/1479661/4407fb0b-e2d4-4420-a735-f20c95d9eac3">

dreambit:
Guys any updates? tried many dark themes, same issue

0xBigBoss:
<img width="995" alt="Screenshot 2024-11-20 at 18 10 20" src="https://github.com/user-attachments/assets/16809980-f1f9-4b60-a1d0-aaf2f5006fe2">

just want to confirm this is impacting me as well. I'm using the default Monokai theme. `@id:vscode.theme-monokai`

PR #3019, opened 2024-11-21
fix getTheme to respect preferred mode

0xBigBoss:
previously the getTheme function would only inspect the workbench.colorTheme. This would not work for users who use window.autoDetectColorScheme. Now, we inspect this value before deciding which color theme to use.

Shuold fix https://github.com/continuedev/continue/issues/1458

### DIFF
--- a/extensions/vscode/src/ContinueGUIWebviewViewProvider.ts
+++ b/extensions/vscode/src/ContinueGUIWebviewViewProvider.ts
@@ -168,7 +168,15 @@ export class ContinueGUIWebviewViewProvider
 
     const currentTheme = getTheme();
     vscode.workspace.onDidChangeConfiguration((e) => {
-      if (e.affectsConfiguration("workbench.colorTheme")) {
+      if (
+        e.affectsConfiguration("workbench.colorTheme") ||
+        e.affectsConfiguration("window.autoDetectColorScheme") ||
+        e.affectsConfiguration("window.autoDetectHighContrast") ||
+        e.affectsConfiguration("workbench.preferredDarkColorTheme") ||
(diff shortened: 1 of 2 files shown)


sestinj (approved):
Very nice, thanks for taking the time to do this!

Patrick-Erichsen:
Hey @0xBigBoss , we're getting [reports from some users](https://discord.com/channels/1108621136150929458/1108621136830398496/1314313326133772339) (myself included) are no longer getting proper syntax highlighting in the sidebar in VS Code after this commit (confirmed via `git bisect`). 

Here's an example of some Python syntax highlighting before and after these changes:

**Before**
Screenshot 2024-12-05 at 12 40 24 PM


**After**
Screenshot 2024-12-05 at 12 40 34 PM

Mind taking a look into this?

0xBigBoss:
Hey @Patrick-Erichsen thanks for reporting. I did a little investigation and found some things. Good news is the theming looks really close, but I'm unsure what happened to the sidebar chat between the time I merged that PR. There seems to be a regression on master that removes all styling from the sidebar, I'm having trouble confirming on latest master.

All this PR aimed to do was properly switch based on the `Window: Auto Detect Color Scheme` property. I didn't update any other theme logic as far as which CSS was applied. 

I did a little investigation and found that there is something else going on with the built-in theming added by continue and it's reproducible by checkout out the commit before this PR and the commit after this PR was merged. My guess was the previous behavior was buggy beforehand with the built-in themes. Anyways, I had great success with these changes locally. You can see a PR here where we can continue the conversation. https://github.com/continuedev/continue/pull/3210

PR #3034, opened 2024-11-22
fix: Looped requests cause page crashes

lkk214:
## Description

Looped `jetbrains/isOSREnabled` requests cause page crash in JetBrains.

### DIFF
--- a/core/protocol/ideWebview.ts
+++ b/core/protocol/ideWebview.ts
@@ -92,7 +92,7 @@ export type ToWebviewFromIdeProtocol = ToWebviewFromIdeOrCoreProtocol & {
   setTheme: [{ theme: any }, void];
   setColors: [{ [key: string]: string }, void];
   "jetbrains/editorInsetRefresh": [undefined, void];
-  "jetbrains/isOSREnabled": [boolean, void];
+  "jetbrains/jcefOSREnabled": [boolean, void];
   addApiKey: [undefined, void];
   setupLocalConfig: [undefined, void];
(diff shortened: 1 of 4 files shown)


sestinj:
@lkk214 this looks like just a rename, can you elaborate on the solution?

lkk214:
@sestinj Okay, I changed this because the following steps would result in an infinite loop of requests.

To elaborate further, the steps involved are as follows:
1. There is an attribute isOSREnabled in TipTapEditor, which sends a request with a messageType of `jetbrains/isOSREnabled` to the ide when it is rendered, and **registers an eventListener about `jetbrains/isOSREnabled` in the webview.**
2. The ide receives this request and processes it, after which the ide sends a response with a messageType of `jetbrains/isOSREnabled` to the web.
3. The eventListener receives the result of `jetbrains/isOSREnabled` from the ide, and executes `ideMessenger.respond` to return a request with a messageType of `jetbrains/isOSREnabled` to the ide.

Obviously, they are in an endless loop, which eventually causes page freezes, slow LLM output, and even page crashes.

Additionally, It is worth mentioning that each dialog box will request multiple repeated `jetbrains/isOSREnabled`. If there are multiple dialog boxes on the page, there will be too many unnecessary requests.

ncoquelet:
This fix works very well in my case.
No more freeze, no more excessive CPU usage
Must be considered

Patrick-Erichsen:
Closing this in favor of merging https://github.com/continuedev/continue/pull/3134

PR #3039, opened 2024-11-22
refactor: Improve the diff block logic for inline chat

lkk214:
## Description

1. When the type of the first diff line is `old`, the inlay box is removed.
2. When the type of the first diff line is `new`, it is displayed above the inlay box.
3. Fixed `IndexOutOfBoundsException` when deleting the last row.

## Screenshots

bug1

bug2

## Testing

fix1
fix2


Patrick-Erichsen:
Nice, these are awesome fixes! This has been bothering me too so I appreciate the contribution. The GIFs are also super helpful.

Patrick-Erichsen (approved):

PR #3054, opened 2024-11-24
Improve Autocomplete Filtering

ruediste:
## Description

Often continue with Codestral would propose autocompletions which just reproduce what is already below the autocomplete position in the editor (suffix). This PR tries to address this by stopping when a part of the suffix is encountered

## Checklist

## Screenshots

na

## Testing

Added unit tests, cleaned up some existing tests and reenabled tests in `lineStream.test.ts`


ruediste:
Still working out the situation with the unit tests. There are currently failing tests in the codebase :-(

PR #3059, opened 2024-11-25
Temporary fix for ollama onboarding 

RomneyDa:
## Description

Handle llm/listInfo returning empty object

Doesn't fix underlying issue of message type not being recognized

per @Patrick-Erichsen will probably be solved by strict null checks


RomneyDa:
noting for future reference that underlying bug is that ideMessenger is returning an empty object when based on types that shouldn't be possible. So removing type casting and upcoming strict null checks should help long term

PR #3061, opened 2024-11-25
Performance and redux audit

RomneyDa:
## Description
- Fix initial config load race condition logic
- Move useSetup to a parallel component to prevent unnecessary rerenders of pages
- Move VscThemeProvider and SubtextmenuContextProviders into a more isolated provider-hook paradigm
- Fix some subtext menu context provider logic that was causing unnecessary renders
- useSelector audit - cleanup/fix selectors that were causing renders on every run
- fix logic that was causing `config.showChatScrollbar` to be ignored


sestinj (approved):

sestinj:
Nice, things seem to work well!

I'm notorious for doing this myself, but as we lean more heavily on bisecting to find issues, probably good to attempt smaller commits during any refactor

PR #3066, opened 2024-11-25
Improve filtering

ruediste:
## Description

I often have the case that the LLM repeats the suffix of the query, which is then proposed. This is quite confusing. This PR detects this and stops the completion when a part from the start of the suffix is found.

In addition, I fixed some broken unit tests, did some test cleanup. Furthermore I disabled `stopNCharsAfterClosingBracket`, since it caused some unit tests to fail and seems to make little sense.

@tomasz-stefaniak: Can you explain the idea behind `stopNCharsAfterClosingBracket`? And provide some samples/test cases for it?

## Checklist

na

## Screenshots

na

## Testing

unit tests


tomasz-stefaniak:
Hi @ruediste, thanks for the PR and for pointing this out. The answer is that `stopNCharsAfterClosingBracket` isn't what it's supposed to be, it was a hotfix and it ended up not working too well. I opened a PR with a fix to the filtering logic with an approach that is, I believe, similar to yours: https://github.com/continuedev/continue/pull/3044/files

Where we differ is that your approach recognizes duplicated suffixes in any case, and in my case it only handles it after a closing bracket. I added [some tests](https://github.com/continuedev/continue/pull/3044/files#diff-929bf04dcb5dfd8bc1aeb02110c8efe7c8a8c499ef739e7f882f7d792d00db48) also.

I need to figure out with @sestinj what the next steps should be and then I'll give your PR a proper review.

core/autocomplete/filtering/streamTransforms/charStream.test.ts
```diff
@@ -166,11 +144,73 @@ describe("stopAtStopTokens", () => {
...
+      const result = stopAtStartOf(mockStream, suffix);
+
+      const resultStr = await streamToString(result);
+      if (resultStr !== removed) {
+        throw new Error(
+          `i=${i} result:\n${resultStr}\n\nremoved:\n${removed}\n\nsuffix:\n${suffix}`,
+        );
+      }
+    }
+  });
```
tomasz-stefaniak, line 194: I'm going to have a better look at these in a bit but my gut feeling is that the tests are too complex. Ideally, it should be visible at a glance what the expected result is. Is there a way to rewrite this test to use an approach similar to the one in `testCases.ts`, where we have a lot of hardcoded strings but little processing logic inside the test?

tomasz-stefaniak:
@ruediste I reviewed your code and applied my tests on top of it - everything is working fine and the tests are passing 🙌 

Could you pull the changes from [this branch](https://github.com/continuedev/continue/pull/3067) into this PR? 

The only remaining thing I'd like to adjust is that we simplify `charStream.test.ts`. Once that's done the PR should be good to go.

Thanks again for the contribution!

tomasz-stefaniak (approved):
I'm merging this PR faster than anticipated to unblock us on some other fronts. Great work!

PR #3104, opened 2024-11-27
Null check warnings for GUI

RomneyDa:
## Description

Strict null checks were already on for vscode and core since we had `strict: true`
Also, eslint wasn't set up for GUI

So I set up _only_ strict null checks with warnings in GUI


RomneyDa:
when we want to move from warnings to errors, we can delete `tsconfig.strictNullChecks.json` and set `strict: true` in `tsconfig.json`

Patrick-Erichsen (approved):

Patrick-Erichsen:
We should add eslint in the near future too. I have a big PR coming from the redux updates but after that I think I will add it.

sestinj (approved):

PR #3115, opened 2024-11-27
Set up temperature experiment

tomasz-stefaniak:
## Description

Set up temperature experiment.

## Checklist

## Screenshots

## Testing


core/util/posthog.ts
```diff
@@ -2,10 +2,28 @@ import os from "node:os";
 
 import { TeamAnalytics } from "../control-plane/TeamAnalytics.js";
 import { IdeInfo } from "../index.js";
+import { PostHog } from "posthog-node";
```
tomasz-stefaniak: @sestinj I'm assuming we had a good reason to have Posthog imported inline. Is moving it here a potential problem?

sestinj: Yes, we have it there so that we can be absolutely sure that PostHog doesn't send any preemptive requests when the user has telemetry disabled

sestinj: Can we just always get it from the Telemetry singleton class?

tomasz-stefaniak: We can, this is for typing reasons though - without this the type has to be `any` and it kind of sucks 🥲

PR #3122, opened 2024-11-28
Minor indexing fixes

RomneyDa:
## Description

Fix a bug that causes auto pausing on start that I introduced in #2972 
Add a cancelled state because "disabled" in case of abort signal is confusing for user

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -154,10 +154,10 @@ export class Core {
       // Index on initialization
       void this.ide.getWorkspaceDirs().then(async (dirs) => {
         // Respect pauseCodebaseIndexOnStart user settings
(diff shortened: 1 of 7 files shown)


core/core.ts
```diff
@@ -154,10 +154,10 @@ export class Core {
       // Index on initialization
       void this.ide.getWorkspaceDirs().then(async (dirs) => {
         // Respect pauseCodebaseIndexOnStart user settings
-        this.indexingPauseToken.paused = true;
         if (ideSettings.pauseCodebaseIndexOnStart) {
+          this.indexingPauseToken.paused = true;
```
RomneyDa, line 158: this is the pause fix

sestinj (approved):

PR #3130, opened 2024-11-29
fix: Add context/getSymbolsForFiles endpoint in JetBrains and handle …

lkk214:
## Description

1. Add `context/getSymbolsForFiles` endpoint in JetBrains to support forwarding requests.
2. Handle parsing exceptions to avoid no response results.


sestinj (approved):
thanks for the fix!

PR #3134, opened 2024-11-29
fix: intellij sidebar chat freezing

0xBigBoss:
## Description

Reverts useIsOSREnabled to use the single use event listener instead of subscribing to `useWebviewListener`.

### DIFF
--- a/core/protocol/ideWebview.ts
+++ b/core/protocol/ideWebview.ts
@@ -43,7 +43,7 @@ export type ToIdeFromWebviewProtocol = ToIdeFromWebviewOrCoreProtocol & {
   insertAtCursor: [{ text: string }, void];
   copyText: [{ text: string }, void];
   "jetbrains/editorInsetHeight": [{ height: number }, void];
-  "jetbrains/isOSREnabled": [undefined, void];
+  "jetbrains/isOSREnabled": [undefined, boolean];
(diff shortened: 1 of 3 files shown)


lkk214:
Similar PR https://github.com/continuedev/continue/pull/3034, but it has not been merged. 
I think this problem is very serious and needs to be solved before the next version is built

K-Mistele:
Waiting on a fix for this! I keep running into this problem and it's making Continue borderline unusable.

sestinj:
@K-Mistele getting close to a release for this! I think we should have an update in the next couple of days. We will absolutely merge prior to said release, there is no question there

ncoquelet:
> PR #3034

Hi,
The PR https://github.com/continuedev/continue/pull/3034 by @lkk214 fix my sidebar freesing with OSR enabled

ncoquelet:
> @K-Mistele getting close to a release for this! I think we should have an update in the next couple of days. We will absolutely merge prior to said release, there is no question there

Great to hear the release is close! Just a heads-up—regressions in the IntelliJ plugin have been a bit of a challenge, especially for daily dev work. That said, I appreciate the hard work, and I’m excited to see the update land!

Patrick-Erichsen:
This all makes sense, thanks for the contribution here @0xBigBoss ! Going to merge this in favor of https://github.com/continuedev/continue/pull/3034

Issue #925, opened 2024-03-04
Add a loading spinner to indicate command has been sent to the model and awaiting response


mavwolverine:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Installed `ollama` locally with `deepseek-coder:33b-instruct`, gave it a test prompt `create a dockerfile for fastapi`.

Nothing happened. 

I was about to reenter the prompt, then suddenly code started showing up in file.

### Solution

An working indicator like a loading spinner or infinity progress bar would be a good idea to indicate, still waiting for response.

hobyte:
maybe this can be similar to the approach taken by gitlab. They show a gitlab logo with a spinner next to the line number

hobyte:
this can be very nice for tab autocomplete. showing the spinner next to the line number instead of the bar at the bottom of VSCode. This makes it way easier to see if the AI model is done with generating.

PR #3145, opened 2024-11-30
VsCode: Add loading spinner during autocomplete request on current line

ruediste:
## Description

Add loading spinner during autocomplete request on current line. Can be switched off in the vscode extension configuration.

#925 

## Checklist

## Screenshots

https://github.com/user-attachments/assets/96a94935-2e9c-4755-b6c7-d3e5d08c3289

## Testing
Edit some code, watch the loading spinner.
Disable the spinner in the settings, validate the absence of the spinner.


sestinj (changes requested):
@ruediste This is super cool! Makes autocomplete feel more predictable. I made a few small adjustments because I think it's really important that this not be too noisy (removed background and underline, and set false by default)

There are two other things I'd probably like to see before merging:
- An animation that feels less noticeable—both the white color and the spinning are a lot of motion. I think that a light gray (HEX #888888) dot that maybe just very slowly pulses would do the trick
- Right now the spinner is flashing a lot when there's a quick completion. I think that we should only show the spinner if ~200ms have passed and the completion is still loading, and probably should show the spinner for a minimum of ~250ms

ruediste:
I added the delay and two icons, now with Synfig Studio, an open source 2D drawing/animation software. I did the dot, but the animation continues running in the background, so some times the dot pops up in it's large state, sometimes in it's small state. So I added a slow spinner as well. To me the spinner looks better. Just replace the `gif` files to see both versions.

Happy to adjust the icons if you have suggestions, I'm not much of a designer ...

ruediste:
The minimum display time is present as well.

sestinj:
@ruediste I'm going to close this for now—given the amount of attention we'd want to put into the UI for this in order to ship it, I just don't think we're going to be able to prioritize it in the near-term. I'm mostly trying to avoid a situation where we ship something without polishing, have a few users use it but share negative feedback, and then we have a maintenance burden we aren't able to prioritize.

If you have any thoughts to the contrary, I'm open to hearing. In the future if/when we get back to this, I'm confident this PR will still be a great reference!

PR #3156, opened 2024-12-02
Remove getIdeSnippets()

ruediste:
## Description

The auto completion had dead code for getting snippets from the IDE via LSP, which the LSP does not support. I removed the code.

## Checklist

## Screenshots

na

## Testing

na


sestinj (commented):
@ruediste thanks for looking out for dead code

in this case, we are experimenting a lot with different autocomplete strategies, so this was dead at the time, but no longer

PR #3160, opened 2024-12-03
include unstaged by default

sestinj:


DanielusG:
Hi, I would like to know the motivation for this change. This because in the past weeks I always used the /commit command by adding the file in stage, but not always I want to commit all changes. With this merge, the LLM considers all changes and I haven't found a way to pass the option for not including the unstaged changes.

I think (for my specific use case) it is easier to choose which files the LLM should consider for building the commit message.

The only useful thing I found to solve this problem is custom slash commands, but that seems a bit too invasive.

Issue #3081, opened 2024-11-26
Word wrap/Line length setting


rosmur:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Currently, the code blocks in the LLM output dont word wrap automatically and thus require cumbersome horizontal scrolling. Eg: ![](https://media.discordapp.net/attachments/1309133128098385980/1309133128228540508/image.png?ex=6746677c&is=674515fc&hm=39ff10a78b28598d70f25c8d8fe8f7a92cb7dd6fc53cb57a16925a1ff506b751&=&format=webp&quality=lossless&width=1404&height=1052)  

### Solution

Request to apply word wrap to code blocks in LLM output as well (uniform enforcement). And/or provide functionality to configure line length.

sestinj:
@rosmur definitely a good idea! I'm tagging this as `good-first-issue` as I don't know that we'll prioritize it, but would be great for contribution. It's probably as straightforward as one line of CSS, and then the button to toggle

mantosh-mariner:
Hi, I’m Mantosh , and I’m interested in contributing to this project. I came across number of issue  and would like to work on it. Could you please assign it to me? 

Thank you for your time and for maintaining this project. Looking forward to your reply!

sestinj:
Just assigned @mantosh-mariner

RomneyDa:
@mantosh-mariner just double checking that you saw @HD787 snagged!

RomneyDa:
This is now an option in user settings

<img width="354" alt="Image" src="https://github.com/user-attachments/assets/4c1b33dc-6b82-4e67-90a3-266a0fc88399" />

PR #3173, opened 2024-12-03
added wrap code to llm response

HD787:
## Description
solves issue #3081

Added a button to wrap code from LLM response in the same way it shows up in the chat box from cmd+l 

hover effect made dynamic changes to css not persist, so I used redux.

I saw this was assigned to someone after I had finished, apologies for any overlap.

## Screenshots

image

## Testing

manually tested on VSCode


RomneyDa:
this is great functionality to add

I'm wondering if the trigger should be in the codeblock toolbar as a more "stateful" kind of toggle rather than in the actions popup

Or perhaps as a config.ui param

HD787:
@RomneyDa thanks for the feedback! I'm a little confused on the difference between the 'actions popup' and the 'codeblock toolbar', the ui config option 'codeBlockToolbarPosition' and the component 'StepContainerPreActionButtons' seem to indicate that they are the same thing. Unless you just meant to use an icon the represents the state of the code wrap.

that being said I implemented the feature as a config.ui parameter and it totally works, if you think this is preferable i'll just push this.

image

RomneyDa:
@HD787 sorry for the confusion, you are right, the codeblocktoolbar is the same thing. I meant some kind of additional toggle in the codeblock viewer itself but on second glance I think its preferable outside. I'm leaning towards the config.json param because I feel like generally a user will want it always on or always off.

What if we did a local storage variable in React Context instead? UI config like this seems like it shouldn't be in config profiles that are designed to be shareable.

In this case maybe your original commit would be a good temporary solution and then we could move to a simple local storage settings UI in the more page or something

@sestinj @Patrick-Erichsen @tomasz-stefaniak thoughts?

RomneyDa:
@HD787 I'd go ahead and push your commit for the config option, worst case can revert it but I think a persisted setting is the move here.

I think there's a strong case for removing most `config.ui` params and moving them to local persisted redux with a simple config UI in the more page. This might be a perfect starting case for that.

Approach:
- Add `wrapCodeBlocks` or similar to the ui slice in `gui/src/redux/slices/uiSlice`
- ensure it is persisted in the [saveSubsetFilters](https://github.com/continuedev/continue/blob/main/gui/src/redux/store.ts#L33-L37) here 
- Add a super simple toggle switch setting in the more page that updates this value
- In the codeblock, check the redux `wrapCodeBlocks` value rather than `config.wrapCode` (remove `config.wrapCode`)

Let me know if you'd have time to give that a shot

HD787:
Yeah I'll give it a shot. Would you want all of the config.ui options moved into a menu on the more page?

RomneyDa:
@HD787 
just chatted with @sestinj, since 
- it might be awkward to have some ui params in config.json and some in more page
- moving all to config will break people's settings and be more work for you and make more page noisier

I think we can just leave it at your second commit!

I've opened an issue here for the future changes which might be held off for a couple months
https://github.com/continuedev/continue/issues/3178

sestinj (approved):

PR #3198, opened 2024-12-05
Core Utils Tests Part 2

RomneyDa:
Adds a bunch of core util tests to improve util coverage

Initial utils coverage 18%
New utils coverage 63%

Note 1/3 ish of this increase is from moving messengers and messageIde out of utils

To get a proper initial coverage report on utils, I
- set `collectCoverage` in jest config to _all_ `ts` files, not just the ones touched by tests. That way it includes zero-coverage files
- run tests specifically in `core/utils` to get coverage for `core/utils`. This is because running global tests might show e.g. 35% coverage in `core/utils`, but _none_ of the coverage from outside calls to `core/utils` should count towards core util unit testing

## Utils before/after
image

image

## Global before/after
image

image


sestinj:
@RomneyDa is this good to merge other than the failing check?

PR #3210, opened 2024-12-05
fix: major fixes to getTheme

0xBigBoss:
Since the update to detect the correct color scheme, users have noticed that the `include` logic has been broken causing incorrect styling.

This fix introduces cleaner theme logic since we assume that the vscode extension will have the theme files locally. The include logic now includes nested themes.

It also fixes finding the default themes correctly by comparing both theme ID and theme label.

---

> [!IMPORTANT]
>  There seems to be a regression on main preventing this from being fully tested. I only tested this PR against the following commits.
>
> `3d2342b2e`
> `5c938c0e5`


Patrick-Erichsen:
Thanks for the quick follow up here @0xBigBoss !

I'm getting the same error locally as we are here in CI: https://github.com/continuedev/continue/actions/runs/12188817022/job/34002678726?pr=3210#step:15:168

Patrick-Erichsen:
cc @sestinj here too, I'm not familiar with the files we're removing in this PR.

0xBigBoss:
Happy to help when I can, yeah I only tested this in local development mode on vscode. I'll fix the CI, but unsure the full impact of removing the built in themes as far as intellij goes. More testing is warranted, the styles disappearing on the sidebar chat completely is also concerning. unsure if other users are reporting that in the pre-release version.

sestinj:
@0xBigBoss I believe that there was just a separate bug causing the styles not to work. Patrick bisected and I think will put up a fix

It's important that we keep these, as they are used in situations where the user doesn't have a theme that we can autodetect the colors for

sestinj:
I didn't realize that Patrick had already messaged you and this was intended to be the resolution 😅

It is unfortunately still true though that these files serve an important purpose, so we'll need a more precise solution. It seems like the problem is black text on black background, so maybe the wrong default theme is being loaded?

0xBigBoss:
I can look more tomorrow at a better solution, but it was strange that it did seem to be related to these builtin themes.

0xBigBoss:
@sestinj @Patrick-Erichsen I have update the PR to correctly find the vscode themes. For some reason, when a user has the default themes e.g. `Default Dark Modern`, it is not enough to compare by theme label and checking by theme ID will find the theme files. I also updated the theme parsing logic to correctly include the local files thus merging the grammars until we get the final theme.

This PR should fix the color scheme parsing logic for you @Patrick-Erichsen and the other users.

https://github.com/user-attachments/assets/e27c4a53-4bd5-41e7-b35f-9969bb7a634c

@sestinj  I can try to include the built-in themes fallback later, but so far I am unable to reproduce a scenario where the extension was not able to find the default themes or an installed extension that includes custom themes.

> where the user doesn't have a theme that we can autodetect the colors for

can you expand on this? How do we reproduce this?

sestinj:
Ok yeah this looks solid—I think that it covers the cases I was concerned with but I'm going to triple check. If I can find a repro, will share, otherwise, will assume we're good to go.

sestinj:
In the process of triple-checking right now and for some reason not getting syntax highlighting at all though 

Screenshot 2024-12-09 at 5 53 45 PM

0xBigBoss:
> In the process of triple-checking right now and for some reason not getting syntax highlighting at all though

I haven't tried in a few days, but there was a regression on master that removes all styling @sestinj . I only tested by cherry picking against the commits in the PR description.

Any updates on finding the root cause for the missing styles, @Patrick-Erichsen ?

sestinj:
Finally figured out why syntax highlighting in general was broken, and now got to try your PR—it works great! Thank you for being patient as we untangled the separate problems

sestinj (approved):

sestinj:
Hi @0xBigBoss, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

Issue #2796, opened 2024-11-03
Chat View History: display current conversation


Astlaan:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Currently, if I start a new Chat (Ctrl+L) and start talking with the LLM, and then click "View History", the current chat will not appear in the History. Only if I, for example, select a past chat and click View History again, will it appear on the top.

### Solution

I think it would be more clear if history displayed not only past conversations, but also the latest created, ongoing conversation.

sestinj:
Yeah this definitely makes sense!

I'm marking this as good-first-issue on the chance that anyone finds it and is interested in contributing before we get to it, but I think this feels like a high bang to buck that we might fix in the next ~week

I think the best way of doing this is probably to auto-save the chat at the end of each message

PR #3215, opened 2024-12-06
History will now show the current chat

HD787:
## Description

added an autosave when a new LLM response completes and the history page will update automatically. solves #2796.

## Testing

tested locally on vscode


HD787:
I've been exploring how to ensure fullscreen mode opens on the current chat. This seems tricky but one approach is to load the current session whenever a chat window gets rendered. If this approach is sound, this would be a good pre-req PR.

Gabomerino:
> ## Description
> 
> 
> 
> added an autosave when a new LLM response completes and the history page will update automatically. solves #2796.
> 
> 
> 
> 
> 
> 
> 
> ## Testing
> 
> 
> 
> tested locally on vscode
> 
>

sestinj (approved):

PR #3233, opened 2024-12-06
Adds support for preprompt (systemMessage) on code edition/creation

AnnoyingTechnology:
## Description

the `systemMessage` in `config.json` was only used in pure LLM Chat mode.
When editing code files, or creating code file, this (potentially) crucial context was not injected in the prompt.
This PR inject the systemMessage when editing or creating code too.  

## Checklist

The current documentation states that systemMessage is not used for completion. Which assumes that it used everywhere else, which it was not. My PR just aligns the actual behavior with the documentation.

## Testing

I have tested with a Claude model on Bedrock, but you are invited to try using a different model, since there are different preprompt. I modified both gpt and claude TS files. 

To test, define a systemMessage then just CMD-I/CTRL-I on a new file, check a the prompt that is sent.
Do the same with a CMD-I/CTRL-I of a piece of existing code.


core/llm/templates/edit/claude.ts
AnnoyingTechnology, line 1: btw, afaik this is not the preffered syntax for talking to claude. If I recall correctly, anthropics recomends using `<somePieceOfCode> ... </somePieceOfCode>` to describe blocks of code.

core/llm/templates/edit/claude.ts
```diff
@@ -12,11 +14,28 @@ Above is the file of code that the user is currently editing in. Their cursor is
 "${otherData.userInput}"
 
 Please generate this code. Your output will be only the code that should replace the "[BLANK]", without repeating any of the prefix or suffix, without any natural language explanation, and without messing up indentation. Here is the code that will replace the "[BLANK]":`;
+    // if the LLM has been configured with a systemMessage
```
sestinj, line 17: I know it may take a bit more shuffling of code, but it's pretty important that the system message be displayed as a system message rather than at the top of the last user message like here

sestinj (changes requested):

core/llm/templates/edit/claude.ts
AnnoyingTechnology: After digging thru the code, I'm not able to figure out how to inject the `systemMessage` at the proper place for edits. It's obvious for the chat mode, but cannot figure it out for edits.

**You'll probably have to do it.**

sestinj: This is fair enough. I'm sorry this ended up being a temporary dead-end. I'll try to do this in the near future when we revisit loading the system message in general

sestinj: Thanks for pointing this out anyway, and making the attempt to solve

PR #3258, opened 2024-12-09
Symbol Thunks and Hovers

RomneyDa:
- Fix symbols not loading on startup
- Move symbols to thunks to speed up
- Only show symbol and file links for context items from this or previous history item
- Show information on hover over links

image

image


sestinj:
This is kinda sick

I think we should
- have a delay before they show up so a normal mouse movement doesn't accidentally trigger it
- try to avoid the background opacity if possible (I realize this is on all of our tooltips right now, and that should probably change across the board)

PR #3261, opened 2024-12-09
config yaml

sestinj:


adam-ah:
I don't understand the point of this PR. You had a perfectly fine JSON config, which now has been moved to YAML for unclear reasons. That directly lead to config reading problems for many users, such as https://github.com/continuedev/continue/issues/5400
This feels like doing it for the sake of doing something

PR #3267, opened 2024-12-09
vscode: Make esbuild-watch script fire when extension.js disappears

halfline:
When running the VS Code extension, the prepackage script cleans the output directory in preparation for it to get populated with new content:

    // Clear folders that will be packaged to ensure clean slate
    ...
    rimrafSync(path.join(__dirname, "..", "out"));
    fs.mkdirSync(path.join(__dirname, "..", "out", "node_modules"), {
      recursive: true,
    });

That, unfortunately, means extension.js can get blown away after it's been built, leading to an error message when launching the test environment:

  Activating extension 'Continue.continue' failed: Cannot find module
  '.../extensions/vscode/out/extension.js'

As part of the build pipeline, there is a esbuild-watch script that watches for changes to the input source to rebuild extension.js when they change.

This commit addresses the above problem by leveraging the esbuild-watch script to also rebuild extension.js when it disappears.

### DIFF
--- a/extensions/vscode/scripts/esbuild.js
+++ b/extensions/vscode/scripts/esbuild.js
@@ -1,4 +1,6 @@
 const esbuild = require("esbuild");
+const fs = require("fs");
+const path = require("path");
 
 const flags = process.argv.slice(2);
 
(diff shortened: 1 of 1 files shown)


RomneyDa:
Linking related PR here
https://github.com/continuedev/continue/pull/3268

found the specific build-completed callback

sestinj (approved):
this is great, thanks!

As Dallin mentioned in Discord, we're happy to accept either this PR with the updated changes, or go ahead and just merge the other, lmk!

halfline:
let's close this one, because it's less root cause-y

halfline:
So Dallin's fix wasn't right, I think.

I posted https://github.com/continuedev/continue/pull/3276 as an alternative, that's a little closer to the meat of the matter.

PR #3268, opened 2024-12-09
Use esbuild onEnd callback

RomneyDa:
Problem was that esbuild.context doesn't wait for build to finish, so activate extension was failing because extension host task was launched before build was complete

Now it won't think the task is complete before the build has finished

Credit @ ray strode [halfline] on discord!

### DIFF
--- a/extensions/vscode/scripts/esbuild.js
+++ b/extensions/vscode/scripts/esbuild.js
@@ -20,27 +20,29 @@ const esbuildConfig = {
   inject: ["./scripts/importMetaUrl.js"],
   define: { "import.meta.url": "importMetaUrl" },
   supported: { "dynamic-import": false },
+  plugins: [
+    {
+      name: "on-end-plugin",
+      setup(build) {
+        build.onEnd((result) => {
+          if (result.errors.length > 0) {
+            console.error("Build failed with errors:", result.errors);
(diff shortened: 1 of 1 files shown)


RomneyDa:
@ray strode [halfline]
https://discord.com/channels/1108621136150929458/1108621136830398496/1315729225126510707

halfline:
so trying this, it doesn't actually fix the problem. I think the issue is if the `vscode-extension:esbuild` task is running in the background from a previous invocation, vscode isn't going to start it again even if it's called for (since it's already running).  So if you have `vscode-extension:esbuild` generate extensions.js then have `vscode-extension:continue-ui:build` call `prepackage.js` which removes extensions.js, calling `vscode-extension:esbuild` again later is going to be a noop, I think, and not regenerate the file (since it's still running in the background)

halfline:
i posted https://github.com/continuedev/continue/pull/3276 to address this I think.

RomneyDa:
@halfline the `onEnd` callback pretty black-and-white solved _first_ launch activation for me, I think we might have been solving two different problems - yours solves restarts, mine solves premature endPattern triggers. Or can you re-iterate which problem it didn't solve? Thanks for the contribution

halfline:
oh if it solves a real problem ignore me

PR #3276, opened 2024-12-10
vscode: Ensure extension.js is rebuilt when deleted

halfline:
Right now, the vscode-extension:continue-ui:build task deletes extension.js as part of its prepackage.js script.

This wouldn't normally be a problem, because extension.js is generated by the vscode-extension:esbuild, and
vscode-extension:continue-ui:build is a dependency of vscode-extension:esbuild, implying it should always get run first.

But, vscode-extension:esbuild is a background task, that stays running after it's been started once. Future calls to run that task become no ops, because it's already active. This means running

vscode-extension:esbuild ->
  vscode-extension:continue-ui:build ->
    vscode-extension:esbuild

will lead to the last vscode-extension:esbuild not regenerating extension.js.

In order to resolve that problem, this commit adds a new foreground task into the build pipeline:

  vscode-extension:touch-source

that tickles the source so the running background
vscode-extension:esbuild task will "wake up" and regenerate extension.js

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -28,6 +28,9 @@
         "vscode-extension:continue-ui:build",
         // // To bundle the code the same way we do for publishing
         "vscode-extension:esbuild",
+        // To ensure extension.js is regenerated even if
+        // vscode-extension:esbuild was already running in background
+        "vscode-extension:esbuild-notify",
         // Start the React app that is used in the extension
         "gui:dev"
       ],
@@ -35,6 +38,13 @@
         "kind": "build",
         "isDefault": true
       }
+    },
+    {
+      "label": "vscode-extension:esbuild-notify",
+      "dependsOn": ["vscode-extension:esbuild"],
+      "type": "npm",
+      "script": "esbuild-notify",
+      "path": "extensions/vscode"
     },
     {
       "label": "vscode-extension:esbuild",
--- a/extensions/vscode/scripts/esbuild.js
+++ b/extensions/vscode/scripts/esbuild.js
@@ -1,4 +1,5 @@
 const esbuild = require("esbuild");
+const fs = require("fs");
 
 const flags = process.argv.slice(2);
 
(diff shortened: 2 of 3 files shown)


halfline:
other ideas:

- Add the file touching to prepackage.js
- Use a nonbackground task for doing the esbuild

.vscode/tasks.json
```diff
@@ -19,6 +22,12 @@
         "isDefault": true
       }
     },
+    {
+      "label": "vscode-extension:touch-source",
+      "type": "shell",
+      "command": "touch",
```
fbricon: I don't think this'll work on windows, `touch` is linux/mac specific

halfline: right, i guess we could add a `--poker-watcher` to `esbuild.js` and call that instead of touching the file. under the hood it would just call `context.rebuild()` to wake up the background task. that's probably nicer than conditionally calling into powershell and munging timestamps wasn't that nice to begin with tbh

halfline: okay, did this (though i called it `--notify` not `--poke-watcher`)

.vscode/tasks.json
```diff
@@ -36,6 +39,13 @@
         "isDefault": true
       }
     },
+    {
+      "label": "vscode-extension:esbuild-notify",
+      "dependsOn": ["vscode-extension:continue-ui:build"],
```
halfline: this should be `vscode-extension:esbuild`

extensions/vscode/scripts/esbuild.js
```diff
@@ -52,6 +52,15 @@ const esbuildConfig = {
   if (flags.includes("--watch")) {
     const ctx = await esbuild.context(esbuildConfig);
     await ctx.watch();
+  } else if (flags.includes("--notify")) {
+    try {
+      const ctx = await esbuild.context(esbuildConfig);
+      await ctx.rebuild();
```
halfline: digging a little more , I don't think this actually works the way I thought it would. Since the contexts aren't shared between notifier and watcher, I now believe the rebuild call is essentially just a build call, and not notifying the watcher.

halfline: i guess i'll go back to the touch approach but do it here

sestinj (approved):
cool! this seems like a fine solution to me. will definitely make development just a bit nicer!

PR #3307, opened 2024-12-11
Remove reranker expand snippet code

RomneyDa:
Remove unused expand snippet reranker code
Left a comment pointing to this PR


sestinj:
Gonna keep this stuff around—we'll always be running new experiments for autocomplete, which might mean dead code from time to time. I should think about a better way to signal this

PR #3314, opened 2024-12-11
AutocompleteTemplate: Use modelfile template for granite3

halfline:
granite3 is currently using the "hole filler" template since it lacks FIM support. This template is getting sent to ollama raw, so the template from the model file is not getting applied to it.

This commit ensures the proper format is fed to the model, by moving the hole filler system message to be a global constant, and using it in the existing holeFillerTemplate and a new granteHoleFillerTemplate that also applies the appropiate model specific formatting.

### DIFF
--- a/core/autocomplete/templating/AutocompleteTemplate.ts
+++ b/core/autocomplete/templating/AutocompleteTemplate.ts
@@ -237,10 +237,8 @@ Fill in the blank to complete the code block. Your response should include only 
   completionOptions: { stop: ["\n"] },
 };
 
-const holeFillerTemplate: AutocompleteTemplate = {
-  template: (prefix: string, suffix: string) => {
-    // From https://github.com/VictorTaelin/AI-scripts
-    const SYSTEM_MSG = `You are a HOLE FILLER. You are provided with a file containing holes, formatted as '{{HOLE_NAME}}'. Your TASK is to complete with a string to replace this hole with, inside a <COMPLETION/> XML tag, including context-aware indentation, if needed.  All completions MUST be truthful, accurate, well-written and correct.
+// From https://github.com/VictorTaelin/AI-scripts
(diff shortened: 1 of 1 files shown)


halfline:
this is marked draft because i haven't tested it yet, so haven't confirmed it doesn't regress results.

I just sketched it out a few days ago after noticing the problem when looking at ollama debug output.

I will test eventually, but for now, I just want to get it off my machine so I don't lose track of it.

core/autocomplete/templating/AutocompleteTemplate.ts
```diff
@@ -380,11 +391,14 @@ export function getTemplateForModel(model: string): AutocompleteTemplate {
...
     lowerCaseModel.includes("gpt") ||
     lowerCaseModel.includes("davinci-002") ||
-    lowerCaseModel.includes("claude") ||
-    lowerCaseModel.includes("granite3")
+    lowerCaseModel.includes("claude")
   ) {
     return holeFillerTemplate;
   }
 
+  if (lowerCaseModel.includes("granite3")) {
```
fbricon, line 399: > lowerCaseModel.includes("granite3") && !lowerCaseModel.includes("-code")

future proofing the next models that'll support FIM

sestinj:
@halfline have you had any time to take a second look at this?

halfline:
> @halfline have you had any time to take a second look at this?

not yet and the latest granite release has a different template. will get back to this sokn

core/autocomplete/templating/AutocompleteTemplate.ts
fbricon: given -code models won't be released for 3.2, you might want to use :

```typescript
lowerCaseModel.match(/^granite3(?:\.1)?-/)
```

sestinj:
I think it may make sense to close this PR for now, especially as we're trying to clean up a large backlog of PRs. @halfline if you get the chance to come back to this let me know and we're of course happy to re-open!

PR #3344, opened 2024-12-12
full screen now opens the current session

HD787:
## Description

clicking on the full screen button will either open the current session or a new one (the old behavior of the last thing that was on the fullscreen window showing up felt weird).

this also helps address an issue where chats are getting duplicated in history due to old chats getting pulled up when the html renders and then resaved under a new session id.

## Screenshots

https://github.com/user-attachments/assets/bf228655-5b90-4e77-a5cd-5aad084a20bd

## Testing

tested on vscode


sestinj (approved):
A clean PR, thanks for the contribution!

Issue #3403, opened 2024-12-16
Error: 400 Invalid 'tools': empty array. Expected an array with minimum length 1, but got an empty array instead.


niquedegraaff:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 0.8.63
- IDE version: VSCode 1.96.0
- Model: gpt-4o-mini and chatgpt-4o-latest
- config.json:
  
{
  "models": [
    {
      "model": "claude-3-5-sonnet-latest",
      "provider": "anthropic",
      "apiKey": "",
      "title": "Claude 3.5 Sonnet"
    },
    {
      "model": "qwen2.5-coder:14b",
      "title": "CodeQwen 2.5 24B (remote)",
      "systemMessage": "You are an expert software developer. You give helpful and concise responses.",
      "apiBase": "MY PRIVATE AI SERVER WAS HERE",
      "provider": "ollama"
    },
    {
      "model": "deepseek-coder-v2:latest",
      "title": "Deepseek coder V2 (remote)",
      "systemMessage": "You are an expert software developer. You give helpful and concise responses.",
      "apiBase": "MY PRIVATE AI SERVER WAS HERE",
      "provider": "ollama"
    },
    {
      "model": "gpt-4o-mini",
      "title": "GPT 4o Mini",
      "systemMessage": "You are an expert software developer. You give helpful and concise responses.",
      "apiKey": "SECRET API WAS HERE",
      "provider": "openai"
    },
    {
      "model": "chatgpt-4o-latest",
      "title": "GPT-4o",
      "systemMessage": "You are an expert software developer. You give helpful and concise responses.",
      "apiKey": "SECRET API WAS HERE",
      "provider": "openai"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Codestral",
    "provider": "mistral",
    "model": "codestral-latest",
    "apiKey": ""
  },
  "customCommands": [
    {
      "name": "test",
[...]

PR #3418, opened 2024-12-17
Don't pass empty or undefined tools body param

RomneyDa:
Don't allow tools to be passed unless there are tools available


RomneyDa:
Fixes https://github.com/continuedev/continue/issues/3403

PR #3456, opened 2024-12-19
ConfigHandler: make updateIdeSettings work (better)

owtaylor:
## Description

`ConfigHandler.updateIdeSettings()` was largely ineffective because the `ideSettingsPromise` was stored in the ProfileLoader object, so updating to a new `ideSettingsPromise` resulted in the config being reloaded with the old settings. Fix this by passing the `ideSettingsPromise` as a parameter to `IProfileLoader.doLoadConfig()`.

I came up with this while working on a larger change (for automatic model selection), so I don't have a great example of when this would matter, but it does seem like this would affect the use of `ideSettings.userToken` in `core/llm/llms/index.ts` - 

```ts
  if (desc.provider === "continue-proxy") {
    options.apiKey = ideSettings.userToken;
    if (ideSettings.remoteConfigServerUrl) {
      options.apiBase = new URL(
        "/proxy/v1",
        ideSettings.remoteConfigServerUrl,
      ).toString();
    }
  }
```

This patch does not fix the case where changing the settings affects what profiles (changes to `ideSettings.remoteConfigServerUrl`)

## Checklist

### DIFF
--- a/core/config/ConfigHandler.ts
+++ b/core/config/ConfigHandler.ts
@@ -51,7 +51,6 @@ export class ConfigHandler {
     // Set local profile as default
     const localProfileLoader = new LocalProfileLoader(
       ide,
-      ideSettingsPromise,
       controlPlaneClient,
       writeLog,
     );
@@ -111,7 +110,6 @@ export class ConfigHandler {
             workspace.name,
             this.controlPlaneClient,
             this.ide,
-            this.ideSettingsPromise,
             this.writeLog,
             this.reloadConfig.bind(this),
           );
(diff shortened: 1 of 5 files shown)


owtaylor:
Rebased to fix conflicts

sestinj:
@owtaylor I appreciate this fix! I agree that it cleans things up, and so it's kind of awkward to say this (especially 3 weeks after your PR was opened), but given that we're going to be removing the need for these remoteConfigServer features, I'm going to close this in favor of the larger change we'll make soon. Before then it feels wise to avoid any kind of shake up in such a core part of the codebase

PR #3469, opened 2024-12-20
Improve article processing and chunking

AnnoyingTechnology:
- Enhance extractTitlesAndBodies function to handle all heading levels
- Improve stringToArticle function with fallback for Readability failures

## Description

Now support h1-6 instead of only h2, which would fail on some documentations. 

Please review carefully before merging.

### DIFF
--- a/core/indexing/docs/article.ts
+++ b/core/indexing/docs/article.ts
@@ -1,3 +1,5 @@
+// Article processing and chunking module
+
 import { Readability } from "@mozilla/readability";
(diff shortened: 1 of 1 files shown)


sestinj:
Before merging let's change h1, h2, etc... to just "h1, h2". I'm worried that this would cause too small of chunks

AnnoyingTechnology:
Right. I'm going to rework this in depths anyways. There are other issues with the chunking that should be addressed (Mozilla's Readability removing the h1, etc.).

PR #3494, opened 2024-12-21
refactor: Adjust inlay rendering and improve completion logic

lkk214:
## Description

- Single-line completion sometimes does not work
- There is a blank area after the cursor when completing multiple lines
- Incompatible line breaks cause accept to fail

## Screenshots
fix14

## Testing
fix15


extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteService.kt
```diff
@@ -66,23 +83,20 @@ class AutocompleteService(private val project: Project) {
 
         // Request a completion from the core
         val virtualFile = FileDocumentManager.getInstance().getFile(editor.document)
+        val line = editor.caretModel.primaryCaret.logicalPosition.line
         val column = editor.caretModel.primaryCaret.logicalPosition.column
         val input = mapOf(
             "completionId" to completionId,
-            "filepath" to virtualFile?.url,
+            "filepath" to virtualFile?.path,
```
sestinj: why change url to path?

sestinj: (was this the solution to a clear problem?)

sestinj (commented):
This works great for me, I'm just curious if the url -> path was the solution to something that was previously broken

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteService.kt
lkk214: I was just about to comment on it. I didn't pay attention to https://github.com/continuedev/continue/pull/3247, need to revert it here.

Because using `File(URI(filepath))` will cause an error in Windows, I made temporary changes locally and submitted this by mistake.

lkk214: Uh, maybe it's better to fix it than revert it.

lkk214:
Also, add directory checks and optimize token usage.
- Added check to skip directories when reading file content.
- Optimized token usage by removing unnecessary `\r` characters.

sestinj:
Looks good! @Patrick-Erichsen for a second pass, then good to merge

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/utils/Utils.kt
```diff
@@ -77,4 +78,6 @@ fun getMachineUniqueID(): String {
 
 fun uuid(): String {
     return UUID.randomUUID().toString()
-}
\ No newline at end of file
+}
+
+fun VirtualFile.toUriOrNull(): String? = fileSystem.getNioPath(this)?.toUri()?.toString()?.removeSuffix("/")
```
Patrick-Erichsen, line 83: @RomneyDa any thoughts here? I'm wondering if `.removeSuffix("/")` is going to not work properly for non-macOS users

Patrick-Erichsen:
LGTM 🫡

AfterStories:
@lkk214 @Patrick-Erichsen @sestinj 
 I still found new similar problems, I think it is also caused by the same bug, but it has not been fixed:
 When I select a piece of code and use the CTRL+I shortcut key, as shown in the screenshot, you can see that the model name is not displayed, only show a null
 
 And at the same time, the log in the Run Extension window:

```
Error handling message: {"messageType":"config/getSerializedProfileInfo","data":{"result":{"config":{"allowAnonymousTelemetry":false,"models":[{"provider":"anthropic","model":"Gosu Model","title":"Gosu Model"... ....(After here, it's my config. json content) 
java.lang.NullPointerException: null cannot be cast to non-null type kotlin.collections.Map<kotlin.String, kotlin.Any>
Registered handler for editor

```

image
image

lkk214:
From the log you provided, it seems that it is a response parsing error. It is not the same issue. 
I have not updated the code, so it works normally locally. The relevant code for response parsing:https://github.com/continuedev/continue/blob/7af1090828db96820260733edf516b81c77b3d2d/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/InlineEditAction.kt#L131-L135

AfterStories:
> From the log you provided, it seems that it is a response parsing error. It is not the same issue. I have not updated the code, so it works normally locally. The relevant code for response parsing:
> 
> https://github.com/continuedev/continue/blob/7af1090828db96820260733edf516b81c77b3d2d/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/InlineEditAction.kt#L131-L135




Yes, I also found this code, and I saw that it made a "config/getSerializedProfileInfo" request to get the content of the config,
but maybe there was a problem in the kt code post message to the “core”  code ，so failed to get the correct response.
Or the core code failed to parse the config.json file and return its content

PR #3500, opened 2024-12-21
shorter more descriptive chat titles

HD787:
## Description

this prompt produces titles which are shorter, more descriptive and less likely to be truncated, this reliably produces more descriptive titles on smaller models, the changes are less pronounced on larger models but still prevents things such as starting with a verb in the present participle ("summarizing", "describing", "discussing")  which don't contribute meaning to the title.

## Screenshots

image

this shows the old summaries alongside new ones, I think it is pretty easy to tell which is which.

## Testing instructions

tested mainly with llama 3.1 8b and chatgpt 4o


sestinj:
This is nice!

PR #3536, opened 2024-12-25
fix: Skip duplicate checks at document end or blank text

lkk214:
## Description

Prevent autocomplete text from being truncated when the cursor is at the end of the document or the text after the cursor is blank.


Patrick-Erichsen:
Hoping this will also solve https://github.com/continuedev/continue/issues/3577

lkk214:
> Hoping this will also solve #3577

Looks like this should fix it.

sestinj (approved):

PR #3581, opened 2025-01-01
Fix examples in docs so that base model of Qwen2.5-coder is used for ollama instead of instruct

rogeroberg:
## Description

Ollama.com models repository uses the Instruct models as default, but for autocomplete the base models should be used to avoid potentially issues. E.g. in order to use the base model you need to add the -base to the model you want to use.
I saw that the autocomplete model in the onboarding.ts had been fixed in the  PR https://github.com/continuedev/continue/pull/3443
This PR updates the documentation when using the Qwen2.5-coder models for autocomplete with Ollama so that the base model is used instead.

## Checklist

### DIFF
--- a/core/autocomplete/README.md
+++ b/core/autocomplete/README.md
@@ -7,7 +7,7 @@ Continue now provides support for tab autocomplete in [VS Code](https://marketpl
 We recommend setting up tab-autocomplete with a local Ollama instance. To do this, first download the latest version of Ollama from [here](https://ollama.ai). Then, run the following command to download our recommended model:
 
 ```bash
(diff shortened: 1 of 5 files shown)


SlavikCA:
I know that this MR is merged already.

But does anyone knows, is the BASE model really that better for autocomplete than INSTRUCT?

I'm currently setup with `qwen2.5-coder:32b` to do chat, edit and autocomplete. The big advantage is that one model used for everything, so it's not unloaded / reloaded to VRAM for different tasks. Will `qwen2.5-coder-BASE` be that much better for autocomplete?

Issue #3610, opened 2025-01-05
Default git diff is Included, contradict with config schema. Please set it back to false


retry51776:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:Mac
- Continue version: 0.8.66
- IDE version: VS code
- Model: codellama
- config.json:
```

### Description

According to schema, "git diff" default value should be false.
https://github.com/continuedev/continue/blob/aa02e0bd630fa5700d8cb48f17b5e624b940f095/extensions/vscode/config_schema.json#L1805C21-L1805C181

But in source code 
https://github.com/continuedev/continue/blob/aa02e0bd630fa5700d8cb48f17b5e624b940f095/core/commands/slash/commit.ts#L8
Default "git diff is true"

### To reproduce

Any autocomplete output will show git diff

`
==========================================================================
==========================================================================
##### Completion options #####
{
  "contextLength": 8096,
  "stop": [
    "<PRE>",
    "<SUF>",
    "<MID>",
    "<EOT>",
    "[INST]",
    "[/INST]",
    "<<SYS>>",
    "<</SYS>>",
    "<PRE>",
    "<SUF>",
    "<MID>",
    "<EOT>",
    "/src/",
    "#- coding: utf-8",
    "```"
  ],
  "model": "codellama:latest",
  "maxTokens": 4096,
  "temperature": 0.01,
  "raw": true
}

##### Prompt #####
<PRE> # diff --git a/README.md b/README.md
# index 9dd61f7..df0b019 100644
# --- a/README.md
# +++ b/README.md
# @@ -1,18 +1,20 @@
#  # notes
# +
#  [![Codacy Badge](https://app.codacy.com/project/badge/Grade/03049fab007a4e098dc046095b4f5928)](https://www.codacy.com/gh/retry51776/notes/dashboard?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=retry51776/notes&amp;utm_campaign=Badge_Grade)
[...]

PR #3633, opened 2025-01-07
fix: set `includeUnstaged` default to `false`

Patrick-Erichsen:
Resolves https://github.com/continuedev/continue/issues/3610


Z-feiyao:
I found a new problem, when adding a new file to the project, the new file is not found in the git diff

PR #3653, opened 2025-01-08
Add the ability to start Ollama when it's stopped

### LINKED ISSUE #3318: When ollama is stopped, propose to start it, instead of downloading it
### Validations

### Problem

When ollama is stopped, you get this error message:

Screenshot 2024-12-09 at 14 47 51

which can be pretty confusing, since well, ollama is already downloaded/installed. We should have a way of starting it instead.

I have some (ASL2) code for checking ollama is installed and for starting it, that I can contribute. I need to do a bit of cleanup first.

### Solution

- if ollama is defined as local (check apiUrl in config.json), then check the output of `ollama -v` to see if it's installed
- there are a few ways to start ollama, depending on the platform:
   * on Mac, run `ollama list`, will start ollama as a service
   * on Linux, run the ollama service if available, else fall back to running `ollama serve`
   * on Windows, AFAIK there's no way to run ollama as a service from CLI, so just do `ollama serve`
   
  So, if ollama is installed, then just replace the `Download Ollama` button with a `Start Ollama` one, hooked into the startup code.

fbricon:
Fixes #3318 

contributing the ollama_start.sh script from [vscode-paver](https://github.com/redhat-developer/vscode-paver/blob/main/start_ollama.sh)

Capture d'écran 2025-01-09 170435

Checked it works from a vsix on all platforms:
- [x] Mac
- [x] Linux
- [x] Windows

### DIFF
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -55,6 +55,7 @@ import {
   toCompleteBody,
   toFimBody,
 } from "./openaiTypeConverters.js";
+import { isOllamaInstalled } from "../util/ollamaHelper.js";
 
 export abstract class BaseLLM implements ILLM {
   static providerName: string;
@@ -414,9 +415,11 @@ export abstract class BaseLLM implements ILLM {
             e.code === "ECONNREFUSED" &&
             e.message.includes("http://127.0.0.1:11434")
           ) {
-            throw new Error(
-              "Failed to connect to local Ollama instance, please ensure Ollama is both installed and running. You can download Ollama from https://ollama.ai.",
-            );
+            const message = (await isOllamaInstalled()) ?
+              "Failed to connect to local Ollama instance. Ollama appears to be stopped. It needs to be running." :
+              "Failed to connect to local Ollama instance, please ensure Ollama is both installed and running. You can download Ollama from https://ollama.ai."
+            ;
+            throw new Error(message);
           }
         }
         throw new Error(e.message);
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -416,8 +416,8 @@ export abstract class BaseLLM implements ILLM {
             e.message.includes("http://127.0.0.1:11434")
           ) {
             const message = (await isOllamaInstalled()) ?
-              "Failed to connect to local Ollama instance. Ollama appears to be stopped. It needs to be running." :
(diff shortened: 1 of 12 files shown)


fbricon:
It was almost ready, until I found a bug in ide.runCommand :-/ It reuses a terminal than might still be busy (eg. tsc:watch task). I want to change the runCommand API to add an optional options object, where I can declare we need to use a specific terminal

fbricon:
@tomasz-stefaniak It should be ready to review. I'll still build a vsix on windows to double check tomorrow, but since it works on Mac and Linux already, I'm fairly confident about it. famous last words ;-)

core/llm/index.ts
```diff
@@ -414,9 +415,11 @@ export abstract class BaseLLM implements ILLM {
             e.code === "ECONNREFUSED" &&
             e.message.includes("http://127.0.0.1:11434")
           ) {
-            throw new Error(
-              "Failed to connect to local Ollama instance, please ensure Ollama is both installed and running. You can download Ollama from https://ollama.ai.",
-            );
+            const message = (await isOllamaInstalled()) ?
+              "Failed to connect to local Ollama instance. Ollama appears to be stopped. It needs to be running." :
```
allanday: Three sentences feels a bit hard to read, and reads a bit abruptly. Reducing it down to two sentences would be easier to digest.

```suggestion
              "Unable to connect to local Ollama instance. Ollama may not be running." :
```

core/llm/index.ts
```diff
@@ -414,9 +415,11 @@ export abstract class BaseLLM implements ILLM {
...
+              "Failed to connect to local Ollama instance, please ensure Ollama is both installed and running. You can download Ollama from https://ollama.ai."
```
allanday: The notification text is a bit long here. Personally I'd drop the download link - it's too much detail for a notification and you'd hope that someone could figure out what to do.

```suggestion
              "Unable to connect to local Ollama instance. Ollama may not be installed or may not running."
```

fbricon:
Ollama detection now works on windows as well.

Updated the error messages as per @allanday's suggestions:
 
Capture d'écran 2025-01-09 170435
Capture d'écran 2025-01-09 170929


@tomasz-stefaniak @sestinj feel free to review

tomasz-stefaniak (dismissed):
Thanks @fbricon

tomasz-stefaniak:
FYI, the tests were failing because of permissions issues but I ran them locally

PR #3673, opened 2025-01-10
Update prompt-files.md

sbalk:
Simplified and added system wide prompt files instruction. 

## Description

In my experience, without the system-tags, it wasn't effective. I don't know anything about the implementation. If you could point me in the right direction, I'm happy to test and edit more. I saw that earlier versions of this file did have the system-tag.

Also, thanks for your work on this project, I love it.

## Checklist

## Screenshots

## Testing instructions


sestinj:
@sbalk We don't implement the system tag any more because .prompt files are invoked as context providers. This means multiple of them can be used in the same prompt, and if multiple of them adjusted the system prompt it wouldn't be clear how these are merged.

So for now I'm going to close this, though I appreciate what I'm taking as feedback: we should reconsider bringing back slash command prompt files

Issue #2311, opened 2024-09-19
"Stop TTS"-Button doesn't stop TTS


Snowman-25:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11 23H2 (Build 22631.4169)
- Continue:   v0.8.52
- IDE:        VSCode 1.92.1 (user setup)
- Model:      n/a
- config.json:
  
{
  "models": [
    {
      "title": "Ollama",
      "provider": "ollama",
      "model": "AUTODETECT"
    }
  ],
  "customCommands": [
    {
      "name": "test",
      "prompt": "{{{ input }}}\n\nWrite a comprehensive set of unit tests for the selected code. It should setup, run tests that check for correctness including important edge cases, and teardown. Ensure that the tests are complete and sophisticated. Give the tests just as chat output, don't edit any file.",
      "description": "Write unit tests for highlighted code"
    },
    {
      "name": "desc",
      "prompt": "{{{ input }}}\n\nWrite a comprehensive comment for each block of the selected code. It should describe what it does and show possible caveats.",
      "description": "Comment the highlighted code"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Starcoder 3b",
    "provider": "ollama",
    "model": "starcoder2"
  },
  "allowAnonymousTelemetry": false,
  "embeddingsProvider": {
    "provider": "transformers.js"
  },
  "contextProviders": [
    {
      "name": "open",
      "params": {
        "onlyPinned": false
      }
     }
  ],
  "experimental": {
    "readResponseTTS": true
  },
  "ui": {
    "showChatScrollbar": true
  },
  "docs": []
}
```

### Description

After activating TTS, while a response is spoken, there's a "Stop TTS"-Button where the 'Cancel' button is when the model is still generating an answer.
It doesn't work. TTS-Output doesn't stop.

### To reproduce

1. Activate TTS by adding readResponseTTS: True to your experimental-section in config.json
2. Enter any prompt
3. While the prompt is being read, press "Stop TTS"

Patrick-Erichsen:
[...]

PR #3695, opened 2025-01-12
New feature added: Integrated WebView with Native Speech and PiperTTS

alpgul:
## Description
- [Native Speech](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisVoice) functionality is integrated within the WebView
- Implemented [PiperTTS ](https://github.com/Mintplex-Labs/piper-tts-web)(Text-to-Speech) support.

## Checklist

## Config
You can get the voice list from[ this link](https://huggingface.co/diffusionstudio/piper-voices/resolve/main/voices.json).

```
...
"readResponseTTS": {"type":"native","lang":"en-US"}
```
```
...
"readResponseTTS": {"type":"piperTTS","lang":"en-US"}
```
```
...
"readResponseTTS": {"type":"piperTTS","lang":"en_US-arctic-medium"}
```
## Fixed 
 Resolved issue with stopping audio playback #2311

## Issue
Usage of worker within WebView only accepts blob URLs. As a temporary solution, a worker has been defined. Vite can't detect the web worker with blobURL, and this issue needs to be resolved.


gui/src/hooks/useSetup.ts
jajibhee, line 1: The `useWebviewListener("setTTSNative")` handler's mixed return types (`true`, `false`, or a `Promise`) can cause inconsistent behavior, making it harder to handle. Standardizing the return type to always be a `Promise` or whatever is ideal ensures predictability and simplifies integration, reducing potential bugs and improving maintainability.

The `Promise` in `setTTSNative` lacks explicit error handling, which could result in uncaught errors during unexpected failures. Try wrapping the logic in a `try-catch` block to ensure errors are logged and handled.

alpgul: > The `useWebviewListener("setTTSNative")` handler's mixed return types (`true`, `false`, or a `Promise`) can cause inconsistent behavior, making it harder to handle. Standardizing the return type to always be a `Promise` or whatever is ideal ensures predictability and simplifies integration, reducing potential bugs and improving maintainability.
> 
> The `Promise` in `setTTSNative` lacks explicit error handling, which could result in uncaught errors during unexpected failures. Try wrapping the logic in a `try-catch` block to ensure errors are logged and handled.

`setTTSNative:[{lang:string, message:string}, boolean];`
In the code above, the return value is defined only as `boolean`, and there isn't `Promise `return type.

Since only the event is defined within the Promise, any error will only affect the event and will not impact the Promise

sestinj:
@alpgul I appreciate the extensive PR and can tell that it was well done, but unfortunately right now we want to avoid adding additional models that are shipped alongside the extension. This leads to a serious maintenance burden and given the current level usage of this feature it isn't something we are willing to take on.

If you have other thoughts about how this could be accomplished we're open to discussing, but it might be best to open an issue to describe the approach before opening a complete PR

PR #3708, opened 2025-01-13
Fix syntax highlighting for .class files

fbricon:
## Description

Syntax highlighting was missing when adding snippets from a Java .class file to the context. This was caused by 2 issues:
- `class` was not mapped to `java`
- the extension mapping would fail because the extension would contain the snippet range, like `class (1-13)`

## Checklist

## Screenshots

Before 
Screenshot 2025-01-13 at 16 22 32

After
Screenshot 2025-01-13 at 16 30 05

## Testing instructions
- install [vscode-java](https://marketplace.visualstudio.com/items?itemName=redhat.java)
- create a java file like 
```java
void main() {
   System.out.println("Hello World!');
}
```
- Ctrl+click on System to open System.class
- highlight some code, Ctrl+L to add to the Chat context
- the snippet should be highlighted as Java


sestinj (approved):
Great, and thanks for adding the test!

PR #3711, opened 2025-01-13
Fix extra slash in embedding for vertex

Lash-L:
## Description

Seems like a bug was introduced when the embeddings were merged into the llm.
https://github.com/continuedev/continue/commit/b14b8c4652c5b132418ad1288444a431e5d81c55#diff-f6056e365b00006d0872276e5078b8e10b940dfeca7f0aa106b1c842e5aa5a14

The embedding class's base url didn't have a slash at the end, while the llm one did. Neither one was 'wrong', but they were different, so when it was merged, it ended up having a // in the url. 

Removing the / before the embedding url fixes this problem.

## Checklist

## Screenshots

## Testing instructions

### DIFF
--- a/core/llm/llms/VertexAI.ts
+++ b/core/llm/llms/VertexAI.ts
@@ -445,7 +445,7 @@ class VertexAI extends BaseLLM {
     }
 
     const resp = await this.fetch(
-      new URL(this.apiBase + `/publishers/google/models/${this.model}:predict`),
+      new URL(this.apiBase + `publishers/google/models/${this.model}:predict`),
       {
         method: "POST",
         body: JSON.stringify({


sestinj (approved):

sestinj (changes requested):
@Lash-L Thanks for making this fix. Any chance we could add a quick test to `llm.test.ts` (or nearby) for this? It should be fast, it would just be good to have a way of knowing going forward that we keep the Vertex provider in a state that works for you

Lash-L:
@sestinj I can try to do so! It may be a bit before i can get around to it though, as i need to re set up my continue dev environment and i'm pretty busy for the next week or so

It may make sense to merge this in and I can try to add the tests as soon as i am able

sestinj:
Cool, no worries, I definitely want to try to be more rigorous than in the past about having tests for new bug fixes, but if I have time before you're back to it I'll go ahead and add the test 

Thanks again for the fix!

Lash-L:
Closing this in favor of: https://github.com/continuedev/continue/pull/4016

As I found another bug. I am not 100% sure what you are looking for re tests, so I sent you a message on discord to try to get some clarify. A response here would work as well!

PR #3723, opened 2025-01-14
bugfix: disable streaming for o1

Patrick-Erichsen:
Addresses https://github.com/continuedev/continue/issues/3698


sestinj (approved):

Patrick-Erichsen:
@sestinj good catch! Copypasta from when that was nested underneath another if-statement in `openai-adapters`.

PR #3724, opened 2025-01-14
Nate/platform

sestinj:


sestinj:
@tomasz-stefaniak finally ready here. Explanation above for why I don't think now is the appropriate time to write tests, but prior to real rollout l want to basically run all of our end to end tests in a matrix with both config.json and config.yaml

sestinj:
Decided to write a couple, WIP

tomasz-stefaniak (dismissed):

Issue #3674, opened 2025-01-10
When Ollama is stopped, the chat window shows an unactionable error


fbricon:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: v0.9.5
- IDE version: vscode 1.96.2
- Model: anything running on local Ollama
- config.json:
  
{
"models":[
        {
            "model": "AUTODETECT",
            "title": "Autodetect",
            "provider": "ollama"
        }
],
...
}
```

### Description

Try asking a question in the chat view, while ollama is stopped, you'll get a cryptic, unhelpful javascript error:

<img width="484" alt="Screenshot 2025-01-10 at 16 55 33" src="https://github.com/user-attachments/assets/27fbc6d5-fa29-4de2-9582-87232475d091" />

Instead, we should have something similar to the notification shown on completion (See #3653)
### Log output

In the "Window", output view, you can see VS Code's log:

```Shell
2025-01-10 17:00:30.100 [error] [Extension Host] Error handling webview message: {
  "msg": {
    "messageId": "3a7f8f52-a829-4110-af4f-18bd9b24f73c",
    "messageType": "llm/streamChat",
    "data": {
      "messages": [
        {
          "role": "user",
          "content": [
            {
              "type": "text",
              "text": "\n\n TextDocumentContentChangeEvent.class (95-95)\n  public void setRangeLength\n```\nwhat's that?"
            }
          ]
        },
        {
          "role": "assistant",
          "content": ""
        }
      ],
      "title": "granite3.1-dense:8b",
      "completionOptions": {}
    }
  }
}

Error: Failed to connect to local Ollama instance. Ollama appears to be stopped. It needs to be running.
```

PR #3725, opened 2025-01-14
Fix "cannot read undefined content" bug

RomneyDa:
- Fix the "trying to read content from undefined" bug
- Flesh out the llm streaming and messaging types to prevent this from happening again

Note was caused on my PR #3597


fbricon:
definitely improves #3674

Screenshot 2025-01-15 at 12 24 14

We'll still need to expose a way to start Ollama, but that can be done separately.

AfterStories:
I hope this PR can be approved quickly.

sestinj (approved):

PR #3755, opened 2025-01-17
pr_checks: Disable the SSH e2e test when not configured

owtaylor:
The SSH e2e test requires a host to be running permanently with SSH_HOST/SSH_KEY set as secrets; if these secrets are not set, then exclude it.

This should fix the problem where PR's from external contributors always have the e2e tests fail - even though only a single test requires the SSH setup, the "Set up SSH" step fails on each vscode-e2e-tests job.

It would be nicer to show the test as skipped rather than just excluding it entirely, but I didn't manage to figure out how to do that.

### DIFF
--- a/.github/workflows/pr_checks.yaml
+++ b/.github/workflows/pr_checks.yaml
@@ -283,8 +283,11 @@ jobs:
   vscode-get-test-file-matrix:
     runs-on: ubuntu-latest
     needs: [install-root, install-vscode]
+    env:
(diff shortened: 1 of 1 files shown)


sestinj:
Ended up making this happen in a separate PR, so going to close, but thanks for raising the issue!

lance-cheng-asus:
hi @sestinj, where is the separate  PR?

sestinj:
I don't know off the top of my head, but it should appear in the git blame of pr_checks.yaml, where I added an env var to skip them

PR #3774, opened 2025-01-20
add geekai to llm provider

geekr-dev:
## Description

add geekai to llm provider, geekai is an openai compatible, unified llms interface like openrouter with lower prices

## Checklist

- modified: core/llm/autodetect.ts
- modified: core/llm/llms/index.ts
- modified: docs/docusaurus.config.js
- modified: extensions/vscode/config_schema.json
- modified: packages/openai-adapters/README.md
- modified: packages/openai-adapters/src/index.ts
- modified: packages/openai-adapters/src/types.ts
- new file: core/llm/llms/GeekAI.ts
- new file: docs/docs/customize/model-providers/more/geekai.md
- new file: docs/i18n/zh-CN/docusaurus-plugin-content-docs/current/customize/model-providers/more/geekai.md


sestinj:
Hi @geekr-dev, I'm thinking it might make sense to just use the `openai` provider for this one, given that it is fully OpenAI-compatible. We're starting to have a really large number of providers and want to be careful about adding all endpoints as their own dedicated class.

I understand that you'll probably still want to make it as easy as possible for users to try your service in Continue, and for this we'll have an even better solution very soon.

I will leave this PR open for some time in order to follow up on the other solution, to make sure it doesn't get forgotten

sestinj:
Hi @geekr-dev, I wanted to share the new solution, which would be to add a model block at hub.continue.dev. This will make it much easier for users to discover your service and even easier to add it within Continue.

If you have any questions or I can help with this, please let me know!

Issue #3786, opened 2025-01-20
Proposed configuration option for better control over hard-coded system messages


ferenci84:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Problem: There are hard-coded system message and there is no way to override.

Why is it a problem:
Reason 1: Some models behave better if we do not add everything possible into the system message. 
Reason 2: If the default system message is wrong (which is the case now, as it breaks code formatting), it would be easier to test a better version if there is a way to override
Reason 3: Developers are said to be smart, let's give them more control

### Solution

There may be plenty of other solutions, this is just one of them:
A config option, named "replaceSystemMessage" (self-descriptive), default false. This would keep the default behavior as it is now, and let us configure the system message for each model separately.

I have already made the modification, pull request will be done in a moment.

ferenci84:
I have already made a pull request. I'll continue to work on it based on suggestions.

ferenci84:
@RomneyDa @sestinj 
Here is an example when you get better results not including instructions about how to format code responses.

Response when default system instructions are included (with all the details how to format code responses):
<img width="603" alt="Image" src="https://github.com/user-attachments/assets/b9c4667f-f773-496a-ad99-a93cf666a0da" />

Response default system instructions omitted:
<img width="604" alt="Image" src="https://github.com/user-attachments/assets/e17ce8ac-64ac-49fa-a727-2e85c8aa02d4" />

There are lot of other use-cases in which formatting instructions are irrelevant (for example writing documentation, composing system instructions, that is programming related, but do not require code output), in those cases user may want to provide custom system instructions specific for that case. My opinion is that user should be able to decide for specific model setup whether it should be included or not.

ferenci84:
@sestinj in reference to the conversation on PR https://github.com/continuedev/continue/pull/3787:

> We've taken step #1 and removed the massive formatting prompt that was causing so many problems. I want to take the time to get more feedback on how this works before completely adding a new field to the config, but in the case that we do, I am heavily leaning toward the following:
> 
> ```yaml
> models:
>   - name: Example
>      provider: anthropic
>      model: claude-3-5-sonnet-latest
>      apiKey: <API_KEY>
>      promptTemplates:
[...]

PR #3787, opened 2025-01-20
replaceSystemMessage model config option

ferenci84:
## Description

https://github.com/continuedev/continue/issues/3786

A model config option, named "replaceSystemMessage" (self-descriptive), default false.

Behavior:
When true, the model's systemMessage property, or the config's root systemMessage property will replace the default hard-coded system message (Something like "When generating new code:\n\n1. Always produce a single code block.\n2. Never separate the code into multiple code blocks.\n3. Only include the code that is being added.\n4. Replace existing code with a \"lazy\" comment like this ...." and also "When using tools, follow the following guidelines:\n- Avoid calling tools unless they are absolutely necessary...")

Setting false or not setting at all, will keep the current behavior.

Minor change:
The existing behavior, that the model config's systemMessage has higher precedence than the root config's systemMesage, was not documented, I added that to make the information in the documentation more obvious about the new config.

## Checklist

## Screenshots

### Testing replaceSystemMessage = true
1. Create a systemMessage in the default Claude model config.
2. Add replaceSystemMessage true in the model config.
3. Send any message with the model
4. Check the Continue LLM Prompt/Completion on the Output tab.
5. The system message in the sent request should now be the just the one given in the model config.
6. Remove the config and send the message again.
7. The system message should be appended to the default one.

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -43,13 +43,13 @@ export interface IndexingProgressUpdate {
   desc: string;
   shouldClearIndexes?: boolean;
   status:
-    | "loading"
-    | "indexing"
-    | "done"
-    | "failed"
-    | "paused"
-    | "disabled"
-    | "cancelled";
+  | "loading"
+  | "indexing"
+  | "done"
+  | "failed"
+  | "paused"
+  | "disabled"
+  | "cancelled";
   debugInfo?: string;
 }
 
@@ -450,6 +450,7 @@ export interface LLMOptions {
   title?: string;
   uniqueId?: string;
   systemMessage?: string;
+  replaceSystemMessage?: boolean;
   contextLength?: number;
   maxStopWords?: number;
   completionOptions?: CompletionOptions;
@@ -650,10 +651,10 @@ export interface IDE {
   getCurrentFile(): Promise<
     | undefined
     | {
-        isUntitled: boolean;
-        path: string;
-        contents: string;
-      }
+      isUntitled: boolean;
+      path: string;
+      contents: string;
+    }
   >;
 
   getPinnedFiles(): Promise<string[]>;
@@ -832,11 +833,11 @@ export interface CustomCommand {
 export interface Prediction {
   type: "content";
   content:
-    | string
-    | {
-        type: "text";
-        text: string;
-      }[];
+  | string
+  | {
+    type: "text";
+    text: string;
+  }[];
 }
 
 export interface ToolExtras {
@@ -903,6 +904,7 @@ export interface ModelDescription {
   template?: TemplateType;
   completionOptions?: BaseCompletionOptions;
   systemMessage?: string;
+  replaceSystemMessage?: boolean;
   requestOptions?: RequestOptions;
   promptTemplates?: { [key: string]: string };
   capabilities?: ModelCapability;
@@ -1157,9 +1159,9 @@ export interface Config {
   embeddingsProvider?: EmbeddingsProviderDescription | ILLM;
   /** The model that Continue will use for tab autocompletions. */
   tabAutocompleteModel?:
-    | CustomLLM
-    | ModelDescription
-    | (CustomLLM | ModelDescription)[];
+  | CustomLLM
+  | ModelDescription
+  | (CustomLLM | ModelDescription)[];
   /** Options for tab autocomplete */
   tabAutocompleteOptions?: Partial<TabAutocompleteOptions>;
   /** UI styles customization */
@@ -1249,9 +1251,9 @@ export type PackageDetailsSuccess = PackageDetails & {
 export type PackageDocsResult = {
   packageInfo: ParsedPackageInfo;
 } & (
-  | { error: string; details?: never }
-  | { details: PackageDetailsSuccess; error?: never }
-);
+    | { error: string; details?: never }
+    | { details: PackageDetailsSuccess; error?: never }
+  );
 
 export interface TerminalOptions {
   reuseTerminal?: boolean;
--- a/core/llm/countTokens.ts
+++ b/core/llm/countTokens.ts
@@ -1,6 +1,6 @@
 import { Tiktoken, encodingForModel as _encodingForModel } from "js-tiktoken";
 
-import { ChatMessage, MessageContent, MessagePart } from "../index.js";
+import { ChatMessage, LLMOptions, MessageContent, MessagePart } from "../index.js";
 
 import { renderChatMessage } from "../util/messageContent.js";
 import {
@@ -27,9 +27,9 @@ class LlamaEncoding implements Encoding {
 }
 
 class NonWorkerAsyncEncoder implements AsyncEncoder {
-  constructor(private readonly encoding: Encoding) {}
+  constructor(private readonly encoding: Encoding) { }
 
-  async close(): Promise<void> {}
+  async close(): Promise<void> { }
 
   async encode(text: string): Promise<number[]> {
     return this.encoding.encode(text);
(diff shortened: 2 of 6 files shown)


sestinj:
@ferenci84 thanks for the thoughtful PR! I'm thinking that there might be a deeper problem, which is just that the default system message isn't satisfactory. Can you share what the biggest motives for replacing it are? If our default was just problematic then I'd want to fix that first before adding additional complexity to config.json. At the end of the day, we're always going to be forced to have some level of opinion about prompting in order to have a functioning product, so a full replace would be somewhat misleading, or we wouldn't do it entirely

ferenci84:
Hi sestinj, as per my experience, especially with Claude, which is the best LLM available in my opinion (not with flaws though), that a small change in system message cause major effect in results. I experimented a lot, starting with a quite complex system message containing many instructions, and I found that the highest quality code was produced when I simplified the most. Also I believe some LLMs may work the best without system message, and one system message may work with one and not work with an other.
The specific problem with the system message was that the sample code has no path, so it doesn't include paths in the response either, however this is not the question in this PR, the question is the ability to customize. For me, the ability to customizing the system message is a deal breaker. I think it's just impossible to craft a system message that works in all the cases (and especially with all different LLMs). A basic default with option to customize should be sufficient I think.
Right now I'm using the customized version of Continue, and I just excluded all instructions about how to format code and how to use the tools, and I'm happy with the result this way.

FYI, my instructions that I found and simplified to the point where it didn't affect the code quality of Claude: "You are intelligent, helpful and an expert developer, who always gives the correct answer and follow instructions. Style your response using Github Flavored Markdown." (In one instance I had to remove "Style your response using Github Flavored Markdown.") I also got amazing results with these additional instructions and adding a complex high-level task description in a prompt file: "User may use the following tags to provide additional context before describing the task: <project-context>Project-level information like tools and versions used, or additional guidelines about the structure to follow when adding features to the project.</project-context>\n<task-context>Background information about the current task we are implementing. Note that the current instructions may implement only a part of the whole task, but it may help to know the whole task in advance.</task-context>\n<step-context>Additional information about the current step within the task, for example user may describe features already implemented and the features we are implementing now.</step-context>\nUse the tags described above in order to find additional context, but only follow the instructions at the end of the user message. Context tags are optional, they may be missing." I also added this section too, but it has no effect at all: "If the user submits a code block that contains a filename in the language specifier, always include the filename along with the path in any code block you generate based on that file. The filename and path should be on the same line as the language specifier in your code block."

I just show this to point to the fact that it's not a one size fits all situation. These instructions work for me, I have made a lot of experiments, and I need to be able to transfer them from one extension to the other to get predictable results. I started with the Genie extension, then used CodeGPT in IntelliJ, and now Continue.

ferenci84:
@sestinj Adding to my previous message: I'm planning to experiment the specific changes on the default system instructions, specifically to adding paths to test files, i.e. "src/test.js" instead of "test.js". I would create a separate PR about this modification.

Also one improvement to this PR may be that instead of replacing allow user to include placeholders like {{modification_instructions}} and {{tools_instructions}}, or just {{default_instructions}}. This way it's possible to prepend, append or insert the default, for example: "You are intelligent, helpful and an expert developer.\n\n{{default_instructions}}\n\nUser may use the following tags to provide additional context...". May be there could be two options: "append" (current behavior) and "placeholder" or just a boolean parameter "systemMessagePlaceholders" with ussage explained in the description of the setting. In addition to giving more freedom, this parameter naming may be better, because it won't raise the direct question whether it's a good idea to replace the system message, but it inherently give the user freedom to do so.

Also thinking that it would be better for user to see the default instructions and change if needed or easily revert back to the default. This would be possible by including it in the VS code settings with a default. This is better than hardcoding, gives the user possibility to look the default, change and revert back to default.

OlhinAS:
Any update on the PR? This feature is kinda useful since default systemMessage has some pretty specific rules:

> When generating new code:
> 
> 1. Always produce a single code block.
> 2. Never separate the code into multiple code blocks.
> 3. Only include the code that is being added.
> 4. Replace existing code with a "lazy" comment like this: "// ... existing code ..."
> 5. The "lazy" comment must always be a valid comment in the current context (e.g. "" for HTML, "// ... existing code ..." for JavaScript, "{/* ... existing code */}" for TSX, etc.)
> 6. You must always provide 1-2 lines of context above and below a "lazy" comment
> 7. If the user submits a code block that contains a filename in the language specifier, always include the filename in any code block you generate based on that file. The filename should be on the same line as the language specifier in your code block.
> ....
> When using tools, follow the following guidelines:
> Avoid calling tools unless they are absolutely necessary. For example, if you are asked a simple programming question you do not need web search. As another example, if the user asks you to explain something about code, do not create a new file.

I literally ran into the problem that I can't turn off "lazy" comments, and periodically prompted the model asking for code block be with complete code output. This worked for 2-3 replies, but then the model start to make "lazy" comments again. "Lazy" comments make "Insert at cursor" tool very unusable - I have to select section of changed code above/below of lazy comment and copy/paste them manually. Actually, such behavior (forcing the systemMessage and it's actual content) is not covered in docs!

I tried to delete it by forcing empty `systemMessage` in root config and customize models sys message, but it doesn't work.

Also, such injection in system message could ruin the users system prompt which, for example, structured for forcing the CoT.

ferenci84:
@OlhinAS Good to know that it's not just me who finds this useful. I'm thinking about an update to make the settings more intuitive and flexible.
It's an interesting point:
> I tried to delete it by forcing empty systemMessage in root config and customize models sys message, but it doesn't work.

I tried the same out of desperation when I got annoyed with the default system message that cannot be removed.

The current behaviour (that is apparent only when you look into the code), that if there is a systemMessage in the model config, the systemMessage in the root config is ignored.

What complicates this, that two defaults that are added, one is the instructions about code changes and other is the instructions about tool use, instructions about the code is not added to every providers (for example if I run Claude via AWS, it's not added), and tool instructions is added only for models/providers with tool support and only if there are any tools enabled. I think the complete solution should allow the user to modify those two default instructions. This may also make the developers of this extension a more flexible way to experiment with different instructions in real-life projects without having to recompile everything.

OlhinAS:
@ferenci84 Thank you for such a detailed response.

In context of systemMessage only: 
Don't You think that instead of adding additional switches and complicating configuration (of course it would be more flexible, but more parameters - more chances to go wrong in config), it would be better to generate the config on initial installation of plugin already with changeable pre-filled root systemMessage? Simply bring it from hardcode to the config.

I am pretty sure that users writing systemMessages doesn't know about additive nature of "Always produce a single code block...." hardcoded instructions, so they think only about their systemMessage. In such case, when they rewrite systemMessage or model systemMessage (it's higher prior) they get only what they wrote.

If you want to add - add, if not - just rewrite completely. Or leave as it.

--- 
In context of ToolUse only:

I think it's not a good idea to completely disable ToolInstructions. Prompting tools behaviour is quite tricky since it could lead to the incorrect interpretation project root path and attempts to write/use files/folders/terminal incorrectly. I already faced this problem and in such case Continue get stuck. Not safety. But keep hardcoded prompt is not flexible.

And i did't get what `disableCodeResponseInstructions` is, sorry(

ferenci84:
@OlhinAS Thank you for your inputs.

@sestinj Would you mind joining the conversation? I think any solution would be good that let us compose the whole system message for specific model setups, if we want to, for any reason.

sestinj:
Thanks for @ 'ing me to make sure I saw this

I agree with the principle of giving you full control. I'm going to take a first step of removing entirely the "lazy comments" system message in a release before next week. The reason I've taken my time in doing that is it's used for apply currently, so we needed a new solution there. Finally have one.

After that, I'd gladly revisit the further levels of control, though my strong preference (to really maximize simplicity) would be to just have a single system message that can be set by the user

sestinj:
We've taken step #1 and removed the massive formatting prompt that was causing so many problems. I want to take the time to get more feedback on how this works before completely adding a new field to the config, but in the case that we do, I am heavily leaning toward the following:


```yaml
models:
  - name: Example
     provider: anthropic
     model: claude-3-5-sonnet-latest
     apiKey: <API_KEY>
     promptTemplates:
       toolSystemMessage: "..."
       
```

We're already using this layout for a custom apply prompt for some models

sestinj:
If possible, I'd like to move this conversation to issues for now

ferenci84:
Replaced by https://github.com/continuedev/continue/pull/4507

PR #3815, opened 2025-01-22
Fix error when opening file outside of any workspace

Flamefire:
## Description

`path.relative` doesn't allow `undefined` as its parameter(s)
However this happens when `getWorkspaceDirs` returns an empty array which causes an error:
> TypeError: The "from" argument must be of type string. Received undefined

So extract the first element and check if it is valid/set before further processing.

## Testing instructions

Open a single file in VSCode and observe that the error message is no longer shown


core/core.ts
```diff
@@ -871,8 +871,10 @@ export class Core {
 
     on("didChangeActiveTextEditor", async ({ data: { filepath } }) => {
       const ignoreInstance = ignore().add(defaultIgnoreFile);
-      let rootDirectory = await this.ide.getWorkspaceDirs();
-      const relativeFilePath = path.relative(rootDirectory[0], filepath);
+      const [rootDirectory] = await this.ide.getWorkspaceDirs();
+      if (!rootDirectory)
+        return; // Not in a workspace
+      const relativeFilePath = path.relative(rootDirectory, filepath);
```
RomneyDa: @Flamefire super delayed review, sorry
`ide.getWorkspaceDirs` returns URIs, not paths, so the path.relative logic will fail. Can you update to use the `findUriInDirs` util in `core/util/uri.ts`?

RomneyDa (changes requested):

core/core.ts
Flamefire: I used the same logic employed in other places. Wasn't able to test this yet, but should be fine.

RomneyDa (dismissed):

Flamefire:
I missed an instance of the renamed variable in the debug message. Fixed

RomneyDa:
Some sweeping changes with ignore logic fixed this as a side effect, and now neither `path.relative` or `findUriInDirs` applies. https://github.com/continuedev/continue/pull/4531

@Flamefire thank you for the contribution and apologies for the delays and misfires.

PR #3851, opened 2025-01-26
fix: JSON schema issue with gemini tool calls

monotykamary:
## Description

Fixes tool calling error that happens when using Gemini. Not actually sure if it's because of MCP servers, but at least now it works.

## Checklist

- [x] Fixed and added JSON schema cleaning for processing tools with Gemini

## Screenshots

image

## Testing instructions

Select any Gemini model and enable tools. Prompt it in such a way that it will invoke the tool and see the results.


RomneyDa:
@monotykamary the schema cleaning is great, should fix some bugs we've been seeing!

Could you remove the 100ms awaits or explain the thought process there beyond the comments? 

There's a good chance it could take way longer than 100ms to process a tool call, and I think in the future continue GUI should handle any needed awaits. Did you find this necessary to work? Also for another edge case I think for now it's reasonable to assume that there will never be function calls and responses within the same message, and if there were, same reasoning could apply. Any other thoughts on your approach?

core/llm/llms/Gemini.ts
```diff
@@ -311,50 +347,77 @@ class Gemini extends BaseLLM {
...
+              yield {
+                role: "assistant",
+                content: "",
+                toolCalls: [part.data],
+              };
+              // Add delay to ensure tool call is processed
+              await new Promise(resolve => setTimeout(resolve, 100));
+            } else if (part.type === 'tool') {
+              // Add delay before tool response to ensure tool call was processed
+              await new Promise(resolve => setTimeout(resolve, 100));
```
RomneyDa: @monotykamary this is great! The schema cleaning is a needed thing I failed to add.
Could you remove the 100ms awaits or explain the thought process there beyond the comments? 

There's a good chance it could take much longer than 100ms to process a tool call, and I think in the future continue GUI should handle any needed waiting/ordering. Did you find this necessary to work? Also for another edge case I think for now it's reasonable to assume that there will never be function calls and responses within the same message, and if there were, same reasoning could apply. Any other thoughts on your approach?

RomneyDa (changes requested):

core/llm/llms/Gemini.ts
monotykamary: I remember adding it for quick testing, but it isn't really needed and definitely true that tool calls can be > 100ms.

RomneyDa (approved):

RomneyDa:
Note, have pulled branch and tested, works great.
Will merge today,

sestinj:
Hi @monotykamary, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

monotykamary:

PR #3924, opened 2025-02-01
fix: Autocomplete not working when lookup is selected

lkk214:


sestinj:
@lkk214 thanks for the fix here! Apologies for being slow to review. Do you have a better example of the problem this is solving? A screenshot or something might help

lkk214:
lookup

If the `def` in lookup is accepted, `isLookupShown` is always **false** because LookupListener.itemSelected is not processed, causing the autocomplete to be skipped.

code related to skip rendering completion:
https://github.com/continuedev/continue/blob/9ab4c735fa54ea9e6bae500ddc99dd6ce130f079/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteService.kt#L233-L234

PR #3926, opened 2025-02-01
Docs: fix duplicate @Clipboard context provider section

jota-codes:
## Description

Found that the documentation for the`@Clipboard` context provider is repeated twice in the documentation: [link](https://docs.continue.dev/customize/context-providers#clipboard-1). Removed the second section since I cannot think of a reason why it should be there twice.

## Checklist

## Screenshots

**Before:**
Screenshot 2025-02-01 at 11 48 53

**After:**
Screenshot 2025-02-01 at 11 49 41

## Testing instructions

Open the [current website for the context provider docs](https://docs.continue.dev/customize/context-providers#clipboard-1) and see that there is two `@Clipboard` sections with the exact same content. In the updated version, you will see only the first of the sections is there.


RomneyDa (approved):

RomneyDa:
@jota-codes thanks for the contribution! Looks like this was fixed at some point, will go ahead and close.

Issue #3900, opened 2025-01-30
More context for autocompletion (context aware autosuggest)


ferenci84:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Even though the documentation suggest that context from other files are added to the prompt sent for autocomplete this is not the case. The result is that you get a completion that takes only the current file into account.

### Solution

This is a complete solution in twinny:
https://github.com/twinnydotdev/twinny/blob/0280f98ba1d6101d57b71feee2cb62fcea8b99bd/src/extension/providers/completion.ts#L554

This is how it works:
   - Collects all open files in VS Code (`textDocuments`) with some exclusions, like it skips untitled files and the current file being edited
   - Takes 200 lines of context (100 lines before and after the average active line), for smaller files includes the entire file content
   - Formats this way:
```
// File: /path/to/file
// Content:
[actual file content or relevant chunk]
```

This is an example of a fim template where it's used:
https://github.com/twinnydotdev/twinny/blob/0280f98ba1d6101d57b71feee2cb62fcea8b99bd/src/extension/fim-templates.ts#L67

For Qwen template, twinny's solution is not the perfect solution as it has it's own separator:
https://github.com/twinnydotdev/twinny/blob/0280f98ba1d6101d57b71feee2cb62fcea8b99bd/src/extension/fim-templates.ts#L84

This is specific instructions:
https://github.com/QwenLM/Qwen2.5-Coder?tab=readme-ov-file#4-repository-level-code-completion

I think the context should be a list of files with name and content, and the fim template should process the files and add it to the context.

Possible solution for Continue:

In streamCompletionWithFilters:
https://github.com/continuedev/continue/blob/57b18d73d67bccbb0726b9170b3d1620c8402dcd/core/autocomplete/generation/CompletionStreamer.ts#L15

In addition to prefix and suffix, an additional context parameter should be added with the list of files.

Context collection should be done in the provideInlineCompletionItems:
https://github.com/continuedev/continue/blob/9edd5e3e241862d8189cf17332f7c3a2933125c4/core/autocomplete/CompletionProvider.ts#L132

[...]

Issue #1952, opened 2024-08-07
Unable to obtain context from other files for tab autocomplete feature


vedants03:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11 23H2
- Continue: v0.8.45
- IDE: VSCode 1.92.0
- Model:
   - Chat: Deepseek-6.7b-instruct
   - Tab Autocomplete: Deepseek-6.7b-instruct

- config.json:
  
{
  "models": [
    {
      "title": "deepseek-6.7b",
      "provider": "openai",
      "model": "deepseek-ai/deepseek-coder-6.7b-instruct",
      "apiBase": "BASE_URL",
      "apiKey": "API_KEY",
      "systemMessage": "You are an expert software developer. You give helpful and concise responses.",
      "template": "deepseek"
    }
  ],
  "customCommands": [
    {
      "name": "test",
      "prompt": "{{{ input }}}\n\nWrite a comprehensive set of unit tests for the selected code. It should setup, run tests that check for correctness including important edge cases, and teardown. Ensure that the tests are complete and sophisticated. Give the tests just as chat output, don't edit any file.",
      "description": "Write unit tests for highlighted code"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Tab Autocomplete",
    "provider": "openai",
    "apiBase": "BASE_URL",
    "model": "deepseek-ai/deepseek-coder-6.7b-instruct",
    "apiKey": "API_KEY"
  },
  "tabAutocompleteOptions": {
    "useOtherFiles": true,
    "multilineCompletions": "always",
    "useFileSuffix": true,
    "maxPromptTokens": 4096,
    "useImports": true,
    "useRecentlyEdited": true,
    "recentlyEditedSimilarityThreshold": 0
  },
  "contextProviders": [
    {
      "name": "code",
      "params": {}
    },
    {
      "name": "docs",
      "params": {}
    },
    {
      "name": "diff",
      "params": {}
    },
    {
      "name": "terminal",
      "params": {}
    },
    {
      "name": "problems",
      "params": {}
    },
    {
[...]

PR #3927, opened 2025-02-01
Autocomplete context improvement for codestral and qwen coder

ferenci84:
## Description

Enhance context inclusion for codestral and qwen coder.

Changes:
1. RecentlyVisitedRangesService from disabled to enabled by default, and numSorroundingLines to 20
2. In core/autocomplete/templating/filtering.ts moved diff to the back to avoid fill up the prompt leaving no room for other context.
3. Created the following experimental options to allow contributors and tech-savvy users experiment with enable/disable and change the order of context before someone concludes to a working model:
```
  // true = enabled, false = disabled, number = enabled with priority
  experimental_includeClipboard: boolean | number;
  experimental_includeRecentlyVisitedRanges: boolean | number;
  experimental_includeRecentlyEditedRanges: boolean | number;
  experimental_includeDiff: boolean | number;
```
4. Changed the templates for codestral and Qwen coder to properly include the context in the prompt

Relates to these issues:
https://github.com/continuedev/continue/issues/3900 (opened by me as a feature request, however I realized that the feature is already there just not working)
https://github.com/continuedev/continue/issues/1952
https://github.com/continuedev/continue/issues/3506

(most likely there are more)

## Checklist

## Testing instructions

1. Run the extension and open the Extension Development Host
2. Close the bottom panel if it's open (do not open the "Output" tab before the end of the test)
3. Set up codestral with remote provider and/or qwen coder with ollama as per the instructions of Continue documentation
4. Add this code into the css test file:
```
.class1 {
  background-color: red;
}

.class1 .myclass2 { /* I would visit this section last */
  background-color: blue;
}
```
Add this code to the html test file:
```
<div class="class1">
    <div class=></div>
  </div>
```
5. Run "Developer: Reload Window"
6. Focus on the css test file somewhere around myclass2
7. Focus on the html test file after `class=` and press `"`
8. It should autocomplete `myclass2`

Possible further improvements on this feature:
- Some code snippets are duplicated. Some logic can be built to resolve overlapping snippets.
- recentlyVisitedRanges also contain contents from other windows like terminal or output if they are visible. We should handle them separately so that we can control their priority and whether they should be included or not.
- Limit diff in case there are a lot of changes
- Test other providers and models


ferenci84 (commented):
Additional note: I changed both qwen and codestral template so that they produce a single prompt instead of separate prefix and suffix.
Explanation:
Creating prefix and suffix will result in a request that has "prompt" and "suffix" in the request body. However, after all, the model itself has to get a single prompt (because LLM models only have one "prompt" as input), so somewhere in the process, the provided "prompt" and "suffix" parameters will be converted to a single "prompt", but may not in a way we want: in case of codestral it's [SUFFIX]{suffix}[PREFIX]{prefix}, that makes it impossible to put the context before the [PREFIX] and [SUFFIX] separators. In contrast, providing a single prompt will let us achieve this format: {context}[PREFIX]{prefix}[SUFFIX]{suffix}

tomasz-stefaniak:
Note to self - refer to this when reviewing:

https://github.com/continuedev/continue/issues/3900#issuecomment-2636182674

tomasz-stefaniak (approved):
Thanks for this quality contribution @ferenci84 !

tomasz-stefaniak:
@sestinj do you mind merging this? I don't have the permissions to do that.

PR #3942, opened 2025-02-03
change qwen2.5-coder:1.5b-base to qwen2.5-coder:1.5b

anvarknian:
## Description

Changed qwen2.5-coder:1.5b-base to qwen2.5-coder:1.5b
Tried qwen2.5-coder:1.5b-base, it returns weird results in russian...
official Qwen models are hosted here: https://ollama.com/library/qwen2.5-coder:1.5b
as you can see there is no '-base' models

## Checklist

## Screenshots

## Testing instructions


sestinj:
@anvarknian Thanks for looking into this! The -base here is important (and the model can be found [here](https://ollama.com/library/qwen2.5-coder/tags)), because otherwise you'll end up with the "instruct" model, which is not the best for autocomplete. If you're seeing repeated problems with this model and can share reproducible problems I think the next best step would be to open an issue that details what is going wrong.

PR #3955, opened 2025-02-04
Fix: getPlatform to return "unknown" when environment isWeb

etherandrius:
## Description

When running VS Code web users will see wrong shortcuts if the remote system is a Linux box while the user is using a Mac. 

See example: The hint on the right is showing CTRL while the text box on the left is showing the CMD key.
image

 
When running VS Code remotely it's not possible to determine users platform and `node.os.platform` should not be trusted.

The change explicitly marks platform as `unknown`  if  `vscode.env.uiKind === Web`. This is relevant in remote workspaces such as Coder/Github codespaces/Project IDX etc. 

If the platform is `unknown` show both CMD and CTRL keys.

VS Code extension API does not provide a way to identify the actual users platform.
- See https://github.com/microsoft/vscode/issues/106525 (the api is not provided)
- And https://github.com/microsoft/vscode/issues/10471 (getContext API which would solve this has not been implemented)

I appreciate this solution is not-ideal. Not sure what the proper solution the code probably needs to leverage the `when` [api](https://code.visualstudio.com/api/references/when-clause-contexts). 

## Checklist

## Screenshots

Hint SVG formatting is still sane after the changes
image
> ^ Platform detected as unknown 

image
> ^ Platform detected as linux 

## Testing instructions

Install continue extension inside github codespaces you will always see linux based shortcuts even when running on a Mac.

### DIFF
--- a/extensions/vscode/src/activation/InlineTipManager.ts
+++ b/extensions/vscode/src/activation/InlineTipManager.ts
@@ -16,9 +16,9 @@ const SVG_CONFIG = {
   leftMargin: 40,
   debounceDelay: 500,
   chatLabel: "Chat",
-  chatShortcut: `${getMetaKeyLabel()}L`,
+  chatShortcut: `${getMetaKeyLabel()}+L`,
   editLabel: "Edit",
-  editShortcut: `${getMetaKeyLabel()}I`,
+  editShortcut: `${getMetaKeyLabel()}+I`,
 
   get fontSize() {
     return Math.ceil(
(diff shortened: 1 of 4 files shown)


RomneyDa (approved):

etherandrius:
After the merge 
image

@RomneyDa let me know if you prefer something other than `Ctrl/⌘`

sestinj:
@etherandrius we should be able to request the platform from the webview, which knows the correct one. Can we do this and then remove the "+" sign?

etherandrius:
@sestinj @RomneyDa friendly bump

RomneyDa (changes requested):
@etherandrius thanks for the bump, running this past product

requested change:  I think gap should still depend on font size? Also related to merge conflict

RomneyDa:
Closing because stale, @etherandrius if you make updates would like to merge, also making note to look into reimplementing

PR #3980, opened 2025-02-05
Enhance processing diff-patterns in single-line completions

ferenci84:
## Description

In the completionProvider when checking single-line completions and the diff is too complicated, it assumes it overwrites existing parts, and it skips the completion.

The change will just allow the completion be inserted, assuming the FIM endpoint works correctly - providing "middle" part

## Explanation

With FIM endpoints it's the desired behavior that the model provides only the "middle" part without including any of the existing parts. In those cases calculating a diff is like comparing apples to bananas, and can produce complicated diff patterns (some equality are usually found, like parentheses, double quotes, or even repeating patterns found at the same line). Showing the whole completion in such a case is more desirable even if it causes some false positive - at least developers will notice it and make corrections in other parts of the software, like required adjustments in the specific model's fim template or additional diff patterns.

## Checklist

No docs or tests should be updated

## Testing instructions

Use the other pull request:
https://github.com/continuedev/continue/pull/3927

It has a corrected version of FIM template for QWEN with this change already merged.

Use this file (add it to the workspace):
https://ferenci84.s3.amazonaws.com/create_token_groups.py

Find this line:
print(f"Total tokens: <place cursor here>")

Press the hotkey for initiating autocomplete (set it up if it doesn't exist).


sestinj:
@ferenci84 thanks for finding this and making the fix! Before merging this, I want to make sure we get the behavior pinned down in a test. Can you add something to our e2e autocomplete tests for VS Code that will capture the problem you solved? `Autocomplete.test.ts` would be the starting point

ferenci84 (commented):
@sestinj I checked what test to add for this, and found that it's more suitable to extract the functionality out to a pure function and create unit test. Also I made a test helper to make the test cases easy to understand. Now each cases in the original function has at least one test case. Also the test helper has it's own tests.

ferenci84:
@sestinj Can you please check?

sestinj:
Yes! Will review tomorrow. It looks really great, I just want to pull down and test some

Issue #3951, opened 2025-02-04
[Jetbrains] In the light theme,button color of Edit feature is too light


AfterStories:
### Validations

- [ ] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Win 10
IntelliJ
Main branch code

Use `./gradlew.bat buildPlugin` command to build the same one install zip package

In Intellij with light theme, the accept and reject buttons look like they are disabled, but they can be clicked

![Image](https://github.com/user-attachments/assets/e3e40092-6392-4878-a401-10704f2143e1)

But, In Intellij with dark theme, the button color looks normal

![Image](https://github.com/user-attachments/assets/3e9edc5d-8c8c-4f8d-bcd6-b2251bf5a730)

### Solution

Need to enhance the color in light theme, I mistakenly thought the button was in disabled state,
And, Which file can I change to the setting color of these buttons, maybe I will want to customize the color to fit my design

AfterStories:
looks like have fixed

PR #4032, opened 2025-02-08
refactor: Replace EmptyBorder with DPI-aware borders

lkk214:
## Description

- `EmptyBorder` is not DPI-aware 
- Removed redundant calls to conversion methods
- Fixed unclear button text in light themes (using default foreground color)

## Screenshots
cut_1

## Testing instructions

This is simple fix, remove custom foreground color and use default foreground color.

cut_2


lkk214:
fixes https://github.com/continuedev/continue/issues/3951

sestinj (dismissed):

PR #4103, opened 2025-02-12
Model Role Updates

RomneyDa:
## Description
- Pass model roles around on config/LLM types
- default to chat if no roles on model
- only show chat models in the model selector
- some naming changes and cleanup for model roles

@sestinj requires `config-yaml` push to work/pass tests


gui/vite.config.ts
```diff
@@ -16,6 +16,14 @@ export default defineConfig({
       },
     },
   },
+  server: {
+    cors: {
+      origin: "*",
```
RomneyDa: duplicate of https://github.com/continuedev/continue/pull/4097

sestinj (dismissed):

PR #4107, opened 2025-02-12
Update README.md

danix800:
Fix gif relative path to project root.

## Description

These gifs cannot be displayed with path relative to project root.

### DIFF
--- a/extensions/vscode/README.md
+++ b/extensions/vscode/README.md
@@ -30,7 +30,7 @@
 
 [Chat](https://continue.dev/docs/chat/how-to-use-it) makes it easy to ask for help from an LLM without needing to leave the IDE
 
-![chat](docs/static/img/chat.gif)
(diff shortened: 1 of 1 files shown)


RomneyDa:
@danix800 If you run the docs server do they show up?
Typically you can't click through to the `gif` if the path isn't relative, but when it builds/actually runs the non-relative paths work.

danix800:
> @danix800 If you run the docs server do they show up? Typically you can't click through to the `gif` if the path isn't relative, but when it builds/actually runs the non-relative paths work.

I cloned this repo and tried to read this README.md in preview mode and these images cannot
be properly rendered, so I thought it might be broken.

This README.md is used in vscode extension gallery and vsce will convert these relative links to
absolute ones like `https://github.com/continuedev/continue/raw/HEAD/docs/static/img/chat.gif`.

I'll close this PR.

PR #4129, opened 2025-02-13
Jetbrains Colors Fixes

RomneyDa:
## Description
- Jetbrains colors now update immediately when the user changes the theme
- Jetbrains color choices are better. Specific major improvements include:
  - uses corresponding UI colors, not just similar editor colors
  - now distinguishes between editor background and sidebar background, same as VS code, which looks super nice
  - other specific fixes include badge color, find match highlight color
- removed dud `vsc-foreground-muted` tailwind class, unused tailwind classes/vsc colors, and moved a couple less-used ones to non-vsc-specific tailwind names

Tested on built-in Intellij/Darcula + several variations of the most popular themes Material and One Dark


RomneyDa:
image
image
image

Issue #4130, opened 2025-02-13
Autocomplete - Git diff is issued at each keystroke - no cache is implemented


ferenci84:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: MacOS
- Continue version: 0.9.264, 0.9.261
- IDE version: 1.97.1
- config.json:

  "tabAutocompleteModel": {
    "title": "Qwen2.5-Coder 1.5B",
    "provider": "ollama",
    "model": "qwen2.5-coder:1.5b-base",
    "completionOptions": {
      "maxTokens": 100
    }
  }
```

### Description

Git diff command is issued at each keystroke. No cache is applied to this, potentially long, operation.

My IDE and my whole system became unresponsive, basically all actions that require any file operation was delayed. Under the hood, git also used the 'fg' command.
The problem disappeared if I did any of this:
- Reloaded the VS code window
- Restarted Extensions

When I disabled the Continue extension, the problem didn't came back. When I reenabled, it came back again immediately, even after a fresh system restart.

Interesting detail is that if I issue 'git diff' or 'git diff --cached', on the same repo, it responds quickly. There may be a locking mechanism or race condition that causes problem when two instances of this git command is issued at the same time.

Few screenshots:

<img width="1359" alt="Image" src="https://github.com/user-attachments/assets/0ec049bd-1f39-4a7f-aaf6-c5da9588bb71" />
<img width="1076" alt="Image" src="https://github.com/user-attachments/assets/9d9d91aa-9df8-4496-8bdb-f4f1c5021c60" />
<img width="1057" alt="Image" src="https://github.com/user-attachments/assets/3efc970c-50f5-49ca-b548-72de9cb41e4f" />

I believe it's caused by this code:

```typescirpt
// core/autocomplete/snippets/getAllSnippets.ts
const getDiffSnippets = async (
  ide: IDE,
): Promise<AutocompleteDiffSnippet[]> => {
  let diff: string[] = [];
  try {
    diff = await ide.getDiff(true);
  } catch (e) {
    console.error("Error getting diff for autocomplete", e);
  }

[...]

PR #4161, opened 2025-02-14
Implement caching mechanism for diff snippets based on file save timestamp

ferenci84:
## Description

Caching mechanism for get diff under VS code.

Bug:
https://github.com/continuedev/continue/issues/4130

## Testing instructions

Watch git log while typing. After modification, `git diff` command should be issued only once after saving any file.

## Note

Equivalent implementation for IntelliJ need to be implemented. For now, there is a fallback to a 10s cache invalidation

## Additional changes fixing underlying issues explained in the above issue

- Added _getRepositories that will find repositories directly from the vscode.git extension. This didn't really contributed to the problem, but that was a fix I tried, and I believe it's better and less resource-intensive (it doesn't even require async) than the original code.
- In crawlTypes I removed Promise.all and implemented sequential processing. This consumes my CPU at the same level and takes the same amount of time to finish in my test project (between 1 and 2 minutes), but doesn't block my resources.
- Increased cache size for executeGotoProvider from 50 to 500. After this change, subsequent runs didn't take minutes to finish, only the first one.
- After I noticed that the ideSnippets is not used at all in the getSnippets from core/autocomplete/templating/filtering.ts, I disabled in a way that it can be enabled later.


ferenci84:
Further changes added to solve the core problem, see my latest comment in https://github.com/continuedev/continue/issues/4130

Changes:
- Added _getRepositories that will find repositories directly from the vscode.git extension. This didn't really contributed to the problem, but that was a fix I tried, and I believe it's better and less resource-intensive (it doesn't even require async) than the original code.
- In crawlTypes I removed Promise.all and implemented sequential processing. This consumes my CPU at the same level and takes the same amount of time to finish in my test project (between 1 and 2 minutes), but doesn't block my resources.
- Increased cache size for executeGotoProvider from 50 to 500. After this change, subsequent runs didn't take minutes to finish, only the first one.
- After I noticed that the ideSnippets is not used at all in the getSnippets from core/autocomplete/templating/filtering.ts, I disabled in a way that it can be enabled later.

sestinj:
@ferenci84 this is great work, thank you for digging so deeply into the problem!

PR #4167, opened 2025-02-14
feat-aws-bedrock-agents-as-provider

chezsmithy:
## Description

This feature is someone experimental, and is one approach to incorporating AWS Bedrock agents as a dedicated provider. The solution allows defining one to many AWS bedrock agents as a dedicated provider and the user can easily toggle between them to perform dedicated tasks.

New agents can easily be added to the configuration as a provider quickly and easily.

For example:
```

    {
      "title": "Gherkin Agent",
      "provider": "bedrockagent",
      "model": "anthropic.claude-3-5-sonnet-20241022-v2:0",
      "region": "us-west-2",
      "profile": "bedrock",
      "agentId": "<AgentID>",
      "agentAliasId": "<AgentAliasId>"
    },
```

Limitations and questions:

* I've not found an easy way to share the taskid with the provider to prime the sessionId for AWS Bedrock Agent Session Memory. I'd love some good ideas how to make this work.
* Is this useful, vs a MCP tools approach which activates the agent on the fly? There may be value in having direct conversations with a specific agent which might also include supervisor tooling. For example, bringing a more educational but less code focused tool into the developer experience instead of forcing the developer to another UI to use those experiences.
* The model is required in the configuration and must be matched to the model used by AWS bedrock. This is a limitation because the prompt formatting and interpretation on the client side is highly dependent on the model and that's not easily exposed from AWS from what I can see.

## Checklist

## Screenshots

## Testing instructions

This new provider shares AWS profile configuration with other AWS models like bedrock and bedrock-import.

Once you have configured a new agent as a provider you can choose the agent from the provider drop down and have a dedicated discussion with the agent in real time.

### DIFF
--- a/core/llm/llms/BedrockAgent.ts
+++ b/core/llm/llms/BedrockAgent.ts
@@ -1,0 +1,195 @@
+import {
+  BedrockAgentRuntimeClient,
+  InvokeAgentCommand,
+  type InvokeAgentCommandInput,
+} from "@aws-sdk/client-bedrock-agent-runtime";
+import { fromIni } from "@aws-sdk/credential-providers";
+
+import {
+  ChatMessage,
+  CompletionOptions as BaseCompletionOptions,
+  LLMOptions,
+  MessageContent,
+} from "../../index.js";
+import { renderChatMessage } from "../../util/messageContent.js";
+import { BaseLLM } from "../index.js";
+
+interface CompletionOptions extends BaseCompletionOptions {
+  agentId: string;
+  agentAliasId: string;
+  sessionId?: string;
+  enableTrace?: boolean;
+}
+
+class BedrockAgent extends BaseLLM {
+  static providerName = "bedrockagent";
+  static defaultOptions: Partial<LLMOptions> = {
+    region: "us-east-1",
+    contextLength: 4096,
+    profile: "bedrock",
+    completionOptions: {
+      maxTokens: 2096,
+      model: "",
+    }
+  };
+
+  constructor(options: LLMOptions) {
+    super(options);
+    if (!options.apiBase) {
+      this.apiBase = `https://bedrock-agent-runtime.${options.region}.amazonaws.com`;
+    }
+    if (options.profile) {
+      this.profile = options.profile;
+    } else {
+      this.profile = "bedrock";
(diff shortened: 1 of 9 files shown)


core/llm/llms/BedrockAgent.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (!agentId || !agentAliasId) {
+      throw new Error(
+        "agentId and agentAliasId are required for Bedrock Agent. Configure them in the config file or provide them in the completion options.",
+      );
+    }
+
+    return {
+      agentId,
+      agentAliasId,
+      sessionId: options.sessionId || Date.now().toString(),
```
chezsmithy, line 149: For reviewers, this is where I would like to pull in the context (id?) of the current task to align it with the AWS bedrock session. This way if you switch tasks your Bedrock Agent Context will follow.

Patrick-Erichsen:
Hey @chezsmithy , apologies for the slow review here, been busy the past few weeks getting ready for our 1.0 launch. Will make an effort to get around to this by end of week 👍

sestinj:
@chezsmithy just merged main into this. I like everything except that we're trying to avoid the explosion of root-level properties that has happened within the LLM class from happening in config.yaml models. The properties here should go into config.yaml's `models[number].env` field, which takes arbitrary key-value pairs

chezsmithy:
Thanks @sestinj 

Curious if you might have an idea how I could obtain or pass the session ID from the chat window to the LLM layer in the most efficient way, so I can pass that to bedrock to hold the interactions together on the server side?

sestinj:
hmm that's definitely a bit awkward. other than passing it all the way through on every call to streamChat nothing comes to mind. Maybe having some kind of global variables or singleton class. But all of these options are kind of gross :/

sestinj:
@chezsmithy What are you thinking on this one? Does it feel like there is another option? If not, is this change something inevitable that you're confident we're going to need to support and so we should just bite the bullet?

chezsmithy:
Closing this. Given the acceleration in the MCP space I believe access to Bedrock agents would be better implemented using a MCP server approach. If someone feels differently in the future we can revisit the solution.

PR #4170, opened 2025-02-15
Docs Fixes and Embeddings Provider Test Run

RomneyDa:
## Description
- Add an embeddings test run to the beginning of docs indexing, and rearrange `indexAndAdd` a bit to account for it
- Fix docs status initialized to zero
- Sort docs by indexing status in More page

PR #4175, opened 2025-02-15
fix: Prevent duplicate context provider addition

lkk214:
## Description

- Add check to avoid adding the same context provider multiple times and conditionally refresh configuration only after a new provider is added
- Add docs/getDetails endpoint in JetBrains to support viewing doc indexes
- Fixed the issue that the progress bar of indexed docs is not displayed after running continue

### DIFF
--- a/core/indexing/docs/DocsService.ts
+++ b/core/indexing/docs/DocsService.ts
@@ -157,7 +157,8 @@ export default class DocsService {
 
       const currentStatus = this.statuses.get(doc.startUrl);
       if (currentStatus) {
-        return currentStatus;
+        this.handleStatusUpdate(currentStatus);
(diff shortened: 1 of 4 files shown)


core/indexing/docs/DocsService.ts
```diff
@@ -157,7 +157,8 @@ export default class DocsService {
 
       const currentStatus = this.statuses.get(doc.startUrl);
       if (currentStatus) {
-        return currentStatus;
+        this.handleStatusUpdate(currentStatus);
```
lkk214, line 160: Duplicate of https://github.com/continuedev/continue/pull/4170

sestinj (approved):

PR #4184, opened 2025-02-15
fix: video-recorder-junit5 should be configured as testImplementation

joffeoja:
## Description

Test class ends up on main class path, probably not intended. If com.automation-remarks:video-recorder-junit5 is only needed for test it should be configured as testImplementation instead of implementation.

Fixes https://github.com/continuedev/continue/issues/4372

## Checklist

## Screenshots

## Testing instructions


Patrick-Erichsen:
@joffeoja apologies for the delay here! I just tried re-running the checks but `jetbrains-tests` appears to still be failing, not sure why it would be related to this change but it appears to be the case.

Issue #3753, opened 2025-01-17
Intermittent Copy/Paste Not Working and Extension Host Crashes in VSCode [specifically for `.sh` files]


topolittle:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- VSCode: 1.96.3
- Continue: 0.8.66
- OS: Fedora 41 (x11)
- Ollama: 0.5.4
- Model: qwen2.5-coder:7b-instruct-q5_K_S
- NVIDIA Drivers: 565.77
- CUDA: 12.6.85
- Other Enabled Extensions: Bash Debug, Bash IDE, ShellCheck

{
  "models": [
    {
      "title": "Qwen2.5-Coder 7b Chat",
      "provider": "ollama",
      "model": "qwen2.5-coder:7b-instruct-q5_K_S"
    }
  ],
  "tabAutocompleteModel": {
    "title": "Qwen2.5-Coder 7b Autocomplete",
    "provider": "ollama",
    "model": "qwen2.5-coder:7b-instruct-q5_K_S"
  },
  "useLegacyCompletionsEndpoint": false,
  "completionOptions": {
    "keepAlive": 900,
    "maxTokens": 2048
  },
    "contextProviders": [
    {
      "name": "code",
      "params": {}
    },
    {
      "name": "docs",
      "params": {}
    },
    {
      "name": "diff",
      "params": {}
    },
    {
      "name": "terminal",
      "params": {}
    },
    {
      "name": "problems",
      "params": {}
    },
    {
      "name": "folder",
      "params": {}
    },
    {
      "name": "codebase",
      "params": {}
    }
  ],
  "slashCommands": [
    {
      "name": "share",
      "description": "Export the current chat session to markdown"
    },
    {
      "name": "cmd",
      "description": "Generate a shell command"
    },
    {
      "name": "commit",
      "description": "Generate a git commit message"
    }
  ]
}
```

### Description

After using the Continue VSCode extension for a few minutes, the Copy/Paste functionality stops working. Additionally, when this happens, the ShellCheck extension begins displaying false syntax errors.

[...]

PR #4250, opened 2025-02-20
Reinstate VSCode copy buffer

RomneyDa:
## Description
Was removed on a hunch that it was causing extension hanging.
Appears to not have solved hanging issues, adding back in to return functionality to the clipboard context provider and autocomplete copy buffer access.

### DIFF
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -1056,7 +1056,7 @@ export function registerAllCommands(
   core: Core,
   editDecorationManager: EditDecorationManager,
 ) {
-  // registerCopyBufferSpy(context, core);
+  registerCopyBufferSpy(context, core);
 
   for (const [command, callback] of Object.entries(
     getCommandsMap(


sestinj (approved):

plohoj:
For me, in versions of the extension where `registerCopyBufferSpy()` is commented out/not called, there are no copy/paste problems. In versions where `registerCopyBufferSpy()` is uncommented/called there are copy paste problems.

If you perform the following actions in 2-3 seconds (quickly):
1. Select the text,
2. Copy the selected text to the clipboard,
3. Move the cursor to another line;

then the entire line where the cursor was placed will be copied to the clipboard.

This problem is interfering with development so much that I am forced to install previous versions of the extension.

Patrick-Erichsen:
@RomneyDa - remind me what the copy buffer is used for, just autocomplete? Given the comment from @plohoj above, and ongoing +1s in https://github.com/continuedev/continue/issues/3753, I lean towards commenting this back out.

PR #4262, opened 2025-02-20
http -> remote context provider and slash command

RomneyDa:
## Description
- Change `http` context provider to `remote` (leave support for `http`)
- Change `/http` slash command to `/remote` (leave support for `/http`)
- Update docs


RomneyDa:
closing for now, sounds like we're not sure "remote" is the right name here and not urgent

sestinj:
i'm flying through PRs and this has accumulated conflicts and I'm not even sure this is the right name, so just going to close

Issue #4124, opened 2025-02-13
RecentlyEditedTracker.insertRange tries to read invalid uri vscode-scm:git/...


ferenci84:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: MacOS
- Continue version: latest
```

### Description

Unable to resolve filesystem provider with relative file path 'vscode-scm:git/scm0/input?rootUri=file%3A%2F%2F%2FUsers%2Fferenci84%2FDev%2FXXXXXXX%2Fxxxxxxx%2Fxxxxxx'
	at Zi.e (/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:112:9922)
	at Object.readFile (/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:112:7650)
	at async _RecentlyEditedTracker.insertRange (vscode-file://vscode-app/Users/ferenci84/.vscode/extensions/continue.continue-0.9.261/out/extension.js:607476:27)

It's same as the bug that I made this PR:
https://github.com/continuedev/continue/pull/3930

I'll provide a new PR that includes correction to this problem as well.

### To reproduce

Happens randomly

### Possibly related issues

https://github.com/continuedev/continue/issues/2110
https://github.com/continuedev/continue/issues/517

PR #3930, opened 2025-02-01
Fix: RecentlyEditedTracker tries to read invalid uri 

ferenci84:
## Description

RecentlyEditedTracked throwed unhandled promise rejection errors due to trying to read file with uri like "output:extension-output-Continue.continue-%231-Continue - LLM Prompt/Completion"

## Testing instructions

In Host:
1. Set up autocompletion
2. Open the Output bottom panel
3. Write something to trigger autocompletion
4. Rightclick on the Output panel and choose clear output
5. Write again something to trigger autocompletion
In main:
6. Check debug console

Before fix, there were unhandled promise rejection errors.

PR #4279, opened 2025-02-21
Modify condition to ignore non-file URIs in RecentlyEditedTracker

ferenci84:
Enhancement of change in https://github.com/continuedev/continue/pull/3930

Issue:
https://github.com/continuedev/continue/issues/4124


RomneyDa (changes requested):
@ferenci84 this is on the right track but we do want to support fairly arbitrary schemes like `wsl`, `vscode-remote`, etc.

Did you see another unsupported scheme come through the logs?

sestinj (approved):

ferenci84:
> @ferenci84 this is on the right track but we do want to support fairly arbitrary schemes like `wsl`, `vscode-remote`, etc.
> 
> Did you see another unsupported scheme come through the logs?

@RomneyDa  What I see is that if you edit anything in the workspace (e.g. clear output, change the git commit message), it triggers this, and it tries to read the "file content" but it's not a file in most cases. I think supporting wsl and vscode-remote is an additional feature, while this PR is just a bugfix.

My recommendation is this: merge this PR (since it's a solution to an bug already present), but keep the PR open, if possible, or open a new issue, for the additional feature to support arbitrary schemes and avoid direct file reading to get the content if possible.

sestinj (approved):

PR #4291, opened 2025-02-22
use local node_modules as cache folder for node-sqlite3 when building binary pkg

welpher:
## Description

use local node_modules as cache folder for node-sqlite3 when building binary pkg always download node-sqlite3 from remote 

## Checklist

## Screenshots

## Testing instructions


sestinj:
@welpher there are a number of other dependencies that we are already manually downloading each time, so I would rather solve them all at once rather than make an exception in this case. I'm going to close this PR, but I do have plans to improve the build process this week and will consider this as an improvement

Issue #3229, opened 2024-12-06
Support MCP tools with Bedrock Claude


michaellatman:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

It appears that MCP does not work when using claw via bedrock. 

### Solution

As far as I'm aware, Bedrock supports tool use. So it would be awesome if continue allowed for tools even when running through Bedrock.

tomasz-stefaniak:
I believe @sestinj is currently working on tool use

prd-tuong-nguyen:
any update on this?

tomasz-stefaniak:
Tool use has been implemented, you can see the list of providers with tool use support here:

https://github.com/continuedev/continue/blob/main/core/llm/toolSupport.ts#L1

We'd be open to merging a PR if any community members feel like extending support to other providers.

chezsmithy:
@prd-tuong-nguyen, @michaellatman I'd love some eyes on my PR.

PR #4346, opened 2025-02-25
Feat: AWS Bedrock Tools Use

chezsmithy:
## Description

Enable tools and MCP use with AWS Bedrock and Claude Sonnet Models

Resolves: https://github.com/continuedev/continue/issues/3229

## Checklist

## Screenshots

## Testing instructions

Ask to make changes to a file.
Observe that the file is updated using file modification tools.

Ask to use a MCP tool.
Note the tool is executed and the results are used to generate new output.


chezsmithy:
@sestinj I'm getting some odd flapping tests. Perhaps you could lend a hand to see if we can stabalize?

sestinj:
Help would be much appreciated. I actually think that a useful partial solution here would be to change the way that we disable the SSH tests.

Right now we are doing so via env var, so some of the (flaky) setup steps still need to happen. But if we did it at the workflow level that wouldn't happen

chezsmithy:
@sestinj I feel like this PR is ready for review. Would love for some other bedrock users to have a look and also test if you have ideas. I'm asking my own team to review as well. Thank you.

sestinj:
@chezsmithy this looks great! One last thing that would probably be worthwhile  if we want to ensure stability for the Bedrock integration would be tests. This should actually be pretty simple to do with our setup in `llm.test.ts`. If you write tests and confirm it works locally for you then I'll go ahead and merge and add the necessary credentials so that it can work in CI from here on out

chezsmithy:
> @chezsmithy this looks great! One last thing that would probably be worthwhile if we want to ensure stability for the Bedrock integration would be tests. This should actually be pretty simple to do with our setup in `llm.test.ts`. If you write tests and confirm it works locally for you then I'll go ahead and merge and add the necessary credentials so that it can work in CI from here on out

@sestinj i've added a unit test for the Bedrock.ts llm provider. I think this will help us as we need to make changes be super clear if we aren't breaking existing structuring for the Converse API. I'd love to get llm.test.ts working, but we would need for there to be a working AWS profile or set of secrets loaded into the pipeline so we could run an E2E type of provider test against bedrock in the CI pipeline. Looks like things are commented out right now as Azure isn't working for example right now due to the ENV not being available. 

If you are agreeable and this can get into preview, I can have my team hammer away on it as well to make sure we haven't introduced any unintended breakage and feed that back.

We are currently using this provider for multiple models every day.

sestinj:
@chezsmithy That sounds like a plan to me! Merging now, thanks for the contribution!

chezsmithy:
@sestinj any idea when this change and the add on change for 3.7 support might get to the release version?

sestinj:
@chezsmithy We make main releases on Fridays, at which point we take the latest prerelease that has been around for at least a week

PR #4401, opened 2025-02-27
Fix renderInlineAs

nanoandrew4:
## Description

Fixed `renderInlineAs` which was not working on the custom context providers or the HTTP context provider.

## Checklist

N/A

## Screenshots

N/A

## Testing instructions

In my case, since I use a custom context provider that makes an API call with the `fullInput` variable, I can see that before it would send the prompt with the context provider name despite me setting `renderInlineAs` to an empty string, and with the change, both my custom context provider and the HTTP context provider now send the value as configured, instead of the null value that was sent before.

### DIFF
--- a/binary/package-lock.json
+++ b/binary/package-lock.json
@@ -48,7 +48,7 @@
         "@aws-sdk/client-sagemaker-runtime": "^3.621.0",
         "@aws-sdk/credential-providers": "^3.620.1",
         "@continuedev/config-types": "^1.0.13",
-        "@continuedev/config-yaml": "^1.0.50",
+        "@continuedev/config-yaml": "^1.0.52",
         "@continuedev/fetch": "^1.0.4",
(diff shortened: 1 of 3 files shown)


RomneyDa (changes requested):
@nanoandrew4 looks great but could you pull main / make sure package locks aren't affected?

sestinj (approved):

sestinj:
nice fix, thank you!

sestinj:
Hi @nanoandrew4, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

PR #4418, opened 2025-02-28
Optional context block name

RomneyDa:
## Description
Add optional `name` to context block


RomneyDa:
replaced by https://github.com/continuedev/continue/pull/4922

Issue #4339, opened 2025-02-24
Add support for Claude 3.7 Thinking & Budget Tokens


Guandor:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Claude 3.7 has added thinking (reasoning) mode. Unlike previous thinking models, they also allow to set a budget for how many tokens the model is allowed to think for. This requires additional thinking parameters in the API call that's not supported in Continue config yet. 

More info at Claude's documentation:
https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking

### Solution

Add "thinking" options to the Continue config. 
```
       "thinking": {
          "type": "enabled",
          "budget_tokens": 16000
        }
```

maqrrr:
I just had claude-code write me a patch to support 3.7 so I could try out the new model today.

This is what it came up with.  Looks OK to me, seems to work with tools support.

I didn't run the modified tests.

Maybe someone who actually works on this project can add the "thinking" feature to this and send a PR.

```diff
Date: Mon, 24 Feb 2025 18:01:12 -0500
Subject: [PATCH] claude-3.7 sonnet support

---
 CLAUDE.md                                     | 21 +++++++++++++++++++
 core/config/default.ts                        | 11 ++++++++++
 core/llm/llm.test.ts                          |  2 +-
 core/llm/llms/Asksage.ts                      |  1 +
 core/llm/llms/FreeTrial.ts                    |  1 +
 core/llm/toolSupport.ts                       |  4 ++--
 extensions/vscode/config_schema.json          |  4 +++-
 .../vscode/e2e/test-continue-yaml/config.yaml |  4 ++--
 .../vscode/e2e/test-continue/config.json      |  4 ++--
 extensions/vscode/package-lock.json           |  4 ++--
 gui/src/pages/AddNewModel/configs/models.ts   | 16 +++++++++++++-
 .../pages/AddNewModel/configs/providers.ts    |  1 +
 .../packages/test-org/claude37sonnet.yaml     |  8 +++++++
 packages/llm-info/src/providers/anthropic.ts  | 12 ++++++++++-
 packages/llm-info/src/providers/bedrock.ts    |  6 ++++++
[...]

PR #3840, opened 2025-01-24
3832: DeepSeek-R1 reasoning display can be toggled

germankrause:
## Related Issue

https://github.com/continuedev/continue/issues/3832

## Description

For DeepSeek-R1: Fixed missing first paragraph of reasoning. Added a spoiler to toggle visibility of reasoning tokens.

## Screenshots

Screenshot 2025-01-24 at 22 35 18
Screenshot 2025-01-24 at 22 35 40

## Testing instructions

tested on model:

```json
{
  "title": "deepseek-r1:14b (local)",
  "model": "deepseek-r1:14b",
  "provider": "ollama",
  "apiBase": "http://localhost:11434"
}
```

PR #4426, opened 2025-03-01
Handle extended thinking with Anthropic provider, process streaming thinking blocks and show them in the output tab

ferenci84:
## Description

This modification allows users to see the human-readable parts of the thinking blocks in the LLM output.

This simple change is a first step on the way of full implementation of handling thinking blocks.

## Checklist

## Screenshots

Screenshot 2025-03-01 at 14 18 00

## Testing instructions

Thinking can be enabled in config this way:
```
   {
      "title": "Anthropic - Claude 3.7 Sonnet (thinking)",
      "model": "claude-3-7-sonnet-latest",
      "provider": "anthropic",
      "apiKey": "xxx",
      "completionOptions": {
        "temperature": 1,
        "topP": 0.999,
        "maxTokens": 8096
      },
      "requestOptions": {
        "extraBodyProperties": {
          "thinking": {
            "type": "enabled",
            "budget_tokens": 2048
          }
        }
      },
      "cacheBehavior": {
        "cacheConversation": true,
        "cacheSystemMessage": true
      }
    },
 ```

 Watch the Continue - LLM Prompt/Completion output tab.
 
 ## Next steps
 
1. The thinking mode doesn't currently work with tool use. The software should add the `thinking` and `redacted_thinking` blocks to the conversation history when sending subsequent requests. Tool use requires this, with no tool use, any thinking blocks will be ignored.

2. human-readable thinking blocks could be added to the UI in form of collapsible boxes. This is an example from TypingMind:
Screenshot 2025-03-01 at 14 29 38

## Edit

Handling thinking blocks and redacted_thinking blocks have been added. Due to these changes, these thinking blocks have also became visible. This may require some visual update, but it's now accessible and functional, and we are closer to a final solution where users can read the thinking blocks if they want.

## Additional testing instructions

The changes are working if you can call tools with no problem.

Redacted thinking can be tested by adding this string in your first prompt:
`ANTHROPIC_MAGIC_STRING_TRIGGER_REDACTED_THINKING_46C9A13E193C177646C7398A98432ECCCE4C1253D5E2D82641AC0E52CC2876CB`

## Edit 2

I have added rendering of thinking blocks as shown in the screenshots below:

Screenshot 2025-03-05 at 11 42 56

Screenshot 2025-03-05 at 11 43 05

Screenshot 2025-03-05 at 11 42 33


chezsmithy:
Would it be possible to add the same support to Bedrock.ts now that I have enabled tools support, and have a seprate PR to enabled claude 3.7 on bedrock as well?

Symfomany:
Deep Research would be a killer-feature in this coding assistant Continue, it's possible to have Perplexity API Key too ?

ferenci84:
@RomneyDa You are the assignee of issue https://github.com/continuedev/continue/issues/4339 and this PR would resolve it, can you please review?

SivasankarMathiyazhagan:
@ferenci84 - is there any calculations behind to set the budget tokens as **"budget_tokens": 2048** ? Could you please suggest how we can determine the appropriate budget token limit for our use case?

ferenci84:
> @ferenci84 - is there any calculations behind to set the budget tokens as **"budget_tokens": 2048** ? Could you please suggest how we can determine the appropriate budget token limit for our use case?

You need to make sure that contextLength is larger than maxTokens and maxTokens is larger than budgetTokens, and budgetTokens is at least 1024. This uses 3200 as budget tokens: https://bigcode-bench.github.io/

sestinj:
@ferenci84 this looks really good, type updates and all. Two ways I think it could be improved:
- In this PR (https://github.com/continuedev/continue/pull/3840) we have a separate UI for displaying thinking tokens. I think we should keep the same UI for both for consistency
- as shown in this screenshot, the toolbar for [trash, copy, thumbs up, thumbs down] shows up between the input and thinking toggle, which causes some awkward blank space

Screenshot 2025-03-09 at 8 38 43 PM

Both of these thoughts lead to a bigger question of how we should combine the previous strategy for handling thinking tokens in Ollama (deepseek literally just outputs <thinking> tags in its output) with this one. It seems to me that parsing the <thinking> tags in `streamChat` in `llm/llms/index.ts` would be the best option

ferenci84:
Yes, the big difference between claude and deepseek is that with anthropic the thinking and is a separate message (even multiple messages in case of redacted thinking) and this should be fed back to the AI in the same way on the next turn (which is required for tool use, otherwise those blocks are just ignored). I was thinking about combining the srategy, but I thought that it would be overly complicated due to the big differences. As of UI change it would be one more week until I could update, but why not let other users start sooner as it is right now?

ferenci84:
The buttons between the first message and thinking block seems to be a mistake as it's not present in other cases.

ferenci84:
Thinking further, there is one corner case in which a shared implementation would be helpful: let's say you start the conversation with anthropic and then swith to deepseek. Not sure how common is that.

ferenci84:
My ide is to have a metadata object with dynamic data structure for assistant messages,  in which thinking data could be added, but also other possibly complex data that help us reconstruct the messages. This could even be any type so that any additional data like signature, multiple redacted thinking that shouldn't be displayed, could be stored. In addition we could add an optional "display" key that would contain data about how to display, for example that deepseek will not show think tag in the main message, we could have sepaeate (optional) display information for main message and thinking message, that is parsed in the core (even better, in the LLM class as it would better help encapsulation), not in the UI as is is now. What do you think?

Any way, I think crucial updates should be done (for example those that support seeing thinking blocks and avoid errors with tools) because it would help the product to succeed.

FallDownTheSystem:
I didn't realize someone else was already working on this. 😅

I've basically implemented the same functionality, with tool support and using the existing Reasoning UI: https://github.com/continuedev/continue/pull/4559

Although there's no support for Bedrock in mine. Also I tried and you can swap between thinking models, unless you use tools, then when you try to swap to DeepSeek you get an error.

Anyway, I'm happy with either or any solution being merged into Continue. 😄

ferenci84:
@FallDownTheSystem Checked your solution, I also wanted to go the same path of adding a new type of MessagePart, before I realized that it would require change in lot of seemingly irrelevant code. What do you think about a longterm solution to simplify as I explained above? Not sure if it was clear. For short term, I think just one of the working solution should be merged so that ordinary users can start to use the new feature of Claude, they will not care as long as it's working well.

FallDownTheSystem:
> @FallDownTheSystem Checked your solution, I also wanted to go the same path of adding a new type of MessagePart, before I realized that it would require change in lot of seemingly irrelevant code. What do you think about a longterm solution to simplify as I explained above? Not sure if it was clear. For short term, I think just one of the working solution should be merged so that ordinary users can start to use the new feature of Claude, they will not care as long as it's working well.

It does feel like some kind of refactoring would be in order. Currently the message are converted into suitable format for the UI and then converted back for the API. Ideally you'd be able to keep the original data intact and have LLM specific methods that are responsible for converting that to a format that's suitable for the UI.

I can see that the code base has gotten more complex as support for new formats have been added. I'm sure originally everything was pretty simple when the output was just a string.

ferenci84:
@sestinj I have changed it to a button display and also added seconds:
Screenshot 2025-03-11 at 22 19 51

It looks like this when seconds is not available:


When opened:
Screenshot 2025-03-11 at 22 23 22

It's also part of the PR with Bedrock support: https://github.com/continuedev/continue/pull/4453

Issue #3994, opened 2025-02-06
Autocomplete: completion is truncated (stops streaming) at the wrong place


ferenci84:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:
- Continue version: 0.9.260
- IDE version:
- Model: qwen2.5-coder:1.5b-base (ollama)
- config.json:
  
"tabAutocompleteModel": {
    "title": "Qwen2.5-Coder 1.5B",
    "provider": "ollama",
    "model": "qwen2.5-coder:1.5b-base",
    "completionOptions": {
      "maxTokens": 256
    }
  },
```

### The problem

The log shows that the completion is OK, but it only partially appears. The model completes `"1__tabAutocompleteModel": {` but only `"1__` appears.

### To reproduce

Open the testing sandbox, add this file:
```
{
  "1__tabAutocompleteModel": {
    "title": "Codestral",
    "provider": "mistral",
    "model": "codestral-latest",
    "apiKey": "",
    "apiBase": "https://codestral.mistral.ai/v1",
    "completionOptions": {
      "maxTokens": 256
    }
  },

  "tabAutocompleteModel": {
    "title": "Qwen2.5-Coder 3B",
    "provider": "ollama",
    "model": "qwen2.5-coder:1.5b-base",
    "completionOptions": {
      "maxTokens": 256
    }
  },
  "3__tabAutocompleteModel": {
    "title": "Together Qwen2.5 Coder",
    "provider": "together",
    "model": "Qwen/Qwen2.5-Coder-32B-Instruct",
    "apiKey": ""
  }
}
```
(A bit awkward, but this is a way I switch between different autocomplete model, and this is the place where the error happens)

Save to a file (test_config.json), reload the window. Place the cursor at the empty line and trigger autocomplete with a hotkey.

For me, the completion is:
```
"2__
```

In the output, the completion is:
```
"2__tabAutocompleteModel": {<|cursor|>
}
```

## What causes the problem

I debugged, and found that the problem is caused by this line in core/autocomplete/filtering/streamTransforms/StreamTransformPipeline.ts:
```
[...]

PR #4448, opened 2025-03-03
default maxTokens setting for autocomplete

ferenci84:
## Description

Default maxTokens at 256 for autocomplete if there is no overriding user setting for the model. Added `autoCompleteMaxTokens`

Change from this comment:
https://github.com/continuedev/continue/issues/3994#issuecomment-2649166970

## Checklist

- [ ] The relevant docs, if any, have been updated or created - let's create a separate issue for this once the PR is approved and `autoCompleteMaxTokens` setting is kept in the completion options.
- [X] The relevant tests, if any, have been updated or created - no relevant tests as far as I know

## Testing instructions

I tested by directly putting a log message into the Ollama._streamFim() method.

### DIFF
--- a/core/llm/llms/index.ts
+++ b/core/llm/llms/index.ts
@@ -111,8 +111,7 @@ export async function llmFromDescription(
   ideSettings: IdeSettings,
   writeLog: (log: string) => Promise<void>,
   completionOptions?: BaseCompletionOptions,
-  systemMessage?: string,
-  isAutocomplete = false
+  systemMessage?: string
 ): Promise<BaseLLM | undefined> {
   const cls = LLMClasses.find((llm) => llm.providerName === desc.provider);
 
--- a/core/llm/llms/index.ts
+++ b/core/llm/llms/index.ts
@@ -19,7 +19,6 @@ import Cohere from "./Cohere";
 import DeepInfra from "./DeepInfra";
 import Deepseek from "./Deepseek";
 import Fireworks from "./Fireworks";
-import NCompass from "./NCompass";
 import Flowise from "./Flowise";
 import FreeTrial from "./FreeTrial";
 import FunctionNetwork from "./FunctionNetwork";
@@ -35,7 +34,9 @@ import Mistral from "./Mistral";
 import MockLLM from "./Mock";
 import Moonshot from "./Moonshot";
 import Msty from "./Msty";
+import NCompass from "./NCompass";
 import Nebius from "./Nebius";
+import Novita from "./Novita";
 import Nvidia from "./Nvidia";
 import Ollama from "./Ollama";
 import OpenAI from "./OpenAI";
@@ -49,7 +50,6 @@ import ContinueProxy from "./stubs/ContinueProxy";
 import TestLLM from "./Test";
 import TextGenWebUI from "./TextGenWebUI";
 import Together from "./Together";
-import Novita from "./Novita";
 import VertexAI from "./VertexAI";
 import Vllm from "./Vllm";
 import WatsonX from "./WatsonX";
@@ -112,6 +112,7 @@ export async function llmFromDescription(
   writeLog: (log: string) => Promise<void>,
   completionOptions?: BaseCompletionOptions,
   systemMessage?: string,
+  isAutocomplete = false
 ): Promise<BaseLLM | undefined> {
   const cls = LLMClasses.find((llm) => llm.providerName === desc.provider);
 
@@ -137,6 +138,11 @@ export async function llmFromDescription(
       maxTokens:
         finalCompletionOptions.maxTokens ??
         cls.defaultOptions?.completionOptions?.maxTokens,
+      autoCompleteMaxTokens:
+        finalCompletionOptions.autoCompleteMaxTokens ??
+        finalCompletionOptions.maxTokens ??
+        cls.defaultOptions?.completionOptions?.autoCompleteMaxTokens ??
+        256
     },
     systemMessage,
     writeLog,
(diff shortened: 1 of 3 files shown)


sestinj (changes requested):
@ferenci84 rather than adding even more options to the config, I would much rather allow users to set `maxTokens` in the `completionOptions` section of their config for the model

core/llm/llms/index.ts
```diff
@@ -137,6 +138,11 @@ export async function llmFromDescription(
       maxTokens:
         finalCompletionOptions.maxTokens ??
         cls.defaultOptions?.completionOptions?.maxTokens,
+      autoCompleteMaxTokens:
```
ferenci84, line 140: We need this extra key in the LLMOptions, if we want to set different maxTokens limit for autocomplete than for any other use, because we cannot determine here whether this model is used for autocomplete or chat.

core/llm/llms/index.ts
```diff
@@ -137,6 +138,11 @@ export async function llmFromDescription(
...
+        finalCompletionOptions.autoCompleteMaxTokens ??
```
ferenci84, line 141: If we remove this line, we disallow users to use this setting.

ferenci84:
@sestinj Please look at this: these are possible places to set defaults for maxTokens for autocomplete:

This is setting for global default:
Screenshot 2025-03-16 at 21 45 35

There may also be defaults for individual models:
Screenshot 2025-03-16 at 21 44 05

This is how the final maxTokens is set when making the request. As you can see, maxTokens user setting will always take precedence, the user don't even have to know about the extra key in the BaseCompletionOptions type:
Screenshot 2025-03-16 at 21 49 10

There may be a misunderstanding, I think it's better if the additional key exist just in the type, but not open to users (i.e. hidden, not documented), however they should be used for setting individual defaults for providers, there may be fast and possibly more capable providers that can output more lines of completion, while slower, or those that tend to go into repetition, should be limited.

I believe that adding this key to the BaseCompletionOptions type is a good way to keep it simple for us, developers. If you want to keep it simple for users too, and limit their options, there is a way for us to ignore this additional setting that comes from the user config:
Screenshot 2025-03-16 at 22 02 18

Any way, please let me know if you have an idea to make it better.

halfline:
out of curiosity, do you know how many tokens on average you're using for autocomplete? Autocomplete pulls in snippets from various parts of the code and it can get kind of lengthy (especially with the hole filler models). on my system, just doing a completion in the quickstart tutorial uses north of 800 tokens.

ferenci84:
@halfline I do not have such stats. You are talking about the contextLength setting. This modification is about maxTokens that is the limit on the output tokens.

RomneyDa (changes requested):
@ferenci84 let's hold on changes since not sure about this parameter yet, but some notes:
- Would need to add YAML support (add to `defaultCompletionOptions` in YAML schema). See previous [JSON-only PR](https://github.com/continuedev/continue/pull/4453) and [resulting issue](https://github.com/continuedev/continue/issues/4918)
- It seems like with this implementation `maxAutocompleteTokens` is redundant of `maxTokens`, since `maxTokens` will be ignored if `maxAutocompleteTokens` is present. E.g. it's not detecting if the model is currently being used for autocomplete so isn't different than just using `maxTokens` (correct me if I'm wrong?)
- The default should probably live in BaseLLM or `constants.js`

ferenci84:
> * E.g. it's not detecting if the model is currently being used for autocomplete so isn't different than just using `maxTokens` (correct me if I'm wrong?)

@RomneyDa, maxAutocompleteTokens is only used when the model is used for autocomplete. See that the maxAutocompleteTokens value is used only in the provideInlineCompletionItems function.

The bottom line: This modification is NOT to add a new parameter, but to keep the default maxToken parameter low, in the following conditions: 
1. The user doesn't set maxTokens explicitly
AND
2. The model is used for autocomplete

I was asked to submit a PR for this issue:
https://github.com/continuedev/continue/issues/3994#issuecomment-2649166970

The reason to add a new key to the BaseCompletionOptions, and store the default there, is to avoid ovecomplicating things with new types. It can be a different decision whether it's good idea to add a new parameter or not (I would vote NOT), it's not part of this modification, so json or yaml config schema shouldn't be changed.

> The default should probably live in BaseLLM or constants.js

I agree. It's possible to add a default per provider. Just as we have default cls.defaultOptions?.completionOptions?.maxTokens, we also have default cls.defaultOptions?.completionOptions?.autoCompleteMaxTokens, except that I added 256 fallback if it's not set (as it will not be set for most providers anyway). We can move the 256 default to any other place.

Please let me know if we are on the same page about the other two questions before I make any further modifications.

sestinj:
```
The bottom line: This modification is NOT to add a new parameter, but to keep the default maxToken parameter low, in the following conditions:

The user doesn't set maxTokens explicitly
AND
The model is used for autocomplete
```

^ given that this is the goal, I think the better approach would be to add some constant in `parameters.ts` or something that can be used as a fallback in `CompletionProvider.ts`. We don't need to introduce new configuration to accomplish the thing you describe here, and I happen to be very averse to new configuration when not necessary

sestinj:
@ferenci84 I'm going to close this PR and would welcome a new one. As mentioned in the comment just above, I think this is a great improvement, but I don't want to add additional config. We should try to accomplish this with something like [here](https://github.com/continuedev/continue/blob/5d5cf0226d0007231af1f2d6b5bcf8a9cc116e5b/core/autocomplete/CompletionProvider.ts#L72-L74)

PR #4449, opened 2025-03-03
Nate/fix-remount

sestinj:


owtaylor:
Confirmed to fix the problem for me, changes look correct as far as I can tell.

sestinj:
glad to hear that! merging and releasing a pre-release in the next few hours

PR #4452, opened 2025-03-03
Set telemetry to disabled by default

jonvargason:
## Description

Disabling telemetry by default.  Why: Some users and corporations may not want to have an opt-out model but would rather have an opt-in to avoid accidentally sending potentially company sensitive data by accident upon install. Full disclosure I don't believe anything this plugin sends is sensitive but companies are very concerned about data egress. I personally think we should enable it for academic and personal usage but enterprise adoption - this may be frowned on or likely would be blocked anyway by firewalls.

## Checklist

## Screenshots

image

image

After install:
image

## Testing instructions

I'll be honest and say I could not get this to build on my work network but I suspect it was a proxy thing and VS issue as even though I made what I believe are the correct modifications and pointed to my instance of VS 2022 it still failed on sqlite and I am not sure why.  I was able to pull the pipeline build at [https://github.com/continuedev/continue/actions/runs/13640223539/job/38128264263?pr=4452](https://github.com/continuedev/continue/actions/runs/13640223539/artifacts/2684470943). 

I believe to test this you basically follow the instructions on [contributing](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md).

1. Make sure you have the correct version of npm (v20.11.0)
2. Pull down my PR
3. Install all dependencies in VS Code
4. Build and Run Extension locally
5. Validate Telemetry is disabled by default (settings -> unchecked Telemetry Enabled)

Alternatively if you want to skip the build like I did locally:

1. Grab that binary at [https://github.com/continuedev/continue/actions/runs/13640223539/job/38128264263?pr=4452](https://github.com/continuedev/continue/actions/runs/13640223539/artifacts/2684470943).
2. Then Open VS Code and Install VSIX
3. Validate Telemetry is disabled by default (settings -> unchecked Telemetry Enabled)

### DIFF
--- a/docs/docs/json-reference.md
+++ b/docs/docs/json-reference.md
@@ -374,7 +374,7 @@ Example:
 
 ### `allowAnonymousTelemetry`
 
-When `true`, anonymous usage data is collected using Posthog, to improve features. Default is `true`.
+When `true`, anonymous usage data is collected using Posthog, to improve features. Default is `false`.
 
 ### `userToken`
 
--- a/docs/docs/telemetry.md
+++ b/docs/docs/telemetry.md
@@ -24,7 +24,7 @@ The following usage information is collected and reported:
 - **System Information:** The name of your operating system (OS) and integrated development environment (IDE).
 - **Pageviews:** General pageview statistics.
 
-## How to Opt Out
+## How to Opt In
 
 You can enable anonymous telemetry by visiting the [User Settings Page](./customize/settings.md) and toggling "Allow Anonymous Telemetry" on.
 
--- a/docs/docs/telemetry.md
+++ b/docs/docs/telemetry.md
@@ -1,12 +1,12 @@
 ---
 title: Telemetry
 description: Learn how Continue collects anonymous usage information and how you can opt out.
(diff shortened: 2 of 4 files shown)


sestinj:
@jonvargason I appreciate you taking the time to make a quality PR here. At least as of now, opt-out telemetry is not something that we are looking to change, as it's a core part of our being able to understand and improve the product for all users. We aim to make sure that it's perfectly clear how to turn off telemetry for those who need to, and have even added a more easily accessible toggle in the sidebar UI recently. And I can confirm that, as you noted, none of the telemetry is sensitive: no code, no prompts, no personal information are ever sent.

However I can understand the tough situation where you might not reliably be able to have every new employee who downloads Continue remember immediately to turn it off. Depending on what sort of environment Continue is being used in we may be able to come to a creative solution for this. Some examples that I can think of immediately:
- If hub.continue.dev is an option for you, you would be able to centrally turn off telemetry for all signed in users in your org
- If you have the ability to manage files on computers, a configuration file that disables telemetry could be placed there
- This traffic could be configured to be blocked at the network level, which provides the most possible security

Feel free to reach out if you want to discuss any of these possibilities

PR #4482, opened 2025-03-05
fix(tools): Terminal command returns stdout but not stderr after error detected

chezsmithy:
## Description

Fixes https://github.com/continuedev/continue/issues/4503

Some terminal commands like running cucumber tests seem to result in an error, but the results are returned to stdout vs stderr. This causes the terminal tool to not see the cucumber failures and is unable to fix the errors automatically.

## Checklist

## Screenshots

## Testing instructions

Run cucumber tests using javascript command npm run cucumber. The tests fail, and no results are observed in the tools terminal call. When setting a breakpoint stderr is empty but stdout is populated.

Curious if we should somehow concat the results so that the LLM can can see all the terminal content to process it to resolve issues.

### DIFF
--- a/core/tools/implementations/runTerminalCommand.ts
+++ b/core/tools/implementations/runTerminalCommand.ts
@@ -1,8 +1,8 @@
 import childProcess from "node:child_process";
 import util from "node:util";
 
-import { ToolImpl } from ".";
 import { fileURLToPath } from "node:url";
+import { ToolImpl } from ".";
 
 const asyncExec = util.promisify(childProcess.exec);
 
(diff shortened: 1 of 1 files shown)


chezsmithy:
@sestinj thoughts on this PR? After enabling tools for bedrock noticing it's a pretty significant barrier to tools coding with some of the test frameworks.

sestinj (approved):

sestinj:
This is great @chezsmithy , and if you are finding other barriers please let me know. We're starting to focus a lot on tool use

chezsmithy:
@sestinj the terminal refactor you are working on should keep this work in place as well.

PR #4497, opened 2025-03-05
Fix vllm setup, json.data is an array and is accessed as such. Also a…

NghiVo21:
## Description
 
- add auto detect behaviour
- fix bug with accessing the array instead of the object

## Checklist

No change from what i can tell.

## Testing instructions

Configure continue to use vllm, simply perform simple tasks with continue like edit or chat to encounter this bug.


sestinj (approved):

sestinj:
thanks for the PR @NghiVo21 !

PR #4505, opened 2025-03-06
Filter for files ending with .prompt in global Prompt Files

iandolge:
## Description

added a check that files end with ".prompt" when searching for files in the global prompt directories. Before this change, all files are collected for `@Prompt Files`

## Checklist

## Screenshots

N/A

## Testing instructions

1. Add files ending with .prompt and .txt in ~/.continue/prompts
2. Write a new prompt using the `@Prompt Files` context.
3. Observe .txt file in the list before this change and only the .prompt file after this change.


RomneyDa (approved):
Nice, I noticed this one this week

PR #4506, opened 2025-03-06
Refactor system message composition for better code proximity & allow user to enforce empty system message as required by Deepseek R1 series

ferenci84:
## Description

1. Moved systemMessage composition to a single place - makes it easy to make changes discussed in issue https://github.com/continuedev/continue/issues/3786
2. When user defined systemMessage is an explicit empty string, no system message will be created - this resolves https://github.com/continuedev/continue/issues/3786#issuecomment-2701614971

Note: https://github.com/continuedev/continue/pull/4507 is built on top of this one, that contains a more complete, yet experimental solution for the issue https://github.com/continuedev/continue/issues/3786

## Checklist

- [ ] The relevant docs, if any, have been updated or created
- [x] The relevant tests, if any, have been updated or created

## Testing instructions

If you set empty systemMessage in the model config or in the top-level config, no systemMessage should be used (model config takes precedence even if it's an empty string)

PR #4507, opened 2025-03-06
System_message_composition

ferenci84:
## Description

Continuation of https://github.com/continuedev/continue/pull/4506

Replacement of https://github.com/continuedev/continue/pull/3787

On top of PR 4506 it adds experimental configuration systemMessageComposition. Changed documentation explains how it works.

Legacy logic remains working if user do not change the configuration.

## Checklist

## Screenshots

## Testing instructions

Experiment with config:
```
  "experimental": {
    "systemMessageComposition": "prepend"
  }
```

In the model description, you can also set:
```
     "promptTemplates": {
        "codeBlockInstructions": "My own instructions for code blocks",
        "toolUseRules": "My tool use rules"
      },
```

### DIFF
--- a/core/config/load.ts
+++ b/core/config/load.ts
@@ -369,7 +369,7 @@ async function intermediateToFinalConfig(
               ideSettings,
               writeLog,
               config.completionOptions,
-              config.systemMessage,
+              config.systemMessage
             );
 
             if (llm?.providerName === "free-trial") {
@@ -397,8 +397,8 @@ async function intermediateToFinalConfig(
       (config.contextProviders || [])
         .filter(isContextProviderWithParams)
         .find((cp) => cp.name === "codebase") as
-        | ContextProviderWithParams
-        | undefined
+      | ContextProviderWithParams
+      | undefined
     )?.params || {};
 
   const DEFAULT_CONTEXT_PROVIDERS = [
@@ -991,5 +991,6 @@ export {
   finalToBrowserConfig,
   intermediateToFinalConfig,
   loadContinueConfigFromJson,
-  type BrowserSerializedContinueConfig,
+  type BrowserSerializedContinueConfig
 };
+
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -1047,6 +1047,7 @@ declare global {
     useChromiumForDocsCrawling?: boolean;
     useTools?: boolean;
     modelContextProtocolServers?: MCPOptions[];
+    systemMessageComposition?: "legacy" | "append" | "prepend" | "placeholders";
   }
   
   interface AnalyticsConfig {
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -43,13 +43,13 @@ export interface IndexingProgressUpdate {
   desc: string;
   shouldClearIndexes?: boolean;
   status:
-    | "loading"
-    | "indexing"
-    | "done"
-    | "failed"
-    | "paused"
-    | "disabled"
-    | "cancelled";
+  | "loading"
+  | "indexing"
+  | "done"
+  | "failed"
+  | "paused"
+  | "disabled"
+  | "cancelled";
   debugInfo?: string;
 }
 
@@ -675,10 +675,10 @@ export interface IDE {
   getCurrentFile(): Promise<
     | undefined
     | {
-        isUntitled: boolean;
-        path: string;
-        contents: string;
-      }
+      isUntitled: boolean;
+      path: string;
+      contents: string;
+    }
   >;
 
   getLastFileSaveTimestamp?(): number;
@@ -862,11 +862,11 @@ export interface CustomCommand {
 export interface Prediction {
   type: "content";
   content:
-    | string
-    | {
-        type: "text";
-        text: string;
-      }[];
+  | string
+  | {
+    type: "text";
+    text: string;
+  }[];
 }
 
 export interface ToolExtras {
@@ -1134,6 +1134,7 @@ export interface ExperimentalConfig {
    */
   useChromiumForDocsCrawling?: boolean;
   modelContextProtocolServers?: MCPOptions[];
+  systemMessageComposition?: "legacy" | "append" | "prepend" | "placeholders";
 }
 
 export interface AnalyticsConfig {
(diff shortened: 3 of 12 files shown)


sammyvoncheese:
very cool change, I have been looking at doing something similar to customize/override all of the hidden system messages sent with each request.

sestinj:
@ferenci84 I'm a bit concerned about all of the complexity that this adds. I'd like to get to some agreement about a plan that is easier to reason about before starting more work.

With something like the below you don't need to think about prepending, composing, etc.. It's a field that can be safely removed or altered at some point, and it seems to satisfy the needs of removing and replacing the default system prompt:

```yaml
  - name: gpt-4o-mini
    provider: openai
    model: gpt-4o-mini
    apiKey: sk-xxx
    roles:
      - chat
      - edit
      - apply
    promptTemplates:
      defaultRules: ""
  ```

If defaultRules is set, then we would just entirely replace our defaults with this, which would effectively remove them if you set it to "". If you want to write your own tool use rules, then you can do so in the rules section as usual.

What do you think?

**Comment by ferenci84** [comment]
Do you want to remove systemMessage alltpgether? If not removed/deprecated, how would that work if any value is set?

**Comment by ferenci84** [comment]
> If you want to write your own tool use rules, then you can do so in the rules section as usual

This is not clear to me. The tool use rule is added dynamically, only when tools are enabled. The rules section contains messages that are added always. Or, do I miss something?

Looking at the rules config option, it raises further questions:
```
rules:
  - uses: myprofile/my-mood-setter
    with:
      MOOD: happy
  - Always annotate Python functions with their parameter and return types
  - Always write Google style docstrings for functions and classes
 ```
Would it appear before or after the "defaultRules" specified in the model, or if "defaultRules" are defined, it replaces everything? What is the advantage of having a list of rules over having a single systemMessage?

ferenci84:
@sestinj I had an idea yesterday, that may be simple to average users, but would provide additional tools for power user, let me explain:

I was reading couple of weeks ago in Claude docs that they recommend using the system message for role prompting, and put any other instructions (like how to format code) in the user message instead.

My idea is to have have promptTemplates with the keys defaultRules and toolUseRules, but also systemMessageTemplate AND userMessageTemplate, both with placeholders using the same keys as in promptTemplates, and with default already set for the user (it could be just systemMessage if we want to avoid changing the old name, but if there may be placeholders, template may be more appropriate).

This would have the following advantages:
- Average or new users can change the defaultRules and toolUseRules without having to understand them more (but as they are not editing a config named "systemMessage" they may not be surprised so much that something is appended to their inputted text)
- Users who want control over the whole systemMessage, can just put their own text and nothing will be added to it, since if there is no placeholder found, there will be no replacement, they work as simple string
- Users who actually read the deep-dive section of the documentation, can alter the systemMessageTemplate and userMessageTemplate so that they can add/remove parts
- Specifically for Claude, it would be possible to add the a standard defultRules and the toolUseRules (that would be added only when the user actually enables any of the tools) right within the user message and keep the system message contain only the role.

The following example elaborates what I thought about:

Simple user-defined default rules, default toolUseRules would be appended:
```
  - name: gpt-4o-mini
    provider: openai
    model: gpt-4o-mini
    promptTemplates:
      defaultRules: "Custom prompt"
      
 - name: gpt-4o-mini-1
    provider: openai
    model: gpt-4o-mini
    promptTemplates:
      defaultRules: "Custom prompt"
      toolUseRules: "Custom tool use rules"
```

User may overwrite the whole system message:
```
  - name: gpt-4o-mini
    provider: openai
    model: gpt-4o-mini
    systemMessageTemplate: ""

  - name: gpt-4o-mini-1
    provider: openai
    model: gpt-4o-mini
    systemMessageTemplate: "You are an expert developer"

# this has same effect as setting both prompt templates:
 - name: gpt-4o-mini
    provider: openai
    model: gpt-4o-mini
    promptTemplates:
      defaultRules: ""
      toolUseRules: ""
      
 - name: gpt-4o-mini-1
    provider: openai
    model: gpt-4o-mini
    promptTemplates:
      defaultRules: "You are an expert developer"
      toolUseRules: ""

```

When user make edit in the config, these defaults would appear (Note: this is a draft, we can talk about the best format.):
```
 - name: gpt-4o-mini-1
    provider: openai
    model: gpt-4o-mini
    systemMessageTemplate: "{defaultRules}{toolUseRules}"
    userMessageTemplate: "{context}{userMessage}"
    promptTemplates:
      defaultRules: "Always include the language and file name in the info string when you write code blocks, for example '```python file.py'."
      toolUseRules: "When using tools, follow the following guidelines:\n- Avoid calling tools unless they are absolutely necessary. For example, if you are asked a simple programming question you do not need web search. As another example, if the user asks you to explain something about code, do not create a new file."
```

For specific models, the placement of "rules" may be changed (we may also allow arbitrary keys in promptTemplates as demonstrated here):
```
 - name: anthropic-claude
    provider: anthropic
    model: anthropic-claude-3.7
    systemMessageTemplate: "{role}"
    userMessageTemplate: "{context}{userMessage}{defaultRules}{toolUseRules}"
    promptTemplates:
      role: "You are an expert developer"
      defaultRules: "Always include the language and file name in the info string when you write code blocks, for example '```python file.py'."
      toolUseRules: "When using tools, follow the following guidelines:\n- Avoid calling tools unless they are absolutely necessary. For example, if you are asked a simple programming question you do not need web search. As another example, if the user asks you to explain something about code, do not create a new file."
```

Additional cool features may also be added:
- A top-level promptTemplates config may let users reuse messages, and use those keys in more than one model setup.
- For some models, systemMessageTemplate and userMessageTemplate may be used in the model defaults, for example we can set default empty system message for Deepseek R1 as recommended by their readme, and put the default and tool use rules in the user message for Claude models.
- There may be rules that could be put only in the first user message instead of repeating it in each. For example if I instruct Claude the output the whole edited code in one of the messages, it will behave the same way in every subsequent messages. (Btw if I do the same in system message, it simply doesn't follow the instructions)
- The method could be reused with just a few lines of code, for example if o1 requires providing developer message instead of system message.

The placeholder part is already done and working, and there is a smart handling of whitespaces before/after the replaced parts (this complexity is clearly encapsulated in a single function). Also with the change in this PR, the final system message is constructed in a single place, so we can put everything about that into a single file and make changes in that single place, instead of having to change several places in the code-base.

Let me know what do you think of it as direction and if you have any idea. I don't think it should be done in one-shot in a full-featured, single change, rather it should be broken down to multiple smaller changes, maybe first doing changes only about the system message (most parts are already done for this), then, if working well, working on the user message template (maybe after involving other users into the talk), and then adding extra features described above and possibly implementing sensible defaults for specific models.

ferenci84:
Became obsolete. New PR was submitted resolving the same problem: https://github.com/continuedev/continue/pull/5254

PR #4515, opened 2025-03-06
Fixed unwantted selecting behavior for send-to-chat

Jazzcort:
## Description

The original behavior selects the entire first line even though it's only partially selected in the editor. This fix is to get rid of the extra selection from the first line and only send the exact part that gets selected in the editor to the chat section. https://github.com/Granite-Code/granite-code/issues/43

## Testing instructions

Partially select the first line and then Cmd-L, check if code snippet appearing in the chat section display the exact part you select in the editor.

### DIFF
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -123,9 +123,20 @@ function getRangeInFileWithContents(
       return null;
     }
 
-    // adjust starting position to include indentation
-    const start = new vscode.Position(selection.start.line, 0);
-    const selectionRange = new vscode.Range(start, selection.end);
+    let selectionRange = new vscode.Range(selection.start, selection.end);
+    const document = editor.document;
+    // Select the context from the beginning of the selection start line to the selection start position
+    const beginningOfSelectionStartLine = selection.start.with(undefined, 0);
+    const textBeforeSelectionStart = document.getText(
+      new vscode.Range(beginningOfSelectionStartLine, selection.start),
+    );
+    // If there are only whitespace before the start of the selection, include the indentation
+    if (textBeforeSelectionStart.trim().length === 0) {
+      selectionRange = selectionRange.with({
+        start: beginningOfSelectionStartLine,
+      });
+    }
+
     const contents = editor.document.getText(selectionRange);
 
     return {


owtaylor:
Hmm, this entirely bypasses the code in `getRangeInFileWithContents()` that is causing the problem:

```
    // adjust starting position to include indentation
    const start = new vscode.Position(selection.start.line, 0);
    const selectionRange = new vscode.Range(start, selection.end);
```

But I think that code *does* have a point - if you select with shift-downarrow, it's easy to get a selection that includes the first line without the whitespace, and the rest of the lines with the whitespace:

image

In that case, we don't want to send to the model:

```
multiply(number) {
    this.result *= number;
    return this;
  }
```

Probably that code needs to be fancier - to check if the text at the beginning of the first line is all whitespace and only *then* adjust the first line boundary to be the start of the line.

extensions/vscode/src/commands.ts
```diff
@@ -150,6 +183,10 @@ function getRangeInFileWithContents(
 async function addHighlightedCodeToContext(
   webviewProtocol: VsCodeWebviewProtocol | undefined,
 ) {
+  const editor = vscode.window.activeTextEditor;
+  if (!editor) {
+    return;
```
owtaylor: This doesn't look necessary - getRangeInFileWithContents() already handles that.

extensions/vscode/src/commands.ts
```diff
@@ -123,9 +123,28 @@ function getRangeInFileWithContents(
...
-    // adjust starting position to include indentation
-    const start = new vscode.Position(selection.start.line, 0);
-    const selectionRange = new vscode.Range(start, selection.end);
+    let selectionRange = new vscode.Range(selection.start, selection.end);
+    const document = editor.document;
+    // Select the context from the beginning of the selection start line to the selection start position
+    const beginningOfSelectionStartLine = new vscode.Position(
+      selection.start.line,
+      0,
+    );
```
owtaylor: There's a little compression we can do here:

```
const beginningOfSelectionStartLine = selection.start.with(undefined, 0);
```

extensions/vscode/src/commands.ts
```diff
@@ -123,9 +123,28 @@ function getRangeInFileWithContents(
...
+    const rangeBeforeSelectionStart = new vscode.Range(
+      beginningOfSelectionStartLine,
+      selection.start,
+    );
+    const contextBeforeSelectionStart = document.getText(
```
owtaylor: I think this is just "text" not "context".  I also think it would be clearer not to have the rangeBeforeSelectionStart variable and just create the range inline.

extensions/vscode/src/commands.ts
```diff
@@ -123,9 +123,28 @@ function getRangeInFileWithContents(
...
+      rangeBeforeSelectionStart,
+    );
+    // If there are only whitespace before the start of the selection, include the indentation
+    if (contextBeforeSelectionStart.trim().length === 0) {
+      selectionRange = new vscode.Range(
+        new vscode.Position(selection.start.line, 0),
+        selection.end,
+      );
```
owtaylor: Writing it this way makes it not clear that we're just adjusting the start position of selectionRange -range also has a `with()` -

```
selectionRange = selectionRange.with({ start: beginningOfSelectionStartLine });
```

would be clearer.

owtaylor (commented):
This looks good to me now, but I think there are a few changes additional changes that could be done to tighten up the code and make it more readable.

owtaylor:
Looks good to me now!

Patrick-Erichsen (approved):
Nice catch!

PR #4517, opened 2025-03-06
edit mode code highlighting fixed to not override cursor selection

panyamkeerthana:
## Description

When switching to the edit mode and we make a selection of code, it needs to appear above the highlighting not be obscured by it. In this fix, instead of using an opaque background colour for the highlighting decoration, we use a highlighting with opacity so that the selected code is visible below it

## Screenshots
Screenshot 2025-03-05 at 3 46 25 PM
Screenshot 2025-03-05 at 3 31 22 PM

Screenshot 2025-03-05 at 3 48 26 PM
Screenshot 2025-03-05 at 3 50 01 PM

Link: https://github.com/Granite-Code/granite-code/issues/47

### DIFF
--- a/extensions/vscode/src/quickEdit/EditDecorationManager.ts
+++ b/extensions/vscode/src/quickEdit/EditDecorationManager.ts
@@ -6,7 +6,10 @@ class EditDecorationManager {
   constructor(context: vscode.ExtensionContext) {
     this.decorationType = vscode.window.createTextEditorDecorationType({
       backgroundColor: new vscode.ThemeColor(
-        "editor.inactiveSelectionBackground",
+        // "editor.selectionHighlightBackground" requires partial transparency.
+        // This ensures the highlight does not completely obscure the selection,
+        // making it useful for repurposing here.
+        "editor.selectionHighlightBackground",
       ),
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -6,7 +6,7 @@ class EditDecorationManager {
   constructor(context: vscode.ExtensionContext) {
     this.decorationType = vscode.window.createTextEditorDecorationType({
       backgroundColor: new vscode.ThemeColor(
```
owtaylor, line 8: I'd include a comment here explaining *why* the new color is a good choice, something like:

```
// editor.selectionHighlightBackground is documented as always needing to be
// partially transparent. This property is useful when we repurpose it here,
// because it means it won't entirely obscure the selection
```

owtaylor (commented):
This seems to be a workable fix to me, as long as the change in appearance is considered OK. (IMO, it's an improvement.)

Patrick-Erichsen (approved):

Patrick-Erichsen:
This looks great, nice catch!

PR #4524, opened 2025-03-07
embeddings: add support for prefixes in embeddings

eliransin:
Some of the embedding models (e.g  nomic-embed-text) are trained to embed for several purposes, this on the flip side, means that if not given the purpose for the embedding the retrieval results might and probably will be suboptimal. This change aims to deal with it by adding a configuration option for chunk and query prefixes with the option to support more embedding purposes in the future (i.e clustering).

Some refferences to justify this change:
1. https://huggingface.co/nomic-ai/nomic-embed-text-v1.5#task-instruction-prefixes The nomic-embed-text prefixes section
2. https://www.youtube.com/watch?v=76EIC_RaDNw (Don’t Embed Wrong! by Matt Williams)

## Description

## Checklist

## Screenshots

## Testing instructions

Nothing special just adding some prefixes and making sure that a different index is being created. Some better retrievals might be observed but I guess it depend on the size of the index (the smaller the index the less the effect I assume).

### DIFF
--- a/docs/docs/customize/model-providers/top-level/ollama.mdx
+++ b/docs/docs/customize/model-providers/top-level/ollama.mdx
@@ -84,7 +84,11 @@ We recommend configuring **Nomic Embed Text** as your embeddings model.
     {
       "embeddingsProvider": {
         "provider": "ollama",
-        "model": "nomic-embed-text"
+        "model": "nomic-embed-text",
+        "embeddingPrefixes" : {
+          "chunk": "search_document: ",
+          "query": "search_query: "
+        }
       }
     }
     ```
--- a/core/config/yaml/models.ts
+++ b/core/config/yaml/models.ts
@@ -80,6 +80,9 @@ async function modelConfigToBaseLLM(
   if (model.embedOptions?.maxChunkSize) {
     options.maxEmbeddingChunkSize = model.embedOptions.maxChunkSize;
   }
+  if (model.embedOptions?.embeddingPrefixes) {
+    options.embeddingPrefixes = model.embedOptions.embeddingPrefixes;
+  }
 
   // These are params that are at model config level in JSON
   // But we decided to move to nested `env` in YAML
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -139,7 +139,7 @@ export interface ILLM extends LLMOptions {
     options?: LLMFullCompletionOptions,
   ): Promise<ChatMessage>;
 
-  embed(chunks: string[]): Promise<number[][]>;
+  embed(chunks: string[], embedding_task: EmbeddingTasks): Promise<number[][]>;
 
   rerank(query: string, chunks: Chunk[]): Promise<number[]>;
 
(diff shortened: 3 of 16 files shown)


sestinj (changes requested):
@eliransin thanks for the great work here! Everything written so far looks good to me. There's one thing we should add, which is support for the config.yaml format. This would be added in the `config-yaml/src/schemas/models.ts` `embedOptionsSchema`, and then should be passed through here: https://github.com/continuedev/continue/blob/c1699c975fc252d2ed14b8c4a9770d802852cd3b/core/config/yaml/models.ts

eliransin:
Changes in this version:
1. rebased on main and resolved conflicts
2. added zod definitions as instructed in cr comments
:slightly_smiling_face:

eliransin:
@sestinj I have a bit of a noob qusetion/s, I noticed that ` config-yaml/src/schemas/models.ts embedOptionsSchema` sits in its own package.config-yaml/src/schemas/models.ts embedOptionsSchemathe 
should I bump the version of this package (and probably also bump up the coresponding entry in the core `package.json`) ? If I do, is the dependency installed from local source or should I first open a PR to update this package and have the new version uploaded and registered in package registy and only then do a second PR that uses this package in core?
Tx

eliransin:
Summary:
Figured out I should do the following, but unsure if I need to do it in two separate steps:
1. Make changes to `packages/config-yaml` including bumping up the version
2. Update the dependencies of core to the new `@continuedev/config-yml` version
3. Add the extra assignment from the newly added option in `core/config/yaml/models.ts`
4. All of the current already made changes

My question is:
Should 1 be in a separate PR?

I am going to go ahead and submit one unified PR and will assume that if this is the wrong thing to do, checks will fail :slightly_smiling_face: 

Probably will only do it in a day or so, so maybe I'll get an answer by then...

eliransin:
@sestinj I digged a bit in the previous commits, it seams like I need to first have the package in registry so probably it should occure in 2 separate stages. I will break this PR into two.

eliransin:
@sestinj converted this to draft until the config-yaml package is published and proper adjutments are made as a result.

docs/docs/customize/model-providers/top-level/ollama.mdx
```diff
@@ -84,7 +84,11 @@ We recommend configuring **Nomic Embed Text** as your embeddings model.
     {
       "embeddingsProvider": {
         "provider": "ollama",
```
sestinj, line 86: Let's remove the "embeddingPrefixes" from the JSON in the docs here. I know it's correct, but it ends up looking very complicated for a beginner setup. The way that we will solve this for all users going forward is by having the blocks on hub.continue.dev include the necessary embedding prefixes and then suggest in the docs that users use `uses: ollama/nomic-embed-text` for example

sestinj (changes requested):
Sorry again for being a bit slow to review. All looks good to me at this point except for the one docs change. I would go ahead and make the update myself but since we're in Draft state I want to leave it for you

sestinj:
I know we were really close, but since this is stuck in draft and appears to have gone stale, I'm going to close it for now. Obviously more than happy to jump back in if you have the time to re-open!

PR #4525, opened 2025-03-07
Set VSCODE_VERSION when running 'extest get-chromedriver'

owtaylor:
If we only set the VSCODE_VERSION when running 'extest get-vscode', then we might get a version of chromedriver that is incompatible with the version of vscode we download.

This should fix e2e tests failing with errors like:
```
     SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 132
Current browser version is 128.0.6613.186 with binary path /home/runner/work/continue/continue/extensions/vscode/e2e/storage/VSCode-linux-x64/code
```

e.g. in https://github.com/continuedev/continue/pull/4522 - I think the reason only *some* of the recent pull requests have these failures is caching of the downloads - in other cases we are getting versions downloaded extest was updated for newer vscode/chromedriver.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -594,7 +594,7 @@
     "e2e:compile": "tsc -p ./tsconfig.e2e.json",
     "e2e:build": "npm --prefix ../../gui run build && npm run package",
     "e2e:create-storage": "mkdir -p ./e2e/storage",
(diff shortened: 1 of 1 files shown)


sestinj:
@owtaylor this is great timing! I just happen to be spending today trying to get tests / PRs in order and was planning on looking at this. Thanks for the help!

PR #4533, opened 2025-03-07
Model context protocol prompt fix and roots

alexsmirnov:
## Description

The fix updates prompt ( slash command ) obtained from MCP to properly include prompt context into model call.
Without it, all @... context resources stripped out from user message. The code based on `/review` and custom prompt command code. Noisy comments reflect my investigation into how it works. To confirm correctness, it also includes unit test that repeated all combinations that I collected from debug sessions.

The second change is support for MCP `roots` feature https://modelcontextprotocol.io/docs/concepts/roots.
 Client reports workspace directories as response to request from server. I found it necessary to provide workspace specific resources and tools.
 I do not dare to include my project https://github.com/alexsmirnov/mcp-server-continue as reference to documentation, it's too early stage. But it supports all features.

## Checklist

## Screenshots

## Testing instructions
with https://github.com/alexsmirnov/mcp-server-continue and `uv` https://github.com/astral-sh/uv
config.json:
```
{
"experimental": {
    "modelContextProtocolServers": [
      {
        "transport": {
          "type": "stdio",
          "command": "uv",
          "args": [
            "run",
            "--project",
            "<absolute path project folder>/mcp-server-continue",
            "mcps"
          ]
        }
      }
    ]
  }
}
```
slash command `/echo` with @some.file , or any other resource

### DIFF
--- a/core/commands/slash/mcp.test.ts
+++ b/core/commands/slash/mcp.test.ts
@@ -1,0 +1,311 @@
+import { substituteLastUserMessage } from "./mcp";
+import { ChatMessage, MessagePart, SlashCommand, UserChatMessage } from "../../index.js";
+
+/**
+ * Test substitution of the prompt content into the history.
+ * actual content collected from debugging actual chats.
+ */
+describe("substituteLastUserMessage", () => {
+  const systemMessage: ChatMessage = {
+    role: "system",
+    content: "Your are..."
+  };
+
+  const assistantEmptyMessage: ChatMessage = {
+    role: "assistant",
+    content: ""
+  };
+
+  const mcpMessage: ChatMessage = {
+    role: "user",
+    content: [{ type: "text", text: "provide answer" }]
+  };
+  const mcpHeadMessage: ChatMessage = {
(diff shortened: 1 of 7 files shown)


sestinj:
@alexsmirnov I'm thinking that the cleaner way to solve this replacement problem would be by handling it in `resolveInput.ts`, which is where we replace context providers with their content. That way all of the logic would be centralized in one place

alexsmirnov:
Hi @sestinj , this is that confused me at the first place, how other prompt providers handle resources. I copied functionality from review, and custom prompts. Can you give me directions how to use resolveInput.ts ? I'll take a look on my own by the way

sestinj:
@alexsmirnov I'm sorry for being slow, but I finally got back around to reviewing this again. I think that the prompt fix is unfortunately just adding too much complexity for us to merge, even with the testing.

I however do think that the addition of `roots` is done very well. Would you be interested in opening up a fresh PR for this? I want to clean up this PR since it has gone relatively stale, but we'd be very excited to add roots!

PR #4559, opened 2025-03-09
Add support for reasoning in the UI

FallDownTheSystem:
## Description

This PR adds support for both showing thinking tokens in the chat as well as controlling the reasoning effort for supported models. I'm opening this PR to get feedback from the Continue team, I'm sure there are some design decisions here that you may disagree with and want to change. I'm also okay with the Continue team taking over this branch and developing on top of it.

I'm adding comments in the PR to explain some of the changes.

- Added UI settings to set reasoning effort / token budget
- Added UI to show reasoning tokens from Anthropic Claude 3.7 Sonnet and DeepSeek R1
    - Supports tool use with `thinking` and `redacted_thinking` message types
- Improved UI scaling on smaller sizes to better fit the new Thinking button
- Added support for requestOptions headers to be passed to the Anthropic provider, so that `"anthropic-beta": "output-128k-2025-02-19"` 128k maxOutput can be enabled.

## Checklist

    - Updated yaml/json references for new config options.
    - Couldn't get tests to run on Windows 11 machine at all, regardless if it was this or the main branch. But the tests on the PR passed.
    - If there are new tests that should be created, let me know.

## Screenshots

This shows most of the changes.

https://github.com/user-attachments/assets/6b9596df-f381-4599-8bf1-71cc81cd94bf

## Testing instructions

Test the following models:
- `"provider": "deepseek", "model": "deepseek-reasoner"`
- `"provider": "openai", "model": "o3-mini"`
- `"provider": "openai", "model": "o1"`
- `"provider": "anthropic", "model": "claude-3-7-sonnet-20250219"`
- A non-thinking model like `"provider": "openai", "model": "gpt-4o"`

`completionOptions` should have schema'd definitions for reasoning_effort for the OpenAI models and `thinking` for Anthropic
```
"thinking": {
    "type": "enabled",
    "budget_tokens": 4096
}
```

Tests:
- The Thinking button is not disabled for these models.
- The thinking output is shown for Anthropic Claude 3.7 Sonnet and DeepSeek Reasoner
- Thinking can only be toggled off for Sonnet 3.7
- Sonnet 3.7 works along with tool use for both thinking and redacted thinking, in multi-turn conversations.
    - For redacted thinking, prompt the API with this magic string `ANTHROPIC_MAGIC_STRING_TRIGGER_REDACTED_THINKING_46C9A13E193C177646C7398A98432ECCCE4C1253D5E2D82641AC0E52CC2876CB`
- Test that the thinking and redacted_thinking works even if `stream` is set to `false` for Sonnet 3.7. (Tool use is not supported in Continue when not streaming afaik)
- Thinking options popover shows up and shows the correct settings for Anthropic and OpenAI
- The UI scales nicely even at the smallest view sizes
    - This incles Tool use popover, which now stacks the options when below `xs` breakpoint
-  Test that non-thinking models still work as expected, specifically in the UI considering the numerous changes made to `sessionSlice`
    - Because of the new Message API types, there were some type annotations added to `FreeTrial`, `Gemini`, and `WatsonX` core/llm's, so those should be tested
- DeepSeek reasoner no longer uses promptTemplates because that limits the completions to `string` only, meaning both `content` and `reasoning_content` couldn't be passed down to the UI.
- Test that the Reasoning tokens are still shown even for non-thinking models if you ask the AI to put some text inside think tags like: 
```
<think>Put something here</think>
Rest of the message here
```
- Test that json.config autocompletes the completion options correctly based on the schema definitions.

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -324,7 +324,22 @@ export type ImageMessagePart = {
   imageUrl: { url: string };
 };
 
-export type MessagePart = TextMessagePart | ImageMessagePart;
+export type ThinkingMessagePart = {
+  type: "thinking";
+  thinking: string;
+  signature: string;
+};
+
+export type RedactedThinkingMessagePart = {
+  type: "redacted_thinking";
+  data: string;
+};
+
+export type MessagePart =
+  | TextMessagePart
+  | ImageMessagePart
+  | ThinkingMessagePart
+  | RedactedThinkingMessagePart;
 
 export type MessageContent = string | MessagePart[];
 
@@ -360,6 +375,7 @@ export interface UserChatMessage {
 export interface AssistantChatMessage {
   role: "assistant";
   content: MessageContent;
+  reasoning_content?: string;
   toolCalls?: ToolCallDelta[];
 }
 
@@ -921,11 +937,17 @@ export interface BaseCompletionOptions {
   prediction?: Prediction;
   tools?: Tool[];
   toolChoice?: ToolChoice;
+  thinking?: {
+    type: "enabled" | "disabled";
+    budget_tokens?: number;
+  };
+  reasoning_effort?: "high" | "medium" | "low";
 }
 
 export interface ModelCapability {
   uploadImage?: boolean;
   tools?: boolean;
+  thinking?: boolean;
 }
 
 export interface ModelDescription {
--- a/core/llm/autodetect.ts
+++ b/core/llm/autodetect.ts
@@ -174,7 +174,8 @@ function autodetectTemplateType(model: string): TemplateType | undefined {
     lower.includes("pplx") ||
     lower.includes("gemini") ||
     lower.includes("grok") ||
-    lower.includes("moonshot")
+    lower.includes("moonshot") ||
+    lower.includes("deepseek-reasoner")
   ) {
     return undefined;
   }
@@ -373,11 +374,45 @@ function autodetectPromptTemplates(
   return templates;
 }
 
+const PROVIDER_SUPPORTS_THINKING: string[] = ["anthropic", "openai", "deepseek"];
+
+const MODEL_SUPPORTS_THINKING: string[] = [
+  "claude-3-7-sonnet-20250219",
+  "claude-3-7-sonnet-latest",
+  "o3-mini",
+  "o3-mini-2025-01-31",
+  "o1",
+  "o1-2024-12-17",
+  "deepseek-reasoner",
+];
+
+function modelSupportsThinking(
+  provider: string,
+  model: string,
+  title: string | undefined,
+  capabilities: ModelCapability | undefined,
+): boolean {
+  if (capabilities?.thinking !== undefined) {
+    return capabilities.thinking;
+  }
+
+  if (!PROVIDER_SUPPORTS_THINKING.includes(provider)) {
+    return false;
+  }
+
+  const lower = model.toLowerCase();
+  return MODEL_SUPPORTS_THINKING.some(
+    (modelName) => lower.includes(modelName) || title?.includes(modelName),
+  );
+}
+
 export {
   autodetectPromptTemplates,
   autodetectTemplateFunction,
   autodetectTemplateType,
   llmCanGenerateInParallel,
   modelSupportsImages,
-  modelSupportsTools,
+  modelSupportsThinking,
+  modelSupportsTools
 };
+
--- a/core/llm/countTokens.ts
+++ b/core/llm/countTokens.ts
@@ -90,8 +90,15 @@ async function countTokensAsync(
     const promises = content.map(async (part) => {
       if (part.type === "imageUrl") {
         return countImageTokens(part);
+      } else if (part.type === "thinking") {
+        return (await encoding.encode(part.thinking ?? "")).length;
+      } else if (part.type === "redacted_thinking") {
+        // For redacted thinking, don't count any tokens
+        return 0;
+      } else if (part.type === "text") {
+        return (await encoding.encode(part.text ?? "")).length;
       }
-      return (await encoding.encode(part.text ?? "")).length;
+      return 0;
     });
     return (await Promise.all(promises)).reduce((sum, val) => sum + val, 0);
   }
@@ -106,12 +113,17 @@ function countTokens(
   const encoding = encodingForModel(modelName);
   if (Array.isArray(content)) {
     return content.reduce((acc, part) => {
-      return (
-        acc +
-        (part.type === "text"
-          ? encoding.encode(part.text ?? "", "all", []).length
-          : countImageTokens(part))
-      );
+      if (part.type === "text") {
+        return acc + encoding.encode(part.text ?? "", "all", []).length;
+      } else if (part.type === "imageUrl") {
+        return acc + countImageTokens(part);
+      } else if (part.type === "thinking") {
+        return acc + encoding.encode(part.thinking ?? "", "all", []).length;
+      } else if (part.type === "redacted_thinking") {
+        // For redacted thinking, don't count any tokens
+        return acc;
+      }
+      return acc;
     }, 0);
   } else {
     return encoding.encode(content ?? "", "all", []).length;
@@ -469,5 +481,6 @@ export {
   pruneLinesFromTop,
   pruneRawPromptFromTop,
   pruneStringFromBottom,
-  pruneStringFromTop,
+  pruneStringFromTop
 };
+
--- a/core/llm/llms/Anthropic.ts
+++ b/core/llm/llms/Anthropic.ts
@@ -1,4 +1,11 @@
-import { ChatMessage, CompletionOptions, LLMOptions } from "../../index.js";
+import {
+  ChatMessage,
+  CompletionOptions,
+  LLMOptions,
+  RedactedThinkingMessagePart,
+  TextMessagePart,
+  ThinkingMessagePart,
+} from "../../index.js";
 import { renderChatMessage, stripImages } from "../../util/messageContent.js";
 import { BaseLLM } from "../index.js";
 import { streamSse } from "../stream.js";
@@ -36,7 +43,15 @@ class Anthropic extends BaseLLM {
             name: options.toolChoice.function.name,
           }
         : undefined,
+      thinking: options.thinking,
     };
+
+    if (
+      finalOptions.thinking?.type === "disabled" &&
+      "budget_tokens" in finalOptions.thinking
+    ) {
+      delete finalOptions.thinking.budget_tokens;
+    }
 
     return finalOptions;
   }
@@ -53,15 +68,54 @@ class Anthropic extends BaseLLM {
           },
         ],
       };
-    } else if (message.role === "assistant" && message.toolCalls) {
-      return {
-        role: "assistant",
-        content: message.toolCalls.map((toolCall) => ({
+    } else if (message.role === "assistant") {
+      // Start with an empty content array
+      const content: any[] = [];
+
+      // 1. Add thinking blocks first (if present in array content)
+      if (Array.isArray(message.content)) {
+        const thinkingBlocks = message.content.filter(
+          (part) =>
+            part.type === "thinking" || part.type === "redacted_thinking",
+        );
+
+        if (thinkingBlocks.length > 0) {
+          content.push(...thinkingBlocks);
+        }
+      }
+
+      // 2. Add tool calls (if present)
+      if (message.toolCalls?.length) {
+        const toolUseBlocks = message.toolCalls.map((toolCall) => ({
           type: "tool_use",
           id: toolCall.id,
           name: toolCall.function?.name,
           input: JSON.parse(toolCall.function?.arguments || "{}"),
-        })),
+        }));
+
+        content.push(...toolUseBlocks);
+      }
+
+      // 3. Add text parts (if present)
+      if (Array.isArray(message.content)) {
+        const textBlocks = message.content.filter(
+          (part) => part.type === "text",
+        );
+        if (textBlocks.length > 0) {
+          content.push(...textBlocks);
+        }
+      } else if (typeof message.content === "string" && message.content) {
+        content.push({
+          type: "text",
+          text: message.content,
+          ...(addCaching ? { cache_control: { type: "ephemeral" } } : {}),
+        });
+      }
+
+      // Return full assistant message with combined content
+      return {
+        role: "assistant",
+        content: content.length > 0 ? content : "", // Handle empty content case
       };
     }
 
@@ -79,19 +133,34 @@ class Anthropic extends BaseLLM {
       return chatMessage;
     }
 
-    return {
-      role: message.role,
-      content: message.content.map((part, contentIdx) => {
-        if (part.type === "text") {
-          const newpart = {
-            ...part,
-            // If multiple text parts, only add cache_control to the last one
-            ...(addCaching && contentIdx == message.content.length - 1
-              ? { cache_control: { type: "ephemeral" } }
-              : {}),
-          };
-          return newpart;
-        }
+    // Filter out empty thinking blocks before mapping
+    const filteredContent = Array.isArray(message.content)
+      ? message.content.filter(
+          (part) =>
+            !(
+              part.type === "thinking" &&
+              (!part.thinking || part.thinking.trim() === "") &&
+              (!part.signature || part.signature.trim() === "")
+            ),
+        )
+      : message.content;
+
+    const convertedContent = (
+      Array.isArray(filteredContent) ? filteredContent : [filteredContent]
+    ).map((part, contentIdx) => {
+      if (part.type === "text") {
+        const newpart = {
+          ...part,
+          // If multiple text parts, only add cache_control to the last one
+          ...(addCaching &&
+          contentIdx ==
+            (Array.isArray(filteredContent) ? filteredContent.length : 1) - 1
+            ? { cache_control: { type: "ephemeral" } }
+            : {}),
+        };
+        return newpart;
+      }
+      if (part.type === "imageUrl") {
         return {
           type: "image",
           source: {
@@ -100,8 +169,32 @@ class Anthropic extends BaseLLM {
             data: part.imageUrl?.url.split(",")[1],
           },
         };
-      }),
+      }
+      if (part.type === "thinking") {
+        // Make sure to preserve the thinking and signature fields
+        return {
+          type: "thinking",
+          thinking: part.thinking,
+          signature: part.signature,
+        };
+      }
+      if (part.type === "redacted_thinking") {
+        // Make sure to preserve the data field
+        return {
+          type: "redacted_thinking",
+          data: part.data,
+        };
+      }
+      // Pass through other blocks as-is
+      return part;
+    });
+
+    const result = {
+      role: message.role,
+      content: convertedContent,
     };
+
+    return result;
   }
 
   public convertMessages(msgs: ChatMessage[]): any[] {
@@ -158,30 +251,38 @@ class Anthropic extends BaseLLM {
     );
 
     const msgs = this.convertMessages(messages);
+
+    // Merge default headers with custom headers
+    const headers: any = {
+      "Content-Type": "application/json",
+      Accept: "application/json",
+      "anthropic-version": "2023-06-01",
+      "x-api-key": this.apiKey as string,
+      ...this.requestOptions?.headers,
+    };
+
+    // Handle the special case for anthropic-beta
+    this.setBetaHeaders(headers, shouldCacheSystemMessage);
+
+    // Create the request body
+    const requestBody = {
+      ...this.convertArgs(options),
+      messages: msgs,
+      system: shouldCacheSystemMessage
+        ? [
+            {
+              type: "text",
+              text: this.systemMessage,
+              cache_control: { type: "ephemeral" },
+            },
+          ]
+        : systemMessage,
+    };
+
     const response = await this.fetch(new URL("messages", this.apiBase), {
       method: "POST",
-      headers: {
-        "Content-Type": "application/json",
-        Accept: "application/json",
-        "anthropic-version": "2023-06-01",
-        "x-api-key": this.apiKey as string,
-        ...(shouldCacheSystemMessage || this.cacheBehavior?.cacheConversation
-          ? { "anthropic-beta": "prompt-caching-2024-07-31" }
-          : {}),
-      },
-      body: JSON.stringify({
-        ...this.convertArgs(options),
-        messages: msgs,
-        system: shouldCacheSystemMessage
-          ? [
-              {
-                type: "text",
-                text: this.systemMessage,
-                cache_control: { type: "ephemeral" },
-              },
-            ]
-          : systemMessage,
-      }),
+      headers,
+      body: JSON.stringify(requestBody),
       signal,
     });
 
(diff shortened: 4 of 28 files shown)


core/llm/autodetect.ts
```diff
@@ -174,7 +174,8 @@ function autodetectTemplateType(model: string): TemplateType | undefined {
     lower.includes("pplx") ||
     lower.includes("gemini") ||
     lower.includes("grok") ||
-    lower.includes("moonshot")
+    lower.includes("moonshot") ||
+    lower.includes("deepseek-reasoner")
```
FallDownTheSystem, line 178: This is to avoid deepseek-reasoner using _streamComplete in `core/llm/index.ts` so that the ChatMessage with `content` and `reasoning_content` are both preserved.

core/llm/autodetect.ts
```diff
@@ -373,11 +374,45 @@ function autodetectPromptTemplates(
   return templates;
 }
 
+const PROVIDER_SUPPORTS_THINKING: string[] = ["anthropic", "openai", "deepseek"];
+
+const MODEL_SUPPORTS_THINKING: string[] = [
```
FallDownTheSystem, line 379: I think support for other proxy providers like OpenRouter could be added as well. I haven't looked into it.

core/llm/autodetect.ts
```diff
@@ -373,11 +374,45 @@ function autodetectPromptTemplates(
...
+  "deepseek-reasoner",
+];
+
+function modelSupportsThinking(
+  provider: string,
+  model: string,
+  title: string | undefined,
+  capabilities: ModelCapability | undefined,
+): boolean {
+  if (capabilities?.thinking !== undefined) {
```
FallDownTheSystem, line 395: I'm not sure if the capabilities: thinking is necessary. Thinking support needs to be hardcoded in some places anyway, so I don't know if there's a reasonable way to try to force it to be enabled.

core/llm/countTokens.ts
```diff
@@ -90,8 +90,15 @@ async function countTokensAsync(
     const promises = content.map(async (part) => {
       if (part.type === "imageUrl") {
         return countImageTokens(part);
+      } else if (part.type === "thinking") {
+        return (await encoding.encode(part.thinking ?? "")).length;
+      } else if (part.type === "redacted_thinking") {
+        // For redacted thinking, don't count any tokens
+        return 0;
```
FallDownTheSystem, line 97: "All extended thinking tokens (including redacted thinking tokens) are billed as output tokens and count toward your rate limits."

But they would have to be counted from the API's response:
```
"usage": {
    "input_tokens": 2095,
    "output_tokens": 503
}
  ```

File: core/llm/llms/FreeTrial.ts [hunk 5]
```diff
@@ -124,7 +125,7 @@ class FreeTrial extends BaseLLM {
       }
       return {
         type: "text",
-        text: part.text,
+        text: (part as TextMessagePart).text,
```
FallDownTheSystem, line 128: The new Message API has `thinking` and `redacted_thinking` types now, so wherever the types were causing errors, I just assumed they'd be `TextMessagePart`s as they've previously been.

core/llm/openaiTypeConverters.ts
```diff
@@ -1,21 +1,32 @@
...
-import {
-  ChatMessage,
-  CompletionOptions,
-  MessageContent,
-  TextMessagePart,
-} from "..";
+import { ChatMessage, CompletionOptions, TextMessagePart } from "..";
+
+// Extend OpenAI API types to support DeepSeek reasoning_content field
+interface DeepSeekDelta {
```
FallDownTheSystem, line 14: The types for OpenAI's messages are imported from an external library, so to support DeepSeek's `reasoning_content` I needed to create those interfaces elsewhere. I'm not sure if this is the best place for them, but it works.

core/util/messageContent.ts
```diff
@@ -17,12 +17,19 @@ export function stripImages(messageContent: MessageContent): string {
     .join("\n");
 }
 
+export function stripThinking(content: string): string {
```
FallDownTheSystem, line 20: think tags are handled differently now. They're always included in the message, but stripped from the UI

docs/docs/json-reference.md
```diff
@@ -36,6 +36,8 @@ Each model has specific configuration options tailored to its provider and funct
 - `engine`: Engine for Azure OpenAI requests.
 - `capabilities`: Override auto-detected capabilities:
   - `uploadImage`: Boolean indicating if the model supports image uploads.
+  - `tools`: Boolean indicating if the model supports tool use.
```
FallDownTheSystem: Tools was missing so I added that along with the new thinking capability

docs/docs/json-reference.md
```diff
@@ -59,6 +61,19 @@ Example:
       "title": "GPT-4o",
       "provider": "openai",
       "apiKey": "<YOUR_OPENAI_API_KEY>"
+    },
+    {
```
FallDownTheSystem, line 65: A new example showcasing a model with thinking capabilities

gui/src/components/Layout.tsx
```diff
@@ -284,6 +285,8 @@ const Layout = () => {
           />
 
           <GridDiv className="">
+            {/* Initialize model-specific settings when model changes */}
+            <ModelSettingsInitializer />
```
FallDownTheSystem, line 289: Kind of a hack. I needed the UI to fetch the `reasoning_effort` and `thinking` options on the initial load, so that they get set in the UI based on the user's config, but so that the user could still change them without changing the config. AI generated this and put them here. It worked but it might be a silly place to do something like this.

gui/src/components/StepContainer/StepContainer.tsx
```diff
@@ -92,11 +92,11 @@ export default function StepContainer(props: StepContainerProps) {
           </pre>
         ) : (
           <>
-            <Reasoning {...props}/>
+            <Reasoning {...props} />
 
             <StyledMarkdownPreview
               isRenderingInStepContainer
-              source={stripImages(props.item.message.content)}
+              source={renderChatMessage(props.item.message)}
```
FallDownTheSystem, line 99: renderChatMessage calls stripImages but now also strips think tags.

gui/src/components/mainInput/InputToolbar.tsx
```diff
@@ -106,123 +119,123 @@ function InputToolbar(props: InputToolbarProps) {
       <StyledDiv
         isHidden={props.hidden}
         onClick={props.onClick}
-        className="find-widget-skip flex"
+        className="find-widget-skip flex flex-col"
```
FallDownTheSystem, line 122: Model selection / enter button are now on their own row, to make room for the rest of the buttons.

gui/src/components/mainInput/InputToolbar.tsx
```diff
@@ -106,123 +119,123 @@ function InputToolbar(props: InputToolbarProps) {
...
+                Add context (files, docs, urls, etc.)
+              </ToolTip>
+            </HoverItem>
+          )}
+
+          <ToggleToolsButton disabled={!toolsSupported} />
+          <ToggleThinkingButton disabled={!thinkingSupported} />
+        </div>
+        <div className="-mb-1 flex w-full items-center gap-2 whitespace-nowrap">
+          <ModelSelect />
```
FallDownTheSystem, line 171: The model select now takes the full width of the remaining space in the container, meaning that there's no need to try to set a reasonable max width.

gui/src/components/mainInput/InputToolbar/PopoverNoMoveTransition.tsx
```diff
@@ -0,0 +1,26 @@
...
+  afterLeave,
+}: {
+  children: React.ReactNode;
+  show?: boolean;
+  afterLeave?: () => void;
+}) {
+  return (
+    <Transition
+      show={show}
+      enter="transition duration-100 ease-out"
```
FallDownTheSystem, line 15: This is a copy of the `PopoverTransition` component. I made the popovers on thinking and tool use toggle buttons be relative to the Chat box instead of the buttons, so that they can be easily positioned and can fit better on small view sizes, but the scale transform can't calculate the position while scaling, causing them to jump, so I removed the scaling animation from those buttons.

gui/src/components/mainInput/InputToolbar/ToggleThinkingButton.tsx
FallDownTheSystem, line 1: Basically a copy of ToggleToolsButton

gui/src/components/mainInput/InputToolbar/ToggleThinkingButton.tsx
```diff
@@ -0,0 +1,301 @@
...
+  const thinkingSettings = useAppSelector((state) => state.ui.thinkingSettings);
+  const defaultModel = useAppSelector(selectDefaultModel);
+  const [showAbove, setShowAbove] = useState(false);
+  const isInEditMode = useAppSelector(selectIsInEditMode);
+
+  const ThinkingIcon = useThinking ? LightBulbIconSolid : LightBulbIconOutline;
+
+  // Get provider from default model
+  const provider = defaultModel?.provider || "";
+  const hasThinkingOptions = provider !== "deepseek";
```
FallDownTheSystem, line 44: There are some provider/model specific logic in here that could probably be placed somewhere else. Basically some models can't be toggled off, whilst some models don't have configuration options, so certain elements/interactions are disabled based on the provider/model.

gui/src/components/mainInput/InputToolbar/ToggleToolsButton.tsx
```diff
@@ -83,93 +83,93 @@ export default function ToolDropdown(props: ToolDropdownProps) {
         {useTools && !isDisabled && (
           <>
             <span className="hidden align-top sm:flex">Tools</span>
```
FallDownTheSystem, line 85: Mostly formatting changes seemingly from Prettier. There were a few changes like the hover events are now bound on the parent container and not the icon/content, before you could hover the edges and not see the background change.

gui/src/components/modelSelection/ModelSelect.tsx
```diff
@@ -312,16 +310,16 @@ function ModelSelect() {
...
         <StyledListboxButton
           data-testid="model-select-button"
           ref={buttonRef}
           className="h-[18px] overflow-hidden"
           style={{ padding: 0 }}
           onClick={calculatePosition}
         >
-          <div className="flex max-w-[33vw] items-center gap-0.5 text-gray-400 transition-colors duration-200">
-            <span className="truncate">
+          <div className="flex w-fit min-w-0 items-center gap-0.5 text-gray-400 transition-colors duration-200">
```
FallDownTheSystem, line 319: These changes allow the model select to scale to the parent container

gui/src/pages/gui/Chat.tsx
```diff
@@ -433,54 +433,60 @@ export function Chat() {
                   contextItems={item.contextItems}
                   toolCallId={item.message.toolCallId}
                 />
```
FallDownTheSystem, line 389: This change makes sure that if the API returns a message along with tool use, that both are shown.

gui/src/redux/slices/sessionSlice.ts
```diff
@@ -315,26 +315,23 @@ export const sessionSlice = createSlice({
             parsedArgs,
           };
         }
-
         for (const message of action.payload) {
           const lastItem = state.history[state.history.length - 1];
           const lastMessage = lastItem.message;
+          // Simplified condition to keep thinking blocks and tool calls together in the same message
```
FallDownTheSystem, line 321: This is the major change. Session slice now should handle all Message API types, and keep collecting different parts to the same Assistant message, so that thinking and tool use work together properly, since Anthropic requires that you send back the thinking message along with tool use.

gui/src/redux/slices/sessionSlice.ts
```diff
@@ -358,29 +355,216 @@ export const sessionSlice = createSlice({
           } else {
             // Add to the existing message
             if (message.content) {
-              const messageContent = renderChatMessage(message);
-              if (messageContent.includes("<think>")) {
+              // Check if the message content is an array with parts
+              if (
+                Array.isArray(message.content) &&
```
FallDownTheSystem, line 360: This part is basically handling the content as parts, aka the Messages API

gui/src/redux/slices/sessionSlice.ts
```diff
@@ -358,29 +355,216 @@ export const sessionSlice = createSlice({
...
+                  }
+                }
+
+                // Only if we couldn't handle all parts, fall back to the default handling
+                if (handledAllParts) {
+                  continue;
+                }
+
+                // For other content types, use renderChatMessage
+                const messageContent = renderChatMessage(message);
```
FallDownTheSystem, line 480: This part handles what's more typical of OpenAI / DeepSeeks APIs

gui/src/redux/slices/sessionSlice.ts
```diff
@@ -358,29 +355,216 @@ export const sessionSlice = createSlice({
...
+                }
+
+                // Add the message content to the existing content
+                lastMessage.content += messageContent;
+
+                // Current full content
+                const fullContent = lastMessage.content as string;
+
+                // If we find <think> tags, extract the content for the reasoning field
+                if (
```
FallDownTheSystem, line 506: Lastly we handle think tags

core/llm/llms/Anthropic.ts
```diff
@@ -79,19 +133,34 @@ class Anthropic extends BaseLLM {
...
-            ...part,
-            // If multiple text parts, only add cache_control to the last one
-            ...(addCaching && contentIdx == message.content.length - 1
-              ? { cache_control: { type: "ephemeral" } }
-              : {}),
-          };
-          return newpart;
-        }
+    // Filter out empty thinking blocks before mapping
+    const filteredContent = Array.isArray(message.content)
```
FallDownTheSystem: I'm not a 100% sure if this is still needed. It was something AI added as an attempted fix, but I don't think we should be getting empty thinking messages from the API.

core/llm/llms/Anthropic.ts
```diff
@@ -158,30 +251,38 @@ class Anthropic extends BaseLLM {
...
+    const headers: any = {
+      "Content-Type": "application/json",
+      Accept: "application/json",
+      "anthropic-version": "2023-06-01",
+      "x-api-key": this.apiKey as string,
+      ...this.requestOptions?.headers,
+    };
+
+    // Handle the special case for anthropic-beta
+    this.setBetaHeaders(headers, shouldCacheSystemMessage);
```
FallDownTheSystem, line 253: This change allows the config to add headers to the request and intelligently merge them with beta headers that continue adds for caching.

FallDownTheSystem:
Resolves #4339

core/llm/llms/Anthropic.ts
FallDownTheSystem: Removed this change

core/index.d.ts
```diff
@@ -360,6 +375,7 @@ export interface UserChatMessage {
 export interface AssistantChatMessage {
   role: "assistant";
   content: MessageContent;
+  reasoning_content?: string;
```
Patrick-Erichsen, line 378:
```suggestion
  reasoningContent?: string;
```

Patrick-Erichsen: After going through this PR I realized that these properties are `snake_case` to match what the Anthropic API is expecting - however, I'd still prefer to keep them `camelCase` in our TS interfaces, both here and elsewhere in the PR.

core/index.d.ts
```diff
@@ -921,11 +937,17 @@ export interface BaseCompletionOptions {
   prediction?: Prediction;
   tools?: Tool[];
   toolChoice?: ToolChoice;
+  thinking?: {
+    type: "enabled" | "disabled";
+    budget_tokens?: number;
+  };
+  reasoning_effort?: "high" | "medium" | "low";
```
Patrick-Erichsen, line 944: Could we make this into an enum instead? Bit of a nitpick but will make any future refactors easier.

core/llm/autodetect.ts
```diff
@@ -373,11 +374,45 @@ function autodetectPromptTemplates(
   return templates;
 }
 
+const PROVIDER_SUPPORTS_THINKING: string[] = ["anthropic", "openai", "deepseek"];
```
Patrick-Erichsen, line 377: Could we pull this logic out into a separate file, e.g. `core/llm/thinking.ts`? Bit of a nitpick/we should have done this with other parts of this file, but trying to tidy up the codebase more recently.

core/llm/autodetect.ts
Patrick-Erichsen: Yeah I think your intuition here is accurate. It's not quite as simple as image uploads. 

If you're reasonably confident, I would be onboard to remove it.

core/llm/countTokens.ts
Patrick-Erichsen: I'm a little confused here - mind linking to the docs you're referencing? 
Is your point that `redacted_thinking` is actually billed, but we're unable to properly count it here since this is for input tokens, not output?

core/index.d.ts
```diff
@@ -921,11 +937,17 @@ export interface BaseCompletionOptions {
   prediction?: Prediction;
   tools?: Tool[];
   toolChoice?: ToolChoice;
+  thinking?: {
+    type: "enabled" | "disabled";
+    budget_tokens?: number;
```
Patrick-Erichsen, line 942: Is this supposed to be the budget for the number of tokens to allocate to thinking? If so I think a more declarative `tokenBudget` would be clearer, took me a second to understand.

core/llm/autodetect.ts
```diff
@@ -373,11 +374,45 @@ function autodetectPromptTemplates(
...
+
+const MODEL_SUPPORTS_THINKING: string[] = [
+  "claude-3-7-sonnet-20250219",
+  "claude-3-7-sonnet-latest",
```
Patrick-Erichsen, line 381: This didn't work for me despite using https://hub.continue.dev/anthropic/claude-3-7-sonnet?view=config which has `    model: claude-3-7-sonnet-latest` - could you try as well?

core/llm/llms/Anthropic.ts
```diff
@@ -53,15 +68,54 @@ class Anthropic extends BaseLLM {
           },
         ],
       };
-    } else if (message.role === "assistant" && message.toolCalls) {
-      return {
-        role: "assistant",
-        content: message.toolCalls.map((toolCall) => ({
+    } else if (message.role === "assistant") {
```
Patrick-Erichsen, line 71: This function is quite large - could we pull it out into a util and break it down to make it easier to read/test? Again, another scenario where the surrounding code isn't particularly clean, but trying to keep things more maintainable going forwards

docs/docs/json-reference.md
Patrick-Erichsen: Appreciate the docs update 👌

docs/docs/yaml-reference.md
```diff
@@ -391,6 +395,18 @@ models:
       temperature: 0.3
       stop:
         - "\n"
+        
+  - name: Claude 3.7 Sonnet
```
Patrick-Erichsen, line 399: Reminder to myself to update the default Claude 3.7 Sonnet block on the Hub with this same YAML

gui/src/components/Layout.tsx
Patrick-Erichsen: Thanks for the explanation!

I think what we should do is just call `useModelThinkingSettings();` from `Chat.tsx`. I think it should achieve the same outcome without adding an unnecessary UI component here.

gui/src/components/mainInput/InputToolbar.tsx
Patrick-Erichsen: This is a good idea, the toolbar was getting quite cluttered already - but, @sestinj has an incoming PR that adds a new "notch" above the input, similar to what we have in Edit mode, to address this. Amongst other changes it moves the "Tool usage" button up there. So let's keep it as is without the `flex-col`.

gui/src/components/mainInput/InputToolbar.tsx
Patrick-Erichsen: I think this is also unnecessary based on my comment above

gui/src/components/mainInput/InputToolbar/ToggleThinkingButton.tsx
```diff
@@ -0,0 +1,301 @@
...
+    if (listboxButtonRef.current) {
+      listboxButtonRef.current.click();
+    }
+  };
+
+  // Get OpenAI icon color based on reasoning effort
+  const getIconColor = () => {
+    if (provider === "openai" && useThinking && !props.disabled) {
+      const effort = thinkingSettings.openai.reasoningEffort;
+      if (effort === "low") return "text-red-400";
```
Patrick-Erichsen, line 97: Chatted with the team on the colors here - I think we should just use `text-lightgray` in all cases to avoid the somewhat unnecessary color indicators.

gui/src/components/mainInput/InputToolbar/ToggleThinkingButton.tsx
```diff
@@ -0,0 +1,301 @@
...
+                                  reasoning
+                                </p>
+                                {provider === "anthropic" && (
+                                  <p>
+                                    Budget tokens: How many tokens to dedicate
+                                    to thinking
+                                  </p>
+                                )}
+                                {provider === "openai" && (
+                                  <p>
```
Patrick-Erichsen, line 216: I think the "Reasoning effort: Amount of effort to put into reasoning" is redundant and can be removed

gui/src/components/mainInput/InputToolbar/ToggleThinkingButton.tsx
```diff
@@ -0,0 +1,301 @@
...
+                                      setOpenAIReasoningEffort(
+                                        level as "low" | "medium" | "high",
+                                      ),
+                                    );
+                                    closeDropdown();
+                                    setIsHovered(false);
+                                  }}
+                                >
+                                  <div
+                                    className={`h-2 w-2 rounded-full ${level === "low" ? "bg-red-400" : level === "medium" ? "bg-yellow-400" : "bg-green-400"} mr-2`}
```
Patrick-Erichsen, line 280: Similarly here, let's just use `text-lightgray`

gui/src/components/mainInput/InputToolbar/ToggleThinkingButton.tsx
Patrick-Erichsen: For models that can't turn off reasoning, I think the current implementation feels a bit noisy.  I'm not sure what the solution here is though - maybe with @sestinj 's new notch UI we have some more options.

gui/src/hooks/useModelThinkingSettings.ts
```diff
@@ -0,0 +1,69 @@
...
+import { useAppDispatch, useAppSelector } from '../redux/hooks';
+import { selectDefaultModel } from '../redux/slices/configSlice';
+import { setAnthropicBudgetTokens, setOpenAIReasoningEffort, setUseThinking } from '../redux/slices/uiSlice';
+
+/**
+ * This hook initializes thinking settings when the model changes.
+ * It runs at a high level in the component tree to ensure model-specific 
+ * settings are loaded only once per model selection.
+ */
+export const useModelThinkingSettings = () => {
```
Patrick-Erichsen, line 11: All of this logic should probably live in a thunk that responds to changes in the `config.defaultModelTitle` property. Then we can get rid of both this hook and the UI component that triggers it.

gui/src/redux/slices/uiSlice.ts
```diff
@@ -21,6 +21,16 @@ type UIState = {
...
   shouldAddFileForEditing: boolean;
   useTools: boolean;
+  useThinking: boolean; // New toggle for thinking
+  thinkingSettings: {
+    // Settings for different providers
+    anthropic: {
+      budgetTokens: number; // Min 1024, max is below maxTokens
+    };
+    openai: {
+      reasoningEffort: "low" | "medium" | "high";
```
Patrick-Erichsen, line 31: We should probably re-use the types from `core` here

Additionally, I don't think we want to be too provider specific here. If there a way we could structure this instead to just have both `budgetTokens` and `reasoningEffort` as options, or some other approach that isn't opinionated/structured according to particular providers?

gui/src/redux/slices/sessionSlice.ts
Patrick-Erichsen: This all looks solid but there is just so much logic here that I think we really need to break this out into some utils that are smaller/easier to read/more testable. Again, the code around isn't the cleanest, but it would be really helpful for maintenance/debugging.

Patrick-Erichsen (changes requested):
Hey @FallDownTheSystem , apologies for the slow review here. 

This is an awesome contribution! I've been curious to play around with different thinking efforts but haven't been able to since we're missing this feature, so it was cool to finally get to try it out while testing.

Things look solid overall but my main concern with merging this is that the logic it touches it quite sensitive, and there is a lot of new logic being introduced here that is in quite large functions. I mentioned this in a few comments but the existing codebase in these areas wasn't that clean to begin with, but I think given that nobody on the team will be familiar with this new logic, it's important to make sure it's especially maintainable.

Lastly, I'd recommend just ignoring the merge conflicts any UI changes for now. There will be more conflicts occur because of @sestinj 's upcoming PR, and I think we'll want to update the UI in this PR based on those changes.

Feel free to DM me/Nate on Discord if you'd prefer to go back and forth more easily 👍

Patrick-Erichsen:
Also, I'd recommend just ignoring the merge conflicts for now. There will be more that occur because of @sestinj 's upcoming PR, and after that goes through, we can probably just resolve these conflicts for you.

FallDownTheSystem:
I'll close this PR for now. https://github.com/continuedev/continue/pull/4426 added support for showing thinking tokens already so half of this PR would conflict with that, and now that the Chat UI has been largely reworked, the thinking UI settings need to be reworked as well to fit the new UI.

I may do so in the future, but I'll be pretty unavailable for the next few weeks. If someone from the team wants to pick up whatever might be useable of the UI portion, feel free to do so. For now I'm happy thinking tokens are supported in the UI.

I'll open another PR just to allow custom headers to be set for Anthropic so that we can use the beta headers for 128k output tokens.

FallDownTheSystem:
See https://github.com/continuedev/continue/pull/4785 for the Anthropic headers.

Patrick-Erichsen:
Got it, that makes total sense. This will be a great base for us to go off of so thanks for doing all the hard work!

Issue #3106, opened 2024-11-27
Add a request rate limit for embeddings model


lakshits11:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 0.8.57
- IDE version: VS Codium
- Model: Mistral Embed API
```

### Description

So my embeddings settings look something like this:

config.ts
```ts
mistral_embeddings: {
    model: "mistral-embed",
    apiKey: apiKeys.MISTRAL,
    maxChunkSize: 2048,
    maxBatchSize: 16,
    requestOptions: {
      timeout: 1001
    },
    provider: "mistral",
    apiBase: "https://api.mistral.ai/v1"
  },
```
  
  When using this configuration to generate embeddings, I get this console msg:
  ```log
  [Extension Host] error when indexing:  Error: Failed to generate embeddings for 129 chunks with provider: MistralEmbeddingsProvider::mistral-embed::2048: Error: {"message":"Requests rate limit exceeded"}
	at LanceDbIndex.getEmbeddings (c:\Users\Lakshit\.vscode-oss\extensions\continue.continue-0.8.57-win32-x64\out\extension.js:175314:17)
	at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
	at async LanceDbIndex.computeRows (c:\Users\Lakshit\.vscode-oss\extensions\continue.continue-0.8.57-win32-x64\out\extension.js:175262:28)
	at async LanceDbIndex.update (c:\Users\Lakshit\.vscode-oss\extensions\continue.continue-0.8.57-win32-x64\out\extension.js:175418:24)
	at async CodebaseIndexer.indexFiles (c:\Users\Lakshit\.vscode-oss\extensions\continue.continue-0.8.57-win32-x64\out\extension.js:540456:39)
	at async CodebaseIndexer.refresh (c:\Users\Lakshit\.vscode-oss\extensions\continue.continue-0.8.57-win32-x64\out\extension.js:540328:30)
[...]

PR #4581, opened 2025-03-10
Add missing custom fetch use cases

RomneyDa:
## Description
- Some uses of fetch were not using the custom fetch that accounts for request options, etc.
- HuggingFaceTEI had nested `withExponentialBackoff`

Effects
- All OpenAI and OpenAI-compatible `embed` and `autocomplete` calls that use the OpenAI adapter
- Continue Proxy rerank
- Voyage rerank
- HuggingFaceTEI rerank and embed


RomneyDa:
Probably fixes https://github.com/continuedev/continue/issues/3106

PR #4586, opened 2025-03-11
fix(intellij): IdeProtocalClient handleMessage subprocess

nangua:
## Description

Logs output:
```plaintext
Unknown message type: subprocess
```

## Checklist

## Screenshots

Not involved.

## Testing instructions

Not involved.

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IdeProtocolClient.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IdeProtocolClient.kt
@@ -526,6 +526,16 @@ class IdeProtocolClient(
                         respond(null)
                     }
 
+                    "subprocess" -> {
+                        val params = Gson().fromJson(
+                            dataElement.toString(),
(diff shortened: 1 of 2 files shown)


sestinj:
@nangua I think we are going to probably deprecate this message soon along with config.ts. I assume you are using `config.ts`, but could you share what your use case is for "subprocess"?

nangua:
@sestinj Yes, I am indeed using `config. ts`，I read the official doc [Custom Context Providers with Submenu or Query](https://docs.continue.dev/customize/tutorials/build-your-own-context-provider#custom-context-providers-with-submenu-or-query), which claims that `ide. subprocess` can be used in the `Custom Context Provider`. However, I happened to need to use some lightweight MCP like logic in the provider that also requires interaction with the operating system through `ide. subprocess`, so I used it.

For example, I need to analyze the risk of dependency components in Maven projects. In order to reduce tokens, I hope to first generate a list of dependency components for the entire project through `mvn dependency:list` in the `Custom Context Provider`, and then parse it into a compact string in `getContextitems()`. Although there may be other more elegant solutions, I believe the current approach is the most convenient.

PR #4587, opened 2025-03-11
Walkdir caching first pass

RomneyDa:
## Description
Cache listDir results and ignore file rectified ignores for directories
- for 30 seconds OR
- till invalidated
 
Any file changes, creations, deletions, or index requests will invalidate the cache, so the _only_ effect this has for now is that parallel walk dirs will be much more efficient. It will not speedup e.g. walkdir for a single instance of codebase reindexing.

Before:
image

After:
image

Startup CPU load in `react` repo before/after:
image

### DIFF
--- a/core/indexing/walkDir.ts
+++ b/core/indexing/walkDir.ts
@@ -12,9 +12,13 @@ import {
 export interface WalkerOptions {
   onlyDirs?: boolean;
   returnRelativeUrisPaths?: boolean;
+  source?: string;
 }
 
 type Entry = [string, FileType];
+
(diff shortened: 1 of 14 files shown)


core/indexing/walkDir.ts
```diff
@@ -103,19 +198,24 @@ class DFSWalker {
             continue;
           }
         }
-        let shouldInclude = true;
+        let shouldIgnore = false;
         for (const ig of ignoreContexts) {
+          if (shouldIgnore) {
+            continue;
```
RomneyDa, line 205: This is another improvement - if one ignore comes out as true, skip the rest.

Issue #1364, opened 2024-05-26
The contextLength setting is not effective


kit1858644:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 12.7.4
- Continue: 0.8.31
- IDE: Vscode
Model: gpt-4o
```

### Description

I set the contextLength to 4096 in the config (as shown below), but the post request sent by continue is 1024:

continue config:
>     {
      "title": "GPT-4o",
      "model": "gpt-4o",
      "provider": "openai",
      "contextLength": 4096,
      "apiKey": "sk-xxx",
      "apiBase": "https://api.xxx.com/v1",
      "systemMessage": "You are an expert software developer. You give helpful and concise responses."
    }

continue post request:

> {
  "messages": [
    {
      "role": "system",
      "content": "You are an expert software developer. You give helpful and concise responses."
    },
    {
      "role": "user",
      "content": "hello"
    }
  ],
  "model": "gpt-4o",
  "max_tokens": 1024,
  "stream": true
}

sealad886:
Can confirm this happens with `llama.cpp` as well.

config file:
```JSON
    {
      "title": "Codellama-34b Instruct (llama.cpp)",
      "provider": "llama.cpp",
      "model": "models/quants/CodeLlama-34b-Instruct-q4_k.gguf",
      "apiBase": "http://localhost:8080",
      "contextLength": 4096,  # <-- this should get passed as n_ctx into llama.cpp but doesn't for some reason
      "completionOptions": {
        "temperature": 0.2,
        "mirostat": 2
      }
    },
```

output from `/.server <with options>`:
```shell
{"tid":"0x1f9303ac0","timestamp":1716850340,"level":"INFO","function":"update_slots","line":1851,"msg":"slot context shift","id_slot":0,"id_task":1026,"n_keep":0,"n_left":511,"n_discard":255,"n_ctx":512,"n_past":511,"n_system_tokens":0,"n_cache_tokens":0}
[...]

PR #4602, opened 2025-03-11
Added contextLength attribute to LLMOptions which was missing

Jazzcort:
## Description

The contextLength attribute exists under both the completionOptions and defaultCompletionOptions attributes, but not at the root level. As a result, the reference in core/llm/llms/Ollama.ts is undefined. To fix this, we simply copied the contextLength from defaultCompletionOptions to the root attribute of LLMOptions.

## Testing instructions

Edit the value of the contextLength attribute under the defaultCompletionOptions section in the config.yaml file. These changes should be reflected in both the 'Continue - LLM Prompt/Completion' output channel and the --ctx-size when checking the information from the Ollama runner.


owtaylor:
To clarify the bug here: the problem is that with config.yaml `contextLength` moved inside `model.defaultCompletionOptions`, but that wasn't handled when converting the YAML config to LLMOptions; it ended up being entirely ignored.

Patrick-Erichsen (dismissed):
Thanks for the contribution here @Jazzcort ! It looks like a merge conflict popped up, if you don't mind resolving that we can get this merged afterwards 👍

Patrick-Erichsen (approved):

bitbottrap:
This looks like the better place for this command. Using 1.1.12 on Linux VSCode today and this does not appear to be fixed. #4602 was merged into pre-release a few builds ago. My yaml entry for the model:

```
- name: QwQ 32B
   provider: llama.cpp
   model: qwq-32b
   apiBase: http://sv-ai:8081/
   requestOptions:
     headers:
       X-Auth-Token: NONE
   contextLength: 100000
   defaultCompletionOptions:
     contextLength: 100000
     maxTokens: 8192
   roles:
     - chat
     - edit
     - apply
```

Wasn't sure where contextLength should go as I've seen it in these two places so I put it in both.

Get the same error as before: "maxTokens (8192) is too close to contextLength (8096), which doesn't leave room for response. Try increasing the contextLength parameter of the model in your config.json."

Jazzcort:
I might be wrong, but it seems like there are some structural error in your config.yaml file. Check out the reference here: https://docs.continue.dev/yaml-reference.

LLukas22:
Can confirm that this is still broken, the contextLength is not applied.

Jazzcort:
Can you show your config.yaml file? So I can try to reproduce your case.

bitbottrap:
```
name: my-configuration
version: 0.0.1
schema: v1

models:
  - name: QwQ 32B
    provider: llama.cpp
    model: qwq-32b
    apiBase: http://sv-ai:8081/
    requestOptions:
      headers:
        X-Auth-Token: NONE
    defaultCompletionOptions:
      contextLength: 100000
      maxTokens: 4096
    roles:
      - chat
      - edit
      - apply
```

This edit to the model section, reducing the maxTokens works. Increasing maxTokens to anything ~8k+ makes it fail.

If you say I have other errors, I believe you. The yaml documentation for models doesn't appear to be complete ... so some of this is by looking around. But I can definitely say that this is giving me a problem.

LLukas22:
Same issue when im trying to use openrouter: 

```
- name: reka
  provider: openai
  model: rekaai/reka-flash-3:free
  apiKey: "[KEY]"
  apiBase: https://openrouter.ai/api/v1
  defaultCompletionOptions:
    contextLength: 65536
    maxTokens: 12000
  roles:
    - chat
    - edit
    - apply
```

Jazzcort:
Some backend providers, such as OpenRouter and Llama.cpp, do not reflect changes to the `contextLength` in the `config.yaml` file because the context length can only be set on the backend side for these providers. The fix implemented here addresses the issue where the `contextLength` attribute was missing when parsing the `config.yaml` file into `LLMOptions`, allowing Ollama to correctly detect changes in the context length.

Given this, I suggest closing this issue and opening a new one to address the fact that some backend providers do not support changing the context length through external requests. Additionally, it would be helpful for the "Continue" feature to provide a heads-up when users attempt to set the context length for providers that don't support it.

bitbottrap:
So where is the chat interface getting the 8096 contextLength value from? It seems like it the value from config.yaml should be returned instead of 8096? (The config.yaml value would then just be advisory in nature and not go through to all LLM APIs.) Or, if the value isn't specified it should not be checked by the chat interface.
Instituting the check in the chat interface in the first place was probably not the best way forward as it broke several model providers.
There are still several modes of failure here that would need to be addressed and I'm not even sure how to open an issue to describe the fix.

RomneyDa:
@Jazzcort reinstated your fix here with you as commit author https://github.com/continuedev/continue/pull/4871
@bitbottrap 8096 is the safe default context length if it's not set and not found in our `llm-info` package

George-Const:
> Some backend providers, such as OpenRouter and Llama.cpp, do not reflect changes to the `contextLength` in the `config.yaml` file because the context length can only be set on the backend side for these providers. The fix implemented here addresses the issue where the `contextLength` attribute was missing when parsing the `config.yaml` file into `LLMOptions`, allowing Ollama to correctly detect changes in the context length.
> 
> Given this, I suggest closing this issue and opening a new one to address the fact that some backend providers do not support changing the context length through external requests. Additionally, it would be helpful for the "Continue" feature to provide a heads-up when users attempt to set the context length for providers that don't support it.

Are those models hardcoded or the ones which allow the changing the context length through external requests in Continue? I am asking because I am using a model in ollama, mistral-small3.5:latest that gives the issue but using curl or others it works setting up larger context length through external request.

George-Const:
> So where is the chat interface getting the 8096 contextLength value from? It seems like it the value from config.yaml should be returned instead of 8096? (The config.yaml value would then just be advisory in nature and not go through to all LLM APIs.) Or, if the value isn't specified it should not be checked by the chat interface. Instituting the check in the chat interface in the first place was probably not the best way forward as it broke several model providers. There are still several modes of failure here that would need to be addressed and I'm not even sure how to open an issue to describe the fix.

I tried figuring it out here: https://github.com/continuedev/continue/issues/1364#issuecomment-2799910883

Jazzcort:
#4602 was somehow reverted and later brought back through #4929 which was merged on April 1st, which is later than the release date of version 1.0.5 (Mar. 27th). The later version should have this issue fixed.

PR #4621, opened 2025-03-12
feat: check for `nvm` version in install script

Patrick-Erichsen:


owtaylor (commented):
This seems like a good idea to me

scripts/install-dependencies.sh
```diff
@@ -5,6 +5,23 @@
...
+    
+    # Remove 'v' prefix from versions for comparison
+    required_version=${required_node_version#v}
+    current_version=${current_node_version#v}
+
+    if [ "$required_version" != "$current_version" ]; then
+        echo "⚠️  Warning: Your Node.js version ($current_node_version) does not match the required version ($required_node_version)"
+        echo "Please consider switching to the correct version using: nvm use"
+        echo "Continuing with installation anyway..."
+        echo
```
owtaylor, line 26: This could easily scroll off, maybe something like:

```
if [ -t 0 ] ; then
     read "Hit return to continue "
else
     echo "Continuing with installation anyway..."
fi
```

No idea how to do that in Powershell...

Patrick-Erichsen: Not sure either but this was what 3.5 Sonnet generated 🤷 

```ps1
if (Test-Path ".nvmrc") {
    $requiredNodeVersion = Get-Content ".nvmrc"
    $currentNodeVersion = node -v

    # Remove 'v' prefix from versions for comparison
    $requiredVersion = $requiredNodeVersion.TrimStart('v')
    $currentVersion = $currentNodeVersion.TrimStart('v')

    if ($requiredVersion -ne $currentVersion) {
        Write-Host "`n⚠️  Warning: Your Node.js version ($currentNodeVersion) does not match the required version ($requiredNodeVersion)" -ForegroundColor Yellow
        Write-Host "Please consider switching to the correct version using: nvm use" -ForegroundColor Yellow
        
        # Check if running in interactive mode
        if ([Environment]::UserInteractive -and [Environment]::GetCommandLineArgs().Count -eq 0) {
            Write-Host "Press Enter to continue with installation anyway..." -NoNewline -ForegroundColor Yellow
            $null = $Host.UI.RawUI.ReadKey("NoEcho,IncludeKeyDown")
            Write-Host "`n" # Add newline after key press
        } else {
            Write-Host "Continuing with installation anyway...`n" -ForegroundColor Yellow
        }
    }
}
```

Our CI will at least catch if there are any issues with the install script in a non-interactive mode

RomneyDa (approved):

scripts/install-dependencies.sh
walterCui: Is it possible to add arch detection for nodejs?

Issue #4475, opened 2025-03-04
Development data collection not working


hawkeye1121:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows11
- Continue version: v1.0.2
- IDE version: VSCode 1.97.2
- Model: Ollama llama3.3 / qwen2.5-coder:7b
- config: (only s
  
data:
name: My Config
version: 1.0.0
schema: v1

models:

  - name: Ollama Chat (Remote)
    provider: ollama
    model: AUTODETECT
    roles:
      - chat
      - edit
      - apply
      - summarize
    apiBase: http://localhost:11434
    contextLength: 12288

  - name: Ollama Autocomplete (Remote)
    provider: ollama
    model: qwen2.5-coder:7b
    roles:
      - autocomplete
    apiBase: http://localhost:11434

  - name: Ollama Embeddings (Remote)
    provider: ollama
    model: nomic-embed-text
    roles:
      - embed
    apiBase: http://localhost:11434

rules:
  - Always give concise responses

context:
  - provider: codebase
  - provider: currentFile

data:
  - name: ContinueLogging
    destination: file:///D:/temp/
    # destination: http://127.0.0.1:5000/data
    schema: 0.2.0
```

### Description

I am trying to utilize the feature for collection of development data to track how/when I use Continue.  I expect to see chat, quickEdit, and autocomplete interactions stored to a json file in the directory specified in config.yaml.  When I perform each of these three interaction types, only the chat is added to D:/temp/0.2.0/chatInteraction.jsonl.  There is no autocomplete or quickEdit json file.

When I revert to Continue 0.8.68, I see an autocomplete.jsonl file updated, but the quickEdit.jsonl and 0.2.0/chatInteraction.jsonl files are not updated.

### To reproduce

1. Setup Continue v1.0.2
2. Perform new chat, quick edit, and autocomplete actions
3. Navigate to development data directory and observe quick edit and autocomplete actions are not logged

1. Setup Continue v0.8.68
2. Perform new chat, quick edit, and autocomplete actions
[...]

Issue #4340, opened 2025-02-25
Autocomplete log doesn't record correctly when using JetBrains IDE.


glm1024:
### Before submitting your bug report

- [ ] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:Windows
- Continue version: 0.0.88
- IDE version: JetBrains IDEA 2024.1.7(Community Edition)
- Model: Qwen-Coder-7b
- config:
```

### Description

I found that: the file .continue/logs/prompt.log doesn't record the autocompletion result when using continue with IDEA, VsCode works fine.

### Log output

```Shell

```

Patrick-Erichsen:
Good catch here @glm1024 !

amos42:
@glm1024  I tracked down the problem and found the cause.
Check "jetbranis" as the IDE identifier, but in reality "JETBRAINS" is transmitted, so correct this.
https://github.com/continuedev/continue/pull/4633

PR #4633, opened 2025-03-13
Fixed an issue where the Log data collection feature in IntelliJ would fail (#4340 #4475)

amos42:
Fixed an issue where the Log data collection feature in IntelliJ would fail.
Issue : #4340 #4475

## Description

1. Check "jetbranis" as the IDE identifier, but in reality "JETBRAINS" is transmitted, so correct this. (#4340)
2. Among the fields of the event, the type of timestamp is different between autocomplete and other events, so if they are processed in common, the zod parser fails to check the type. This has been fixed. (#4475)

## Checklist

## Testing instructions

1. Add the following to config.json.
```
  "data": [
    {
      "schema": "0.2.0",
      "destination": "file://c:/tmp/continue_logs/"
    }
  ]
```
2. Enter a chat or activate the autocomplete function.
3. Check if the log files are created normally in the c:/tmp/continue_logs/ directory.

### DIFF
--- a/core/autocomplete/CompletionProvider.ts
+++ b/core/autocomplete/CompletionProvider.ts
@@ -262,7 +262,7 @@ export class CompletionProvider {
 
       // When using the JetBrains extension, Mark as displayed
       const ideType = (await this.ide.getIdeInfo()).ideType;
-      if (ideType === "jetbrains") {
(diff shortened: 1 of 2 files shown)


sestinj (approved):
thanks for the fix @amos42

Issue #4595, opened 2025-03-11
mcp server: Failed to call tool


reigeo:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: 1.0.3
- IDE version: 1.98.1
- Model: llama3.1:latest llama
- config:
  
{
  "experimental": {
    "enabled": true,
    "modelContextProtocolServers": [
      {
        "transport": {
          "type": "stdio",
          "command": "docker",
          "args": ["run", "-i", "--rm", "mcp/fetch"]
        }  
      }
    ]
  },
  "contextProviders": [
    {
      "name": "http",
      "params": {
        "url": "http://localhost:8000/retrieve",
        "title": "http",
        "description": "Custom HTTP Context Provider",
        "displayTitle": "myDocs"
      }
    }
  ],
  "models": [
    {
      "title": "llama",
      "provider": "ollama",
      "model": "llama3.1:latest"
    }
  ]
}
```

### Description

while trying to get MCP up and running within continue I tried (see config) to integrate smaple fetch mcp server called via docker.
also other MCP servers from public sample repo as well as self written servers do not work. All tool calls fail with the error:

Failed to call tool fetch: Failed to call tool: fetch

 ERR [Extension Host] Error handling webview message: {
  "msg": {
    "messageId": "857dcdcb-5f95-43f2-9241-09c1561ff95c",
    "messageType": "tools/call",
    "data": {
      "toolCall": {
        "id": "",
        "type": "function",
        "function": {
          "name": "fetch",
          "arguments": "{\"url\":\"google.com\"}"
        }
      },
      "selectedModelTitle": "AXAI"
    }
  }
}

Error: Failed to call tool: fetch

### To reproduce

add mcp server to your config 
try to call tool

### Log output

```Shell

```

tomasz-stefaniak:
@reigeo thanks for bringing this up, we're working on a bunch of tool fixes this week.

reigeo:
[...]

PR #4636, opened 2025-03-13
4595 remove parameter shadowing

reigeo:
## Description

fix for issue:  https://github.com/continuedev/continue/issues/4595 

renaming parameters to remove parameter shadowing which lead to confusion while tool call

## Checklist

## Screenshots

## Testing instructions

### DIFF
--- a/core/tools/callTool.ts
+++ b/core/tools/callTool.ts
@@ -53,7 +53,7 @@ export function decodeMCPToolUri(uri: string): [string, string] | null {
 
 async function callToolFromUri(
   uri: string,
-  args: any,
+  uriArgs: any,
   extras: ToolExtras,
 ): Promise<ContextItem[]> {
   const parseable = canParseUrl(uri);
@@ -65,7 +65,7 @@ async function callToolFromUri(
   switch (parsedUri?.protocol) {
     case "http:":
     case "https:":
-      return callHttpTool(uri, args, extras);
+      return callHttpTool(uri, uriArgs, extras);
     case "mcp:":
       const decoded = decodeMCPToolUri(uri);
       if (!decoded) {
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
@reigeo seems like this is just a rename? What does this change?

reigeo:
@RomneyDa the The issue is likely related to closure variable capture which gets fixed by this renaming. By using distinct names I eliminated any ambiguity about which variable is being referenced. Each parameter now has a unique name throughout the call chain, preventing any confusion about which value should be used.

RomneyDa:
@reigeo can you test this and show the before/fail and after/pass with a video or something?

reigeo:
here the super short version ;) 

https://github.com/user-attachments/assets/be54a491-1a87-46c5-becf-a89bbc3b1970

RomneyDa:
@reigeo on further review there is no closure on the word args that would be the cause of this issue. I didn't fully understand the video you sent, can you describe what it was trying to convey?

reigeo:
@RomneyDa I try to explain more detailed the problem and why renaming args fix it:

Renaming the parameters helps to avoid issues with variable shadowing and ensures that the correct values are passed to the functions. In JavaScript and TypeScript, if a variable is declared in an outer scope and a variable with the same name is declared in an inner scope, the inner variable shadows the outer one. This can lead to unexpected behavior, such as the outer variable being undefined when accessed within the inner scope.

I know the video is cut down as much as possible but you see two debug sessions. The first one shows distinct set of parameters, which prevents any accidental shadowing and makes the code easier to understand and maintain. So you see with the renamed ags, the value is correctly set, and gets forwarded to the mcp server as expected. Then you see that I revert my changes, meaning back to each functions having same args naming, you see that args is not set in that context, so there are no arguments that can be handed over to the mcp server which is the reason for the error shown at the end of the video.

RomneyDa:
@reigeo took a closer look with @Patrick-Erichsen and we're fairly sure based on our understanding of parameter shadowing that this won't be the source of the issue in this case. The video shows debugging at two different points and is not conclusive that the rename changes anything. Feel free to provide more details or make another case, will close this for now.

PR #4658, opened 2025-03-14
Edit tool

RomneyDa:
## Description

Approach: 
- Use the built in chat generate/apply
- If it's an edit tool response, auto apply


RomneyDa:
reworked here https://github.com/continuedev/continue/pull/4978

PR #4666, opened 2025-03-15
fix: Eliminate Uncaught TypeError in dev tools

lkk214:
## Description

Eliminate the `Uncaught TypeError: window.cefQuery_xxx is not a function` error that keeps appearing in the JS debug console.

## Screenshots

err0315

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt
@@ -11,7 +11,7 @@ import com.intellij.openapi.wm.ToolWindowFactory
 import com.intellij.ui.content.ContentFactory
 import javax.swing.*
 
-const val JS_QUERY_POOL_SIZE = "200"
+const val JS_QUERY_POOL_SIZE = 200
 
 class ContinuePluginToolWindowFactory : ToolWindowFactory, DumbAware {
   override fun createToolWindowContent(project: Project, toolWindow: ToolWindow) {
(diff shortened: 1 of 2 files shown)


sestinj (approved):
this looks great @lkk214 — can we also keep around the dev tools option by adding it to the apply?

lkk214:
@sestinj JS_QUERY_POOL_SIZE? It is just a configuration parameter in jbcef, and maybe it can be configured in the tools settings in JetBrains, but I don't think users will care about it.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/toolWindow/ContinuePluginToolWindowFactory.kt
```diff
@@ -43,11 +43,6 @@ class ContinuePluginToolWindowFactory : ToolWindowFactory, DumbAware {
   class ContinuePluginWindow(project: Project) {
     private val defaultGUIUrl = "http://continue/index.html"
 
-    init {
-      System.setProperty("ide.browser.jcef.jsQueryPoolSize", JS_QUERY_POOL_SIZE)
-      System.setProperty("ide.browser.jcef.contextMenu.devTools.enabled", "true")
```
sestinj, line 48: This is the line I'm referencing. It allows you to right click and "Open Dev Tools" which is very useful for debugging

sestinj (changes requested):

lkk214:
@sestinj Oh, actually it has been an invalid configuration, it will only take effect when jcef is opened when starting the plugin, even if we don't configure it, we can still open it by right clicking and "Open Dev Tools".

devtools

sestinj (approved):

PR #4672, opened 2025-03-16
feat: Openrouter reasoning support

SpeedCrash100:
## Description

Added initial support of thinking retrieval from Openrouter.

Currently. It is required to set up `extraBodyProperties` and `provider` correctly:

Deepseek is working, but some models may not return reasoning even if they are thinking. Can be verified with curl.

```yaml
# Note: openai here, idk why, but with `openrouter` it will not work
  - provider: openai
    apiBase: https://openrouter.ai/api/v1
    apiKey: sk-*
    name: Openrouter Deepseek R1 FREE
    model: deepseek/deepseek-r1:free
    requestOptions:
      extraBodyProperties:
        stream: true
        include_reasoning: true
```

## Checklist

## Screenshots

изображение
изображение

## Testing instructions

1. set up config.yaml to use openrouter DeepSeek R1 model and add  `extraBodyProperties:` from __Description__ section.
2. Send chat request and ensure the `Thinking...` section appeared
3. It should be tested that Ollama Deepseek R1 still working.

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -361,6 +361,7 @@ export interface AssistantChatMessage {
   role: "assistant";
   content: MessageContent;
   toolCalls?: ToolCallDelta[];
+  reasoning?: string;
 }
 
 export interface SystemChatMessage {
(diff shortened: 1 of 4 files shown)


SpeedCrash100:
Will wait for #4559 might be handled here already

PR #4687, opened 2025-03-17
rejected edits trailing whitespaces 

panyamkeerthana:
## Description

Earlier trailing whitespace was stripped in streamDiffLines.ts before generating diffs, causing mismatches when rejecting a diff in the editor.
This fix updates linesMatchPerfectly to compare lines using trimEnd(), ensuring that trailing whitespace differences do not interfere with accurate diff rejection.

## Screenshots

Screenshot 2025-03-14 at 3 14 53 PM

Screenshot 2025-03-14 at 3 14 28 PM

https://github.com/Granite-Code/granite-code/issues/50

### DIFF
--- a/core/diff/streamDiff.test.ts
+++ b/core/diff/streamDiff.test.ts
@@ -223,6 +223,109 @@ describe("streamDiff(", () => {
     await expectDiff("fastapi-tabs-vs-spaces.py");
   });
 
+  test("trailing whitespaces should match ", async () => {
+    const oldLines = ["first item  ", "second arg ", "third param  "];
+
+    const newLines = ["first item", "second arg", "third param "];
+
+    const { streamDiffs } = await collectDiffs(oldLines, newLines);
+
+    expect(streamDiffs).toEqual([
+      { type: "same", line: "first item  " },
+      { type: "same", line: "second arg " },
+      { type: "same", line: "third param  " },
+    ]);
+  });
+
+  //indentation and whitespace handling
+  test.each([false, true])(
+    "ignores indentation changes for sufficiently long lines (trailingWhitespace: %s)",
+    async (trailingWhitespace) => {
+      let oldLines = [
+        " short",
+        "   middle",
+        " a long enough line",
+        "  short2",
+        "indented line",
+        "final line",
+      ];
+
+      let newLines = [
+        "short",
+        "middle",
+        "a long enough line",
+        "short2",
+        " indented line",
+        "final line",
+      ];
+
+      if (trailingWhitespace) {
+        oldLines = oldLines.map((line) => line + " ");
+      }
+
+      const { streamDiffs } = await collectDiffs(oldLines, newLines);
+      const expected = trailingWhitespace
+        ? [
+            { type: "old", line: " short " },
+            { type: "new", line: "short" },
+            { type: "old", line: "   middle " },
+            { type: "new", line: "middle" },
+            { type: "same", line: " a long enough line " },
+            { type: "same", line: "  short2 " },
+            { type: "same", line: "indented line " },
+            { type: "same", line: "final line " },
+          ]
+        : [
+            { type: "old", line: " short" },
+            { type: "new", line: "short" },
+            { type: "old", line: "   middle" },
+            { type: "new", line: "middle" },
+            { type: "same", line: " a long enough line" },
+            { type: "same", line: "  short2" },
+            { type: "same", line: "indented line" },
+            { type: "same", line: "final line" },
+          ];
+
+      expect(streamDiffs).toEqual(expected);
+    },
+  );
+
+  test("preserves original lines for minor reindentation in simple block", async () => {
+    const oldLines = ["if (checkValueOf(x)) {", "   doSomethingWith(x);", "}"];
+    const newLines = ["if (checkValueOf(x)) {", "  doSomethingWith(x);", "}"];
+
+    const { streamDiffs } = await collectDiffs(oldLines, newLines);
+
+    expect(streamDiffs).toEqual([
+      { type: "same", line: "if (checkValueOf(x)) {" },
+      { type: "same", line: "   doSomethingWith(x);" },
+      { type: "same", line: "}" },
+    ]);
+  });
+
+  test("uses new lines for nested reindentation changes", async () => {
+    const oldLines = ["if (checkValueOf(x)) {", "   doSomethingWith(x);", "}"];
+    const newLines = [
+      "if (checkValueOf(x)) {",
+      "   if (reallyCheckValueOf(x)) {",
+      "     doSomethingElseWith(x);",
+      "   }",
+      "}",
+    ];
+
+    const { streamDiffs } = await collectDiffs(oldLines, newLines);
+
+    expect(streamDiffs).toEqual([
+      { type: "same", line: "if (checkValueOf(x)) {" },
+      { type: "new", line: "   if (reallyCheckValueOf(x)) {" },
+      { type: "old", line: "   doSomethingWith(x);" },
+      { type: "new", line: "     doSomethingElseWith(x);" },
+      { type: "old", line: "}" },
+      { type: "new", line: "   }" },
+      { type: "new", line: "}" },
+    ]);
+  });
+
   test("FastAPI example", async () => {
     await expectDiff("fastapi.py");
   });
--- a/core/diff/streamDiff.ts
+++ b/core/diff/streamDiff.ts
@@ -35,7 +35,6 @@ export async function* streamDiff(
 
     let type: DiffLineType;
 
-    let isLineRemoval = false;
     const isNewLine = matchIndex === -1;
 
     if (isNewLine) {
@@ -45,7 +44,6 @@ export async function* streamDiff(
       for (let i = 0; i < matchIndex; i++) {
         yield { type: "old", line: oldLinesCopy.shift()! };
       }
-
       type = isPerfectMatch ? "same" : "old";
     }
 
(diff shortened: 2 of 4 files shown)


extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -6,7 +6,7 @@ class EditDecorationManager {
   constructor(context: vscode.ExtensionContext) {
     this.decorationType = vscode.window.createTextEditorDecorationType({
       backgroundColor: new vscode.ThemeColor(
-        "editor.inactiveSelectionBackground",
+        "editor.selectionHighlightBackground",
```
owtaylor: A stow-away

core/diff/util.ts
```diff
@@ -16,7 +16,11 @@ export type MatchLineResult = {
 };
 
 function linesMatchPerfectly(lineA: string, lineB: string): boolean {
-  return lineA === lineB && lineA !== "";
+
+//checks if two lines match exactly ignoring trailing whitespace differences
+//this prevents false mismatches in diff handling when trailing spaces are stripped
+
+  return lineA.trimEnd() === lineB.trimEnd() && lineA !== "";
```
owtaylor: This fixes up one case where we are comparing oldLines to newLines, but checking over the code, this isn't the only place where we might have issues:

 - in `matchLine()`, after calling `linesMatchPerfectly()`, we make more references to `oldLines[i]` that potentially might need changes.
 - In streamDiff(), we have:
```
         if (oldLinesCopy[0] !== newLine) {
          yield { type: "new", line: newLine };
        } else {
          isLineRemoval = true;
        }
```
   which looks suspicious, but I'm pretty sure can never actually be hit: if `oldLinesCopy[0] === newLine` then `isPerfectMatch` should have been true and the case will be `"same"` not `"old"`. Needs double checking.

But the takeaway for me here is that the test cases need updating to test different cases where lines differ only in trailing whitespace. From the `continue/core` directory, you can run:

```
$ npm test  diff/streamDiff.test.ts
```

to run streamDiff test cases. Add a test case to check the `linesMatchPerfectly` case, add a test case case that exercises:       `// This is a way to fix indentation, but only for sufficiently long lines to avoid matching whitespace or short lines` when the indentation changes *and* there is a change in whitespace.

panyamkeerthana: Thanks for the review, I have made the following changes:

-Added a test case for oldLinesCopy[0] === newLine that confirmed it’s already handled as "same" or "old", so the else statement was unnecessary and removed.
-Trimmed oldLines throughout and added a test to ensure trailing whitespaces are correctly treated as "same", preventing incorrect diffs.
-Added a test to verify that indentation changes in long lines don’t cause unnecessary diffs.

core/diff/streamDiff.test.ts
```diff
@@ -223,6 +223,51 @@ describe("streamDiff(", () => {
...
+
+    expect(streamDiffs).toEqual([
+      { type: "same", line: "first item  " },
+      { type: "same", line: "second arg " },
+      { type: "same", line: "third param" },
+    ]);
+  });
+
+  test("ignores indentation changes for sufficiently long lines", async () => {
+    const oldLines = ["   first item", " second arg", "third param", " fourth val "];
```
owtaylor: I would try to test this more comprehensively - to test that indentation changes on short lines are considered changes, until one line with length > 8 is encountered, then they are ignored, even for short lines.

Importantly, I'd also check here that the same behavior is seen when the old and new lines differ in trailing whitespace - that's the point of adding the check as part of this patch, after all. 

**Something** like:

```
 test.each([false, true])(
   `ignores identation changes for sufficiently long lines",
   async (trailingWhitespace) => {
      [...]
      if (trailingWhitespace)
         oldLines = oldLines.map(line => line + " ");
      [...]
   });
```

I'm sure there's a more stylish way to do that.

core/diff/util.ts
```diff
@@ -62,24 +62,27 @@ export function matchLine(
...
+    //this ensures trailing spaces don't affect matching.
+    const oldLineTrimmed = oldLines[i].trimEnd();
     // Don't match end bracket lines if too far away
     if (i > 4 && isEndBracket) {
       return { matchIndex: -1, isPerfectMatch: false, newLine };
     }
 
-    if (linesMatchPerfectly(newLine, oldLines[i])) {
+    if (linesMatchPerfectly(newLine, oldLineTrimmed)) {
       return { matchIndex: i, isPerfectMatch: true, newLine };
```
owtaylor, line 76: I think this probably should be oldLines[i] not newLine? The goal here should be that when we return `same` - we actually have the old line, not a version that differs in trailing whitespace. First try to make a test case that fails, then change this.

core/diff/util.ts
```diff
@@ -62,24 +62,27 @@ export function matchLine(
...
       if (
-        newLine.trimStart() === oldLines[i].trimStart() &&
+        newLine.trimStart() === oldLineTrimmed.trimStart() &&
         (permissiveAboutIndentation || newLine.trim().length > 8)
       ) {
         return {
           matchIndex: i,
           isPerfectMatch: true,
-          newLine: oldLines[i],
+          newLine: oldLineTrimmed,
```
owtaylor: Similarly, I think this should be oldLines[i]?

core/diff/streamDiff.ts
```diff
@@ -63,8 +63,6 @@ export async function* streamDiff(
 
         if (oldLinesCopy[0] !== newLine) {
           yield { type: "new", line: newLine };
-        } else {
-          isLineRemoval = true;
         }
```
owtaylor: You should remove the if/else entirely and just leave the yield, then since isLineRemoval can never be set, it needs to be removed as well.

core/diff/streamDiff.test.ts
```diff
@@ -223,6 +223,51 @@ describe("streamDiff(", () => {
     await expectDiff("fastapi-tabs-vs-spaces.py");
   });
 
+
+  test(" condition where oldLinesCopy[0]=== newLine", async () => {
```
owtaylor: I think this was useful during development of your patch, but since this code ends up removed in your patches, I'd suggest removing the test case as well.

core/diff/util.ts
panyamkeerthana: The test cases I ran for this scenario passed. In streamDiff():

```
case "same":
    yield { type, line: oldLinesCopy.shift()! };
    break;
```
So, regardless of what matchLine() returns for newLine, the line that is actually output is the one removed from oldLinesCopy (which is the original old line with the trailing whitespaces). But I can change it to oldLine[i] to maintain logic and uniformity

core/diff/util.ts
panyamkeerthana: I'll change that to oldLine[i] here as well

core/edit/streamDiffLines.ts
```diff
@@ -79,7 +79,7 @@ export async function* streamDiffLines(
   }
 
   // Trim end of oldLines, otherwise we have trailing \r on every line for CRLF files
-  oldLines = oldLines.map((line) => line.trimEnd());
+  oldLines = oldLines.map((line) => line);
```
owtaylor: This can just go away, along with the comment

core/diff/streamDiff.ts
```diff
@@ -35,17 +35,14 @@ export async function* streamDiff(
 
     let type: DiffLineType;
 
-    let isLineRemoval = false;
     const isNewLine = matchIndex === -1;
 
     if (isNewLine) {
       type = "new";
     } else {
-      // Insert all deleted lines before match
```
owtaylor: Should this comment be removed?

core/diff/streamDiff.test.ts
```diff
@@ -223,6 +223,68 @@ describe("streamDiff(", () => {
...
+        "a long enough line",
+        "  indented short",
+        "final line",
+      ];
+
+      let newLines = [
+        "short",
+        "middle",
+        "a long enough line",
+        "indented short",
```
owtaylor: "indented short".length is going to be > 8 - this needs to be "short2" or something to actually test that handlng.

core/diff/util.ts
```diff
@@ -62,18 +62,21 @@ export function matchLine(
...
-      return { matchIndex: i, isPerfectMatch: true, newLine };
+    if (linesMatchPerfectly(newLine, oldLineTrimmed)) {
+      return { matchIndex: i, isPerfectMatch: true, newLine: oldLines[i] };
     }
-    if (linesMatch(newLine, oldLines[i], i)) {
+    if (linesMatch(newLine, oldLineTrimmed, i)) {
       // This is a way to fix indentation, but only for sufficiently long lines to avoid matching whitespace or short lines
       if (
-        newLine.trimStart() === oldLines[i].trimStart() &&
+        newLine.trimStart() === oldLineTrimmed.trimStart() &&
```
owtaylor: I don't think we can *assume* that newLine is trimmed at the end, so this should be newLine.trim()? [make sure the test case tests this - one line that is supposed to not reindented should have a trailing space on the new]

owtaylor (commented):
Just a few more comments

core/diff/util.ts
panyamkeerthana: Instead, as discussed I created newLineTrimmed.

` newLineTrimmed.trimStart() === oldLineTrimmed.trimStart()`

core/diff/streamDiff.ts
panyamkeerthana: No I might have accidentally removed it, my bad

core/diff/util.ts
owtaylor: Turned out that this very much needed to be newLine - the calling code relies on this only being different from the passed in newline when the leading indentation changed. Sorry for the confusion!

core/diff/streamDiff.ts
owtaylor: If you add this back, the rest looks great to me!

core/diff/util.ts
panyamkeerthana: All this testing has only made the whitespace and indentation handling more robust!

core/diff/streamDiff.ts
panyamkeerthana: Done!

owtaylor:
Looks good to me now!

tomasz-stefaniak (approved):
Thank you for contributing @panyamkeerthana, and @owtaylor for the thorough review!

PR #4692, opened 2025-03-17
Handle message during tool call

RomneyDa:
## Description
If user sends message during tool call (especially during asking permission), send a tool response notifying that the user cancelled the tool call, and then move forward like a normal message

This PR also removes the `gatherContext` thunk and moves the once-used logic into the streamResponse thunk, in part because it would be nice to move towards no/less thunks

TODO change the approach here - should prob inject a tool response message not a tool response AND user message


sestinj (approved):

chezsmithy:
@RomneyDa I performed a number of tests with this change using both the Bedrock Claude 3.5 and 3.7 models and performaed a number of steps where I cancelled the tools use, and switched directions or then restarted without finding any issues. This is a really great change and a common complaint from our users given to escape from this situation they have to start a new context today. Recommend we merge this as soon as possible.

chezsmithy:
Likely fixes https://github.com/continuedev/continue/issues/4751

chezsmithy:
Likely fixes https://github.com/continuedev/continue/issues/4893

RomneyDa:
@chezsmithy note closing because this approach is flawed. I'll do another tool architecture fix in the next couple of days

Issue #32, opened 2023-05-29
Another iteration on email newsletter input


TyDunn:
1. Make it disappear
2. Make styling fit brand
3. Make it more clear that you should enter your email

PR #4698, opened 2025-03-17
Changed Enter button icon and fixed sensitivity issues

GauravVBhambhani:
## Description

Chat view: Improve look and feel of the enter button

The chat view for continue currently consists of a Send(Enter) button that looks like an enter key icon (↵), which isn't ideal as many keyboards don't use that icon anymore.

Furthermore, there are some sensitivity issues:
 - the button is active sometimes when it shouldn't be (when there's no user input).
 - the styling of the button makes it appear insensitive sometimes, even when it's active.

This commit addresses these issues, by switching to a more modern airplane icon, using vscode theme colors for styling.
Marking the button insensitive when there's no user input.

## Screenshots
Screenshot 2025-03-17 at 3 27 13 PM

Screenshot 2025-03-17 at 3 26 35 PM

[Granite.Code Issue #32](https://github.com/Granite-Code/granite-code/issues/32)

### DIFF
--- a/gui/src/components/mainInput/InputToolbar.tsx
+++ b/gui/src/components/mainInput/InputToolbar.tsx
@@ -1,4 +1,4 @@
-import { AtSymbolIcon, PhotoIcon } from "@heroicons/react/24/outline";
+import { AtSymbolIcon, PaperAirplaneIcon, PhotoIcon } from "@heroicons/react/24/outline";
 import { InputModifiers } from "core";
 import { modelSupportsImages, modelSupportsTools } from "core/llm/autodetect";
 import { useRef } from "react";
@@ -50,7 +50,8 @@ const StyledDiv = styled.div<{ isHidden?: boolean }>`
 
 const EnterButton = styled.button<{ isPrimary?: boolean }>`
   all: unset;
-  padding: 2px 4px;
+  padding: 3px 4px;
+  padding-bottom: 0px;
   display: flex;
   align-items: center;
   background-color: ${(props) =>
@@ -83,6 +84,7 @@ interface InputToolbarProps {
   toolbarOptions?: ToolbarOptions;
   disabled?: boolean;
   isMainInput?: boolean;
+  hasText?: boolean;
 }
 
 function InputToolbar(props: InputToolbarProps) {
@@ -223,10 +225,18 @@ function InputToolbar(props: InputToolbarProps) {
             }}
             disabled={isEnterDisabled}
           >
-            <span className="hidden md:inline">
-              ⏎ {props.toolbarOptions?.enterText ?? "Enter"}
+            <span data-tooltip-id="submit-tooltip">
+              <PaperAirplaneIcon
+                className={`h-4 w-4 ${
+                  props.hasText
+                  ? "text-[var(--vscode-editorForeground)]"
+                  : "text-[var(--vscode-descriptionForeground)]"
+                } transition duration-150 ease-in-out`}
+              />
             </span>
-            <span className="md:hidden">⏎</span>
+            <ToolTip id="submit-tooltip" place="top-middle">
+              Send (Enter)
+            </ToolTip>
           </EnterButton>
         </div>
       </StyledDiv>
--- a/gui/src/components/mainInput/TipTapEditor.tsx
+++ b/gui/src/components/mainInput/TipTapEditor.tsx
@@ -551,6 +551,21 @@ function TipTapEditor(props: TipTapEditorProps) {
     editable: !isStreaming || props.isMainInput,
   });
 
+  const [hasText, setHasText] = useState(false);
+  useEffect(() => {
+    if (!editor) return;
+    // Check if the editor has text
+    const updateEditorHandler = () => {
+      const text = editor.getText().trim();
+      setHasText(text.length > 0);
+    };
+    editor.on("update", updateEditorHandler);
+    updateEditorHandler();
+    return () => {
+      editor.off("update", updateEditorHandler);
+    };
+  }, [editor]);
+
   const [shouldHideToolbar, setShouldHideToolbar] = useState(false);
   const debouncedShouldHideToolbar = debounce((value) => {
     setShouldHideToolbar(value);
@@ -1015,6 +1030,7 @@ function TipTapEditor(props: TipTapEditorProps) {
             });
           }}
           disabled={isStreaming}
+          hasText={hasText}
         />
       </div>
 


gui/src/components/mainInput/InputToolbar.tsx
```diff
@@ -223,10 +225,18 @@ function InputToolbar(props: InputToolbarProps) {
             }}
             disabled={isEnterDisabled}
           >
-            <span className="hidden md:inline">
-              ⏎ {props.toolbarOptions?.enterText ?? "Enter"}
-            </span>
-            <span className="md:hidden">⏎</span>
+            <div data-tooltip-id="submit-tooltip">
```
halfline: is there a reason this needs to be a div and not a span?

gui/src/components/mainInput/TipTapEditor.tsx
```diff
@@ -1015,6 +1015,7 @@ function TipTapEditor(props: TipTapEditorProps) {
             });
           }}
           disabled={isStreaming}
+          hasText={editor ? editor.getText().trim().length > 0 : false}
```
halfline: instead of doing this here, can you just put `hasText={hasText}`  and then do:

```
  const [hasText, setHasText] = useState(false);
  useEffect(() => {
...
   if (!editor) return;
   const updateEditorHandler = () => {
        const text = editor.getText().trim();
        setHasText(text.length > 0);
    };
    const clearUpdateEditorHandler = () => {
        editor.off("update", clearUpdateHandler);
    }
    updateEditorHandler();
    return clearUpdateEditorHandler;
  }
...
```

or something like that, so it tracks state more robustly?

gui/src/components/mainInput/InputToolbar.tsx
```diff
@@ -223,10 +225,18 @@ function InputToolbar(props: InputToolbarProps) {
...
+              <PaperAirplaneIcon
+                className={`h-4 w-4 ${
+                  props.hasText 
+                  ? "text-[var(--vscode-editorForeground)]"
```
halfline: you have some trailing whitespace here I think.

halfline (commented):
So, I haven't tried it yet, but to me, in a cursory review this looks pretty good.
Beefing up the state management for hasText seems like good idea, but don't know what Dallin/Nate's takes are.

I would improve the commit message a bit...maybe something like:

chat: Improve look and feel of enter button

The chat view for continue currently features a submit button that looks like an enter key icon.

This metaphor isn't ideal, since many keyboards these days don't use that icon. Furthermore, there are some sensitivity issues:
1. The styling of the button makes it sometimes appear insensitive when when it's active
2. The button is active sometimes when it shouldn't be (when there is no user input).

This commit addresses those issues, by switching to a more modern airplane icon instead of a return key icon, using vscode theme colors for styling, and marking the button insensitive when there is no user input, respectively.

gui/src/components/mainInput/InputToolbar.tsx
GauravVBhambhani: That's just to make it more readable.

gui/src/components/mainInput/TipTapEditor.tsx
GauravVBhambhani: Sure! I'll look into it.

gui/src/components/mainInput/InputToolbar.tsx
GauravVBhambhani: I chose a div in this case because it naturally behaves as a block-level element so the tooltip is correctly positioned. But, I can use a span as well. I might have to add a inline-block style for proper layout. I'll test it out.

GauravVBhambhani: Nevermind, works the same without any additional styling. Changed to span.

gui/src/components/mainInput/InputToolbar.tsx
halfline: you already have line breaks for readability... i'm talking about after `prop.hasText` there's a space before the line break. same with at the end of the `? "text-..."` line

halfline:
this looks good to me. only other change I would recommend (aside from the whitespace nit above) is squashing the two commits together and using something like my proposed commit message for clarity.

```
❯ git reset HEAD~2

❯ git commit -a -m "chat: Improve look and feel of enter button

The chat view for continue currently features a submit button that looks like an enter key icon.

This metaphor isn't ideal, since many keyboards these days don't use that icon. Furthermore, there are some sensitivity issues:

    The styling of the button makes it sometimes appear insensitive when when it's active
    The button is active sometimes when it shouldn't be (when there is no user input).

This commit addresses those issues, by switching to a more modern airplane icon instead of a return key icon, using vscode theme colors for styling, and marking the button insensitive when there is no user input, respectively."

❯ git push --force 
```

or similar

halfline:
lgtm

sestinj:
@GauravVBhambhani thanks for taking the time to make this PR, and @halfline for the preliminary review! The change in button icon and size is not something we will be willing to merge, as it appears a bit clunkier than we would wish for, however the sensitivity issues would be important to solve. Would you be interested in making those updates in a separate PR, or just adjusting this one to remove the change in appearance?

halfline:
Obivously, your call, Nate, but just for completeness I want to make my case and try to change your mind 😉 

My arguments:
- From a UX perspective it's better if the button represents what the action does, i.e.g, "send out the message", not what keybinding is associated with the action. 
- ⏎ is an old metaphor that rarely appears on keyboards anymore (not quite as archaic as using 💾 for save icon, but close)
- None of Co-Pilot, ChatGPT, meta.ai, gemini, molmo. or claude.ai use the ⏎ . they all use either a straight arrow or an airplane. (cursor follows suit though)

What would you think about using a straight arrow instead? Or what about if a clear background was used?

halfline:
ping, just looking for a yay or nay to know next steps

allanday:
Changing the send icon was originally my idea. For me the main motivations for the change are:

* The return symbol isn't a great icon visually. It's relatively indistinct - just thin lines, without a clear shape. The vertical alignment is also off when I look at it here (admittedly that could be fixed).
* A consistent visual treatment for this buttons helps when communicating sensitivity, since we can use the same colors for sensititive/insensitive. The arrow icon better supports this due to its greater visual presence.
* As @halfline said, the send arrow is a convention on other AI assistants, so makes it easier for people to understand what they are looking at.

But I agree that the sensitivity changes are the main thing here, so maybe we shouldn't get too sidetracked by the icon discussion. :slightly_smiling_face:

allanday:
Oh, the other obvious point about the "send" icon that I should have mentioned: the arrow is the convention from pretty much every chat app, and this is chat. Using the standard icon for the domain of communication helps users to orientate themselves to the interaction model.

RomneyDa:
Fixing sensitivity issues would be great but the Icon is a significant UI change that we wouldn't slip in with it. For now we feel the arrow is a bit clunky.

brief glance at some products in the same space, looks like copilot and cline use the icon you're proposing, and cody uses a similar triangle. But other major one don't, e.g. cursor, windsurf, and major chat apps like claude, chatgpt, lovable, bolt, vercel's popular ai chat frontend, etc. all use various thin arrows

allanday:
Thanks @RomneyDa , that makes sense

RomneyDa:
Closing for now, feel free to reopen with updates!

PR #4709, opened 2025-03-18
Improve search result

cuining:
## Description
When i search something with the middle part of one filename， is cannot return the items.

for example, i have a file `SmartStringUtil.java`, when i search @stringutil, it returns null


sestinj (approved):
really nice fix, thanks a lot @cuining !

PR #4725, opened 2025-03-19
Improved text selection by allowing multiple edit ranges to be highlighted

GauravVBhambhani:
## Description

Text Selection: Decorate all active edit ranges

If a user selects multiple chunks of code with cmd-I, only the most recent chunk of code is decorated as a selection.

This commit addresses that problem by changing setDecorations to addDecorations, and carefully tracking which ranges are currently active in the EditDecorationManager. 

Additionally, overlapping or adjacent ranges are now merged seamlessly to prevent redundant highlights.

## Screenshots
Screenshot 2025-03-19 at 12 44 14 PM

[Granite.Code Issue #49](https://github.com/Granite-Code/granite-code/issues/49)
@halfline

### DIFF
--- a/extensions/vscode/src/quickEdit/EditDecorationManager.ts
+++ b/extensions/vscode/src/quickEdit/EditDecorationManager.ts
@@ -2,6 +2,7 @@ import vscode from "vscode";
 class EditDecorationManager {
   private _lastEditor: vscode.TextEditor | undefined;
   private decorationType: vscode.TextEditorDecorationType;
+  private activeRangesMap: Map<string, vscode.Range> = new Map(); // Store unique active ranges
 
   constructor(context: vscode.ExtensionContext) {
     this.decorationType = vscode.window.createTextEditorDecorationType({
@@ -23,18 +24,55 @@ class EditDecorationManager {
     );
   }
 
-  setDecoration(editor: vscode.TextEditor, range: vscode.Range) {
-    if (this._lastEditor) {
-      this._lastEditor.setDecorations(this.decorationType, []);
+  // Converts each range to a unique string for storing in the map
+  private rangeToString(range: vscode.Range): string {
+    return `(${range.start.line},${range.start.character})-(${range.end.line},${range.end.character})`;
+  }
+
+  // Checks if two ranges are adjacent or overlapping
+  private rangesCoincide(range1: vscode.Range, range2: vscode.Range): boolean {
+    return range1.start.isEqual(range2.start)
+    || range1.end.isEqual(range2.start)
+    || range2.end.isEqual(range1.start)
+    || !!range1.intersection(range2);
+  }
+
+  // Merges new range with existing ranges in the map
+  private mergeNewRange(newRange: vscode.Range): void {
+    let mergedRange = newRange;
+    const rangesToPrune: string[] = [];
+
+    for (const [key, existingRange] of this.activeRangesMap.entries()) {
+      if (!this.rangesCoincide(mergedRange, existingRange)) continue; 
+      mergedRange = mergedRange.union(existingRange);
+      rangesToPrune.push(key);
     }
-    editor.setDecorations(this.decorationType, [range]);
+
+    for (const key of rangesToPrune) this.activeRangesMap.delete(key);
+    this.activeRangesMap.set(this.rangeToString(mergedRange), mergedRange);
+  }
+
+  // Adds a new decoration to the editor and merges it with existing ranges
+  addDecorations(editor: vscode.TextEditor, ranges: vscode.Range[]): void {
+    if (this._lastEditor?.document !== editor.document) {
+      this.clear(); // Clear previous decorations if editor has changed
+    }
     this._lastEditor = editor;
+
+    for (const range of ranges) this.mergeNewRange(range);
+
+    const activeRanges = Array.from(this.activeRangesMap.values());
+    if (activeRanges.length === 0) return; // No ranges to highlight
+
+    // Update active ranges and apply decorations
+    editor.setDecorations(this.decorationType, activeRanges);
     this.updateInEditMode(true);
   }
 
   clear() {
     if (this._lastEditor) {
       this._lastEditor.setDecorations(this.decorationType, []);
+      this.activeRangesMap.clear();
       this._lastEditor = undefined;
       this.updateInEditMode(false);
     }
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -569,7 +569,7 @@ const getCommandsMap: (
       const range =
         args?.range ?? new vscode.Range(startFromCharZero, endAtCharLast);
 
-      editDecorationManager.setDecoration(editor, range);
+      editDecorationManager.addDecorations(editor, [range]);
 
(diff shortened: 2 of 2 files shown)


extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -2,6 +2,8 @@ import vscode from "vscode";
 class EditDecorationManager {
   private _lastEditor: vscode.TextEditor | undefined;
   private decorationType: vscode.TextEditorDecorationType;
+  private activeRanges: vscode.Range[] = []; // Store active ranges to be highlighted
+  private activeRangeSet: Set<string> = new Set(); // Store active ranges as strings to avoid duplicates
```
halfline: I think it would be a little cleaner to combine these into a single map instead

```
    private activeRangesMap: Map<string, vscode.Range> = new Map();
```

extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -23,18 +25,40 @@ class EditDecorationManager {
     );
   }
 
-  setDecoration(editor: vscode.TextEditor, range: vscode.Range) {
-    if (this._lastEditor) {
-      this._lastEditor.setDecorations(this.decorationType, []);
+  // converting each range to a unique string for storing in set
```
halfline: // Converts each range to a unique string for storing in the map

extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -23,18 +25,40 @@ class EditDecorationManager {
...
+  private rangeToString(range: vscode.Range): string {
+    return `${range.start.line}:${range.start.character}-${range.end.line}:${range.end.character}`;
```
halfline: I feel like some slightly different punctuation could make this clearer in console.log output or whatever...e.g, you have
```
1:1-4:80
```
but
```
(1,1)-(4,80)
```
would look a little clearer. doesn't really matter, but just food for though for later debugging sanity :-)

extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -23,18 +25,40 @@ class EditDecorationManager {
...
+  }
+
+  setDecoration(editor: vscode.TextEditor, ranges: vscode.Range[]): void {
```
halfline: "set" usually means "replace" but this function augments not replaces now.should probably be renamed `addDecorations`

extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -23,18 +25,40 @@ class EditDecorationManager {
...
+      }
+      this.activeRangeSet.add(rangeKey);
+      return true;
+    });
+
+    if (newRanges.length === 0) return; // No new ranges to highlight
+
+    // Update active ranges and apply decorations
+    this.activeRanges.push(...newRanges);
+    editor.setDecorations(this.decorationType, this.activeRanges);
```
halfline: This doesn't really handle overlapping ranges right I think.

I think you need some code like this (assuming we switch to the map as suggested above):
```
  private rangesCoincide(a: vscode.Range, b: vscode.Range): boolean {
    if (a.end.isEqual(b.start)) return true;
    if (b.end.isEqual(a.start)) return true;
    if (a.intersection(b)) return true;
    return false;
  }

  private mergeNewRange(newRange: vscode.Range): void {
    let mergedRange = newRange;
    const rangesToPrune: string[] = [];

    for (const [key, existingRange] of this.activeRangesMap.entries()) {
      if (!this.rangesCoincide(merged, existingRange)) continue;
      mergedRange = mergedRange.union(existingRange);
      rangesToPrune.push(key);
    }

    for (const key of rangesToPrune) this.activeRangesMap.delete(key);

    this.activeRangesMap.set(this.rangeToString(mergedRange), mergedRange);
  }
```
and then here like
```
    for (const range of ranges) this.mergeNewRange(range);

    const activeRanges = Array.from(this.activeRangesMap.values());
    if (activeRanges.length === 0) return; // No new ranges to highlight
    editor.setDecorations(this.decorationType, activeRanges);
```
or something like that.

halfline (commented):
Hey thanks for working on this. I love that you're just going through our todo list and clearing it out ! :-) You're very quick.

I haven't tried this yet, but I made a few suggestions. As always Nate and Dallin get final call, but here are my thoughts just doing a quick read through.

The commit message also needs work. The top line of a git commit message is supposed to be a summary line that doesn't get much longer than 50 characters.

I'd suggest something like:

Decorate all active edit ranges 

If a user selects multiple chunks of code with cmd-I, only the most recent chunk of code is decorated as a selection.

This commit addresses that problem by changing setDecorations to addDecorations, and carefully tracking which ranges are currently active in the EditDecorationManager.

extensions/vscode/src/quickEdit/EditDecorationManager.ts
GauravVBhambhani: You pretty much fixed the whole thing😂, but thank you! Got to learn a few things. Appreciate it! 😊

extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -23,18 +24,55 @@ class EditDecorationManager {
     );
   }
 
-  setDecoration(editor: vscode.TextEditor, range: vscode.Range) {
-    if (this._lastEditor) {
-      this._lastEditor.setDecorations(this.decorationType, []);
+  // Converts each range to a unique string for storing in the map
+  private rangeToString(range: vscode.Range): string {
+    return `(${range.start.line},${range.start.character}-${range.end.line},${range.end.character})`;
```
halfline: i would still bracket it differently... like:
```
 return `(${range.start.line},${range.start.character})-(${range.end.line},${range.end.character})`;
```
so dev reading logs knows immediately it's `(a,b)`-`(c,d)`, not (`a`, `b-c`, `d`)

extensions/vscode/src/quickEdit/EditDecorationManager.ts
```diff
@@ -23,18 +24,55 @@ class EditDecorationManager {
...
+      mergedRange = mergedRange.union(existingRange);
+      rangesToPrune.push(key);
+    }
+
+    for (const key of rangesToPrune) this.activeRangesMap.delete(key);
+    this.activeRangesMap.set(this.rangeToString(mergedRange), mergedRange);
+  }
+
+  // Adds a new decoration to the editor and merges it with existing ranges
+  addDecoration(editor: vscode.TextEditor, ranges: vscode.Range[]): void {
```
halfline: i would make this plural since it's takes an array of ranges and could end up with more than one

halfline:
sorry for not mentioning this earlier, but this looks good to me

sestinj:
thanks for the great PR!

fbricon:
I can't get this feature to work in the latest pre-release build. @RomneyDa could it be that https://github.com/continuedev/continue/commit/bb2bae3e106ae42c42150e5111f40dd295068c12#diff-83cecd4ec0e810627c89f4a442a85005401de10f5c282be8c840cc4080ff2ddb broke it?

PR #4753, opened 2025-03-21
Implement caching for diff in IntelliJ extension

ferenci84:
## Description

Caches diff for IntelliJ plugin.

Issue: https://github.com/continuedev/continue/issues/4130

## Checklist

## Testing instructions

Use autocomplete in a testing environment under IntelliJ.

IDE Logs tab contains logs about caching:
Screenshot 2025-03-21 at 15 25 42

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/GitService.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/GitService.kt
@@ -1,0 +1,94 @@
+package com.github.continuedev.continueintellijextension.`continue`
+
+import com.github.continuedev.continueintellijextension.services.ContinuePluginService
+import com.github.continuedev.continueintellijextension.utils.toUriOrNull
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.project.guessProjectDir
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.BufferedReader
+import java.io.File
+import java.io.InputStreamReader
+import java.net.URI
+
+class GitService(
+    private val project: Project,
+    private val continuePluginService: ContinuePluginService
+) {
+    // Add a simple cache for diff results
+    private data class DiffCache(
+        val timestamp: Long,
+        val diffs: List<String>
+    )
+
+    // Cache the last diff result
+    private var diffCache: DiffCache? = null
+    private var lastFileSaveTimestamp: Long = System.currentTimeMillis()
+
+    /**
+     * Updates the timestamp when a file is saved
+     */
+    fun updateLastFileSaveTimestamp() {
+        lastFileSaveTimestamp = System.currentTimeMillis()
+    }
+
+    /**
+     * Returns the git diff for all workspace directories
+     */
+    suspend fun getDiff(includeUnstaged: Boolean): List<String> {
+        // Check if we have a valid cache entry
+        if (diffCache != null && diffCache!!.timestamp == lastFileSaveTimestamp) {
+            return diffCache!!.diffs
+        }
+
+        // If no cache hit, compute the diff
+        val workspaceDirs = workspaceDirectories()
+        val diffs = mutableListOf<String>()
+
+        for (workspaceDir in workspaceDirs) {
+            val output = StringBuilder()
+            val builder = if (includeUnstaged) {
+                ProcessBuilder("git", "diff")
+            } else {
+                ProcessBuilder("git", "diff", "--cached")
+            }
+            builder.directory(File(URI(workspaceDir)))
+            val process = withContext(Dispatchers.IO) {
+                builder.start()
+            }
+
+            val reader = BufferedReader(InputStreamReader(process.inputStream))
+            var line: String? = withContext(Dispatchers.IO) {
+                reader.readLine()
+            }
+            while (line != null) {
+                output.append(line)
+                output.append("\n")
+                line = withContext(Dispatchers.IO) {
+                    reader.readLine()
+                }
+            }
+
+            withContext(Dispatchers.IO) {
+                process.waitFor()
+            }
+
+            diffs.add(output.toString())
+        }
+
+        // Cache the result
+        diffCache = DiffCache(lastFileSaveTimestamp, diffs)
+        return diffs
+    }
+
+    fun workspaceDirectories(): Array<String> {
+        val dirs = this.continuePluginService.workspacePaths
+
+        if (dirs?.isNotEmpty() == true) {
+            return dirs
+        }
+
+        return listOfNotNull(project.guessProjectDir()?.toUriOrNull()).toTypedArray()
+    }
+
+}
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/GitService.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/GitService.kt
@@ -81,7 +81,7 @@ class GitService(
         return diffs
     }
 
-    fun workspaceDirectories(): Array<String> {
+    private fun workspaceDirectories(): Array<String> {
         val dirs = this.continuePluginService.workspacePaths
 
         if (dirs?.isNotEmpty() == true) {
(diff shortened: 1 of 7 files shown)


extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
```diff
@@ -66,6 +77,14 @@ class IntelliJIDE(
             Paths.get(pluginPath.toString(), "ripgrep", "bin", "rg" + if (os == OS.WINDOWS) ".exe" else "").toString()
     }
 
+    /**
+     * Updates the timestamp when a file is saved
+     */
+    override fun updateLastFileSaveTimestamp() {
+        println("Updating last file save timestamp")
```
Patrick-Erichsen:
```suggestion
```

I think this will be too noisy of a log

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
```diff
@@ -110,6 +129,16 @@ class IntelliJIDE(
     }
 
     override suspend fun getDiff(includeUnstaged: Boolean): List<String> {
+        println("Getting diff, cache timestamp: ${diffCache?.timestamp}, current timestamp: $lastFileSaveTimestamp")
```
Patrick-Erichsen:
```suggestion
```

Similarly, I think this will be too noisy

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
```diff
@@ -110,6 +129,16 @@ class IntelliJIDE(
...
+
+        // Check if we have a valid cache entry
+        if (diffCache != null && diffCache!!.timestamp == lastFileSaveTimestamp) {
+            println("Using cached diff")
```
Patrick-Erichsen:
```suggestion
```

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
```diff
@@ -110,6 +129,16 @@ class IntelliJIDE(
...
+            return diffCache!!.diffs
+        }
+
+        // If no cache hit, compute the diff
+        println("Computing fresh diff")
```
Patrick-Erichsen:
```suggestion
```

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IdeProtocolClient.kt
```diff
@@ -202,6 +206,11 @@ class IdeProtocolClient(
                         respond(null)
                     }
 
+                    "updateLastFileSaveTimestamp" -> {
```
Patrick-Erichsen: This isn't very clear in our codebase, but the messages in this class are intended for things sent from GUI <> IDE or IDE <> Core

In this case, since it's just a service class that we're calling for our diff caching, we shouldn't be adding a new message, we can just call the logic directly.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
```diff
@@ -51,8 +51,19 @@ import java.nio.file.Paths
 class IntelliJIDE(
     private val project: Project,
     private val continuePluginService: ContinuePluginService,
+    private var lastFileSaveTimestamp: Long = System.currentTimeMillis()
+
+) : IDE {
+
+    // Add a simple cache for diff results
+    private data class DiffCache(
```
Patrick-Erichsen: This class is very simple currently but I'd still prefer to pull it out into it's own file.

Some other misc thoughts:
- There should be a max size on this so that it can't grow indefinitely
- ^ Ideally there is a lib we can pull in for this to avoid having to maintain the logic ourselves
- Rather than a "DiffCache" class, perhaps we build a "DiffService" that allows us to move all of the `getDiff` logic out of this IDE class where it doesn't belong, and just move the caching logic into that service as well
- This will also make it easier to test

Patrick-Erichsen (changes requested):
Hey @ferenci84 , thanks for the additional contribution here! 

I'm curious what prompted you to build this out - have you been having issues with the current `getDiff`? Looking at the code I can see how it would be quite expensive to compute, so a cache here makes sense to me, but just curious if you've been having issues or if it's just a part of the codebase you noticed could be improved.

Regarding my review - I was a bit nitpicky here because this feels like an important part of the codebase to get right, so I think if we want to build out a cache for diffs we should make sure it's rock solid since we're introducing more complexity.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
ferenci84: OK, I made a new class.

ferenci84:
Hi @Patrick-Erichsen, Thank you for thoughtful review. I have made the requested modifications.

Regarding cache size, this stores only the very last value, so the size is already limited, it's not a full-featured cache, anything more complicated would be overkill for this case.

> I'm curious what prompted you to build this out - have you been having issues with the current getDiff? Looking at the code I can see how it would be quite expensive to compute, so a cache here makes sense to me, but just curious if you've been having issues or if it's just a part of the codebase you noticed could be improved.

I made a resolution of issue https://github.com/continuedev/continue/issues/4130 for VS code with a default 10s cache for platforms with no implementation, and recently I'd been getting a message every few days that the problem presents (too many git processes are started at the same time), and it's so serious that they cannot use the extension at all. This is not something that comes up with simple test scenarios, but in real-world case where you have lot of staging modifications and potentially large diff (think complied js files that are technically not binary, but are very large and diff computation is slow).

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/GitService.kt
```diff
@@ -0,0 +1,94 @@
...
+
+            diffs.add(output.toString())
+        }
+
+        // Cache the result
+        diffCache = DiffCache(lastFileSaveTimestamp, diffs)
+        return diffs
+    }
+
+    fun workspaceDirectories(): Array<String> {
```
Patrick-Erichsen: This feels like it shouldn't live in the `GitService` as it isn't `git` related

Patrick-Erichsen (changes requested):
Hey @ferenci84 , thanks for your patience with this review. 

Left a small comment on the `GitService`, but after discussing with @sestinj and @RomneyDa we realized there is a larger issue with the current implementation, both here and in https://github.com/continuedev/continue/pull/4161/files#diff-3dfc62a1eec0c52efd9c760f8719dc5c7bd66e04808ceedf350d631b2391991eR46-R53

It feels like we should probably be handling the last saved file cache on the `core`, rather than in both `vscode` and `jetbrains`. 

What are your thoughts?

ferenci84:
> Hey @ferenci84 , thanks for your patience with this review.
> 
> Left a small comment on the `GitService`, but after discussing with @sestinj and @RomneyDa we realized there is a larger issue with the current implementation, both here and in https://github.com/continuedev/continue/pull/4161/files#diff-3dfc62a1eec0c52efd9c760f8719dc5c7bd66e04808ceedf350d631b2391991eR46-R53
> 
> It feels like we should probably be handling the last saved file cache on the `core`, rather than in both `vscode` and `jetbrains`.
> 
> What are your thoughts?

@Patrick-Erichsen @sestinj My opinion that it doesn't matter at all whether we do this simple functionality in the core or in the plugin itself, so I think the simpler is better, and it's simpler this way. In contrast to the VS plugin, the IntelliJ plugin's native language is not javascript, the methods written in java cannot be called directly from the core. Adding one more method call between core and the IntelliJ plugin, to save the last save timestamp in the core that was generated in the plugin, just to then give this information back to the plugin again, seems to be overly complicated and unnecessary with no advantage. Or am I wrong?

Patrick-Erichsen (approved):

Patrick-Erichsen:
Chatted this over more with the team - I think the one other benefit to having the actual cache in `core` is that it would eliminate an extra IDE request on a cache hit. But this really isn't that big of a deal so I'd say the simplicity argument carries about as much weight.

Will merge, thanks for taking the time to both implement + engage with our feedback! Really appreciate all the great contributions you've made recently 👌

slr1997:
I updated the version of continue in jetbrains to 1.0.10, but when I use the auto-complete, it still keeps generating the 'Git For Windows' progress.

PR #4759, opened 2025-03-21
feat: Improve ProtocolClient initialization and GUI loading

lkk214:
## Description

- ContinueBrowser adds LoadHandler before loading the GUI
- Load the url after protocolClient is initialized

Some messages will be lost if the "Continue" tool window is open during JetBrains IDE startup, as this can cause the GUI to initialize without data.

Affected message types: `jetbrains/isOSREnabled`, `jetbrains/getColors`, `jetbrains/onLoad`, etc.

Currently, it is obvious that you can't right-click and "Open Dev Tools", and the theme color of the GUI is unexpected. These issues occur specifically when the "Continue" tool window is open during JetBrains IDE startup.


AfterStories:
hi @lkk214 
I noticed your PR and it seems that it may solve a problem I previously discovered. When using plugins in IntelliJ, it is possible that colors cannot be obtained, and many components in the UI do not have colors rendered.
Screenshots and descriptions can be found here:
https://discord.com/channels/1108621136150929458/1346780535200157716/1352193700888969270

I would like to ask if you think this color rendering issue is related to your PR this time. If it is, I think I should merge your fix code this time

sestinj:
@lkk214 I like the changes here! I think this is important enough that we should probably attempt to write a quick e2e test to capture the behavior. I want to make sure that this doesn't cause a loop where the tool window is waiting infinitely, and it's an opportunity to make sure that the colors actually arrive.

I think many different tests could suffice, but one example would be checking that the tool window correctly gets the colors.

lkk214:
@sestinj I can add an e2e test for it, but I see that the code for the GUI test is commented. I don't understand why it is commented, maybe the GUI test cannot pass normally.I can add some e2e tests after reusing the GUI tests.

This PR will not cause the tool window to wait indefinitely. It just ensures that the URL is not loaded before the ideProtocolClient is initialized, ensuring that `ideProtocolClient?.handleMessage` is executed as expected.

Issue #4774, opened 2025-03-23
Chat Title Displays as [object Object]


sh4shii:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows
- Continue version: 1.1.13
- IDE version: VSCode 1.98.1
- Model: Gemini 2.0 Flash
- config:
  {
  "models": [
    {
      "title": "Gemini 2.0 Flash",
      "model": "gemini-2.0-flash",
      "contextLength": 1000000,
      "apiKey": "",
      "provider": "gemini"
    }
  ],
  "contextProviders": [
    {
      "name": "code",
      "params": {}
    },
    {
      "name": "docs",
      "params": {}
    },
    {
      "name": "diff",
      "params": {}
    },
    {
      "name": "terminal",
      "params": {}
    },
    {
      "name": "problems",
      "params": {}
    },
    {
      "name": "folder",
      "params": {}
    },
    {
      "name": "codebase",
      "params": {}
    }
  ],
  "slashCommands": [
    {
      "name": "share",
      "description": "Export the current chat session to markdown"
    },
    {
      "name": "cmd",
      "description": "Generate a shell command"
    },
    {
      "name": "commit",
      "description": "Generate a git commit message"
    }
  ],
  "data": []
}

  
  OR link to assistant in Continue hub:
```

### Description

When using the chat feature in the sidebar, the generated title for the chat displays as [object Object] instead of the expected text. This issue can be seen in the attached screenshot.

![Image](https://github.com/user-attachments/assets/8d3ea038-b9bc-42a8-8903-e9fe1dd9021f)

### To reproduce

1. Initiate a chat in the sidebar.
2. Observe the generated title for the chat.

### Log output

```Shell

```

PR #4775, opened 2025-03-23
Fix: Incorrect chat title generation

sh4shii:
## Description
fixes: #4774 
The chat title was previously being generated as `[object Object][object Object]`. This issue has been fixed by ensuring the correct string representation is used for the title.

## Checklist

## Screenshots
As you can see in the attached screenshot, the chat title is now generated correctly.
image

## Testing instructions

### DIFF
--- a/core/util/chatDescriber.ts
+++ b/core/util/chatDescriber.ts
@@ -30,19 +30,14 @@ export class ChatDescriber {
     completionOptions.maxTokens = ChatDescriber.maxTokens;
 
     // Prompt the user's current LLM for the title
-    const titleResponse = await model.chat(
-      [
-        {
-          role: "user",
-          content: ChatDescriber.prompt + message,
-        },
-      ],
+    const titleResponse = await model.complete(
+      ChatDescriber.prompt + message,
       new AbortController().signal,
       completionOptions,
     );
 
     // Set the title
-    return removeQuotesAndEscapes(titleResponse.content.toString());
+    return removeQuotesAndEscapes(titleResponse);
   }
 
   //   // TODO: Allow the user to manually set specific/tailored prompts to generate their titles


RomneyDa (changes requested):
@sh4shii would it be possible to implement the same fix using `chat` instead of `complete`? Something on the second line you changed is probably the culprit

sh4shii:
> @sh4shii would it be possible to implement the same fix using `chat` instead of `complete`? Something in the latter lines you updated is likely the culprit?
> 
> Specifically wondering what the [object Object] actually is in your case. Is it an error because the model doesn't support chat? Is it not parsed correctly?

@RomneyDa

sh4shii:
@RomneyDa  tried implementing the fix using model.chat, but it's still returning [object Object][object Object] instead of the expected string.

It seems like the response from model.chat is not being parsed correctly, as it returns:
titleResponse 
{role: 'assistant', content: '[object Object][object Object]'}
Whereas model.complete correctly returns a string with the title.

This suggests that either the model's chat response structure differs, or additional parsing is needed when using model.chat. Do you have any insights on why this might be happening? I can explore further based on that.

RomneyDa:
I don't have an immediate answer to why `chat` is returning `{role: 'assistant', content: '[object Object][object Object]'}
` but let's debug that before switching to `complete`. If you put a breakpoint in `chat` for whichever provider class you're using (or BaseLLM) you should be able to walk through and find this. I haven't been able to duplicate on my end, would you be able to try that breakpoint?

RomneyDa:
@sh4shii  looks like could be one of two lines. Probably BaseLLM's chat function:
```
async chat(
    messages: ChatMessage[],
    signal: AbortSignal,
    options: LLMFullCompletionOptions = {},
  ) {
    let completion = "";
    for await (const chunk of this.streamChat(messages, signal, options)) {
      completion += chunk.content;
    }
    return { role: "assistant" as const, content: completion };
  }
```

In `completion += chunk.content`, `chunk.content` is MessageContent, which can be an array of parts. I might try that first

Same issue actually shows up in `ChatDescriber`
>removeQuotesAndEscapes(titleResponse.content.toString());

where `content` can be an array

RomneyDa:
@sh4shii done here https://github.com/continuedev/continue/pull/5425
Added you as coauthor on the commit, thanks for the contribution!

PR #4790, opened 2025-03-24
Forbid file mentioning for large files

Jazzcort:
## Description

When the context items provided by user overflow the context length and are truncated, it is very confusing for the user. This is a first step that catches an common case at the least confusing time - when the user is selecting a file. A more general check that warns whenever the most recent message is truncated is something that can be added later.
https://github.com/Granite-Code/granite-code/issues/22

## Screenshots

large-file-mentioning

## Testing instructions

Select a file that is large enough to exceed `contextLength` - `maxTokens`, then the mention should be canceled and a error dialog should pop up.

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -22,6 +22,7 @@ import { DataLogger } from "./data/log";
 import { streamDiffLines } from "./edit/streamDiffLines";
 import { CodebaseIndexer, PauseToken } from "./indexing/CodebaseIndexer";
 import DocsService from "./indexing/docs/DocsService";
+import { countTokens } from "./llm/countTokens";
 import Ollama from "./llm/llms/Ollama";
 import { createNewPromptFileV2 } from "./promptFiles/v2/createNewPromptFile";
 import { callTool } from "./tools/callTool";
@@ -887,6 +888,26 @@ export class Core {
       }
 
       return { contextItems };
+    });
+
+    on("isItemTooBig", async ({ data: { item, selectedModelTitle } }) => {
+      const { config } = await this.configHandler.loadConfig();
+
+      if (!config) {
+        return false;
+      }
+
+      const llm = await this.configHandler.llmFromTitle(selectedModelTitle);
+
+      // Count the size of the file tokenwise
+      const tokens = countTokens(item.content);
+
+      // File exceeds context length of the model
+      if (tokens > llm.contextLength - llm.completionOptions!.maxTokens!) {
+        return true;
+      }
+
+      return false;
     });
   }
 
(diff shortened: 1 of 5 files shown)


gui/src/components/mainInput/AtMentionDropdown/index.tsx
```diff
@@ -243,7 +302,52 @@ const AtMentionDropdown = forwardRef((props: AtMentionDropdownProps, ref) => {
     }
 
     if (item) {
-      props.command({ ...item, itemType: item.type });
+      if (item.type === "file" && item.query) {
+        console.log(item);
+        isItemTooBig(item.type, item.query).then(([fileExceeds, fileSize]) => {
```
sestinj: I think this callback could be broken out into a separate function to reduce the complexity of this logic

core/core.ts
```diff
@@ -893,6 +894,26 @@ export class Core {
 
       return { contextItems };
     });
+
+    on("isItemTooBig", async ({ data: { item, selectedModelTitle } }) => {
+      const { config } = await this.configHandler.loadConfig();
+
+      if (!config) {
+        return true;
```
sestinj: I want to be very careful that we don't accidentally return `true`, so I think this should default to `false`. The consequences are pretty bad if we accidentally return `true`.

gui/src/components/mainInput/AtMentionDropdown/index.tsx
```diff
@@ -157,6 +178,44 @@ const AtMentionDropdown = forwardRef((props: AtMentionDropdownProps, ref) => {
...
+        name,
+        query,
+        fullInput: "",
+        selectedCode,
+        selectedModelTitle: selectedModelTitle ?? "",
+      },
+    );
+
+    if (contextResult.status === "error") {
+      return [true, -1];
```
sestinj: same here, we should default to `false`

gui/src/components/mainInput/AtMentionDropdown/index.tsx
```diff
@@ -157,6 +178,44 @@ const AtMentionDropdown = forwardRef((props: AtMentionDropdownProps, ref) => {
...
+    const item = contextResult.content[0];
+
+    // Check if the context item exceeds the context length of the selected model
+    const result = await ideMessenger.request("isItemTooBig", {
+      item,
+      selectedModelTitle: selectedModelTitle,
+    });
+
+    if (result.status === "error") {
+      return [true, -1];
```
sestinj: here too

sestinj (changes requested):
Generally looks good. I left some comments that should be pretty quickly resolvable

core/core.ts
Jazzcort: I spent some time considering whether we should return `true` or `false` for these situations and I was still not sure when I submitted this PR. I agree that this might be too aggressive to return `true` by default.

Jazzcort:
Hey @sestinj, I’ve pushed the updated version. Let me know if any further changes are needed for this PR.

sestinj (approved):

sestinj:
It looks good to me, thanks!

PR #4812, opened 2025-03-25
Use fetchwithRequestOptions for all calls in WatsonX.

splatch:
## Description

Small adjustment to rely on fetchWithRequestOptions rather than native node fetch call.
Not sure if it was left purposely, but it seems to be a miss from overall intention to rely on @continuedev/fetch package for all LLM calls.

## Checklist

## Screenshots

No UI changes

## Testing instructions

While this PR should not change the behavior of system, I am not quite sure how to test it as I have no access to WatsonX.

### DIFF
--- a/core/llm/llms/WatsonX.ts
+++ b/core/llm/llms/WatsonX.ts
@@ -21,7 +21,7 @@ class WatsonX extends BaseLLM {
     if (this.apiBase?.includes("cloud.ibm.com")) {
       // watsonx SaaS
       const wxToken = await (
-        await fetch(
+        await this.fetch(
           `https://iam.cloud.ibm.com/identity/token?apikey=${this.apiKey}&grant_type=urn:ibm:params:oauth:grant-type:apikey`,
           {
             method: "POST",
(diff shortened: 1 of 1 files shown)


RomneyDa:
Nice, somehow missed this on https://github.com/continuedev/continue/pull/4581

Patrick-Erichsen (approved):

Patrick-Erichsen:
This looks good to merge, but we also don't have any watsonx accounts to test on.

@ajshedivy - mind pulling this down to verify?

ajshedivy (commented):
@Patrick-Erichsen @splatch  This looks good to me! We use @continuedev/fetch package for the LLM call in `streamChat`: 

https://github.com/continuedev/continue/blob/075b785e192a9f992b14533908c6bade6fa540a1/core/llm/llms/WatsonX.ts#L230

We opted to use native fetch for the authentication process since it was unique to watsonx and didn't need the additional capabilities of Continue custom fetch. But the changes here help keep fetch calls consistent in the extension which is a better design.

PR #4813, opened 2025-03-25
Carry over requestOptions from config to completion modles

splatch:
## Description

The configuration handling of models and completion models is fairly similar, however they differ in carrying over request options. This PR attempts to align both parts to have common logic which accepts config level request options to be taken into account.

## Checklist

## Screenshots

No UI changes.

## Testing instructions

With this PR request options set at the config level should be considered also at tabCompletionModel level.


sestinj:
@splatch before looking deeper into this PR, it would be helpful to understand how you're using requestOptions—right now we don't support top-level requestOptions in `config.yaml` (https://docs.continue.dev/reference) and we don't currently have plans to. If that continues to be the case, I don't think we would want to merge this, given we are eventually going to move toward YAML format

splatch:
Hello @sestinj, reason for above changes was inability to get `requestOptions` for completion models. If you will have a look again on the change set there is one extra line: `applyRequestOptionsToModels(tabAutocompleteModels, config);`.
Without it you can't pass any options to completion models. So even if that part of configuration relies on same `ModelDescription` as models, it doesn't work.

I was stuck with 1.0.2 release of jetbrains plugin which worked with config.json, thus I started my patchwork from that place.

sestinj (approved):

splatch:
Hah, nothing special. Got a ollama instance behind nginx with self signed ssl certificate in front. Intellij have had hard times reaching it without `requestOptions` set.

sestinj:
Hi @splatch, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

splatch:
I have read the CLA Document and I hereby sign the CLA.

PR #4816, opened 2025-03-25
Show warning message when last user input get pruned

Jazzcort:
## Description

If the user's last input is pruned due to context overflow, a warning message will be displayed in the chat section, alerting them that some details may have been lost. As a result, the response they receive might be incomplete or inaccurate due to the truncated input.
https://github.com/Granite-Code/granite-code/issues/22

## Screenshots

show-warning-in-chat

## Testing instructions

Set the model’s context length to a small value (e.g., 512) and continue asking questions until the limit is reached. Once exceeded, a warning message will appear at the bottom of the chat section, indicating that some input may have been truncated. Deleting previous messages will remove the warning.

### DIFF
--- a/gui/src/redux/thunks/streamNormalInput.ts
+++ b/gui/src/redux/thunks/streamNormalInput.ts
@@ -9,6 +9,7 @@ import {
   addPromptCompletionPair,
   selectUseTools,
   setToolGenerated,
+  setWarning,
   streamUpdate,
 } from "../slices/sessionSlice";
 import { ThunkApiType } from "../store";
@@ -60,6 +61,7 @@ export const streamNormalInput = createAsyncThunk<
     );
 
     // Stream response
+    let warningMessage = "";
     let next = await gen.next();
     while (!next.done) {
       if (!getState().session.isStreaming) {
@@ -67,9 +69,21 @@ export const streamNormalInput = createAsyncThunk<
         break;
       }
 
-      dispatch(streamUpdate(next.value));
+      const filteredMessages: ChatMessage[] = [];
+
+      next.value.forEach((msg) => {
+        if (msg.role === "warning") {
+          warningMessage += msg.content;
+        } else {
+          filteredMessages.push(msg);
+        }
+      });
+
+      dispatch(streamUpdate(filteredMessages));
       next = await gen.next();
     }
+
+    dispatch(setWarning(warningMessage));
 
     // Attach prompt log
     if (next.done && next.value) {
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -321,7 +321,8 @@ export type ChatMessageRole =
   | "assistant"
   | "thinking"
   | "system"
-  | "tool";
+  | "tool"
+  | "warning";
 
 export type TextMessagePart = {
   type: "text";
@@ -385,12 +386,18 @@ export interface SystemChatMessage {
   content: string;
 }
 
+export interface WarningChatMessage {
+  role: "warning";
+  content: string;
+}
+
 export type ChatMessage =
   | UserChatMessage
   | AssistantChatMessage
   | ThinkingChatMessage
   | SystemChatMessage
-  | ToolResultChatMessage;
+  | ToolResultChatMessage
+  | WarningChatMessage;
 
 export interface ContextItemId {
   providerTitle: string;
--- a/core/llm/countTokens.ts
+++ b/core/llm/countTokens.ts
@@ -27,9 +27,9 @@ class LlamaEncoding implements Encoding {
 }
 
 class NonWorkerAsyncEncoder implements AsyncEncoder {
-  constructor(private readonly encoding: Encoding) { }
+  constructor(private readonly encoding: Encoding) {}
 
-  async close(): Promise<void> { }
+  async close(): Promise<void> {}
 
   async encode(text: string): Promise<number[]> {
     return this.encoding.encode(text);
@@ -243,12 +243,14 @@ function pruneChatHistory(
   chatHistory: ChatMessage[],
   contextLength: number,
   tokensForCompletion: number,
-): ChatMessage[] {
+): [ChatMessage[], boolean] {
   let totalTokens =
     tokensForCompletion +
     chatHistory.reduce((acc, message) => {
       return acc + countChatMessageTokens(modelName, message);
     }, 0);
+
+  let shouldWarn = false;
 
   // 0. Prune any messages that take up more than 1/3 of the context length
   const longestMessages = [...chatHistory];
@@ -275,6 +277,7 @@ function pruneChatHistory(
       content,
     );
     totalTokens -= delta;
+    shouldWarn = true;
   }
 
   // 1. Replace beyond last 5 messages with summary
@@ -327,9 +330,10 @@ function pruneChatHistory(
       tokensForCompletion,
     );
     totalTokens = contextLength;
+    shouldWarn = true;
   }
 
-  return chatHistory;
+  return [chatHistory, shouldWarn];
 }
 
 function messageIsEmpty(message: ChatMessage): boolean {
@@ -367,6 +371,7 @@ function chatMessageIsEmpty(message: ChatMessage): boolean {
         !message.toolCalls
       );
     case "thinking":
+    case "warning":
     case "tool":
       return false;
   }
@@ -381,11 +386,16 @@ function compileChatMessages(
   prompt: string | undefined = undefined,
   functions: any[] | undefined = undefined,
   systemMessage: string | undefined = undefined,
-): ChatMessage[] {
+): [ChatMessage[], boolean] {
   let msgsCopy = msgs
     ? msgs
-      .map((msg) => ({ ...msg }))
-      .filter((msg) => !chatMessageIsEmpty(msg) && msg.role !== "system")
+        .map((msg) => ({ ...msg }))
+        .filter(
+          (msg) =>
+            !chatMessageIsEmpty(msg) &&
+            msg.role !== "system" &&
+            msg.role !== "warning",
+        )
     : [];
 
   msgsCopy = addSpaceToAnyEmptyMessages(msgsCopy);
@@ -445,7 +455,7 @@ function compileChatMessages(
     }
   }
 
-  const history = pruneChatHistory(
+  const [history, shouldWarn] = pruneChatHistory(
     modelName,
     msgsCopy,
     contextLength,
@@ -459,7 +469,7 @@ function compileChatMessages(
 
   const flattenedHistory = flattenMessages(history);
 
-  return flattenedHistory;
+  return [flattenedHistory, shouldWarn];
 }
 
 export {
@@ -470,6 +480,5 @@ export {
   pruneLinesFromTop,
   pruneRawPromptFromTop,
   pruneStringFromBottom,
-  pruneStringFromTop
+  pruneStringFromTop,
 };
-
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -195,11 +195,11 @@ export abstract class BaseLLM implements ILLM {
         options.completionOptions?.maxTokens ??
         (llmInfo?.maxCompletionTokens
           ? Math.min(
-            llmInfo.maxCompletionTokens,
-            // Even if the model has a large maxTokens, we don't want to use that every time,
-            // because it takes away from the context length
-            this.contextLength / 4,
-          )
+              llmInfo.maxCompletionTokens,
+              // Even if the model has a large maxTokens, we don't want to use that every time,
+              // because it takes away from the context length
+              this.contextLength / 4,
+            )
           : DEFAULT_MAX_TOKENS),
     };
     this.requestOptions = options.requestOptions;
@@ -766,7 +766,18 @@ export abstract class BaseLLM implements ILLM {
 
     completionOptions = this._modifyCompletionOptions(completionOptions);
 
-    const messages = this._compileChatMessages(completionOptions, _messages);
+    const [messages, shouldWarn] = this._compileChatMessages(
+      completionOptions,
+      _messages,
+    );
+
+    if (shouldWarn) {
+      yield {
+        role: "warning",
+        content:
+          "The context has reached its limit. This may lead to less accurate or incomplete answers.",
+      };
+    }
 
     const prompt = this.templateMessages
       ? this.templateMessages(messages)
@@ -840,7 +851,6 @@ export abstract class BaseLLM implements ILLM {
             signal,
             completionOptions,
           )) {
-
             if (chunk.role === "assistant") {
               completion += chunk.content;
               yield chunk;
@@ -948,7 +958,7 @@ export abstract class BaseLLM implements ILLM {
     );
   }
 
-  protected async * _streamComplete(
+  protected async *_streamComplete(
     prompt: string,
     signal: AbortSignal,
     options: CompletionOptions,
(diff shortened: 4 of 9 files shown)


RomneyDa:
@Jazzcort this could be a great addition, could you explore solutions that avoid injecting a new `warning` message type into the chat messages, along with a more subtle warning UI?

RomneyDa (changes requested):
@Jazzcort this could be a great addition, could you explore solutions that avoid injecting a new warning message type into the chat messages, along with a more subtle warning UI?

Jazzcort:
@RomneyDa I'll try to find another way to send the warning message back to the webview.

Jazzcort:
I've already implemented this second approach in the following branch: [Jazzcort/warn-when-truncate-last-msg-v2](https://github.com/Granite-Code/continue-for-granite/tree/Jazzcort/warn-when-truncate-last-msg-v2).

Instead of sending the warning through the stream, I used the messenger system to deliver the warning message. With this approach, I make the pruning behavior occur before calling `streamChat` so I can reach the messenger reference. The advantage of this approach is that users receive the warning message before the streamed response rather than after it has finished.

Regarding the "warning:" field, are you suggesting adding it to AssistantChatMessage? I think that could work as well! I'm open to either approach—whichever aligns better with the project's design. We can also discuss the UI implementation afterward.

owtaylor:
Another idea would be to extend @Jazzcort's last approach and have two separate calls from webview => core.
 
1. llm/pruneChat => {prunedMessages, warning?: string} 
2. llm/streamChat

or something like that. (llm.streamChat could still prune itself when it's not being invoked from the chatview.)

This would avoid having to worry about the interaction between warnings and streaming. It could also be potentially useful for some other things:

 * prefix-caching sensitive pruning (project writeup: https://github.com/Granite-Code/granite-code/issues/96)
 * having some way to reveal the pruned messages in the UI. I'm not at all sure that this is a good idea - the user can look at logs - but I do feel that it can be deceptive to have a rich chat history with no indication that only a tiny fraction of it might be actually be sent to the model.

Jazzcort:
I agree with @owtaylor's suggestion. Calling llm/pruneChat before llm/streamChat not only helps manage context length but also provides control over whether llm/streamChat is called at all. Users might not focus too much on the last response when they see the warning message.

Additionally, leveraging [Context stability - taking advantage of prefix caching](https://github.com/Granite-Code/granite-code/issues/96) is a great strategy. It can enhance the user experience by reducing response time when the context limit is reached. @RomneyDa If everything sounds great, I'll start working on it.

RomneyDa:
@Jazzcort @owtaylor after some thought would agree that two separate calls is a good approach, because it won't effect all the other uses of `streamChat` in core, etc. @Jazzcort would agree with `llm/pruneChat`, and I'd be interested in approaches to this that don't do the pruning twice, but also use the same `streamChat` function. Perhaps some kind of `alreadyPruned` boolean passed to `llm.streamChat`. Counting tokens can be a bit expensive but not bad.

@sestinj tagging since touches core streaming

Jazzcort:
I'm planing to  implement `llm/compileChat` which call `compileChatHistory` to prune the chat messages we pass and return the compiled chat messages and a boolean that indicates whether the compiled chat messages are pruned or not. So, we can use this boolean value to decide if we need to warn users about the pruning. I'll also add a boolean parameter to `llm/streamChat` so we won't do the pruning twice. How do you guys think? @owtaylor @RomneyDa @sestinj

core/core.ts
```diff
@@ -474,6 +482,48 @@ export class Core {
...
+
+      options.log = undefined;
+      const completionOptions: CompletionOptions = mergeJson(
+        model.completionOptions,
+        options,
+      );
+
+      if (completionOptions.model === "o1") {
+        completionOptions.stream = false;
+      }
```
owtaylor: Hmm, duplicating this type of logic into two places is going to make things difficult to maintain in the future. I would suggest making model._compileChatMessages() public [and not _-prefixed] and using that. You could make it call this._modifyCompletionOptions() itself as well, since calling that multiple times shouldn't hurt.

gui/src/redux/thunks/streamNormalInput.ts
```diff
@@ -39,6 +40,34 @@ export const streamNormalInput = createAsyncThunk<
...
+    if (res.status === "error") {
+      throw new Error(res.error);
+    }
+
+    const { compiledChatMessages, lastMessageTruncated } = res.content;
+
+    if (lastMessageTruncated) {
+      dispatch(
+        setWarningMessage(
+          "The context has reached its limit. This may lead to less accurate answers.",
```
owtaylor: "the context has reached its limit" sounds like the entire chat is too long. Maybe:

 The provided context items are too large. They have been truncated to fit within the model's context length.

It's a bit verbose but:

 - Say more specifically what happened
 - Contain the phrase "context length" so that if the user searches they can find out what the context length is.

owtaylor (commented):
Looks pretty good to me. Just two suggestions.

core/core.ts
```diff
@@ -474,6 +476,24 @@ export class Core {
       }
     });
 
+    on("llm/compileChat", async (msg) => {
+      const { title: modelName, messages, options } = msg.data;
+      const model = await this.configHandler.llmFromTitle(modelName);
+
+      options.log = undefined;
+      const completionOptions: CompletionOptions = mergeJson(
```
owtaylor: Hmm, thinking about it, it's probably better to have a public `llm.compileChatMessages()` that takes the LLMFullCompletionOptions, and have that and streamChat() call `_compileChatMessages()` - sorry for not suggesting that on the previous round.

owtaylor (commented):
Code looks good to me. @RomneyDa - do you still think the warning needs to be more subtle? From my perspective, when the last message is truncated, the results are unlikely to be good.

Patrick-Erichsen:
Not to make this a design by committee, but my only feedback is that I think the warning should be a yellow rather than red. Red feels a bit too scary for the severity of this warning imo.

Jazzcort:
@Patrick-Erichsen I’ve updated the warning message color to yellow—thanks for the feedback!
@RomneyDa  Let me know if you'd like to tweak the UI further. We can either make those changes now or go ahead and merge this PR and revisit the design later.

RomneyDa:
@Jazzcort we've made several changes to pruning and message compilation recently that conflict with this. I will take a look and see if I can pull your branch and resolve this.

Acknowledging a long delay on this one, it's such a sensitive part of the code that I've put it off till I have time to properly dig in

Jazzcort:
@RomneyDa No problem! Let me know if you find anything that is unclear for this feature or you need a hand on resolving the conflicts! If the conflict is too complex to resolve, I can try to re-implement this base on the newest commit. 😊

sestinj:
@Jazzcort I took a look at this and planned to fix the conflicts, but I think it's going to be much safer for either you to re-implement on top, or for Dallin to handle the merge, and he is out until start of next week. Both directions would be okay, just a trade-off of speed vs. amount of work

Jazzcort:
I'll try to reimplement this on top this week! Will ping you when I'm done. 😄

sestinj:
sounds great! thank you 🙏

Jazzcort:
I found that with the newly implemented `compileChatMessages` function, the last user input will never get deleted or truncated. For the condition that the last user input can not fit in the context length, this hard error will be raised instead.

Screenshot 2025-05-07 at 12 54 26 PM

@allanday What do you think about this?

Jazzcort:
@RomneyDa Do you think we can separate take `lastMessagesTokens` out of the calculation of `inputTokenAvailable` in core/llm/countTokens.ts, so we can handle the situation where the user's last input gets removed differently?

Jazzcort:
@RomneyDa @sestinj If we can take the last user input out of the calculation of `inputTokenAvailable`, we can have a different display for it. Here is the demo video! Let me know what you guys think about this!

https://github.com/user-attachments/assets/60799c80-9c7a-47fe-a847-e92d95ed208d

I also tried to place this display right above the input box, but since we are using relative display, there will be a gap between the message history and the input box which looks a bit buggy. I also tried to change the user input box to an absolute display, but it cause another issue that the input box might block the delete button of some messages which I think is not acceptable.


https://github.com/user-attachments/assets/2e5bfd0f-acb0-4989-8e7e-55dd72480fbc


P.S. This implementation is based on a fairly recent commit of Continue, so no complicated conflicts I think.

Jazzcort:
@RomneyDa @sestinj So we end up thinking the way we notifying users when the last user input is deleted in previous proposal is not strong enough and we probably need to provide more guidance to users so we can prevent them from getting the inaccurate answer. I think we can change the color of the delete button for the last user input and maybe add a little animation to it so this could serve as a clearer suggestion for users to delete that message.
  
  

https://github.com/user-attachments/assets/7cd111ac-f1f6-47e8-becf-3829eeabf272

RomneyDa:
@Jazzcort, sorry about the further delay on this (one of the oldest PRs at this point!!)
I've gone back and forth on this, I think that it's important that we add some indication that pruning is occurring, I think the additional message compilation message is the right approach. I think that this is such a common occurrence that I'm tentative about yellow warnings, but I do think your most recent iteration with the little yellow dot is pretty nice. 

The message shown to the user should make three points clear
- inform: context exceeds window, can't include all
- hint at temporary solution: that it only prunes the back side of the conversation
- call to action: recommend they start a new chat, maybe even give them a new session button

e.g. maybe "Chat history exceeds model's context length (XXX tokens). Old messages will not be included. Start a new session [clickable]"

Or even more concise, "Long chat, old messages are being excluded. New Session [clickable]" or similar

I like the more subtle warning. Let's keep brainstorming on ways to handle the error after system message, last message, and tools were all made non-negotiable. What that change actually did was just remove the ability for system message lines to be pruned, and account for tools tokens. The user message text content behavior was the same before, and it should probably present some kind of error but I do agree that a popup could be annoying. But it should be a bit of an edge case like huge files etc. Personally I don't think highlighting the trash icon is ideal, but I like the direction. Maybe we should just pop up the same little yellow warning with modified text "can't submit"

Thoughts?

owtaylor:
Hi Dallin - thanks for the reply here! If I understand correctly, what you are saying is:

 * You think that the expandable-yellow-dot approach is pretty good for when the top of the conversation is pruned
 * But you think that when the last message is pruned there should be something that is more clearly an error message
 * To avoid an annoying popup, that error message could maybe share formatting with the expanded yellow dot.

? If that's a correct interpretation, would you see it looking more error like - red instead of yellow, etc.? 

I'm also not a big fan of highlighting the trash can, because I think what the user will almost always want to do is edit the provided context, rather than starting over with a new message.

RomneyDa:
@owtaylor I think a red dot that auto shows in case of non-negotiable tokens exceeding context length and a yellow dot that only shows when clicked would be great. Any other solutions/thoughts in mind? Appreciate the iterations!

Jazzcort:
@RomneyDa Red dot for the  non-negotiable token exceeding case and yellow dot for the case when we start to prune the chat message from the top sound good to me. And we can display different error messages and instructions for users to deal with the situations.

@owtaylor If this also sounds great to you, I'll go ahead and implement this feature!

Jazzcort:
@RomneyDa Sorry for the delay. I've been waiting for the CLA thing to be cleared. I think @sestinj will manually check that out for us.
For this PR, I did tweak it a little bit. We think the expendable dot might not be that intuitive for users to click on, so now the warning messages will directly be displayed at the bottom of the last user input box. 
I added the recommended action for each situation (open config file for the non-negotiable case, start a new session for the pruning from top case). We can add more actions in the future if needed.
Let me know what do you think!

https://github.com/user-attachments/assets/b68b0731-5a02-458a-a6bc-4b2c71013832

Jazzcort:
https://github.com/user-attachments/assets/ba5ef5df-1ee6-466c-bbe7-448694ee29ff

@RomneyDa Sorry for the frequent changes. This should be the final tweak I guess. 😄
Directly showing warning messages when Continue start to prune the chat messages from top might be a little intrusive so we decide to add the expendable dot back for just this situation. For the non-negotiable case, we just show the error message directly without expendable feature.

RomneyDa:
@Jazzcort got some feedback from prod/eng. Summary:
- The red error saying "warning" seems incorrect, it's an error that will prevent proceeding and should be dismissable with an x or as a popup dialog. Feels a bit weird that it could be scrolled out of view
- Liked the warning attached to the bottom input/maybe in the notch more than at the bottom of the messages, which would probably involve keeping the non-negotiable as an error dialog. Or leave non-negotiable in messages but attach the warning super subtly to the notch/bottom. Since the warning applies to pruning of ALL messages, it should probably not be between any messages, although it kind of makes sense during streaming. Probably best just in or attached to the notch/main input
- The warning is still a bit noisy, e.g. the bouncing of the icon. This is something that would be expected to show up in many agent conversations after a while, and should be more of a tip/"just so you know" than a "warning"
- Start new session should be a button that you can click to start a new session (it is implicitly our recommendation), and the "recommended action" text should be removed/feels duplicate

Maybe lets set up a time to review this live?

Jazzcort:
@RomneyDa Thanks for the feedback! I'm wondering if we can set up a google meet to finalize the details so we don't have to go back and forth here. 😄 @allanday and I are thinking about Wednesday or Thursday at 9 or 10 am (Pacific Time) next week. What do you think?

RomneyDa:
@Jazzcort how about 10 am PST Thursday/Tomorrow?

Jazzcort:
@RomneyDa Sounds good! I just sent you an invitation! Let me know if you received it. Looking forward to seeing you! 😄

RomneyDa:
@Jazzcort since it's hard to iterate on UI design remotely, we're thinking let's merge this so that you don't have to keep fixing merge conflicts and then we can do a quick iteration on the easier UI choice parts and then get a more effective feedback loop going.

Would you be able to get tests passing/conflicts fixed on this one more time and then we can go ahead and merge?

I can also chat briefly tomorrow morning. I did not receive an invite, can you double check?

Jazzcort:
@RomneyDa Sounds like a good idea! I'll open another issue for us to iterate the UI design for this. There seems to be some flaky errors happening for the tests, but I'm not 100 % sure if it's okay or not since I'm not an expert for this. Let's me know if anything need to be done to pass the failing tests. 

I sent the invitation to the email that you use for git commit. I hope that's the right email. 😅 You should see the invitation from the my redhat corporate email. Let me know if you still can't find the invitation.

Jazzcort:
@RomneyDa removed the bouncing effect. It was very nice talking to you! Thanks for joining the meeting. 😄

RomneyDa:
@Jazzcort test seems to be failing persistently, any ideas? Thanks for opening the UI issue!

Jazzcort:
@RomneyDa Looking into it rn, I’ll update you soon! 😄

Jazzcort:
@RomneyDa I found the issue — it looks like we also need to declare a dummy "llm/compileChat" handler in `MockIdeMessenger.ts` to avoid errors when clicking the send button during tests. This happens because "llm/compileChat" is wrapped in every stream input. I'll make this change and push it shortly.

RomneyDa:
@Jazzcort would like to get this merged, looks like some more conflicts came up, could you take a look? I am not able to push to this branch to help out.

Jazzcort:
@RomneyDa No problem! Will ping you when I'm done

Jazzcort:
@RomneyDa Everything looks good to me—just a heads-up that `jetbrains-test` is still failing. 😄

RomneyDa:
@Jazzcort I think latest main should have a fix for the jetbrains failing test

Jazzcort:
@RomneyDa It's ready to be merged now! 😄

RomneyDa (approved):
Addition of a pruning step that allows warning and showing context window

PR #4820, opened 2025-03-25
Add gemini-2.5-pro to model list

tgmerritt:
## Description

Added gemini-2.5-pro to the model list

## Testing instructions

Add gemini-2.5-pro to your model list
Select it from the drop down in the Continue chat tray
Make a model request (you will need Pro access which requires a paid subscription)


tgmerritt:
image

Google's API is currently saying model not found, but this was announced only a few hours ago by them, so we expect it will be added shortly to their model availability list.

RomneyDa:
@tgmerritt looks like I broke the json when trying to quick fix merge conflicts. taking another look now

RomneyDa (approved):

Issue #4890, opened 2025-03-28
Bring back "Insert at cursor" on the new codeblock tool bar


sestinj:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

On the alternate/older toolbar, there are 3 options: copy, insert at cursor, and apply. The newer one is missing "insert at cursor" and some users have requested for it to be added back

Patrick-Erichsen:
+1

PR #4936, opened 2025-04-01
feat: add insert at cursor functionality with tooltip

aadarshkt:
## Description

Fixes : #4890
Added insert at cursor with help of existing codebase.

## Checklist

## Screenshots
Screenshot 2025-04-01 at 23 53 27
Screenshot 2025-04-01 at 23 53 38

## Testing instructions


Patrick-Erichsen:
Thanks for taking this on @aadarshkt ! 

Some thoughts on the layout here:
- Of the three options, apply is what we want to push people towards
- Copy/Insert at cursor are both essentially backups/fallbacks

With that in mind, I think the following UI would work well:
- Order of the buttons, from left to right, is "Copy", "Insert at cursor", "Apply"
- Only "Apply" gets a label, "Copy" and "Insert at cursor" are both just icon buttons

If we go this route, we should probably re-order the the other action buttons (can't recall where they are in the GUI folder) to also follow this same order (see screenshot below)

Screenshot 2025-04-03 at 9 02 15 AM

Lmk your thoughts 👍

aadarshkt:
@Patrick-Erichsen, Thanks for the review, 

Yeah seems right to me as well, users will remember more what was first and last in the icon list, thus it is better to keep frequently used icons there.

I will make a change at both places.

aadarshkt:
Made a update in the UI, 
Screenshot 2025-04-04 at 10 06 49

Will look into tooltipplacement, it is controlled by global state it seems.

gui/src/components/markdown/StepContainerPreToolbar/StepContainerPreToolbar.tsx
```diff
@@ -79,6 +86,12 @@ export default function StepContainerPreToolbar(
     selectApplyStateByStreamId(state, streamIdRef.current),
   );
 
+  const uiConfig = useAppSelector(selectUIConfig);
+  const isBottomToolbarPosition =
+    uiConfig?.codeBlockToolbarPosition == "bottom";
+
+  const toolTipPlacement = isBottomToolbarPosition ? "top" : "bottom";
```
tomasz-stefaniak, line 101: Is all of this logic necessary given that the toolbar position is either top or bottom?
https://github.com/aadarshkt/continue/blob/fix-insert-at-cursor/core/index.d.ts#L1099

How about something like the following instead?

```
              <HeaderButtonWithToolTip
                text="Insert at cursor"
                style={{ backgroundColor: vscEditorBackground }}
                onClick={() =>
                  ideMessenger.post("insertAtCursor", {
                    text: codeBlockContent,
                  })
                }
                tooltipPlacement={uiConfig?.codeBlockToolbarPosition ?? "top"}
              >
 ```

Patrick-Erichsen:
Closing this out as we merged in similar logic with https://github.com/continuedev/continue/pull/5161 - thanks for the help @aadarshkt , we pulled in some of your work!

PR #4957, opened 2025-04-02
LS Tool

RomneyDa:
## Description
- Adds LS tool which lists files/directories in a given relative directory, respecting continue ignore rules
- Removes repo map and viewSubdirectory tools for now
- adds `recursive` option to `walkDir` and changes `onlyDirs` to be able to output "both", "dirs", or "files"
- rearrange tool policies into notch toolbar folder
- Fix bug where boolean tool args were not displaying


sestinj (approved):
thumbs up as long as claude is reliably able to use this in some testing

PR #4978, opened 2025-04-04
Edit tool + Create tool tweaks

RomneyDa:
## Description
- Adds lazy apply with `updateApplyState` messages for Jetbrains
- Edit tool with custom tool call UI and accept/reject buttons
- Disables submitting tip tap editors while this apply states are pending (for now)
- error if attempting to create file that already exists. This is important because right now it overwrites in a way that is not undo-able, and this is bad regardless of if an edit tool is available
- Both create and edit tool call UIs hide the codeblock toolbar apply button
- Fixes generating logic error causing multiple codeblocks to show generating

Originally (see [4683a20](https://github.com/continuedev/continue/pull/4978/commits/4683a20c2fbc0c70fe5457221743bd6a15749a3b)) I enabled core to initiate `applyToFile` within a core-side edit tool implementation, and generated a stream id on GUI that was passed to core and then around from there. The main issue with this was a lack of stateful awareness of diffs in core before passing tool output back to model. I found this to be a bad user experience.

I ended up deciding that it's okay to have a custom Edit tool implementation in GUI that bypasses the other tools. This opens up channels for potential tools like a "new session" tool that would also be GUI-only. I added `toolCallId` to Apply state so that apply streams can optionally be associated with tool calls.


core/tools/callTool.ts
```diff
@@ -138,6 +138,9 @@ export async function callTool(
   switch (uri) {
     case BuiltInToolNames.ReadFile:
       return await readFileImpl(args, extras);
+    // Custom GUI handling for edit
```
Patrick-Erichsen: Delete?

core/tools/implementations/createNewFile.ts
```diff
@@ -1,15 +1,34 @@
...
 export const createNewFileImpl: ToolImpl = async (args, extras) => {
   const resolvedFileUri = await inferResolvedUriFromRelativePath(
     args.filepath,
     extras.ide,
   );
   if (resolvedFileUri) {
+    const exists = await extras.ide.fileExists(resolvedFileUri);
+    if (exists) {
+      throw new Error(
+        `File ${args.filepath} already exists. Use the edit too to edit this file`,
```
Patrick-Erichsen:
```suggestion
        `File ${args.filepath} already exists. Use the edit tool to edit this file`,
```

gui/src/pages/gui/ToolCallDiv/CreateFile.tsx
```diff
@@ -7,9 +7,13 @@ interface CreateFileToolCallProps {
 }
 
 export function CreateFile(props: CreateFileToolCallProps) {
-  const src = `\`\`\`${getMarkdownLanguageTagForFile(props.relativeFilepath ?? "test.txt")} ${props.relativeFilepath}\n${props.fileContents ?? ""}\n\`\`\``;
+  const src = `\`\`\`${getMarkdownLanguageTagForFile(props.relativeFilepath ?? "output.txt")} ${props.relativeFilepath}\n${props.fileContents ?? ""}\n\`\`\``;
 
   return props.relativeFilepath ? (
-    <StyledMarkdownPreview isRenderingInStepContainer={true} source={src} />
+    <StyledMarkdownPreview
+      isRenderingInStepContainer={true}
```
Patrick-Erichsen, line 14: Nitpick but in general I prefer 
```suggestion
      isRenderingInStepContainer
```

gui/src/pages/gui/ToolCallDiv/EditFile.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (!props.relativeFilePath) {
+    return null;
+  }
+
+  return (
+    <>
+      <StyledMarkdownPreview
+        isRenderingInStepContainer={true}
+        source={src}
+        disableManualApply={true}
```
Patrick-Erichsen, line 34: It feels like a codesmell that in the general  StyledMarkdownPreview we're passing down props that are specific to the toolbar. Not sure how to get around this but I worry about this growing more and more unwieldly

Patrick-Erichsen (approved):

sestinj (approved):

PR #4979, opened 2025-04-04
Create tool tweaks

RomneyDa:
- error if attempting to create file that already exists. This is important because right now it overwrites in a way that is not undo-able, and this is bad regardless of if an edit tool is available
- hide apply button


RomneyDa:
done on #4978

PR #4982, opened 2025-04-04
feat: prompt blocks

Patrick-Erichsen:


RomneyDa (changes requested):
when doing e.g. command delete the prompt-block node isn't removed along with the slash command one

image

appears model doesn't ever call tools when using slash command, as well as same codeblocks ignored bug you found. 

`.continue/rules` not supported 

imo should always insert `prompt-block` at index 0, inserting at end feels off because slash command is anchored to beginning of editor so prompt-block should be too, and adding additional codeblocks inserts them in the middle which feels odd

image

RomneyDa (approved):

alicezhangyuan:
is Da's issue solved? I meet the same problem. put some prompt in file, and use slash command, but it seems the model do  not receive the info in prompt file. @Patrick-Erichsen

PR #4999, opened 2025-04-05
fix pr_checks.yaml linting

uinstinct:
## Description

comment out the empty lists in github actions pr_check.yaml config's `require-all-checks-to-pass` job. so that it becomes a valid config

## Checklist

#### before

image

#### after

image

## Testing instructions

### DIFF
--- a/.github/workflows/pr_checks.yaml
+++ b/.github/workflows/pr_checks.yaml
@@ -632,6 +632,6 @@ jobs:
       - name: Decide whether the needed jobs succeeded or failed
         uses: re-actors/alls-green@release/v1
         with:
-          allowed-failures:
-          allowed-skips:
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen:
Are you sure there isn't some side effect of having those as empty values?

If so, let's just delete them altogether rather than commenting out.

uinstinct:
> Are you sure there isn't some side effect of having those as empty values?
> 
> If so, let's just delete them altogether rather than commenting out.

yes, there will be no side effects. by default those options are empty arrays.

Patrick-Erichsen (approved):

Patrick-Erichsen:
Great, thanks for the followup and for addressing this!

PR #5013, opened 2025-04-06
Config Loading and Org/Profile Selection Fixes

RomneyDa:
## Description
- Changes org/profile loading to a safer single-vertical cascading pattern
- Fixes various bugs with assistant and org selection race conditions etc.
- all orgs/profiles loaded on startup unless manually refreshed
- With new data structure, unifies org/profile redux slices
- Adds refresh all button to assistant select list
- Clicking the org in the assistant listbox cycles orgs rather than navigating away


sestinj (approved):
Works great. I looked closely mostly at `ConfigHandler`, and spent the rest of my time testing. Only odd case I found is when you click the org name in the assistant switcher popover and it cycles to an org without profiles, then the popover disappears. Obviously doesn't need to be fixed for this PR

PR #5014, opened 2025-04-07
Support fnm in the install-all-dependencies task.

shermanhuman:
### Problem
I use fnm and even though I have node 20.11 installed the script doesn't respect that and attempts to use my default node version.  

### Proposed Solution
Add support for fnm to the install-all-dependencies script.

## Checklist

- []~~The relevant docs, if any, have been updated or created~~
- [] ~~The relevant tests, if any, have been updated or created~~

## Testing instructions

Run the install-all-dependencies task in VS Code.  I've tested it as an FNM user, with `.node-version`.  Haven't tested it with nvm and `.nvmrc`.  

We don't have to commit .node-version, we could add it to .gitignore instead.  That way maintainers don't have to think about updating it in two places, and just leave the logic in place for fnm users.


Patrick-Erichsen (approved):
Neat, I hadn't heard of `fnm` before - installed it myself! 👌

PR #5022, opened 2025-04-07
Hub SDK

sestinj:


sestinj:
@Patrick-Erichsen this is old and overwritten by the auto-generated SDK that Tomasz built, but tagging you here in case anything is useful as reference

PR #5046, opened 2025-04-08
fix code block preview needing to click twice to hide

uinstinct:
## Description

When selecting a code block and clicking on the hide icon, we see that code block re-renders and does not actually hide (attached video named _before_).

This is happening because we are resetting the `hidden` state again when the user toggles. We actually do not need the `userToggled` state because we only want to reset when our props change (which happens as soon as the `ExpandableToolbarPreview` component initially renders) and not when the user toggles.

## Checklist

### Before

https://github.com/user-attachments/assets/3e3a7865-ee27-45e8-92cb-af6baecbfded

### After

https://github.com/user-attachments/assets/e4d870e7-b775-4808-afb9-b46e9452b773

## Testing instructions

- Add any code to highlight in the chat section
- Click on the eye icon to hide the code
- See that code still stays shown

### DIFF
--- a/gui/src/components/mainInput/TipTapEditor/components/ExpandableToolbarPreview.tsx
+++ b/gui/src/components/mainInput/TipTapEditor/components/ExpandableToolbarPreview.tsx
@@ -118,6 +118,7 @@ export function ExpandableToolbarPreview(props: ExpandableToolbarPreviewProps) {
     props.itemId,
     props.initiallyHidden,
   ]);
+
   const [hidden, setHidden] = useState(calculatedInitiallyHidden);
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen:
Nice catch, this was a recent update that I made so thanks for cleaning up my buggy code 😅

Patrick-Erichsen (approved):

uinstinct:
No problem. I like developing and contributing to continue.

Patrick-Erichsen:
@uinstinct - our Netlify CI testing logic seems to be hanging. I'm going to try closing and re-opening to see if that fixes, but if it doesn't, the lazy solution is just to open this in a new PR we can merge.

PR #5058, opened 2025-04-08
feat: fallback to chat model from apply

Patrick-Erichsen:
## Overview
In the case a user has a misconfigured Apply model, e.g. an API key is missing, fall back to a Chat model.

image

## Testing

Add a misconfigured apply model to your local config.yaml, e.g.

```yaml
  - name: relace
    provider: relace
    model: relace
    roles: [apply]
```

- Verify that your apply model selector has fallen back to the first chat model 
- Add an `apiKey` property to the apply model yaml
- Verify we are now using the configured apply model

## Notes
This is only configured/used for Apply models at the moment. To implement the logic for other roles, we'd need to handle those cases individually.


sestinj (dismissed):
this looks good once tests are passing

RomneyDa (commented):
The logic in the listbox button that shows "Using chat model" might break if invalid models are included in list? Shows if models.length === 0

RomneyDa:
@Patrick-Erichsen from our convo

 1. add check for is valid in rectifySelectedModelsFromGlobalContext
 2. Keep it where you put it in config loading, but also add if there are zero apply models. Or even just always add (if not already in apply list)
 3. then can remove fallback logic from IDEs yay
 3. optionally find some way to indicate that it came from fallback in UI

RomneyDa (approved):

PR #5073, opened 2025-04-09
Mitigate registerCopyBufferSpy failures

fbricon:
## Description

In some cases registerCopyBufferSpy fails to register due to 

>Error activating extension: Error: command 'editor.action.clipboardCopyAction' already exists"

Screenshot 2025-04-07 at 09 49 50

And clicking on the "View Logs" button doesn't work because the catastrophique failure occurs before all commands are registered.

This apparently can happen if Continue is installed on both the host and a devcontainer, or when a Continue fork, such as Granite.Code, is used.

The hack used to hook into editor.action.clipboardCopyAction is pretty ugly, but an [alternative solution using the Paste API](https://github.com/continuedev/continue/commit/449d0b52532e1862d6b8b0eb281de80b535a3232 ) doesn't work as well as the existing solution: its support is limited to text documents. You can't intercept text copied from a webview, eg. the chat view, since it's not a TextDocument.

This PR sort of mitigates the failure:
- registerCopyBufferSpy call is moved after registering commands, so view logs still works
- registerCopyBufferSpy failure is caught and a specific message is displayed, hopefully downgrading the perceived severity of the problem. 
- the rest of the activation  can proceed after this specific failure.

However, I'm not completely happy with it as it would still show up 100% of the time, provided the conditions stay the same. So I'm inclined to add a "Don't show again" button. It could then store a flag in the global state (without an easy way to revert it), or in the user settings (but creating such setting feels wrong). We could also decide to always fail silently and only keep a trace in the logs. 

## Checklist

## Screenshots
Screenshot 2025-04-09 at 15 29 51

## Testing instructions

- Install [Granite.Code](https://marketplace.visualstudio.com/items?itemName=redhat.granitecode)
- Open Granite.Code's chat view
- Install Continue
- Restart  

cc @owtaylor @allanday

### DIFF
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -382,6 +382,7 @@ const getCommandsMap: (
       range,
     );
   }
+
   return {
     "continue.acceptDiff": async (newFileUri?: string, streamId?: string) =>
       processDiff(
@@ -950,7 +951,7 @@ const getCommandsMap: (
           if (core.configHandler.currentProfile?.profileDescription.id) {
             core.invoke("config/updateSelectedModel", {
               profileId:
-                core.configHandler.currentProfile?.profileDescription.id,
+              core.configHandler.currentProfile?.profileDescription.id,
               role: "autocomplete",
               title: selectedOption,
             });
@@ -1129,7 +1130,7 @@ export function registerAllCommands(
     //Non-critical error, it needs to be intercepted and not prevent the extension from starting
     console.log("Error registering CopyBufferService: ", e);
     Telemetry.capture(
-      "vscode_extension_activation_error",
+      "vscode_extension_copy_buffer_failure",
       {
         stack: extractMinimalStackTraceInfo(e.stack),
         message: e.message,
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/commands.ts
```diff
@@ -1124,4 +1123,28 @@ export function registerAllCommands(
...
     );
   }
+
+  try {
+    registerCopyBufferSpy(context, core);
+  } catch (e: any) {
+    //Non-critical error, it needs to be intercepted and not prevent the extension from starting
+    console.log("Error registering CopyBufferSpy: ", e);
+    Telemetry.capture(
+      "vscode_extension_activation_error",
```
fbricon: we'd likely need a different event id here

sestinj (dismissed):

sestinj (commented):
While we're here, I think it would be reassuring to name this as something less ominous for what it is, perhaps `CopyBufferService`

fbricon:
> While we're here, I think it would be reassuring to name this as something less ominous for what it is, perhaps `CopyBufferService`

done

RomneyDa:
@fbricon 
>This apparently can happen if Continue is installed on both the host and a devcontainer, or when a Continue fork, such as Granite.Code, is used.

Interesting if you've had any other ideas on how to hook into clipboard events in a way that this won't break?

RomneyDa (dismissed):

fbricon:
A "safer" (probably), but uglier (definitely) way to track clipboard changes would be to register a function that'd poll the clipboard content every X (500?)ms. That's what https://github.com/arjun-g/electron-clipboard-extended does.

sestinj:
I made a quick update so taht the warning message isn't shown as I don't think this is critical enough to warrant that. Will merge once tests are passing!

PR #5086, opened 2025-04-10
docs page for source control

sestinj:


docs/docs/hub/source-control.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Source Control
+description: Source Control
+keywords: [source, control, github, workflow, action, version, yaml]
+sidebar_label: Source Control
+---
+
+# Source Control
+
+When managing your custom assistants within an organization, you might want to take advantage of your usual source control workflows. Continue makes this easy with a [GitHub Action](https://github.com/continuedev/continue-publish-action) that automatically syncs your YAML files with hub.continue.dev.
```
TyDunn, line 10: What about many organizations who use GitLab, BitBucket, Gitee, etc. Can we call out somewhere that we plan to add support and to let us know in Discord if you are looking for this?

docs/docs/reference.md
```diff
@@ -49,6 +49,28 @@ models:
     provider: ollama
 ```
 
+### Local Blocks
+
+It is also possible to define blocks locally in a `.continue` folder. This folder can be located at either the root of your workspace (these will automatically be applied to all assistants when you are in that workspace) or in your home directory at `~/.continue` (these will automatically be applied globally).
```
amanda-wee, line 54: Could you clarify if a local block at workspace-level will override a local block at global-level, and if so, how will the override be identified (e.g., by file name or by block name)?

sestinj: They won't be overridden at all. Instead, both of the blocks will be added to the assistant. For example, if you have ~/.continue/rules/my-rule.yaml and /path/to/workspace/.continue/rules/my-rule.yaml they will both show up in your assistant

PR #5090, opened 2025-04-10
Sort models alphabetically in Chat view

fbricon:
## Description

I have a *ton of models installed on Ollama, and selecting one of them in the Chat view is a PITA, because they're not sorted alphabetically. This PR fixes that, while maintaining the existing sort where models without an API key are listed at the end.
Also applied to the Models section.

## Checklist

## Screenshots

Before: &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; After:
Screenshot 2025-04-10 at 10 19 08&nbsp; &nbsp; &nbsp; Screenshot 2025-04-10 at 10 29 41

In Models section:
Screenshot 2025-04-10 at 10 53 54

## Testing instructions

Just have several models to choose from, they should be listed alphabetically in the Chat view

### DIFF
--- a/gui/src/components/modelSelection/ModelSelect.tsx
+++ b/gui/src/components/modelSelection/ModelSelect.tsx
@@ -131,8 +131,9 @@ function ModelSelect() {
 
   // Sort so that options without an API key are at the end
   useEffect(() => {
-    const enabledOptions = options.filter((option) => option.apiKey !== "");
-    const disabledOptions = options.filter((option) => option.apiKey === "");
+    const alphaSort = options.sort((a, b) => a.title.localeCompare(b.title));
+    const enabledOptions = alphaSort.filter((option) => option.apiKey !== "");
+    const disabledOptions = alphaSort.filter((option) => option.apiKey === "");
 
     const sorted = [...enabledOptions, ...disabledOptions];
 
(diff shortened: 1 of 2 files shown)


RomneyDa:
What if we sorted by LRU first and alphabetical second?

fbricon:
Are we keeping track of each model usage?

Patrick-Erichsen:
> What if we sorted by LRU first and alphabetical second?

This sounds nice in theory but in practice I feel it might be a bit too clever. I think alphabetical is a good lazy solution, but if we want to take it further, I would probably push for something like "drag to reorder" or pinning that isn't as dynamic as LRU.

fbricon:
> I think alphabetical is a good lazy solution that would be an improvement over current state imo, but if we want to take it further, I would probably push for something like "drag to reorder" or pinning that isn't as dynamic as LRU.

that seems a bit over-engineered to my taste. I'd rather stick to "lazy" (i'd say low-hanging fruit)

RomneyDa:
quick win for now sounds good!

PR #4565, opened 2025-03-09
first pass at terminal tool output ui

sestinj:

PR #5130, opened 2025-04-12
feat: enhanced terminal tool and ui

chezsmithy:
## Description

I've taken the great work that was started in https://github.com/continuedev/continue/pull/4565 and enhanced it.

* I've added new UI elements to note the outcomes of the terminal commands with color coding.
* I've enhanced the terminal tool with a waitForCompletion option so that commands can be started in the background. This supports server commands like running express servers for example when you want to run tests against them. This status is provided to the terminal tool so the context is aware the process went into a background state.
* If a long running process is started in the foreground there is now a way to continue it in the background. The child process is not in a detached state so it is then marked as moved to the background and any further events from it are ignored. This solves for example if the terminal command is killed from a terminal by the user. Any events are ignored as they are no longer relevant.
* Terminal processes now stream their output into the terminal containers, including long running processes.
* If you use the run option to move a terminal command to the terminal in vscode just the command is copied rather than the output in the window.
* When a terminal command generates output in a streaming way to either stdout or stderr that output is accumulated into the terminal container. This supports commands that stagger their output like jest for example. It also solves for commands that return both stdout and stderr like cucumber. 

## Checklist

## Screenshots

https://github.com/user-attachments/assets/e545850e-dd45-4c41-9515-9395a162f237

## Testing instructions

Testing an endless loop.
1. Prompt - Create an endless loop using shell commands and run it.
2. You should observe the endless shell command get run in the background
3. The it should be killed.
4. Prompt- Now run the endless loop in the foreground
5. You should observe the endless shell command get run in the foreground (*Note the Continue in background option)
6. The terminal window will continue streaming.
7. Click the Continue in background option
8. The terminal will detach and you should see the interaction with the LLM continue
9. You can ask the llm to kill the loop or it may offer to do it automatically.
10. Note the endless loop is stopped with no furhter action.

Testing an endless server
1. In the sandbox ask the llm to build an express server to host @test.js
2. Once the server is built as the LLM to start it.
3. You should observe the server gets started in the background
4. You can ask the llm to stop it.
5. Now ask the LLM to start the server in the foreground
6. You should observe the "Continue in background" option is provided.
7. Click continue in background.
8. Now to to a terminal window and kill the express server
9. Note that the chat window continues normally even tho the non-detached child process exits in the background


core/core.ts
```diff
@@ -672,6 +677,14 @@ export class Core {
         throw new Error("No chat model selected");
       }
 
+      // Define a callback for streaming output updates
+      const onPartialOutput = (params: {
+        toolCallId: string,
+        contextItems: any[]
```
Patrick-Erichsen: Should this be `contextItems: ContextItem[]`?

core/core.ts
```diff
@@ -672,6 +677,14 @@ export class Core {
...
+      }) => {
+        this.messenger.send("toolCallPartialOutput" as any, params);
```
Patrick-Erichsen: Similarly here, can we define the object for this message instead of passing `any`?

core/core.ts
```diff
@@ -715,6 +730,21 @@ export class Core {
 
       return false;
     });
+
+    // Process state handlers
+    on("process/markAsBackgrounded", async ({ data: { toolCallId } }) => {
+      const { markProcessAsBackgrounded } = await import("./tools/implementations/processState");
```
Patrick-Erichsen: In both this and `on("process/isBackgrounded"`), I think it would be a bit simpler to just import the helper fns at the top of the file instead of dynamically since they aren't expensive to import.

core/protocol/coreWebview.ts
```diff
@@ -2,7 +2,10 @@ import { ToCoreFromIdeOrWebviewProtocol } from "./core.js";
 import { ToWebviewFromIdeOrCoreProtocol } from "./webview.js";
 
 export type ToCoreFromWebviewProtocol = ToCoreFromIdeOrWebviewProtocol & {
-  didChangeSelectedProfile: [{ id: string }, void];
-  didChangeSelectedOrg: [{ id: string; profileId?: string }, void];
+  didChangeSelectedProfile: [{ id: string | null }, void];
```
Patrick-Erichsen: Is there a reason these are changing in this PR?

core/tools/implementations/processState.ts
```diff
@@ -0,0 +1,17 @@
+// Track which processes have been backgrounded
+const backgroundedProcesses = new Map<string, boolean>();
```
Patrick-Erichsen: I think `processTerminalBackgroundStates` (or something similar) would be a bit clearer name for this file.

core/tools/implementations/runTerminalCommand.ts
```diff
@@ -3,35 +3,247 @@ import util from "node:util";
 
 import { fileURLToPath } from "node:url";
 import { ToolImpl } from ".";
+import { isProcessBackgrounded, removeBackgroundedProcess } from "./processState";
 
 const asyncExec = util.promisify(childProcess.exec);
 
+// Export the process state management functions from the shared module
```
Patrick-Erichsen: Is there a reason we need to do re-export from this file? Why not just let other files import directly from `processState`?

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/constants/MessageTypes.kt
```diff
@@ -69,6 +69,8 @@ class MessageTypes {
             "getWebviewHistoryLength",
             "getCurrentSessionId",
             "didCloseFiles",
+            "toolCallPartialOutput",
```
Patrick-Erichsen, line 72: Thanks for updating JetBrains as well!

gui/src/pages/gui/Chat.tsx
```diff
@@ -272,6 +273,22 @@ export function Chat() {
     [mainTextInputRef],
   );
 
+  // Handle partial tool call output for streaming updates
+  useWebviewListener(
+    "toolCallPartialOutput",
+    async (data) => {
+        console.log("toolCallPartialOutput", data);
```
Patrick-Erichsen:
```suggestion
```

gui/src/pages/gui/ToolCallDiv/RunTerminalCommand.tsx
```diff
@@ -1,16 +1,128 @@
...
+  }
+
+  return {
+    commandOutput: output,
+    statusMessage: null
+  };
 }
 
 export function RunTerminalCommand(props: RunTerminalCommandToolCallProps) {
+  const dispatch = useDispatch<ThunkDispatch<any, any, AnyAction>>();
```
Patrick-Erichsen:
```suggestion
  const dispatch = useAppDispatch();
```

gui/src/redux/slices/sessionSlice.ts
```diff
@@ -637,6 +637,23 @@ export const sessionSlice = createSlice({
 
       toolCallState.output = action.payload;
     },
+    updateToolCallOutput: (
```
Patrick-Erichsen, line 640: This seems like a broadly useful piece of logic to have around, thanks for adding this!

gui/src/pages/gui/ToolCallDiv/RunTerminalCommand.tsx
```diff
@@ -1,16 +1,128 @@
...
+          {isRunning && props.toolCallId && (
+            <BackgroundLink
+              onClick={() => {
+                // Dispatch the action to continue the command in the background
+                dispatch(continueTerminalCommand({
+                  toolCallId: props.toolCallId as string
+                }));
+              }}
+            >
+              Continue in background
```
Patrick-Erichsen: This is a bit of a nitpick but I think we should try to avoid using the phrase `Continue` to avoid confusion. What if it was "Move to background"?

gui/src/redux/thunks/continueTerminalCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+import { streamResponseAfterToolCall } from "./streamResponseAfterToolCall";
+
+/**
+ * This thunk is used to continue running a terminal command in the background
+ * when the user clicks the "Continue" button in the UI
+ * 
+ * It marks the command as visually complete without re-running it or stopping
+ * the already running process
+ */
+export const continueTerminalCommand = createAsyncThunk<
```
Patrick-Erichsen: Similar to my other comment, I think using the prefix `continue` will lead to confusion here. What about `moveProcessToBackground` or something equivalent?

gui/src/redux/thunks/continueTerminalCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+  
+  const toolCall = toolCalls[0].toolCallState;
+  if (!toolCall) {
+    console.error("Tool call state is missing");
+    return;
+  }
+  
+  // Update the tool call output to indicate it's running in background
+  const currentOutput = toolCall.output?.[0]?.content || "";
+  const updatedOutput = currentOutput + "\n[Command moved to background. Further output will be ignored.]";
```
Patrick-Erichsen: Why do we format the output in braces here? I'm wondering if this should perhaps be part of a more general util fn that does whatever conversion is required here.

Patrick-Erichsen (commented):
Wow, this is an incredible PR! Not only does this add some really useful functionality to the terminal tool but the attention to detail on the UX here is also excellent. Excited to get this merged!

We are planning to add some more functionality to the "notch" feature in our UI, and I think this will make a nice top level button once we do that to show the users an active list of what processes they have running. E.g. below is a similar implementation from Windsurf

image
image

Only other thing is that if there's any obvious ways you think we could get more test coverage in this PR, that would also be appreciated.

core/tools/implementations/processState.ts
RomneyDa, line 1: `implementations` folder is just tool implementations, maybe add a util folder or add to existing utils

core/core.ts
```diff
@@ -715,6 +730,21 @@ export class Core {
...
+      markProcessAsBackgrounded(toolCallId);
+    });
+
+    on("process/isBackgrounded", async ({ data: { toolCallId }, messageId }) => {
+      const { isProcessBackgrounded } = await import("./tools/implementations/processState");
+      const isBackgrounded = isProcessBackgrounded(toolCallId);
+      // Need to manually send response since we can't return value directly
+      this.send("process/isBackgroundedResponse", { isBackgrounded }, messageId);
```
RomneyDa: rather than doing ide.post("process/isBackgrounded") and then having the GUI receive "process/isBackgroundedResponse" later, you can just do ide.request and it will get the response back. Do it in a thunk if it shouldn't interrupt other UI

core/tools/definitions/runTerminalCommand.ts
RomneyDa, line 1: tell the model that wait for completion means it can get the output

core/protocol/coreWebview.ts
chezsmithy: I belive this got changed as I was working due to a typescript error. I've reverted it now.

gui/src/redux/slices/sessionSlice.ts
chezsmithy: Giving @sestinj the credit on this one :)

gui/src/redux/thunks/continueTerminalCommand.ts
chezsmithy: @Patrick-Erichsen it's to make the output in the command window consistent with how it's being updated in runTerminalCommand.ts and to allow the parsing to be consistent here: https://github.com/chezsmithy/continue/blob/573ce851b670e64bf4203f43273aa8b07ff0fe59/gui/src/pages/gui/ToolCallDiv/RunTerminalCommand.tsx#L60

If you might have some better ideas how to share this with whats also in https://github.com/chezsmithy/continue/blob/573ce851b670e64bf4203f43273aa8b07ff0fe59/core/tools/implementations/runTerminalCommand.ts#L98 for example, and put it into a common formatter?

core/tools/definitions/runTerminalCommand.ts
chezsmithy: @RomneyDa check me how I made this change?

chezsmithy: @RomneyDa I think I got it dialed in. I did a bit of tuning to also make it so that when a command enters the background that the LLM suggests formatted terminal commands only to clean it up, and doesn't suggest pressing Ctrl-C for example. See what you think.

gui/src/redux/thunks/continueTerminalCommand.ts
chezsmithy: @Patrick-Erichsen I refactored and added a status to the ContextItem to carry the data to display for each command. Cleans up the braces, and removes parsing that could become messy later. I added some tests to help keep things verified.

chezsmithy:
@Patrick-Erichsen and @RomneyDa this should be ready for another look.

core/tools/implementations/runTerminalCommand.test.ts
Patrick-Erichsen, line 1: Thanks for the tests here, very much appreciated!

chezsmithy: @Patrick-Erichsen as a watch list item, I attempted to make sure tests would run on windows, linux and macOS as they are spawning real commands. I attempted mocking but it just became pretty useless. Let me know if you get feedback for other contributors or your github actions with these tests and I can jump in to help.

Patrick-Erichsen: That makes sense, and will do!

RomneyDa:
@chezsmithy just a note looks like prettier didn't format code in this PR to match our rules, not critical at all but could double check that on your end. Thanks!

PR #5148, opened 2025-04-14
feat: Provide workspace path to HttpContextProvider

s-h-a-d-o-w:
## Description

While I appreciate that you generally want to discuss ideas beforehand, I figured that with things this simple, it's easier to just show the code.

I recently implemented my own indexing/search service and soon realized that unless the server knows which project is being worked on, it's not going to be usable. And so this would provide that information.

(I have restricted it to local servers because while I can't envision what an attacker might be able to do with workspace paths, I've seen it as generally frowned upon to publicly reveal information about local filesystems.
I would be fine with removing this restriction. Or to tweak things to e.g. provide only the "base name" of the workspace when the server isn't local - for possible other use cases where differentiating by project would still be beneficial.)

## Checklist

## Testing instructions

- Locally run a server of your choosing that can be used with this provider (maybe the [example](https://github.com/continuedev/continue/blob/main/core/context/providers/context_provider_server.py)?) and log the received request body.
- Whenever you use the context, you should see the workspace path getting sent - according to whatever workspace might currently be active.

### DIFF
--- a/core/context/providers/HttpContextProvider.ts
+++ b/core/context/providers/HttpContextProvider.ts
@@ -1,3 +1,4 @@
+import isLocalhost from "is-localhost-ip";
 import {
   ContextItem,
   ContextProviderDescription,
@@ -18,13 +19,8 @@ class HttpContextProvider extends BaseContextProvider {
   private async getWorkspacePath(ide: IDE, url: URL) {
     try {
       const currentFile = await ide.getCurrentFile();
-      const hostname = url.hostname.toLowerCase();
-      const isLocalServer= hostname === "localhost" ||
-             hostname === "127.0.0.1" ||
-             hostname.startsWith("192.168.") ||
-             hostname === "[::1]";
-
-      return isLocalServer ?
+      // `isLocalhost` actually also returns true for other local addresses, not just localhost
+      return await isLocalhost(url.hostname) ?
              (await ide.getWorkspaceDirs()).find(workspaceDirectory => {
                return currentFile?.path.startsWith(workspaceDirectory)
              }) : undefined
--- a/core/context/providers/HttpContextProvider.ts
+++ b/core/context/providers/HttpContextProvider.ts
@@ -2,6 +2,7 @@ import {
   ContextItem,
   ContextProviderDescription,
   ContextProviderExtras,
+  IDE,
 } from "../../index.js";
 import { BaseContextProvider } from "../index.js";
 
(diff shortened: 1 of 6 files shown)


uinstinct:
reference discussion: https://discord.com/channels/1108621136150929458/1359790258333421689/1359790258333421689

i can work on this

core/context/providers/HttpContextProvider.ts
```diff
@@ -14,6 +15,24 @@ class HttpContextProvider extends BaseContextProvider {
...
   };
 
+  private async getWorkspacePath(ide: IDE, url: URL) {
+    try {
+      const currentFile = await ide.getCurrentFile();
+      const hostname = url.hostname.toLowerCase();
+      const isLocalServer= hostname === "localhost" ||
+             hostname === "127.0.0.1" ||
+             hostname.startsWith("192.168.") ||
+             hostname === "[::1]";
```
uinstinct: does it work if devices are on same LAN?
i sometimes rename a device on my LAN like `http://secondmachine.local:11434`

s-h-a-d-o-w: That's a great point! It doesn't. But I already have an idea of how to make it work. Will probably push an update soon.

s-h-a-d-o-w: Done. I've confirmed that the library I switched to supports local names like you mentioned.

Patrick-Erichsen:
Neat idea, thanks for opening the PR!

I think it would likely make more sense to add this as a boolean option to the params for the HTTP provider: https://docs.continue.dev/customize/context-providers#http

That way it's more of an opt-in behavior.

sestinj (approved):

sestinj:
I think the way it's implemented now is fair, and if we find it often used, we could consider a config option. But erring on the side of avoiding new config is usually my preference.

One thing to note is that Model Context Protocol is solving this exact problem with [roots](https://modelcontextprotocol.io/docs/concepts/roots). As far as I can tell MCP can do everything that our HttpContextProvider can (with [resources](https://modelcontextprotocol.io/docs/concepts/resources)). I'm curious whether you found any limitations, or just had the HttpContextProvider set up before, or maybe hadn't considered MCP? @s-h-a-d-o-w

s-h-a-d-o-w:
> One thing to note is that Model Context Protocol is solving this exact problem with [roots](https://modelcontextprotocol.io/docs/concepts/roots). As far as I can tell MCP can do everything that our HttpContextProvider can (with [resources](https://modelcontextprotocol.io/docs/concepts/resources)). I'm curious whether you found any limitations, or just had the HttpContextProvider set up before, or maybe hadn't considered MCP? @s-h-a-d-o-w

I have implemented an MCP server as well. But HttpContextProvider can provide initial context in Chat. Which was my primary goal.

(I actually wasn't aware of `roots` though, that's great to know! That's gonna be helpful. Although—at least based on the docs, you're currently not providing any info there, right?)

sestinj:
Hi @s-h-a-d-o-w, yesterday we shared [some updates](https://discordapp.com/channels/1108621136150929458/1131313996750917835/1371996789967687783) with our contributors about how we're aiming to improve the contribution process. Part of this included the addition of a [Contributor License Agreement](https://github.com/continuedev/continue/blob/main/docs/docs/CLA.md) (CLA) to protect both contributors and the project. We're reaching out to ask that previous contributors sign it.

```
```

s-h-a-d-o-w:

PR #5154, opened 2025-04-14
Small fix in isItemTooBig

Jazzcort:
## Description

1. **Provide model name for tokens counting**:
Although Continue now uses the Llama tokenizer for most models,
it's still important to provide the model name to the
countTokens function. This ensures flexibility in case we
introduce more specialized tokenizers for different models in
the future.

2. **Fix typo in error dialog**: 
Fixed typo from "connot" to "cannot" 

3. **Delete redundant calling of loadConfig**:
Config has already been gathered at the top of this function,
so we can directly use this config instead of calling
loadConfig again.


RomneyDa (dismissed):
👍

RomneyDa:
@Jazzcort looks good, fixed merge conflict, will get this in today

Patrick-Erichsen (approved):

PR #5157, opened 2025-04-14
Nate/new-block-files

sestinj:


core/promptFiles/v2/createNewPromptFile.ts
```diff
@@ -74,3 +76,96 @@ export async function createNewPromptFileV2(
   await ide.writeFile(promptFileUri, PROMPT_FILE);
   await ide.openFile(promptFileUri);
 }
+
+function getContentsForNewBlock(blockType: BlockType): ConfigYaml {
```
Patrick-Erichsen: Two thoughts
1. This probably shouldn't live in `createNewPromptFile.ts`
2. I was thinking we should move our default YAMLs out of the GUI code on Hub and into `config-yaml` package. That was we can reuse them both here, and in the Hub.

gui/src/components/mainInput/Lump/sections/ExploreBlocksButton.tsx
```diff
@@ -30,16 +26,19 @@ export function ExploreBlocksButton(props: { blockType: string }) {
...
-          });
-      }
+      ideMessenger.request("config/addLocalWorkspaceBlock", {
+        blockType: props.blockType as BlockType,
+      });
+      // switch (props.blockType) {
+      //   case "docs":
+      //     dispatch(setShowDialog(true));
+      //     dispatch(setDialogMessage(<AddDocsDialog />));
+      //     break;
```
fbricon, line 36: @sestinj what's the reason to disable the AddDocsDialog component? It's still reachable if you try to add docs from the `@Docs` provider, so it's inconsistent.

PR #5158, opened 2025-04-14
fix package.json

sestinj:
Built on top of 5104


RomneyDa:
@sestinj could update description/title/etc, looks like this PR adds ability to turn rules on/off for chat/agent?
@tomasz-stefaniak might overlap with your current work
I'm wondering how this would work for Edit mode rules, seems like since GUI only it wouldn't, which could be confusing

RomneyDa:
@sestinj looks like non-flaky failing tests and merge conflicts

PR #5189, opened 2025-04-16
Correct the "extensionkind" manifest entry

eramitmittal:
For dev containers, continue vscode extension works best when loaded as "workspace" extension

## Description

For VS code dev containers, the continue dev extension is unable to index codebase when loaded as "ui" extension. Instead it should be loaded as "workspace" extension.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -8,7 +8,6 @@
     "url": "https://github.com/continuedev/continue"
   },
   "extensionKind": [
-    "ui",
     "workspace"
   ],
   "bugs": {


sestinj:
@eramitmittal can you explain more about what the problem is?

The reason we put UI first is that for folks using Remote SSH, if it were a workspace extension they would have no way to connect to a local Ollama for example. I'd need a stronger reason for changing it to the opposite

eramitmittal:
hi, this was false alarm. Please close this. In the version I was running there was some issue writing indexes from devcontainer. However, with latest pre-release I don't see this issue

PR #5212, opened 2025-04-17
Fix battery state not updating

ton-An:
## Description
The method responsible for updating the battery state only gets the actual battery info on it's first call. All subsequent calls of the method use that initial battery info. Resulting in the battery state not changing.

## Checklist

## Testing instructions (needs a laptop)
- Enable pauseTabAutocompleteOnBattery in VSCode settings
- Plug and unplug the laptop to observe the auto-completion pausing and resuming (this doesn’t happen in the production version)

### DIFF
--- a/extensions/vscode/src/util/battery.ts
+++ b/extensions/vscode/src/util/battery.ts
@@ -9,7 +9,6 @@ export class Battery implements Disposable {
   private readonly onChangeLevelEmitter = new EventEmitter<number>();
   private acConnected: boolean = true;
   private level = 100;
-  private readonly batteryStatsPromise = si.battery();
 
   constructor() {
     this.updateTimeout = setInterval(() => this.update(), UPDATE_INTERVAL_MS);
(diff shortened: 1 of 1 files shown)


sestinj:
@ton-An this was changed because it caused problems on certain operating systems. In fact, we are likely going to remove this feature in the near future. Is this something you have been relying on heavily, or were you mostly helping to clean up?

ton-An:
This feature is pretty important for my use case, as I often work on battery and frequently switch between battery and AC.

Have you tried using `si.powerShellStart()` and `si.powerShellRelease()` as mentioned in [systeminformation#616](https://github.com/sebhildebrandt/systeminformation/issues/616#issue-1051048083)? According to the maintainer, that reduced the cpu usage drastically. In addition, updating the state (lets say) only once per minute could help a lot too.

Would you be open to exploring this? I’m happy to make the changes, but I don’t have a Windows machine to test on

sestinj:
I would need to be very confident that this doesn't spawn and fail to stop background processes on Windows, as this has historically caused a lot of frustration for users. Unfortunately I think the risk here is too high if we are unable to validate this.

I'm also wondering whether having a quick keyboard shortcut to disable autocomplete would be at all helpful to you in place of this?

sestinj:
Hey @ton-An, I'm going to close this PR for now since I think it's probably hit a dead-end for the moment. I'm totally open to reconsidering this later and if you were just busy but want to get back to it, please re-open!

PR #5220, opened 2025-04-18
feat: `create_rule_block` tool

Patrick-Erichsen:
image

### DIFF
--- a/core/tools/implementations/createRuleBlock.ts
+++ b/core/tools/implementations/createRuleBlock.ts
@@ -39,24 +39,14 @@ export const createRuleBlockImpl: ToolImpl = async (
     `${safeRuleName}.yaml`,
   );
 
-  try {
-    await extras.ide.writeFile(rulesDirUri, ruleYaml);
-    await extras.ide.openFile(rulesDirUri);
+  await extras.ide.writeFile(rulesDirUri, ruleYaml);
+  await extras.ide.openFile(rulesDirUri);
 
-    return [
-      {
-        name: "Rule Block Created",
-        description: `Created ${args.rule_name} rule`,
-        content: ruleYaml,
-      },
-    ];
-  } catch (error) {
-    return [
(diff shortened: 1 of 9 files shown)


core/tools/implementations/createRuleBlock.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    return [
+      {
+        name: "Rule Block Created",
+        description: `Created ${args.rule_name} rule`,
+        content: ruleYaml,
+      },
+    ];
+  } catch (error) {
+    return [
```
RomneyDa: You can just throw the error it will turn into a context item downstream

gui/src/pages/gui/ToolCallDiv/ToolCallButtonsDiv.tsx
```diff
@@ -79,7 +79,7 @@ export function ToolCallButtons(props: ToolCallButtonsProps) {
               onClick={() => dispatch(callTool())}
               data-testid="accept-tool-call-button"
             >
-              Continue
+              Accept
```
RomneyDa: Remove that last remaining Continue button? this is quite a moment

RomneyDa (changes requested):
Add default permissions in redux (I think UI slice?)

Only other thought is that this tool has super long descriptions, is quite a bit more tokens than any other tool function I think (~500 according to https://platform.openai.com/tokenizer). Probably fine but would like to cut down to maybe 200

RomneyDa:
also could add section to built in tool docs

gui/src/pages/gui/ToolCallDiv/ToolCallButtonsDiv.tsx
Patrick-Erichsen: reverted

RomneyDa (approved):

PR #5224, opened 2025-04-18
do not allow to toggle config page from command palette

uinstinct:
## Description

One can toggle the config page from the command palette using the `continue: open settings` command. This creates some confusion when the config page is already open.

This PR fixes this by separating `continue.openConfigPage` with a new `continue.openConfigPageFromPalette` command.

## Checklist

## Screenshots

https://github.com/user-attachments/assets/afc94bdd-fb29-4728-b9a5-0b6106d2a4c7

https://github.com/user-attachments/assets/9f62f0d9-271f-45c8-a83d-dc703a676521

## Testing instructions


Patrick-Erichsen:
Hey @uinstinct , I've watched the video a few times but I'm still a bit confused. Why do we need both of the comamnds?

uinstinct:
> Hey @uinstinct , I've watched the video a few times but I'm still a bit confused. Why do we need both of the comamnds?

Sure, I will try to explain. Toggling using command palette with `continue: open settings` is somewhat unintuitive. It opens the chat area even if we used the command to _open the settings_ (i.e. `continue: open settings`).
Toggling with the settings button on the gui is intuitive because there is instant feedback between the chat and config page views. User understands we are switching between those 2 views.

RomneyDa:
@uinstinct nice find, I'd consider just removing the toggle functionality from the original command

Patrick-Erichsen:
Yeah +1 to Dallin, we have a back button on the settings page so I don't think the settings icon needs to be overly clever and have toggle functionality.

uinstinct:
We could also remove other commands present both with an icon and in the command palette (those including settings, view history, new session, open in new window).
Or should we just remove the settings?

Patrick-Erichsen:
I think some users find value in being able to quickly use the command palette to navigate instead of clicking around the GUI. I think Dallin's point (that I agree with) is just to keep the logic simple and not have it toggle.

uinstinct:
Here is the updated working:

https://github.com/user-attachments/assets/cee21495-4a1a-4eda-802d-75446c49d83e

sestinj (approved):

PR #5236, opened 2025-04-18
add standalone template parameter in yaml model config

mq200:
## Description

Add support for the `template` parameter in the YAML model configuration. 
It is supported in the JSON configuration.
Allows to bypass template auto-detection when needed.

## Checklist


sestinj (changes requested):
@mq200 this should go under the `promptTemplates` section as a property called `chatTemplate` in my opinion

sestinj:
ah I understand now sorry for the confusion. this looks good and I'll make a quick follow-up edit in another PR

PR #5283, opened 2025-04-22
feat: helper & unHelperful can redux persistence

smmzyl:
## ✨ feat(core): Add Redux persistence for helper/unHelperful states


sestinj:
Hey @smmzyl! I appreciate this fix. It's definitely the right behavior for the product, but this PR touches some important logic and we're not in a good position to test this well. We're working over the course of the next week to make it easier to test the GUI and this would allow us to revisit these changes, but right now the reward probably isn't enough for introducing the extra complexity

smmzyl:
Thank you @sestinj , and have a good life

PR #5290, opened 2025-04-22
add smooth scrolling to chat

uinstinct:
## Description

Recently I was used Gemini Code Assist and the chat smooth scrolling (and last prompt's sticky behaviour) on a new prompt was something I liked. I wanted to add this to Continue as well.

Whenever a new prompt is given, the chat scrolls to the bottom in smoothly.

(I wanted to also add the sticky behaviour of the prompt. It would require changes to how we render the [history](https://github.com/continuedev/continue/blob/b8c92f26cad325ff79eca6b5d01fd4ac47b4a88f/gui/src/pages/gui/Chat.tsx/#L315) items such that all other items remain constant during generation)

## Checklist

## Screenshots

https://github.com/user-attachments/assets/fcc98d54-b984-4a48-a9d5-074ca39f0800

https://github.com/user-attachments/assets/d332e400-bce5-4823-a56e-04ceb89bb2ba

## Testing instructions

1. open the chat messages area
2. add 2 or 3 prompts and long answers
3. see that when you prompt next, the last message is smoothly scrolled into view

### DIFF
--- a/extensions/vscode/e2e/tests/GUI.test.ts
+++ b/extensions/vscode/e2e/tests/GUI.test.ts
@@ -400,7 +400,7 @@ describe("GUI Test", () => {
         GUISelectors.getThreadMessageByText(view, llmResponse),
       );
 
-      await new Promise(resolve=>setTimeout(resolve, 200));
+      await new Promise(resolve=>setTimeout(resolve, 500));
 
       const viewportHeight = await driver.executeScript(
         "return window.innerHeight",
--- a/extensions/vscode/e2e/tests/GUI.test.ts
+++ b/extensions/vscode/e2e/tests/GUI.test.ts
@@ -400,6 +400,8 @@ describe("GUI Test", () => {
         GUISelectors.getThreadMessageByText(view, llmResponse),
       );
 
+      await new Promise(resolve=>setTimeout(resolve, 500));
+
       const viewportHeight = await driver.executeScript(
         "return window.innerHeight",
       );
--- a/extensions/vscode/e2e/tests/GUI.test.ts
+++ b/extensions/vscode/e2e/tests/GUI.test.ts
@@ -400,7 +400,7 @@ describe("GUI Test", () => {
         GUISelectors.getThreadMessageByText(view, llmResponse),
       );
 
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
The main difference here is that currently we just jump to the last prompt, but this instead smooth scrolls?

PlebeiusGaragicus:
instead - I like when the newly submitted prompt "pops" to the top of the window and as the agent reply streams it will again scroll once the window is filled.

uinstinct:
> The main difference here is that currently we just jump to the last prompt, but this instead smooth scrolls?

Yes, we jump to the last prompt using smooth scroll.

uinstinct:
> instead - I like when the newly submitted prompt "pops" to the top of the window and as the agent reply streams it will again scroll once the window is filled.

this will require some modifications to how we are rendering the history items as it was indicated here.

> (I wanted to also add the sticky behaviour of the prompt. It would require changes to how we render the [history](https://github.com/continuedev/continue/blob/b8c92f26cad325ff79eca6b5d01fd4ac47b4a88f/gui/src/pages/gui/Chat.tsx/#L315) items such that all other items remain constant during generation)

Patrick-Erichsen:
cc @RomneyDa here, I think you'll have an opinion.

Patrick-Erichsen (approved):

RomneyDa:
@uinstinct this can look nice especially if there's a long delay I think it ends up looking worse if you're on a tall screen and the model responds quickly, because it's still scrolling while the model starts responding. To make it scroll fast enough we'd have to have the scroll finish in a couple hundred milliseconds which at that point looks jenky.

Also, would need screenshots/more testing on truncating messages, one of the main reasons we abandoned autoscroll in the past was it often felt jenky when you go back and truncate the history at e.g. message 3/20

EDIT this was merged and then reverted https://github.com/continuedev/continue/pull/5408

uinstinct:
Fair points @RomneyDa 
I made a note of these these points and will try to add smooth scrolling once again. I think we will need to tweak how we are rendering the messages the above cases you mentioned and [this one](https://github.com/continuedev/continue/pull/5290#issuecomment-2822946729) as well.

PR #5344, opened 2025-04-24
fix(build): update esbuild to 0.25.3

chezsmithy:
## Description

I've noticed a build problem where the 2 year old version of esbuild is no longer pulling binaries which are required to build locally.

I'm noting a question that needs to be resolved: 

https://github.com/chezsmithy/continue/blob/e96639f17107def6050f265508949d6a071a0222/extensions/vscode/scripts/prepackage.js#L386

Notes that it pulls a pre-build version of esbuild. Would this also need to be updated to use the latest version as well? I'm assuming so?

## Checklist

## Screenshots

## Testing instructions


Patrick-Erichsen:
@chezsmithy did you recently start having issues building? I'm wondering if this PR is the root cause: https://github.com/continuedev/continue/pull/5271

If so, can you try downgrading to Node `20.11.0` and see if it resolves the issue for you?

Patrick-Erichsen:
> Would this also need to be updated to use the latest version as well? I'm assuming so?

This is correct.

@chezsmithy - is this still an issue for you? I'd prefer to avoid bumping esbuild since we're planning to deprecate eventually, and it's quite sensitive logic.

sestinj:
+1 to the comments made here by Patrick—if at all possible we want to avoid this bump (we've bumped in the past and caused other bugs)

chezsmithy:
@Patrick-Erichsen  This is fixed after doing some surgery on my node installation. I had some conflicts between node installed via homebrew and nvm. Removed all the homebrew versions and now things are fixed up. Made sure all the vscode tasks are running with the right node versions as well.

Issue #2061, opened 2024-08-21
IntelliJ: Unable to enable "Full line code completion" with disabled "Tab Autocomplete"


coinzz:
### Validations

- [ ] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Using: 
- IntelliJ: 2024.2.X
- Full Line Code Completion: bundled 242.20224.419
- Continue: 0.0.62

In this scenario I am unable to use the IntelliJ feature even with disabled "Tab Autocomplete":
![continue-autocomplete](https://github.com/user-attachments/assets/483fe9dc-8c49-4a4e-815b-b7a08d6fb63f)
![full-line-completion](https://github.com/user-attachments/assets/6498ef77-faa5-4f43-b36b-eb9912f7fd84)

See also [here](https://youtrack.jetbrains.com/issue/IJPL-81549), same problem is with Copilot.

### Solution

Use the API from IntelliJ as answered [here](https://intellij-support.jetbrains.com/hc/en-us/community/posts/17619324449042-Full-line-code-completion-is-disabled-with-Copilot-installed-but-completions-disabled) by Olga from JetBrains

sestinj:
Thanks for raising this issue, and for all the +1's. With the major caveat that I may simply be missing something (though I have looked quite hard), we're currently waiting on getting a clearer answer from JetBrains as to why our plugin is considered "conflicting". I fully agree that with Continue autocomplete disabled there is no reason for JetBrains AI not to work, and definitely want users to be able to use both. Unfortunately the API referenced in the link above is for classic "Intellisense-style" completions (as far as I have been able to tell—again, the caveat), which would not be a valid substitute for inline completions.

I hope to have more answers soon, and I've sent further follow-up to the JetBrains team. They've typically been very responsive, so I'm quite hopeful we'll get it worked out soon!

kkarnauk:
@sestinj hello! I am from JetBrains, and I would be happy to answer any question you have. We highly encourage plugins that offer inline completions to migrate to our built-in Inline Completion API, as it’s important for us to ensure consistency and integration. Apologies for not reaching out to you sooner.

[...]

PR #5360, opened 2025-04-25
initial commit for InlineCompletionProvider

sestinj:
WIP migrating to the native JetBrains InlineCompletionProvider API. This may not be picked up again for a week or two. If anyone comes across this and has interest in cleaning up more of the details to get it across the finish line earlier, please feel free to help out!


kkarnauk:
Hi! Sorry to bother you, I'm one of the authors of the Inline Completion API in IntelliJ. I noticed that you're using an older version of the API, which was part of a highly experimental stage and is no longer available in the platform. I recommend switching to the version introduced in IntelliJ 2024.1. It's much more stable and production-ready now. Please let me know if you have any questions.

sestinj:
Thanks for reaching out @kkarnauk! The purpose of this PR is to migrate to the new API, which we are very excited to do. We need a bit more time to be able to prioritize finishing this, but plan to get to it very soon

coinzz:
@kkarnauk is the newest API version enabled by using platform and platformVersion set to 2023.3 and documented [here](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html#attaching-sources-in-the-ide)?

kkarnauk:
@coinzz I'd suggest to use the API starting from 2024.1. The one in 2023.3 has lots of deprecated methods/classes. They may be removed at some point of time (since almost 2 years already passed).

Unfortunately, there is no documentation on the IJ page. There are lots of KDocs in the code, though.

coinzz:
> Unfortunately, there is no documentation on the IJ page. There are lots of KDocs in the code, though.

@kkarnauk ah no worries, do you have a link to the current implementation with the KDocs for me then? I have scheduled a meeting with a colleague to take a look on that next friday 🙂

kkarnauk:
@coinzz great to hear, thanks a lot for that!

Basically, you need to implement this [interface](https://github.com/JetBrains/intellij-community/blob/f50abc6062773c82daa76b757ad1bb199ce05b1e/platform/platform-impl/codeinsight-inline/src/com/intellij/codeInsight/inline/completion/InlineCompletionProvider.kt).

A brief overview:
* Implement `isEnabled(event)` to let the platform know whether your provider is applicable in the current context. 
  * The implementation should be fast as it's called on EDT. 
  * Usually, a provider is enabled on `InlineCompletionEvent.DocumentChange` (== a typing in an editor) and `InlineCompletionEvent.DirectCall` (== a user invokes the inline completion by shortcut). 
  * There are also other built-in events. If you lack of the default events, please use `InlineCompletionEvent.ManualCall`. 
  * If the completion of your plugin is disabled by some setting, please return `false` in this case. It's important to let other providers work in case your provider is unavailable.
* Implement `getSuggestion(request)`. 
  * It is called on a background thread, so it may do some heavy stuff/internet access.
  * The result is basically a flow of some elements. If you need to render gray text, please use `InlineCompletionGrayTextElement`
  * There is `InlineCompletionTextElement` to provide colorful suggestions.
  * If you need to jump over a few symbols in the editor (e.g., a suggestion inserts text at several close offsets in the editor), please use `InlineCompletionSkipTextElement`.
  * Please do not create your own implementations of `InlineCompletionElement`. Otherwise, it would not work in the Remote Development setup.
* If you need a custom handling when your suggestion is inserted into the editor, please override `insertHandler`.
* A bonus. There is a native support for the Remote Development setup.

coinzz:
I had some build problems which I figured out now. But @kkarnauk I still was not able to take a deeper look at your brief overview, because I still try to create a testable environment. So far I figured out:

- I guess I need to use:
  -  platformVersion=2024.3
  - key="ORG_GRADLE_PROJECT_platformVersion" value="2024.3.6"
  - I need to install the Jetbrains AI Plugin to test, since the Full Line Code Completion Plugin is an Ultimate Plugin (I did this manually in the sandbox env, because I was not able to add this plugin via gradle yet). Without any Inline Completion Plugin the settings look like this:
2024 3 6_no-plugin

### When I start the intellij sandbox env with version 2024.2.6 and the AI Assistant Plugin I get this message:


### When I start the intellij sandbox env with version 2024.3.6 and the AI Assistant Plugin I get this message:

Only this one is talking about "Built-in local inline completion", therfore I guessed that I need to update this far. But it could be only a textual issue as well? I guess the goal is as described in the ticket (#2061) to disable the "Enable Tab autocompletion" and then being able to enable "Enable cloud completion suggestions" on one of the screenshots. So which version do you think @kkarnauk ?

And @sestinj should I fork this branch and do a draft PR into this branch, because I guess I am not allowed to push directly here and this is your branch?

RomneyDa:
Closing in favor of https://github.com/continuedev/continue/pull/6947

Issue #5340, opened 2025-04-24
Implement Tools Config Caching for Anthropic Claude


chezsmithy:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

I would like to add prompt caching for the tools configuration for Anthropic, AWS Bedrock and Vertex hosted Claude models. Tools config is highly repeatable and caching this configuration limits the number of tokens used. I've validated this with my local configuration using AWS Bedrock with Claude 3.7.

For AWS Bedrock and Claude 3.7 see this page for more information: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html

### Solution

I've identified through local testing that it's possible to add caching support for tools config. This allows the config to enter the prompt cache and then significantly lowers the number of tokens consumed on each subsequent call to the LLM.

I've identified that AWS Bedrock, Anthropic and Vertex seem to all support this feature tho in slightly different ways.

Here is what I would propose as PRs.

1. Create a new configuration item to enable toolsConfigCache under the cacheBehavior key.
2. Using that key have contributors with access to each back end independently add the caching support as a separate PR with updated tests.

### Implementation Questions

1. Should I implement the configuration changes in JSON and YAML?
2. Should the configuration only be enabled for certain models and providers in a specific way?

chezsmithy:
Noting I'm willing to contribute a PR to first implement the configuration changes.

Then once that is approved, I will submit a PR for AWS Bedrock Claude 3.7.

chezsmithy:
Once the Cache Behavior merges then I'll open a PR for the Bedrock provider and Claude model.

chezsmithy:
@Patrick-Erichsen we likely want to do a scan of the other providers implementating caching and make sure they are using the new combined ```promptCaching``` settting and start to deprecate the individual settings.

PR #5371, opened 2025-04-25
feat(config-yaml): Add promptCaching to Default Completion Options and enable Bedrock Tools Caching

chezsmithy:
## Description

Provides inital configuration support for this new feature: https://github.com/continuedev/continue/issues/5340

Adding the ability to specify if the model / provider should enable caching for the tools Configuration. The changes to enable the use of this configuration in providers will be done via separate PRs. The documenation to use this feature will be updated on a per model / provider basis.

Enabling tools caching for AWS Bedrock Claude Models.

Note, I'm not planning to make significant updates to the JSON schema to support this feature.

## Checklist

## Screenshots

## Testing instructions

defaultCompletionOptions:
  promptCaching: true

Add this configuration to a model and validate through debugging that it's loaded and available to the model / provider in question downstream.

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -852,7 +852,6 @@ declare global {
   export interface CacheBehavior {
     cacheSystemMessage?: boolean;
     cacheConversation?: boolean;
-    cacheToolsConfig?: boolean;
   }
   
   export interface ClientCertificateOptions {
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -852,6 +852,7 @@ declare global {
   export interface CacheBehavior {
     cacheSystemMessage?: boolean;
     cacheConversation?: boolean;
+    cacheToolsConfig?: boolean;
   }
   
   export interface ClientCertificateOptions {
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -927,7 +927,6 @@ export interface RequestOptions {
 export interface CacheBehavior {
   cacheSystemMessage?: boolean;
   cacheConversation?: boolean;
-  cacheToolsConfig?: boolean;
 }
 
 export interface ClientCertificateOptions {
@@ -1024,6 +1023,7 @@ export interface BaseCompletionOptions {
   toolChoice?: ToolChoice;
   reasoning?: boolean;
   reasoningBudgetTokens?: number;
+  promptCaching?: boolean;
 }
 
 export interface ModelCapability {
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -927,6 +927,7 @@ export interface RequestOptions {
 export interface CacheBehavior {
   cacheSystemMessage?: boolean;
   cacheConversation?: boolean;
+  cacheToolsConfig?: boolean;
 }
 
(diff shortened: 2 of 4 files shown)


Patrick-Erichsen:
@chezsmithy thanks for splitting up the schema/implementation work here. 

I worry that the `CacheBehavior` is getting a bit too granular. I think the majority of users probably just want everything or nothing to be cached.

What's your take? Do you find value in having it this granular? Is there any scenario where you wouldn't just want the system message, chat, and tool configs to all be cached together?

chezsmithy:
@Patrick-Erichsen agree. If you are going to cache you might as well enable all of them. One pain point to adding a global cache setting is testing for the various providers. I don't have access to them all so this kind of change would require a bit of a breaking change. Maybe we add the individual settings and slowly deprecate them in favor of a single setting that replaces them all?

Where would you think that should live? Under env?

Patrick-Erichsen (approved):

Patrick-Erichsen:
I would agree, better to get this shipped and see how it's used vs trying to figure out all the edge cases making it generalizable to all providers. 

I approved but just realized there's a small merge conflict from your other caching PR, once that's resolved let's merge 👍

chezsmithy:
@Patrick-Erichsen i have the changes ready and tested for the bedrock provider. I can add those to the PR. I don't have access to Vertex or Claude API direct so I'm less inclined to add those as I can't test it.

sestinj:
I actually kind of worry about the same thing as Patrick here. If we add all of these granular permissions, we're going to be stuck supporting them across a very large number of providers. Has anyone encountered cases where multiple different caching strategies are desired for different cases but with the same provider? If not it does seem like the wise thing for now would be to just have a setting to enable/disable caching all at once

chezsmithy:
@Patrick-Erichsen I've pushed the changes for review and have added the Bedrock implmentation as an example.
@sestinj I'll queue up a new change to implement a single place to enable caching. Could you add some ideas to https://github.com/continuedev/continue/issues/5340 related to the best place to add that setting and the name for it? Once we have that ironed out I'll submit a new PR to add in the global setting and make Bedrock work with both the gobal setting and individual settings for backwards compatiblity. Then we can move to deprecate the individual settings over time.

chezsmithy:
@sestinj quick ping on how you would like to proceed?

sestinj:
@chezsmithy Sorry if this didn't come across in the last message—I don't think we should merge a specific setting for caching tools at all if we are just going to add a general cache setting right after and then deprecate it.

I think we should skip straight to the creation of the general caching setting. Probably lives in `defaultCompletionOptions.promptCaching`

tomasz-stefaniak:
Should this PR stay open?

chezsmithy:
@tomasz-stefaniak stand by. I'm going to push an update shortly based on feedback.

chezsmithy:
@tomasz-stefaniak it's ready for review. The e2e tests seem to be failing for all recently pushed PRs for some reason. I noted it on discord.

Patrick-Erichsen (approved):

PR #5375, opened 2025-04-26
feat: add `globs` to create rule tool

Patrick-Erichsen:


docs/docs/customize/deep-dives/rules.md
```diff
@@ -22,7 +22,7 @@ Rules can be added to an Assistant on the Continue Hub. Explore available rules
 :::info Automatically create local rule blocks
 When in Agent mode, you can simply prompt the agent to create a rule for you.
 
-For example, you can say "Create a rule for this", and a rule will be created for you in `~/.continue/rules` based on your conversation.
+For example, you can say "Create a rule for this", and a rule will be created for you in `.continue/rules` based on your conversation.
```
RomneyDa, line 25: Clarify this only works if this tool is not excluded

core/tools/implementations/createRuleBlock.ts
```diff
@@ -45,7 +48,7 @@ export const createRuleBlockImpl: ToolImpl = async (
   return [
     {
       name: "Rule Block Created",
-      description: `Created ${args.rule_name} rule`,
+      description: `Created ${args.name} rule`,
       content: ruleYaml,
```
RomneyDa: Thinking the output of this tool should just be a success message, not the full rule YAML
Instead could add a file uri to the context item to make it click-throughable, which might be better for both user and model

core/tools/definitions/createRuleBlock.ts
```diff
@@ -18,18 +18,23 @@ export const createRuleBlock: Tool = {
       "Not for one-time instructions. Rules are stored as YAML in `.continue/rules`.",
     parameters: {
```
RomneyDa, line 17: tool adds ~250 tokens/request, could we get it down to < 100?

RomneyDa (approved):

PR #5396, opened 2025-04-28
feat: introduce separate agent system message

uinstinct:
## Description

This PR introduces a new `chatOptions.baseAgentSystemMessage`. It provides the flexibility to advanced users to use different system prompts for agent and chat modes.

references: https://discord.com/channels/1108621136150929458/1108621136830398496/1366280234760146954

## Checklist

## Screenshots

https://github.com/user-attachments/assets/be5137c6-72a3-44f1-9c5e-94310a2b8b50

## Testing instructions

1. Inside the config.yaml file, scroll to the model with the chat role
2. Specify the `chatOptions` key and inside it specify a `baseAgentSystemMessage` prompt (for example, it could be `baseAgentSystemMessage: "add comments for each line"`)
3. Switch to agent mode
4. Chat with the mode and see that the given prompt applies


core/config/load.ts
```diff
@@ -609,6 +609,7 @@ function llmToSerializedModelDescription(llm: ILLM): ModelDescription {
     contextLength: llm.contextLength,
     template: llm.template,
     completionOptions: llm.completionOptions,
+    baseAgentSystemMessage: llm.baseAgentSystemMessage,
```
RomneyDa, line 630: would break `systemMessage` in JSON applying to agent mode, could just say `baseAgentSystemMessage: llm.baseChatSystemMessage` with a comment explaining that JSON doesn't have a field for agent system message

RomneyDa (changes requested):
@uinstinct this is a good idea, will put on hold for a bit to make sure I get feedback from folks since it effects the YAML schema.

One effect of this is that users would **have** to specify the agent system message separately from the custom chat system message, since a `baseChatSystemMessage` doesn't apply in agent mode. This could be seen as a feature but I think generally there's a lot of overlap. If agent were to fall back to `baseChatSystemMessage` that would probably also be a bit confusing.

uinstinct:
> One effect of this is that users would have to specify the agent system message separately from the custom chat system message, since a baseChatSystemMessage doesn't apply in agent mode. This could be seen as a feature but I think generally there's a lot of overlap. If agent were to fall back to baseChatSystemMessage that would probably also be a bit confusing.

True. What do you think about falling back to `baseChatSystemMessage` since the same model is being used for both chat and agent? (One point is that we [have mentioned](https://docs.continue.dev/agent/how-to-use-it) agent and chat use the same interface)

RomneyDa:
@uinstinct sorry I was pointing to the wrong part of the code there. For the serialized model description you would want to keep `baseAgentSystemMessage: llm.baseAgentSystemMessage` since it's for both JSON and YAML. My bad

Where I meant to point was in `core/llm/llms/index.ts` llmFromDescription

```
  let baseChatSystemMessage: string | undefined = undefined;
  if (desc.systemMessage !== undefined) {
    baseChatSystemMessage = DEFAULT_CHAT_SYSTEM_MESSAGE;
    baseChatSystemMessage += "\n\n";
    baseChatSystemMessage += await renderTemplatedString(
      desc.systemMessage,
      readFile,
      {},
    );
  }
  
    let options: LLMOptions = {
    ...desc,
    completionOptions: {
      ...finalCompletionOptions,
      model: (desc.model || cls.defaultOptions?.model) ?? "codellama-7b",
      maxTokens:
        finalCompletionOptions.maxTokens ??
        cls.defaultOptions?.completionOptions?.maxTokens,
    },
    baseChatSystemMessage,
    logger: llmLogger,
    uniqueId,
  };
  ```
  
Can add `baseAgentSystemMessage: baseChatSystemMessage` to the options for JSON there

**Comment by RomneyDa** [comment]
Falling back baseAgentSystemMessage -> baseChatSystemMessage -> DEFAULT_CHAT_SYSTEM_ MESSAGE is interesting. I think that might be a bit confusing for users.

**Comment by RomneyDa** [comment]
@uinstinct got some feedback, lets implement this but keep it as a hidden setting with no documentation, only to be used for now in order to experiment and help find a better agent system message

Can merge with the following changes
- JSON correction above
- Fall back `baseAgentSystemMessage` -> `baseChatSystemMessage` -> `DEFAULT_CHAT_SYSTEM_ MESSAGE` in agent mode so that hidden setting doesn't effect `baseChatSystemMessage`
- Comment out the documentation changes with a note that this setting is experimental and should not be documented

**Comment by uinstinct** [comment]
> @uinstinct got some feedback, lets implement this but keep it as a hidden setting with no documentation, only to be used for now in order to experiment and help find a better agent system message
> 
> Can merge with the following changes
> 
>     * JSON correction above
> 
>     * Fall back `baseAgentSystemMessage` -> `baseChatSystemMessage` -> `DEFAULT_CHAT_SYSTEM_ MESSAGE` in agent mode so that hidden setting doesn't effect `baseChatSystemMessage`
> 
>     * Comment out the documentation changes with a note that this setting is experimental and should not be documented

Implemented these changes!

File: gui/src/redux/thunks/streamResponseAfterToolCall.ts [hunk 2]
```diff
@@ -64,10 +64,20 @@ export const streamResponseAfterToolCall = createAsyncThunk<
 
         dispatch(setActive());
 
+        
         const updatedHistory = getState().session.history;
+        const messageMode = getState().session.mode
+
+        let baseChatOrAgentSystemMessage: string|undefined
```
RomneyDa: one more nitpick, just so someone doesn't mess up this fallback logic, could you create a `getBaseSystemMessage(config, mode) => string | undefined` or similar to prevent the logic from being in 3 places?

RomneyDa (changes requested):

gui/src/redux/thunks/streamResponseAfterToolCall.ts
uinstinct: implemented!

RomneyDa (approved):

PR #5420, opened 2025-04-29
Add support for reranking models in Llama.cpp with configuration examples

fradav:
## Description

Implement a custom interface for the llama.cpp reranking response and modify the rerank method to handle the response format correctly. This ensures compatibility with the expected OpenAI format.

## Checklist

## Testing instructions

see the updated .mdx doc

### DIFF
--- a/core/llm/llms/LlamaCpp.ts
+++ b/core/llm/llms/LlamaCpp.ts
@@ -1,6 +1,21 @@
 import { CompletionOptions, LLMOptions } from "../../index.js";
 import { BaseLLM } from "../index.js";
+import { LlmApiRequestType } from "../openaiTypeConverters.js";
 import { streamSse } from "../stream.js";
+
+// Define a custom interface for the llama.cpp reranking response
+interface LlamaCppRerankResponse {
+  model: string;
+  object: string;
+  usage: {
+    prompt_tokens: number;
+    total_tokens: number;
+  };
+  results: Array<{
+    index: number;
+    relevance_score: number;
+  }>;
+}
 
 class LlamaCpp extends BaseLLM {
   static providerName = "llama.cpp";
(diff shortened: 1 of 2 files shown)


core/llm/llms/LlamaCpp.ts
```diff
@@ -52,6 +67,60 @@ class LlamaCpp extends BaseLLM {
...
+  ];
+
+  protected createOpenAiAdapter() {
+    const adapter = super.createOpenAiAdapter();
+    
+    if (adapter) {
+      // Create a wrapper around the original rerank method to handle llama.cpp response format
+      const originalRerank = adapter.rerank.bind(adapter);
+      
+      adapter.rerank = async (params) => {
```
sestinj, line 82: @fradav this should be implemented in the `packages/openai-adapters` package rather than here. If you can make this change the rest of this PR looks excellent

sestinj (changes requested):

core/llm/llms/LlamaCpp.ts
fradav: Thank you for your suggestion to move the llama.cpp reranking implementation to the packages/openai-adapters package to better align with the project's architecture.

I initially tried to fully migrate the implementation, but encountered integration issues with the core module that resulted in runtime errors ("Failed to rerank retrieval results TypeError: Cannot read properties of undefined"). The reranking functionality has complex integration with the BaseLLM class and retrieval pipeline that made a complete migration challenging without broader refactoring.

To address this while respecting your architectural guidance, I've implemented a hybrid approach:

1. I've maintained the working implementation in the core module (core/llm/llms/LlamaCpp.ts) to preserve functionality.

2. I've added a well-documented LlamaCppApi class in the packages/openai-adapters package that properly extends OpenAIApi, following the established pattern of other adapters.

3. I've updated the constructLlmApi function to use this new LlamaCppApi class for the llama.cpp provider.

In the future, a more comprehensive refactoring could fully migrate the implementation, but that would require changing how the core module interacts with adapters - which seems beyond the scope of the current PR.

Would this approach be acceptable as an intermediate step toward the architectural goal you suggested?

sestinj: @fradav I really appreciate the thoughtfulness here. I think it's closer to working than it might appear—probably if you were to set up a debugger it would become obvious where the "Cannot read properties of undefined..." is coming from. Would you be willing to investigate this? As far as I know from our other openai-adapters integrations, this should be able to work without much further adjustment

We won't be willing to merge this PR until the code is in the correct spot in openai-adapters, otherwise it is going to make this untestable and harder to maintain down the road

docs/docs/customize/model-roles/reranking.mdx
```diff
@@ -154,3 +154,47 @@ The `"modelTitle"` field must match one of the models in your "models" array in
     ```
     </TabItem>
 </Tabs>
+
+### Llama.cpp
```
sestinj, line 158: One last piece of feedback: since llama.cpp is more difficult to set up than other providers, I don't think we want to include it on the main page for `reranking.mdx`. Instead, could we move this new section to `docs/docs/customize/model-providers/more/llamacpp.mdx`?

sestinj (changes requested):

sestinj:
Hey @fradav, just wanted to check in here. Do you think you'll have the time to get back to this, or is there any way I could help?

sestinj:
@fradav I'm going to close this for now since I haven't heard back. I'd still love to get this merged, but don't want to leave it stale. Feel free to re-open or let me know if I can help in any way!

PR #5425, opened 2025-04-29
Fix [Object object] chat title generation

RomneyDa:
Wasn't rendering chat message content properly

Co-authored-by: @sh4shii


RomneyDa:
@sestinj fixed - tests were broken because they were mocked with no `role` in the message. Removed mocks
Also removed a different test I felt was extraneous

sestinj (approved):

PR #5461, opened 2025-05-01
fetch secrets from <workspace>/.continue/.env folder before <workspace>/.env

uinstinct:
## Description

Fetch secrets from <workspace>/.continue/.env file before fetching from workspace .env

So now the order of fetching the first env file is `~/.continue/.env, <workspace>/.continue/.env, <workspace>/.env`

https://discord.com/channels/1108621136150929458/1355378099004379248/1367363239792476211

## Checklist

## Screenshots

## Testing instructions

1. Create a .env file inside the .continue directory
2. Add key value pairs
3. Use the secrets in config yaml
4. See them reflected via config


tomasz-stefaniak:
@uinstinct could you add tests for this as documentation and to prevent future regressions?

core/config/yaml/LocalPlatformClient.test.ts
```diff
@@ -0,0 +1,96 @@
+import { FQSN, SecretResult, SecretType } from "@continuedev/config-yaml";
+import { testControlPlaneClient, testIde } from "../../test/fixtures";
+import { LocalPlatformClient } from "./LocalPlatformClient";
+
+jest.mock("../../util/paths.ts", () => ({
+  ...jest.requireActual("../../util/paths"),
+  getContinueDotEnv: jest.fn(),
+}));
```
uinstinct: hey @tomasz-stefaniak

there is some problem when am trying to mock a local file as jest is unable to resolve it. it seems like a problem with how jest is setup. I [tried various ways](https://github.com/continuedev/continue/pull/5461/files#diff-5166dd2f791d972b6c797187e7f9ebefe27745b7d43e3467daf2bab31f82edabR66-R82) to get it working. Currently none of these work.

Any idea what should I do to a mock?

(I need to mock `utils/paths` for imported functions like [`getContinueDotEnv()`](https://github.com/continuedev/continue/blob/fe6d55dd1b4f7f4c5af7dbf71eef590d6b38f907/core/config/yaml/LocalPlatformClient.ts/#L47))

tomasz-stefaniak: @uinstinct have you checked some existing instances of `jest.mock` we have in the codebase?
image

I understand most of them are for node packages but maybe they can point you in the right direction. If you are still stuck, let me know and I can try to pull this into the sprint next week.

uinstinct: Thanks for replying Tomasz! 
I had tried these actually but for some reason jest is unable to resolve it.

For alternative options, we can:
1. use the filesystem env (problem: the test might affect the local users .env)
2. use a conditional (`process.env.TEST` flag) inside `getContinueDotEnv` to produce a key value pair for testing

uinstinct: Spent some more time debugging this. Looks like the module resolution was starting from `core/test` folder since jest was declared as global in `core/test/jest.setup-after-env.js`

So, `../util/paths` would work but not `../../util/paths`.

uinstinct: Okay. Tried both `unstable_mockModule` and `jest.mock('../util/paths')` both which resolve correctly but do not mock implementations for some reason.

tomasz-stefaniak: @uinstinct I'm sorry for the late reply, I haven't had the time to look into this yet but I opened an internal ticket for the team.

uinstinct: Thanks Tomasz!
Happy to assist if you need any help from my side.

tomasz-stefaniak: Update: this is a huge pain, I'm exploring migrating to Vitest 😆

uinstinct: 🥲 Agree!!!
BTW, can i help in some way? I have some experience with migrating jest to vitest.

tomasz-stefaniak: That would be great, actually! Do you think you'd be able to create a basic vitest setup? It could be on this branch, and it would be used to run [LocalPlatformClient.test.ts](https://github.com/continuedev/continue/pull/5461/files#diff-5166dd2f791d972b6c797187e7f9ebefe27745b7d43e3467daf2bab31f82edab). Once this proof of concept works we can migrate other tests to vitest.

This would also require adding an extra step here: https://github.com/continuedev/continue/blob/main/.github/workflows/pr_checks.yaml

Let me know what you think!

uinstinct: implemented!

- vitest works nicely (and with mocks) 🥳
- added vitest running for LocalPlatformClient only in pr_check.yaml


Edit: Should I draft another PR to migrate core tests to vitest?

tomasz-stefaniak (approved):
Thanks! Great job and thanks for introducing vitest. 

> Edit: Should I draft another PR to migrate core tests to vitest?

I'd love that, it's very high on our priority list.

core/config/yaml/LocalPlatformClient.test.ts
tomasz-stefaniak: > Edit: Should I draft another PR to migrate core tests to vitest?

Yeah, that would be great. It's one of our more important tasks at the moment.

PR #5463, opened 2025-05-01
watsonx - add new line to prompt for Codestral

mq200:
## Description

Add `\n` to end of prompt when using Codestral. Without the new line, Codestral sometimes ignores the `[/INST]` token and continues or repeats the user input.

## Checklist

### DIFF
--- a/core/llm/llms/WatsonX.ts
+++ b/core/llm/llms/WatsonX.ts
@@ -224,10 +224,11 @@ class WatsonX extends BaseLLM {
     }
 
     let input = messages[messages.length - 1].content
-    // Prevent Codestral from continuing user input, regardless of chat template
+    // Add \n to prevent Codestral from continuing user input when using chat template
     if (
-        options.model?.toLowerCase()?.includes("codestral") ||
-        this.deploymentId?.toLowerCase()?.includes("codestral")
+        input.toString().endsWith("[/INST]") &&
+        (options.model?.toLowerCase()?.includes("codestral") ||
+        this.deploymentId?.toLowerCase()?.includes("codestral"))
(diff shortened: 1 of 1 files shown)


RomneyDa:
@sestinj not familiar with this, will pass to you for now

sestinj (commented):
@mq200 I would rather not maintain this kind of code in general (i.e. logic that is hyper-specific to a single model for a single provider at a single point in time), though if it is the only solution I will of course merge this. Is there any way that we can bake this into a template in `core/llm/templates/chat.ts`, or solve the problem on the WatsonX side?

Also, I notice that we are only sending the last message to the model in `_streamChat`—do you know whether this is something we will be able to update in the near future, or is this a limitation of WatsonX?

mq200:
@sestinj Thank you for the feedback, agreed that it's best to avoid such hyper-specific logic (I didn't create such a chat template as I was trying to to keep things consistent with codestral being assigned the 'none' template in `core/llm/autodetect.ts`–as it's not needed when using the Mistral messages API, but now realize that was a better approach anyway.)
I just opened a [PR](https://github.com/continuedev/continue/pull/5523) containing a 'codestral' template in `core/llm/templates/chat.ts`

> Also, I notice that we are only sending the last message to the model in _streamChat—do you know whether this is something we will be able to update in the near future, or is this a limitation of WatsonX?

The entire conversation is actually getting passed in this last 'message' (but the logic is hidden in the `_complete` function due to the original implementation). I just opened a [PR](https://github.com/continuedev/continue/pull/5522) with the integration of the watsonx Chat API, in which this part of the code is refactored.

PR #5465, opened 2025-05-01
include rules in edit requests

RomneyDa:
Includes rules in edit streaming requests

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -63,32 +63,54 @@ import { llmStreamChat } from "./llm/streamChat";
 import type { FromCoreProtocol, ToCoreProtocol } from "./protocol";
 import type { IMessenger, Message } from "./protocol/messenger";
 
+// This function is used for jetbrains inline edit and apply
 async function* streamDiffLinesGenerator(
   configHandler: ConfigHandler,
   abortedMessageIds: Set<string>,
   msg: Message<ToCoreProtocol["streamDiffLines"][0]>,
 ): AsyncGenerator<DiffLine> {
-  const data = msg.data;
+  const {
+    highlighted,
+    prefix,
+    suffix,
+    input,
+    language,
+    modelTitle,
+    includeRulesInSystemMessage,
+  } = msg.data;
 
   const { config } = await configHandler.loadConfig();
   if (!config) {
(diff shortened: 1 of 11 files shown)


core/core.ts
```diff
@@ -63,32 +63,54 @@ import { llmStreamChat } from "./llm/streamChat";
...
 
+  // fallback to selected chat model
   if (!llm) {
-    throw new Error("No chat model selected");
+    llm = config.selectedModelByRole.chat;
   }
 
+  if (!llm) {
+    throw new Error("No model selected");
+  }
```
tomasz-stefaniak: Could you make a `getLLM` helper? This would make things more single-responsibility and would prevent future bugs with `let`.

core/edit/streamDiffLines.ts
```diff
@@ -99,38 +99,41 @@ export async function* streamDiffLines({
...
-          prompt,
-          (msg) => msg.role === "user" || msg.role === "tool",
-        ) as UserChatMessage | ToolResultChatMessage | undefined);
-
-  const systemMessage = getSystemMessageWithRules({
-    rules,
-    userMessage: lastUserMessage,
-    baseSystemMessage: undefined,
-  });
+  let systemMessage: string | undefined = undefined;
```
tomasz-stefaniak: Ditto with `let`. How about `const systemMessage = getSystemMesage(...)`? (Or something like that, as long as it doesn't use `let` or if `let` is used in a small function. In my experience `complex function + let = bugs`)

core/edit/streamDiffLines.ts
```diff
@@ -76,7 +76,7 @@ export async function* streamDiffLines({
   language: string | undefined;
   onlyOneInsertion: boolean;
   overridePrompt: ChatMessage[] | undefined;
-  rules: RuleWithSource[];
+  includeRulesAsSystemMessage: RuleWithSource[] | undefined;
```
tomasz-stefaniak: This is named like a boolean but the payload is an array 🤔

core/core.ts
RomneyDa: Nice catch can change to const in this case
This way of getting LLM is unique in app

PR #5472, opened 2025-05-01
Ghost text for removed lines in apply

RomneyDa:
Updates apply behavior to fix linting errors caused by removed diff lines being "physically" present in the file
- Uses "ghost" text to show removed lines, overlayed using `after` styling. The underlying line is kept empty
- Because each red diff line has a unique decorator associated with it, the behavior is different enough that I created a second decoration manager with similar methods.

BEFORE
image

AFTER
image

See comments below

### DIFF
--- a/extensions/vscode/src/diff/vertical/decorations.ts
+++ b/extensions/vscode/src/diff/vertical/decorations.ts
@@ -1,14 +1,19 @@
 import * as vscode from "vscode";
 
-export const redDecorationType = vscode.window.createTextEditorDecorationType({
-  isWholeLine: true,
-  backgroundColor: { id: "diffEditor.removedLineBackground" },
-  color: "#808080",
-  outlineWidth: "1px",
-  outlineStyle: "solid",
-  outlineColor: { id: "diffEditor.removedTextBorder" },
-  rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
-});
+export const redDecorationType = (line: string) =>
+  vscode.window.createTextEditorDecorationType({
+    isWholeLine: true,
+    backgroundColor: { id: "diffEditor.removedLineBackground" },
+    // color: "#808080",
+    outlineWidth: "1px",
+    outlineStyle: "solid",
+    outlineColor: { id: "diffEditor.removedTextBorder" },
+    rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
+    after: {
+      contentText: line,
+      color: "#808080",
+    },
+  });
 
 export const greenDecorationType = vscode.window.createTextEditorDecorationType(
   {
@@ -34,6 +39,13 @@ export const belowIndexDecorationType =
     backgroundColor: "rgba(255, 255, 255, 0.1)",
     rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
   });
+
+function translateRange(range: vscode.Range, lineOffset: number): vscode.Range {
+  return new vscode.Range(
+    range.start.translate(lineOffset),
+    range.end.translate(lineOffset),
+  );
+}
 
 export class DecorationTypeRangeManager {
   constructor(
@@ -82,20 +94,10 @@ export class DecorationTypeRangeManager {
     return this.ranges;
   }
 
-  private translateRange(
-    range: vscode.Range,
-    lineOffset: number,
-  ): vscode.Range {
-    return new vscode.Range(
-      range.start.translate(lineOffset),
-      range.end.translate(lineOffset),
-    );
-  }
-
   shiftDownAfterLine(afterLine: number, offset: number) {
     for (let i = 0; i < this.ranges.length; i++) {
       if (this.ranges[i].start.line >= afterLine) {
-        this.ranges[i] = this.translateRange(this.ranges[i], offset);
+        this.ranges[i] = translateRange(this.ranges[i], offset);
       }
     }
     this.editor.setDecorations(this.decorationType, this.ranges);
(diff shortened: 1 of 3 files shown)


extensions/vscode/src/diff/vertical/handler.ts
```diff
@@ -499,24 +504,27 @@ export class VerticalDiffHandler implements vscode.Disposable {
...
+        .map((line) => line.line)
+        .join("\n") + "\n";
 
-    const diffs = myersDiff(oldFileContent, newFileContent);
+    const myersDiffs = myersDiff(oldFileContent, newFileContent);
 
-    const myersDiffLines = diffs.map((diff) => diff.line).join("\n");
+    const replaceContent = myersDiffs
+      .map((diff) => (diff.type === "old" ? "" : diff.line))
+      .join("\n");
```
RomneyDa, line 507: main change in myers diff is that replaced content has "" for old lines since they will have the ghost text

extensions/vscode/src/diff/vertical/manager.ts
```diff
@@ -344,11 +344,11 @@ export class VerticalDiffManager {
       // startLine += effectiveLineDelta;
       // endLine += effectiveLineDelta;
 
-      existingHandler.clear(false);
+      await existingHandler.clear(false);
     }
 
     await new Promise((resolve) => {
-      setTimeout(resolve, 200);
+      setTimeout(resolve, 150);
```
RomneyDa, line 351: wonder if this promise was required because of the lack of await on the one above. Figured leave it for now

extensions/vscode/src/diff/vertical/decorations.ts
```diff
@@ -1,26 +1,33 @@
 import * as vscode from "vscode";
 
-export const redDecorationType = vscode.window.createTextEditorDecorationType({
+const removedLineDecorationType = (line: string) =>
```
RomneyDa, line 3: this decoration type is dynamically created so each line has a unique decoration type that has the ghost test as `after` content

extensions/vscode/src/diff/vertical/decorations.ts
```diff
@@ -107,5 +106,86 @@ export class DecorationTypeRangeManager {
...
       }
     }
+    this.editor.setDecorations(this.decorationType, this.ranges);
+  }
+}
+
+// Class for managing ghost-text decorations for removed lines (e.g. RED)
+// Behavior is slightly different all around
+// because each line will have a unique decoration type
+export class RemovedLineDecorationManager {
```
RomneyDa, line 116: Bread and butter:
Decided to create a new decoration manager for removed/red lines, since the behavior is pervasively slightly different with unique decorators per line, e.g. can't have ranges with multiple lines

extensions/vscode/src/diff/vertical/handler.ts
```diff
@@ -215,50 +205,62 @@ export class VerticalDiffHandler implements vscode.Disposable {
...
-  public getLineDeltaBeforeLine(line: number) {
-    // Returns the number of lines removed from a file when the diff currently active is closed
-    let totalLineDelta = 0;
-    for (const range of this.greenDecorationManager
-      .getRanges()
-      .sort((a, b) => a.start.line - b.start.line)) {
-      if (range.start.line > line) {
-        break;
-      }
+  // public getLineDeltaBeforeLine(line: number) {
```
Patrick-Erichsen: Can we just remove?

extensions/vscode/src/diff/vertical/handler.ts
```diff
@@ -215,50 +205,62 @@ export class VerticalDiffHandler implements vscode.Disposable {
...
       "setContext",
       "continue.streamingDiff",
       false,
     );
-    const rangesToDelete = accept
-      ? this.redDecorationManager.getRanges()
-      : this.greenDecorationManager.getRanges();
 
-    this.clearDecorations();
+    const deleteRangeLines = async (ranges: vscode.Range[]) => {
```
Patrick-Erichsen: Generally not a big fan of defining function expressions inline like this - main downsides are that we can't test this easily without testing the surrounding fn, can easily run into scope shadowing bugs, etc

extensions/vscode/src/diff/vertical/handler.ts
```diff
@@ -499,24 +504,27 @@ export class VerticalDiffHandler implements vscode.Disposable {
...
+        .filter((line) => line.type === "same" || line.type === "new")
+        .map((line) => line.line)
+        .join("\n") + "\n";
 
-    const diffs = myersDiff(oldFileContent, newFileContent);
+    const myersDiffs = myersDiff(oldFileContent, newFileContent);
 
-    const myersDiffLines = diffs.map((diff) => diff.line).join("\n");
+    const replaceContent = myersDiffs
+      .map((diff) => (diff.type === "old" ? "" : diff.line))
```
Patrick-Erichsen, line 506: Noticed a bug while testing locally - when deleting lines, since we've replaced the actual line content with empty spaces, the codelens buttons aren't aligned with the ghost text.

image

I tried the following:
```ts
      .map((diff) => (diff.type === "old" ? " ".repeat(diff.line.length) : diff.line))
```

But that appears to not work, the codelens aren't showing up at all:

image


Here's the sample code I was using, with the prompt "make subtract  return the result not `this`, add a comment to the return value on `add`, above the return statement":

```tsx
  add(number) {
    return () => {
      return () => {
        return () => {
          return () => {
            return () => {
              this.result += number;
              return this;
            };
          };
        };
      };
    };
  }

  subtract(number) {
    return () => {
      return () => {
        return () => {
          return () => {
            return () => {
              this.result -= number;
              return this.result;
            };
          };
        };
      };
    };
  }```

Patrick-Erichsen (changes requested):
I think the codelens bug I outlined is a blocker to shipping.

extensions/vscode/src/diff/vertical/handler.ts
RomneyDa: Looked into the codelens horizontal issue, there's no way to influence the horizontal position of a vscode codelens above a whitespace-only line as far as I can tell

RomneyDa: got really close by doing leading whitespace + special hidden unicode character e.g. /U200C + trimmed line as ghost text
but vscode highlights the special character

sestinj (approved):

sestinj:
turned out (after Dallin tried some very creative solutions) that there's not a way to adjust the positioning of the codelens where we want it

PR #5478, opened 2025-05-02
Raise edit/apply streaming errors

RomneyDa:
## Description
Fixes regression where edit and apply errors were thrown silently

### DIFF
--- a/extensions/vscode/src/diff/vertical/manager.ts
+++ b/extensions/vscode/src/diff/vertical/manager.ts
@@ -261,7 +261,7 @@ export class VerticalDiffManager {
       this.disableDocumentChangeListener();
       const handled = await handleLLMError(e);
       if (!handled) {
-        let message = "Error streaming diff";
+        let message = "Error streaming diffs";
         if (e instanceof Error) {
           message += `: ${e.message}`;
         }
@@ -473,7 +473,7 @@ export class VerticalDiffManager {
       this.disableDocumentChangeListener();
       const handled = await handleLLMError(e);
       if (!handled) {
-        let message = "Error streaming diff";
+        let message = "Error streaming edit diffs";
         if (e instanceof Error) {
           message += `: ${e.message}`;
         }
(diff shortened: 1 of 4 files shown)


extensions/vscode/src/util/errorHandling.ts
```diff
@@ -2,17 +2,15 @@ import { ILLM } from "core";
...
-function supportSpecialHandling(error: any): boolean {
-  let message: string | undefined = error.message;
-  return message !== undefined && message.toLowerCase().includes("ollama");
-}
-
 /**
  * @param error Handles common LLM errors. Currently only handles Ollama-related errors.
  * @returns true if error is handled, false otherwise
  */
-export async function handleLLMError(error: any): Promise<boolean> {
```
RomneyDa, line 14: changes to remove `any` type from error handling

extensions/vscode/src/diff/vertical/manager.ts
```diff
@@ -259,8 +259,13 @@ export class VerticalDiffManager {
       this.enableDocumentChangeListener();
     } catch (e) {
       this.disableDocumentChangeListener();
-      if (!handleLLMError(e)) {
-        vscode.window.showErrorMessage(`Error streaming diff: ${e}`);
+      const handled = await handleLLMError(e);
```
RomneyDa, line 262: This was the line causing the error - handleLLMError returns a promise, so !handle... was never true

Patrick-Erichsen: I thought we had a eslint rule for this? https://www.npmjs.com/package/eslint-plugin-no-floating-promise

I made a Linear issue for it titled 'Add eslint-plugin-no-floating-promise'

Patrick-Erichsen (approved):

extensions/vscode/src/diff/vertical/manager.ts
RomneyDa: I think the negation makes it not floating somehow

RomneyDa: `no-misused-promises` might be the rule we want

PR #5481, opened 2025-05-02
Streaming diff - error on max tokens reached (+recursive framework)

RomneyDa:
## Description
Throw an error in edit/apply diff streaming when max token limit reached. Prevents truncation of files

I've placed this inside a recursive streaming function that would be great to implement soon but is commented out for now:

- Creates a recursive "Continue where you left off" generator that keeps track of tokens and recurses before hitting max tokens
* not infinite but makes the limiting factor the context length not the maxTokens, which is a 10x for many models
- with this implementation loses prediction abilities after first recursion
- in draft state because seems to be having whitespace issues that cause extra diff lines at recursion boundaries (see pictures with three recursions and a problematic diff at between the 2nd and 3rd group of ~400 tokens

image

### DIFF
--- a/core/edit/recursiveStream.ts
+++ b/core/edit/recursiveStream.ts
@@ -1,0 +1,101 @@
+import { ChatMessage, ILLM, Prediction, PromptLog } from "..";
+import { DEFAULT_MAX_TOKENS } from "../llm/constants";
+import { countTokens } from "../llm/countTokens";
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen (approved):

RomneyDa:

sestinj:
recheck

PR #5503, opened 2025-05-04
open config when clicking on editing the prompt

uinstinct:
## Description

the edit button on the prompts section opens the config page (local or hub) like other sections

## Checklist

## Screenshots

https://github.com/user-attachments/assets/747812d8-8f5c-466f-9cab-bc6a3d39e8c7

https://github.com/user-attachments/assets/04068f52-a4ff-4214-888c-76bf5565dfec

## Testing instructions

### DIFF
--- a/gui/src/components/mainInput/Lump/sections/PromptsSection.tsx
+++ b/gui/src/components/mainInput/Lump/sections/PromptsSection.tsx
@@ -4,6 +4,8 @@ import {
 } from "@heroicons/react/24/outline";
 import { BookmarkIcon as BookmarkSolid } from "@heroicons/react/24/solid";
 import { SlashCommandDescription } from "core";
+import { useContext } from "react";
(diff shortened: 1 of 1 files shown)


uinstinct:
When there are prompts inside the .continue/prompts folder, it does not open the exact file.

Maybe we can refactor [`openConfigProfile`](https://github.com/continuedev/continue/blob/8665a083604f51244f60e0f9b5b11233e628f97e/core/config/ConfigHandler.ts/#L470-L484) to have a file uri as well.
We will need to have a key value of `fileURI` for `prompt` type. This file uri will contain the location of the .prompt file

RomneyDa (approved):
@uinstinct nice find, this is an improvement.
A further improvement would be to open the actual slug for hub blocks, similar to how rules does it. Might have to pass slug around on slashcommand description

PR #5514, opened 2025-05-05
feat: make `assistant` optional on sdk

Patrick-Erichsen:
- Made the `assistant` parameter optional in `ContinueClientOptions`, returning only the API client when omitted.
- Added function overloads to improve type safety when destructuring client properties.

These changes make it so that applications can use the Continue API without needing to pass an assistant slug upfront.


DocMAX:
my local assistans in ~/.continue/assistants/*.yaml are not loaded anymore?!

sestinj (approved):

PR #5519, opened 2025-05-05
Pe/shortcut bg color

Patrick-Erichsen:
## Description

Resolves https://github.com/continuedev/continue/issues/5507

Removes a hardcoded color on the shortcuts table with an appropriate theme based value.


sestinj (approved):
continue-sdk updates probably shouldn't be in here (maybe we can automate them), but otherwise this is very obviously good to go

RomneyDa:
@Patrick-Erichsen looks like no file changes?

PR #5528, opened 2025-05-06
fix(mode-switch)-tools-calls-broken

chezsmithy:
## Description

* Currently if the user is in agent mode and uses tools and swtiches back to chat mode a tools call error occurs.
* Further you can't currently use any tools in chat mode which isn't helpful if you want to lookup information on the web, call MCP tools, etc.
* Further in chat mode, it doesn't really make sense that you can use destructive tools such as edit or create file. You also want to block the use of the terminal tool creating new files etc. I've added a new list of unsafe tools for chat mode as as new const: CHAT_UNSAFE_TOOLS.
* I've added separate default system messages for Chat and Agent mode. I've made changes so that when you attempt to do unsafe operations in Chat mode, you are suggested to switch to Agent mode to continue.

## Checklist

## Screenshots

## Testing instructions

1. Start in Agent mode and add a power function to @test.js.
2. Note the change is done as expected.
3. Now ask to remove the function.
4. Note the change is done as expected.
5. Switch to chat mode.
6. Ask to add the power function to @test.js
7. Note that the edit isn't made, just a suggestion is presented.
8. Ask for the edit to be applied.
9. Note that switching to agent mode is suggested.
10. Switch back to agent mode.
11. Ask for the edit to be applied.
12. Note the edit is applied.

### DIFF
--- a/core/llm/constructMessages.ts
+++ b/core/llm/constructMessages.ts
@@ -8,14 +8,73 @@ import {
 } from "../";
 import { findLast } from "../util/findLast";
 import { normalizeToMessageParts } from "../util/messageContent";
-import { isUserOrToolMsg } from "./messages";
+import { isUserOrToolMsg, messageHasToolCallId } from "./messages";
 import { getSystemMessageWithRules } from "./rules/getSystemMessageWithRules";
+import { CHAT_UNSAFE_TOOLS } from "../tools/builtIn";
 
 export const DEFAULT_CHAT_SYSTEM_MESSAGE_URL =
   "https://github.com/continuedev/continue/blob/main/core/llm/constructMessages.ts";
 
+export const DEFAULT_AGENT_SYSTEM_MESSAGE_URL =
+  "https://github.com/continuedev/continue/blob/main/core/llm/constructMessages.ts";
+
 export const DEFAULT_CHAT_SYSTEM_MESSAGE = `\
 <important_rules>
+  You are in chat mode.
+  Only use functions that are explicitly listed in the toolsConfig provided to you. Ignore any previous tools that might have been used.
+  Before making any function call, verify that the function name exists in your current toolsConfig.
+  Never attempt to use functions you've observed in context if they aren't in your current toolsConfig.
+  Ignore any examples that use unavailable functions.
+  In chat mode, only suggest changes rather than attempting to make them directly. and can't actually edit or create files.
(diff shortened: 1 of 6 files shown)


chezsmithy:
@Patrick-Erichsen @tomasz-stefaniak FYI based on issues I am attempting to resolve. Note this is the current top complaint from my team that uses continue.

Patrick-Erichsen:
Hey @chezsmithy , thanks for the PR!

Chatted this over with @sestinj this morning - we both agreed that the chat/agent system messages are a great addition, but the tool calling in Chat mode we'd like to avoid. I can see the value in having the subset of tools available in Chat, but the problem is that exposing these tools makes the models way too eager to use tools instead of just answering questions directly, which is the intended behavior for Chat. E.g. I tried to ask a simple question with 3.7 Sonnet and instead it searched my codebase, couldn't find the answer, then searched the web, etc

With that being said, a few points of feedback on the system message updates:
- Can we update the rule title in the GUI so that instead of always `Default chat system message`, it says `Default agent system message` when in agent mode?
- In general I think the agent system message could be shortened a bit
- The "lazy edits" part of both sys msgs should be pulled out into a variable and reused in both to make sure they use the same instructions for this

chezsmithy:
@Patrick-Erichsen how about I leave the code in place for tools use in Chat Mode but just disallow all tools for the first pass at the PR. Then later we decide to dial them on one by one with a chat safe approach vs a chat unsafe approach? For my user base they see chat mode like a safe place to do research without affecting code. MCP and tools use is a key requirement to perform research, pull in JIRA tickets, look at files etc. curious on your thoughts?

chezsmithy:
@Patrick-Erichsen ok, I've made the suggested changes and cleaned up. I've got a problem with a failing GUI test. Now that tool calls are explicitly blocked in chat mode I need a way to add the Mode selector to the GUI test. I tried to do that, but can't quite seem to get the syntax right. Could you please review and suggest how I might resolve the GUI test issue? Thx!

Patrick-Erichsen:
@chezsmithy pushed up a fix that will hopefully resolve it, your syntax was correct, I think we just needed to hardcode the `capabilities` for the model to allow for `tool_use` 👍  thanks for updating the tests!

Patrick-Erichsen (approved):
🚀

PR #5543, opened 2025-05-06
Draft: autocompletion rework and granite completion fixes

halfline:
This is a draft merge request that reworks how autocomplete behaves with regard to the intellisense suggestion widget. 

The core idea is that the LLM is the smart thing, so it should be informing the suggestion widget, not the other way around. The top item in the suggestion widget is frequently something the user doesn't want, so including it in autocomplete can make the autocomplete text go off in the wrong direction.

If we defer showing the suggestion widget until the LLM has come back, then we can get a good inline complete and still show the suggestion pop up, but with the LLM suggested menu item preselected.

We can also simplify the completion item range calculation, by doing this, because we're only ever inserting new code, not replacing existing code.

This is a draft pull request, because I've only tested it with granite models at the moment, and there are some hacks for granite models at end of the patchset.

Once I do some more testing (with e.g. qwen2.5-coder or some other non-granite model), and drop some of the bottom commits, I'll drop draft status.


halfline:
So the code seems to work fine with qwen2.5-coder and claude-sonnet-3.7. codestral has some whitespace issues, but I think they may be specific to codestral.

core/autocomplete/postprocessing/index.ts
```diff
@@ -103,22 +104,29 @@ export function postprocessCompletion({
...
-      }
-    }
+    const prefixFragment = new SourceFragment(prefix);
+    const suffixFragment = new SourceFragment(suffix);
+    const completionFragment = new SourceFragment(completion);
+
+    const truncatedCompletion = completionFragment.getAsTruncatedFragment({
+      suffix: suffixFragment,
+      ignoreWhitespace: true
+    });
```
halfline: actually, digging through the code with @panyamkeerthana today, i don't think this should be necessary because the same thing should be happening at a different layer. the StreamTransformPipeline does this:

```
    charGenerator = stopAtStartOf(charGenerator, suffix);
```

here: https://github.com/continuedev/continue/blob/e4dd0241bfee86b4559e4cfadb6ba4ab89ad4758/core/autocomplete/filtering/streamTransforms/StreamTransformPipeline.ts#L36

So I may just be able to drop this commit entirely, or potentially drop this commit and tweak the implementation of `stopAtStartOf` if it's not engaging for some reason. will do some testing.

core/autocomplete/templating/AutocompleteTemplate.ts
```diff
@@ -60,6 +61,86 @@ const stableCodeFimTemplate: AutocompleteTemplate = {
...
+    const suffixFragment = new SourceFragment(suffix);
+    const [firstLineOfSuffix = undefined] = suffixFragment.head(1, {
+      ignoreWhitespace: true,
+    });
+
+    const stopTokenSequences = [
+      ...staticStopTokenSequences,
+      firstLineOfSuffix,
+    ].filter((sequence) => sequence !== undefined);
+
```
halfline: now that I know the design is to stop at the suffix for all LLMs, i should probably drop this and instead put this commit back on the branch:

```diff
commit 1cb51eaf08252c83f41da1760ed1d1591cce0a86
Author: Ray Strode <rstrode@redhat.com>
Date:   Mon Apr 28 15:28:46 2025 -0400

    autocomplete: Use suffix as stop token sequence
    
    Some LLM models (e.g., granite 3.3) run past the suffix in
    their completion output when using fill in the middle support.
    
    This commit adds the suffix as a stop token sequence to prevent
    that from happening.

diff --git a/core/autocomplete/templating/index.ts b/core/autocomplete/templating/index.ts
index 9722ee328..27e91227f 100644
--- a/core/autocomplete/templating/index.ts
+++ b/core/autocomplete/templating/index.ts
@@ -5,4 +5,5 @@ import { AutocompleteLanguageInfo } from "../constants/AutocompleteLanguageInfo"
 import { HelperVars } from "../util/HelperVars";
 
+import { SourceFragment } from "../../util/SourceFragment";
 import { SnippetPayload } from "../snippets";
 import {
@@ -111,9 +112,14 @@ export function renderPrompt({
         );
 
-  const stopTokens = getStopTokens(
+  // Some models don't stop at the suffix and just keep going, so extract the first line
+  // of the suffix and add it as a stop token sequence
+  const suffixFragment = new SourceFragment(suffix);
+  const [suffixStart = ""] = suffixFragment.head(1, { ignoreWhitespace: true });
+
+  const stopTokens = [...getStopTokens(
     completionOptions,
     helper.lang,
     helper.modelName,
-  );
+  ), suffixStart];
 
   return {
```

That way LLMs aren't generating content that we just then throw away.

core/util/SourceFragment.ts
```diff
@@ -0,0 +1,866 @@
...
+    if (bestMatchLineIndex < 0) return fragmentToTruncate;
+
+    // Build a new range that cuts right where overlap starts.
+    const adjustedRange: SourceFragmentRange = {
+      startLine: 0,
+      endLine: bestMatchLineIndex,
+      endLineLimit: bestMatchStartOffset,
+    };
+
+    // Truncate everything before the overlap.
```
halfline, line 370: this should say "Truncate to the non-overlapping section" or something

core/util/SourceFragment.ts
```diff
@@ -0,0 +1,866 @@
...
+    // at the seams in some cases between what is matched and what is left to
+    // to complete. This lets leading indent on the line get counted as part
+    // of the completion even if ignoreWhitespace is true.
+    if (mergeWhitespace) {
+      range = this.findIndentationMergeRange({ ignoreWhitespace });
+      if (range) return this.referenceFragment.getAsFragment({ range });
+    }
+
+    // If the source fragment is empty, we should just return the entire reference fragment
+    // If true,matchingRange at this point covers the entire reference fragment, but no
```
halfline, line 425: missing space after comma. should probably mention that not specifying the range when it is the entire fragment hits a fast path.

extensions/vscode/src/commands.ts
```diff
@@ -794,6 +796,56 @@ const getCommandsMap: (
...
+        "type",
+        async (args) => {
+          const editor = vscode.window.activeTextEditor;
+          if (editor) {
+            await editor.edit((editBuilder) => {
+              editor.selections.forEach((selection) => {
+                editBuilder.insert(selection.active, args.text);
+              });
+            });
+          }
```
halfline: @owtaylor hammered on this pr last week and noticed an odd issue...  If the user is typing code and hits enter, VS Code no longer autoindents on the next line. I suspect the cause is this commit. We're no longer inserting the text into the editor from a "keyboard" source, I believe, so it turns off auto indentation logic. We either need to figure out how to get that logic back on, or drop this hack and come up with a different way of disabling the suggestion widget (change user settings? ugh)

RomneyDa:
@halfline closing for now since this appears to be stale and has many merge conflicts. Feel free to reopen when ready!

PR #5544, opened 2025-05-06
Docs: in JSON migration, adding keepAlive equivalent in YAML

Boute95:
## Description

JSON completionOptions.keepAlive has no equivalent in YAML's defaultCompletionOptions. Instead, user need to use keep_alive in extraBodyProperties.

Added also keep_alive usage in config example below (StarCoder autocomplete model).

### DIFF
--- a/docs/docs/yaml-migration.md
+++ b/docs/docs/yaml-migration.md
@@ -40,7 +40,7 @@ Add all model configurations in `config.json`, including models in `models`, `ta
 
 Model-level `requestOptions` remain, with minor changes. See [YAML Continue Config Reference](./reference.md#models)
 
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen:
Hi @Boute95 , appreciate the PR but I think this is a bit too niche to call out specifically in the migration guide, so closing for that reason.

Boute95:
No problem I was a bit thinking the same :)
I was just confused for a while as to why Ollama was unloading my models, and then found the setting was not equivalent in YAML.

Thank you for your work have a good day !

ZobairQ:
Why wouldn't this be part of the documentation?? I have been searching for this entire day! This is very useful, especially since continue plugin decides to mess with default keep_alive configuration!

PR #5555, opened 2025-05-07
Fix: Prioritize user-defined embed models over Transformers.js built-in

PhilipAD:
## Description
When defining an embed model with `roles: [embed]` in the YAML config file, the extension defaults to using the built-in Transformers.js embedder instead of the user-defined model. This PR fixes this issue by ensuring user-configured embed models in config.yaml are prioritized over the built-in Transformers.js model.

## Changes Made
1. Updated `core/config/yaml/loadYaml.ts` to:
   - Only add the Transformers.js embedder when no other embed models exist
   - Explicitly initialize the selected embed model to prioritize non-Transformers.js models
2. Enhanced `core/config/selectedModels.ts` to prioritize non-Transformers.js models when selecting the initial embed model

## Testing
I've tested these changes with a custom OpenAI-compatible embed model defined in config.yaml with `roles: [embed]`, and confirmed that:
- The custom model is now automatically selected on startup
- The built-in Transformers.js model is still available as a fallback
- The changes don't affect other model roles

## Checklist

### DIFF
--- a/core/config/selectedModels.ts
+++ b/core/config/selectedModels.ts
@@ -3,8 +3,8 @@ import { ModelRole } from "@continuedev/config-yaml";
 import { ContinueConfig, ILLM } from "..";
 import { LLMConfigurationStatuses } from "../llm/constants";
 import {
-  GlobalContext,
-  GlobalContextModelSelections,
+    GlobalContext,
+    GlobalContextModelSelections,
 } from "../util/GlobalContext";
 
 export function rectifySelectedModelsFromGlobalContext(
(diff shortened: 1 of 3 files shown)


sestinj:
Nice find! I want to make sure we are able to test this, but I also understand that the work required to do that would probably blow up the scope of this PR. I'm going to make an attempt at doing this before merging

PhilipAD:
Does this still need merging? @sestinj

sestinj:
looking back at this I realistically don't think it's something we want to change given the amount of feedback we've gotten about this particular ability versus the critical path that it would be updating. we may be able to circle back to this later but especially given the decreased prevalence of embeddings vs. agentic search it doesn't feel high on our priority list. as always open to further feedback if you have it

PR #5564, opened 2025-05-07
fix(files-context)-refresh-on-new-file-and-save

chezsmithy:
## Description

* Traced the refreshes happening when new files are created and saved in the vscode UI. Noted that based on the parameter being sent to refresh the context certain contexts were being ignored noting they needed to wait for indexing to occur, but indexing was already performed before the call so I believe this was in error.
* Updated the create file tool to also make a call into refreshing the indexes.

TODO (Recommend a separate PR):

* Could also solve for adding a file watcher to the workspace to detect changes that might be caused by the terminal tool, or made through running scripts and via that hook update the index, and the contexts.

## Checklist

## Screenshots

## Testing instructions

New File in VsCode
1. Create a new file in the VsCode UI, and save it.
2. Validate the file now shows up in the @files context.

In Agent Mode
1. Ask for a new file to be created.
2. Validate the file now shows up the @files context.

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -735,6 +735,7 @@ export class Core {
         tool,
         toolCallId: toolCall.id,
         onPartialOutput,
+        core: this, // Add a reference to the Core instance
       });
     });
 
@@ -1075,7 +1076,7 @@ export class Core {
     this.indexingCancellationController = undefined;
   }
 
-  private async refreshCodebaseIndexFiles(files: string[]) {
+  async refreshCodebaseIndexFiles(files: string[]) {
     // Can be cancelled by codebase index but not vice versa
     if (
       this.indexingCancellationController &&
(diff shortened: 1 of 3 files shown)


Patrick-Erichsen:
Hey @chezsmithy , chatted with @sestinj on this. I think we'd like to avoid injecting the entire `core` obj as a dependency to `callTool`. 

Instead, we're thinking of creating a more scoped class for indexing related logic, eg `refreshCodebaseIndexFiles`, and then we can just inject that.

I'm planning to do that next week, so once that's merged, I'll push up a commit here to use that new class, and then we can get this merged 👍

chezsmithy:
Note to add tests related to On Save events triggering excessive indexing. 

Note to review JetBrains impacts.

chezsmithy:
@Patrick-Erichsen a quick check in on the core indexing refactor?

chezsmithy:
@Patrick-Erichsen  this is ready for review now. 

Note I caught another await that needed to be moved to a void to avoid blocking the indexing update.

Patrick-Erichsen (approved):

PR #5581, opened 2025-05-08
fix(diff-editor)-performance

chezsmithy:
## Description

When the diff editor was running in vscode it was frustrating that the progress would slowly start from the top of the file and make it's way to the bottom. Sometimes it felt like nothing was happening until the change cursor came into view.

I added code to change the focus in the editor depending on where the current edit is occuring. Interestingly this had the side effect of speeding the whole process up imensely for me. I'm curious if others see the same difference.

I also added code to push the viewport to the top of the file after all the edits are complete to give a visual indication that the process is done. Note, it might be better to focus the editor at the first diff but I didn't see an easy way to determine where that was without a more significant change.

## Checklist

## Screenshots

## Testing instructions

1. Ask for logging to be added to @test.js
2. Observe that after a bit of a wait the editing to the test.js file starts
3. Note that the current line displayed in the editor window follows the changes being made.
4. Observe the speed of applying those edits seems to be significantly improved. (side effect?)

### DIFF
--- a/extensions/vscode/src/diff/vertical/handler.ts
+++ b/extensions/vscode/src/diff/vertical/handler.ts
@@ -127,6 +127,11 @@ export class VerticalDiffHandler implements vscode.Disposable {
   private incrementCurrentLineIndex() {
     this.currentLineIndex++;
     this.updateIndexLineDecorations();
+    const range = new vscode.Range(this.currentLineIndex, 0, this.currentLineIndex, 0);
+    this.editor.revealRange(
+      range,
+      vscode.TextEditorRevealType.Default,
+    );
   }
 
   private async insertTextAboveLine(index: number, text: string) {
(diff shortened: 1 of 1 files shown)


chezsmithy:
@Patrick-Erichsen I'm unclear why this has broken the e2e tests?

Patrick-Erichsen (approved):
Wow, very simple addition but makes a huge difference! I tested this with Edit to see how it would behave on a single-range change (i.e. not a full file) and it also worked great, the jump-to-first-line-of-range logic is quite nice.

Thanks for the attention to detail here!

chezsmithy:
@Patrick-Erichsen thanks! I was a bit shocked by the performance change. I wonder if it has something to do with making changes in the editor outside the currently displayed viewport? None the less it's a quick win.

Patrick-Erichsen:
Let's keep an eye out for user feedback on this one. Since this effectively forces the viewport to follow along with edits it prevents users from pausing to read the code as it's streamed in.

HQ1363:
Jetbrains, it's also slow. Is there room for optimization?

Issue #5542, opened 2025-05-06
Error executing ripgrep tool on PyCharm (Windows)


jnslmk:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 1.0.16
- IDE version: PyCharm 2025.1
- Model: Azure GPT-4o
- config:
  
  - name: Azure Gpt-4 Omni
    provider: azure
    apiBase: https://redacted.openai.azure.com/
    model: gpt-4o
    apiVersion: 2024-08-01-preview
    apiKey: redacted
    requestOptions:
      proxy: http://proxy.local:80
      caBundlePath: C:/path/to/crt.crt
    apiType: azure
    roles:
      - chat
      - edit
  
  OR link to assistant in Continue hub:
```

### Description

When querying my codebase I get the following error:
```text
Error handling message of type getSearchResults: com.intellij.execution.process.ProcessNotCreatedException: Cannot run program "C:\Users\<redacted>\AppData\Roaming\JetBrains\PyCharm2025.1\plugins\continue-intellij-extension\ripgrep\bin\rg.exe" (in directory "D:\<redacted>"): CreateProcess error=2, The system cannot find the file specified
```

The `plugins\continue-intellij-extension\ripgrep\bin\` directory only contains an `rg` file which is not executable on Windows if renamed to `rg.exe`.

Manually adding the ripgrep executable from https://github.com/BurntSushi/ripgrep/releases/ fixes the issue.

### Log output

```Shell

```

Patrick-Erichsen:
Thanks for verifying that adding the binary manually works.

> When querying my codebase I get the following error:

To be clear, this is when you are in Agent mode?

jnslmk:
Yes, this happens in agent mode when the plugin tries to search the codebase with ripgrep

Patrick-Erichsen:
I just confirmed this on IntelliJ too. Will look into it this week!

Patrick-Erichsen:
@jnslmk please let me know if this is resolved in subsequent releases!

PR #5582, opened 2025-05-09
feat: manual ripgrep downloads for JB

Patrick-Erichsen:
## Description

Resolves https://github.com/continuedev/continue/issues/5542

This PR introduces a manual step in the build process for JetBrains to install `ripgrep` for the respective OS/arch.

In our VS Code build process, we don't need to do this because we run builds in CI on the appropriate runners for each OS/arch. But for JB, we need to upload a single zip that contains all the required deps for each OS/arch, so we only run on a single runner (`macOS-latest`), which then causes us to only install the ripgrep binary for that platform.

### DIFF
--- a/binary/build.js
+++ b/binary/build.js
@@ -8,13 +8,15 @@ const {
   execCmdSync,
   autodetectPlatformAndArch,
 } = require("../scripts/util");
+const { downloadRipgrep } = require("./utils/ripgrep");
+const { ALL_TARGETS, TARGET_TO_LANCEDB } = require("./utils/targets");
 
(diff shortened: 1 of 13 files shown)


Patrick-Erichsen:
@uinstinct @fbricon - mind taking a look at this since you both have some context from the recent ripgrep PR.

uinstinct (commented):
Tested and Ripgrep works fine on both vscode and jetbrains (intellij community)!

image

image

(BTW like the new PR size label feature!)

tomasz-stefaniak (approved):

PR #5588, opened 2025-05-09
fix test dirs being created inside core folder

uinstinct:
## Description

Some test directories were being created in the current directory (in our case the `core` folder) instead of the os tmp dir.

This was happening because [we were using the relative file path](https://github.com/continuedev/continue/blob/ec0180bf95bf6c127e78db5d8b02e3f57b7814a0/core/test/testDir.ts/#L46) (instead of the appended os tmp dir). This PR fixes it

## Checklist

## Screenshots

**before**
Screenshot 2025-05-09 at 5 18 40 PM

## Testing instructions

- cd into the core folder
- run `npm run test -- walkDir`
- see that test folders like `xyz` are created
    

---

## Summary by mrge
Fixed test directories being created in the core folder instead of the system temp directory.

### DIFF
--- a/core/test/testDir.ts
+++ b/core/test/testDir.ts
@@ -43,7 +43,7 @@ export function addToTestDir(pathsOrUris: (string | [string, string])[]) {
     if (Array.isArray(p)) {
       fs.writeFileSync(filepath, p[1]);
     } else if (p.endsWith("/")) {
-      fs.mkdirSync(p, { recursive: true });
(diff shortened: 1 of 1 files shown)


uinstinct:
> @uinstinct looks like tests are failing on this one, could you look into this?

Done Dallin!
Looks like my branch was outdated.

RomneyDa (approved):
Fixes using relative path rather than the path in the temp directory for test files

PR #5599, opened 2025-05-09
Revert "Update default local models"

tomasz-stefaniak:
I tested Ollama 3.1 with tool use and Qwen 3 with tool use and Ollama 3.1 seems both faster and more stable when using tools. We should revert to Ollama 3.1.
    

---

## Summary by mrge
Reverted the default local chat model from Qwen3-8B back to Llama 3.1 8B for onboarding. This restores the previous model and title used in the onboarding config.


sestinj (approved):

sestinj (approved):
Looks good. Seems like good reasoning for moving back.

I double-checked to be sure that this constant is the only place where we define the name of the suggested model

Issue #4098, opened 2025-02-11
Continue freezes IDE


cobrafast:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 10 Pro 22H2 64bit
- Continue version: 0.0.88
- IDE version: JetBrains Rider 2024.3.5
- Model: irrelevant?
- config.json: barebones
```

### Description

Randomly Rider will freze completely and hide the mouse cursor. Only recourse is to kill Rider from Task Manager.

* Problem does not happen with Continue plug-in disabled.
* Problem did not happen with Continue `0.0.83`.
* Seems to happen most frequently when editing somehwere in the middle of a line.

### To reproduce

1. Work normally. Problem will occur sooner or later.

### Log output

Continue core.log: https://pastebin.com/7xQ27HcX
JetBrains idea.log (excerpt): https://pastebin.com/172sKtBk

AayushSameerShah:
It happens to me too. In my case, with VS code, After installing and using the extension for a while, and when we close the VS code, the system removes VS code. So no matter what you do, click on VS code icon, or use "Open with VS Code" anything, the system doesn't find the application.

We have to uninstall it and reinstall from scratch.

sestinj:
Thanks for providing the logs. Unfortunately I'm not seeing anything obvious in there, and it's not something we've reproduced ourselves. Is there any chance you could share a bit more information?
- Are any of you on ARM64, or all x64?
- Does this happen on any and all codebases? I'm wondering if it is related to larger codebases and whether this would happen on a tiny codebase
- Does this still happen if you "Disable Indexing" in the settings page?

cobrafast:
It happened to me on a fresh console project with maybe 30 lines.

hitgub-john:
This also happened to me while using PyCharm, but doesn't happen when I'm using VS Code.

Siiiidox:
Can confirm this issue in Rider on x64 Windows. Stops occuring when after I disabled Autocomplete.

kclay:
[...]

PR #5600, opened 2025-05-09
feat(JB): limit to 4 threads in `handleMessage`

Patrick-Erichsen:
## Description

(Hopefully) resolves https://github.com/continuedev/continue/issues/4098

This PR implements advice from @ForNeVeR to limit the number of parallel threads to 4 in our message handling in JetBrains. See #4098 for context on why this is necessary.

    

---

## Summary by mrge
Limited parallel threads in JetBrains message handling to 4 to prevent UI freezing. This helps keep the IDE responsive during heavy message processing.


sestinj (approved):
Looks like this is a strictly safer version of Dispatchers.IO, unless we have a problem with not having enough threads. Will keep an eye on this while in EAP, but this resolves major problems

ForNeVeR (commented):
Yep, this should work.

Issue #5594, opened 2025-05-09
Indexing failed: File name too long


adambarber:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 15.3.2
- Continue version: 1.1.33
- IDE version: VSCode 1.99.3 (Universal)
- Model: n/a
- config: n/a
```

### Description

When working in a VSCode dev container on macOS, code base indexing fails due to "File name too long (os error 63)"

### To reproduce

1. Create a new Ruby on Rails project
2. Open the project in a devcontainer 
3. Manually reindex the code base
4. Indexing fails

(In the log output, I replaced a couple strings with "*" of the same length)

### Log output

```Shell
[...]

PR #5632, opened 2025-05-12
fix: truncate `tagToString` to max filename len

Patrick-Erichsen:
## Description

Resolves https://github.com/continuedev/continue/issues/5594

    

---

## Summary by mrge
Updated the tagToString function to truncate long strings, making sure table names and identifiers stay within safe filename limits. This prevents errors from overly long paths, especially in dev containers.

### DIFF
--- a/core/indexing/chunk/ChunkCodebaseIndex.ts
+++ b/core/indexing/chunk/ChunkCodebaseIndex.ts
@@ -4,7 +4,7 @@ import { RunResult } from "sqlite3";
 
 import { IContinueServerClient } from "../../continueServer/interface.js";
 import { Chunk, IndexTag, IndexingProgressUpdate } from "../../index.js";
-import { DatabaseConnection, SqliteDb, tagToString } from "../refreshIndex.js";
+import { DatabaseConnection, SqliteDb } from "../refreshIndex.js";
 import {
   IndexResultType,
   MarkCompleteCallback,
@@ -13,8 +13,9 @@ import {
   type CodebaseIndex,
 } from "../types.js";
 
-import { chunkDocument, shouldChunk } from "./chunk.js";
 import { getUriPathBasename } from "../../util/uri.js";
+import { tagToString } from "../utils.js";
+import { chunkDocument, shouldChunk } from "./chunk.js";
 
 export class ChunkCodebaseIndex implements CodebaseIndex {
   relativeExpectedTime: number = 1;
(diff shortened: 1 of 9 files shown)


Patrick-Erichsen:
@sestinj I don't believe this should break any existing indexes since it only uses the new logic if the tagName would have thrown an error using the old logic. But worth a sanity check.

core/indexing/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  const maxDirLength = 200; // Leave room for branch and artifactId
+  const dir =
+    tag.directory.length > maxDirLength
+      ? tag.directory.slice(0, maxDirLength)
+      : tag.directory;
+
+  return `${dir}::${tag.branch}::${tag.artifactId}`.slice(
+    0,
```
sestinj, line 42: For uniqueness, should truncate/remove initial characters of dir, not last

core/indexing/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * To handle long paths:
+ * 1. First tries the full string - most backwards compatible
+ * 2. If too long, truncates directory to 200 chars to leave room for branch and artifactId
+ * 3. Finally ensures entire string stays under MAX_TABLE_NAME_LENGTH for OS compatibility
+ *
+ * @param tag The tag containing directory, branch, and artifactId
+ * @returns A string representation safe for use as a table name
+ */
+export function tagToString(tag: IndexTag): string {
```
sestinj, line 28: Add some basic tests in `utils.test.ts`, can probably 1-shot

core/indexing/utils.ts
```diff
@@ -0,0 +1,41 @@
+import { IndexTag } from "..";
+
+// Maximum length for table names to stay under OS filename limits
+const MAX_TABLE_NAME_LENGTH = 240;
```
sestinj, line 4: either both constants inside the function or outside. inside seems to make the most sense given usage

core/indexing/chunk/ChunkCodebaseIndex.ts
```diff
@@ -4,7 +4,7 @@ import { RunResult } from "sqlite3";
 
 import { IContinueServerClient } from "../../continueServer/interface.js";
 import { Chunk, IndexTag, IndexingProgressUpdate } from "../../index.js";
-import { DatabaseConnection, SqliteDb, tagToString } from "../refreshIndex.js";
+import { DatabaseConnection, SqliteDb } from "../refreshIndex.js";
```
sestinj, line 7: nothing to do in your PR here, but reminder that we should add prettier as a CI step to resolve these messy diffs

sestinj (dismissed):
- tests
- trim from end instead of start
- move the constant

RomneyDa (approved):

PR #5637, opened 2025-05-12
fix: hardcode `fontSize` for tool call status msg

Patrick-Erichsen:
## Description

Fixes a UI bug where the `fontSize` for `<ToolCallStatusMessage />` was larger than the default.

Also configured eslint + prettier to format on save.

### Before
image

### After
image

    

---

## Summary by mrge
Set a fixed font size for the tool call status message to match the default UI text size.

- **Bug Fixes**
  - Tool call status message now uses the correct font size for consistency.

### DIFF
--- a/gui/src/context/LocalStorage.tsx
+++ b/gui/src/context/LocalStorage.tsx
@@ -18,6 +18,7 @@ export const LocalStorageProvider: React.FC<{ children: React.ReactNode }> = ({
 }) => {
   const [values, setValues] = useState<LocalStorageType>(DEFAULT_LOCAL_STORAGE);
 
+  // TODO setvalue
   useEffect(() => {
     const isJetbrains = getLocalStorage("ide") === "jetbrains";
     let fontSize = getLocalStorage("fontSize") ?? (isJetbrains ? 15 : 14);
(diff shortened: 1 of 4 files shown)


gui/src/pages/gui/ToolCallDiv/ToolCallStatusMessage.tsx
```diff
@@ -65,7 +66,7 @@ export function ToolCallStatusMessage({
     message = futureMessage;
   }
   return (
-    <div className="block">
+    <div className="block" style={{ fontSize: getFontSize() }}>
```
RomneyDa: Ideally use the `useFontSize` hook so that it doesn't retreive font size on any rerender

gui/src/context/LocalStorage.tsx
```diff
@@ -18,8 +18,6 @@ export const LocalStorageProvider: React.FC<{ children: React.ReactNode }> = ({
 }) => {
   const [values, setValues] = useState<LocalStorageType>(DEFAULT_LOCAL_STORAGE);
 
-  // TODO setvalue
```
RomneyDa, line 21: We should decide on if TODO comments should be left in, I'm usually a fan

RomneyDa (approved):
Is the addition of eslint and prettier files intentional?

Patrick-Erichsen:
> Is the addition of eslint and prettier files intentional?

Yes, should have called that out in the description. It sets up auto format/lint on save as a default behavior.

RomneyDa (approved):

PR #5638, opened 2025-05-13
fix: prevent autoscrolling on new tool msgs

Patrick-Erichsen:
## Description

Closes CON-1920

Prevents chat panel from auto scrolling on new tool messages by only overwriting the user's manual scroll state when we receive new messages of role type "user". 

This solves a bug where a user would scroll up to read some content in Agent mode, and then a new tool response would stream in, which would render as a new message, and force scroll the user to the bottom of the chat.

## Testing
- Enter Agent mode
- Send a message that uses some tools, eg "explore the codebase file by file"
- Wait for a full page of messages to fill up
- Then as messages continue to stream in, scroll up
- Confirm that you aren't force scrolled to the bottom of the chat as new tool responses come in

### DIFF
--- a/gui/src/pages/gui/Chat.tsx
+++ b/gui/src/pages/gui/Chat.tsx
@@ -134,6 +134,8 @@ export function Chat() {
     return isJetBrains();
   }, []);
 
+  useAutoScroll(stepsDivRef, history);
+
   useEffect(() => {
     // Cmd + Backspace to delete current step
     const listener = (e: any) => {
@@ -284,8 +286,6 @@ export function Chat() {
 
   const showScrollbar = showChatScrollbar ?? window.innerHeight > 5000;
 
-  useAutoScroll(stepsDivRef, history);
-
   return (
     <>
       {widget}
@@ -301,7 +301,7 @@ export function Chat() {
           <div
             key={item.message.id}
             style={{
-              minHeight: index === history.length - 1 ? "25vh" : 0,
+              minHeight: index === history.length - 1 ? "200px" : 0,
             }}
           >
             <ErrorBoundary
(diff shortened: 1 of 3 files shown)


gui/src/pages/gui/Chat.tsx
```diff
@@ -301,7 +301,7 @@ export function Chat() {
           <div
             key={item.message.id}
             style={{
-              minHeight: index === history.length - 1 ? "25vh" : 0,
```
RomneyDa, line 304: Realizing maybe I was wrong to suggest this yesterday, that on small screens this would cause it to pop out of view. Maybe 40vh?

gui/src/pages/gui/useAutoScroll.ts
```diff
@@ -1,14 +1,25 @@
...
   const [userHasScrolled, setUserHasScrolled] = useState(false);
+  const [numUserMsgs, setNumUserMsgs] = useState(0);
 
   useEffect(() => {
-    if (history.length) {
+    /**
+     * Only reset scroll state when a new user message is added to the chat.
+     * We don't want to auto-scroll on new tool response messages.
+     */
+    const newNumUserMsgs = history.filter(
```
RomneyDa: could cause jenky behavior if you e.g. went to the history page and returned, because numUserMsgs would be reset to 0
Maybe numUserMsgs = useMemo(count user messages) and use that as a dependency?

gui/src/pages/gui/Chat.tsx
Patrick-Erichsen: Hmm, by "small screens" do you mean height of the screen? 200px is more than small enough for any size computer screen imo

gui/src/pages/gui/useAutoScroll.ts
Patrick-Erichsen: What about doing `const [numUserMsgs, setNumUserMsgs] = useState(history.length);`?

RomneyDa (approved):
Fixes scrolling on new messages in agent mode

Issue #5353, opened 2025-04-25
Automatic bypass for edit tool


chezsmithy:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

When I choose to allow the edit tool to work in automatic mode, I’m still asked to approve each change before the tool proceeds.

### Solution

Update the edit tool, so that when in automatic mode each of the diff apply questions is bypassed and each change is auto applied. This will implement full YOLO mode.

chezsmithy:
I’m noting I’m willing to submit a PR to solve for this issue. I’m also interested in other solutions or suggestions on implementation.

RomneyDa:
@chezsmithy this would be interesting! Thinking to keep uniform behavior for tool policies, this could be a settings toggle in the Settings Page. Potentially could also apply to chat codeblocks that match an existing file.

Top setting name candidates
- Full YOLO mode
- Edit Tool Auto-Apply
- Automatically Apply Agent Edits

I think a good start would be to
1. Add an "Automatically Apply Agent Edits" or similar settings toggle in the settings page that sets a new shared config value "autoApplyAgentEdits" or similar
2. In the Client Tool Implementation of the edit tool, reference the value of `autoApplyAgentEdits` to conditionally accept all pending diffs after applies complete (see accept all button). Then, streaming should automatically continue after complete without changing any other logic!

RomneyDa:
Fixed by https://github.com/continuedev/continue/pull/5657

dy:
Hmm, I don't see that option in my installed version 1.2.22. Quite annoying that z.ai asks every edit the approval command

Foorack:
@dy  Seems like it was removed here? https://github.com/continuedev/continue/pull/8310

Foorack:
I cannot find any information AT ALL why they removed this feature. Makese no sense to me. Luckily I found you can configure Ollama as a provider in Copilot extension.

nateriverdev:
I'm using the Continue VSCode extension and really need a YOLO mode so it can automatically approve all permissions, instead of me having to sit there pressing Enter for every single step 
Additionally, if possible, please add auto-retry when errors occur, along with suggestions and a task list when interacting with it.  :)
[...]

PR #5657, opened 2025-05-13
Setting for auto-accepting agent mode edits

RomneyDa:
@chezsmithy's "YOLO" mode requested in #5353 

If the new `Auto-Accept Agent Edits` setting is on the edit tool implementation automatically accepts diffs generated from applying once applying is complete. This setting is under a new "Experimental Settings" dropdown and shows a warning that this is a risky setting.
    
image

Note: I have decided not to add a note in the Agent mode docs for this as for now it is considered a dangerous setting for power users. But the user settings page docs have been updated


chezsmithy:
@RomneyDa Sweet 🎉

docs/docs/customize/deep-dives/settings.md
```diff
@@ -10,23 +10,24 @@ The **User Settings page** can be accessed by clicking the gear icon in the head
 
 ![slash-commands](/img/header-buttons.png)
 
-Which takes you to this page:
+Which takes you to this page (select the "Settings" tab):
```
Patrick-Erichsen, line 13: nitpick but can we just include the tabs in the screenshot to make this more obvious?

gui/src/components/gui/Switch.tsx
```diff
@@ -1,23 +1,40 @@
...
 import React from "react";
 import { vscButtonBackground } from "..";
+import { ToolTip } from "./Tooltip";
 
 type ToggleSwitchProps = {
   isToggled: boolean;
   onToggle: () => void;
   text: string;
   size?: number;
+  onWarningText?: string;
```
Patrick-Erichsen: I would lean towards a more generic `warningItem` prop here to enable more flexibility for the consuming component to use whatever it wants. 

Although you could make the argument that we actually want to limit variability here and limit it to just the exclamation triangle, but in that case `onWarningText` isn't semantically clear to me since the "on" prefix is typically for event handlers, so I'd call it just `warningText`

gui/src/components/gui/Switch.tsx
```diff
@@ -1,23 +1,40 @@
...
     <div className="flex cursor-pointer select-none items-center justify-between gap-3">
       <span className="truncate-right">{text}</span>
-      <div>
+      <div className="flex flex-row items-center gap-1">
+        {isToggled && !!onWarningText && (
+          <>
+            <ExclamationTriangleIcon
+              data-tooltip-id={`${text}-warning-tooltip`}
+              className="text-yellow-500"
+              style={{
```
Patrick-Erichsen: can we do this in tailwind instead?

gui/src/components/gui/Switch.tsx
```diff
@@ -1,23 +1,40 @@
...
   return (
     <div className="flex cursor-pointer select-none items-center justify-between gap-3">
       <span className="truncate-right">{text}</span>
-      <div>
+      <div className="flex flex-row items-center gap-1">
+        {isToggled && !!onWarningText && (
+          <>
+            <ExclamationTriangleIcon
+              data-tooltip-id={`${text}-warning-tooltip`}
+              className="text-yellow-500"
```
Patrick-Erichsen: no explicit color?

gui/src/util/clientTools/editImpl.ts
```diff
@@ -25,9 +25,24 @@ export const editToolImpl: ClientToolImpl = async (
...
+      filepath: firstUriMatch,
+    });
+    if (out.status === "error") {
+      throw new Error(out.error);
+    }
+    return {
+      respondImmediately: true,
+      output: undefined, // TODO - feed edit results back to model (also in parallel listeners)
+    };
+  } else {
```
Patrick-Erichsen: another nit but since the `if` clause has a return we could remove the `else` clause here and avoid the unnecessary nesting

Patrick-Erichsen (approved):
few minor nits but good to go 👍  will reapprove once we add the experimental section dropdown

gui/src/components/gui/Switch.tsx
RomneyDa: setting up this color on another branch!

docs/docs/customize/deep-dives/settings.md
RomneyDa: done!

gui/src/components/gui/Switch.tsx
RomneyDa: sounds good, I try to avoid e.g. `h-${size}` with dynamic sizing based on a prop but since pulling comp out will use tailwind direct

gui/src/components/gui/Switch.tsx
RomneyDa: good point, changed to `showIftogged` comp passed

Patrick-Erichsen (approved):

PR #5662, opened 2025-05-14
fix: :bug: Remove all history when a user request is cancelled

chezsmithy:
## Description

Updated clearLastEmptyResponse to fully remove all history up until the last user message. This cleans up any dangling thinking, or tools use messages that may have been sent when the cancel button is used.

This should resolve any issues where dangling thinking or tools use messages are left behind when the cancel button is used breaking the ability to continue the session.

TODO: I discovered one edge case that I could use help fixing. When a file is edited there is a small window between when the edit is started and the editor displays the edits, Accept and Reject Buttons where the request can be cancelled. In this case the cancel button can leave things in an odd state with code still highlighted and the Accept / Reject buttons left behind. I wasn't sure how to check if we are in that state, and fully cancel the Edit Session. Effectively, I would like to perform he same operations as clicking the Reject Button. (Remove decorations, and cancel the edit session).

## Checklist

## Screenshots

## Testing instructions

Enable Thinking
In Agent Mode:

1. Ask to update test.js and add a power function.
2. When thinking is displayed hit cancel.
3. Observe that thinking is removed, and you are retunred to the previous command.

In Agent Mode:

1. Ask to update test.js and add a power function.
2. Once thinking ends, and the LLM is starting to stream the next steps hit cancel.
3. Observe the edit step is stopped, and the thinking modes are cancelled.

Outcomes:
* All steps after the last user message should be removed from the UI and the messages should be removed from history. Tool calls in progress should be stopped.

### DIFF
--- a/gui/src/components/mainInput/Lump/LumpToolbar.tsx
+++ b/gui/src/components/mainInput/Lump/LumpToolbar.tsx
@@ -96,6 +96,9 @@ export function LumpToolbar() {
         <StopButton
           className="text-gray-400"
           onClick={() => {
+            if (toolCallState?.status === "calling" || toolCallState?.status === "generating") {
+              dispatch(cancelCurrentToolCall);
+            }
             dispatch(cancelStream());
           }}
         >
--- a/gui/src/redux/slices/sessionSlice.ts
+++ b/gui/src/redux/slices/sessionSlice.ts
@@ -107,15 +107,21 @@ export const sessionSlice = createSlice({
       if (state.history.length < 2) {
         return;
       }
-      const lastMessage = state.history[state.history.length - 1];
 
-      // Only clear in the case of an empty message
-      if (!lastMessage.message.content.length) {
-        state.mainEditorContentTrigger =
-          state.history[state.history.length - 2].editorState;
-        state.history = state.history.slice(0, -2);
-        // TODO is this logic correct for tool use conversations?
-        // Maybe slice at last index of "user" role message?
+      // Find the index of the last user message
+      let lastUserIndex = -1;
+      for (let i = state.history.length - 1; i >= 0; i--) {
+        if (state.history[i].message.role === "user") {
+          lastUserIndex = i;
+          break;
+        }
+      }
+
+      // If we find a user message, set the editor state to the last user messsage
+      // and slice the history to remove it and all the messages after it
+      if (lastUserIndex >= 0) {
+        state.mainEditorContentTrigger = state.history[lastUserIndex].editorState;
+        state.history = state.history.slice(0, lastUserIndex);
       }
     },
     // Trigger value picked up by editor with isMainInput to set its content


gui/src/redux/slices/sessionSlice.ts
```diff
@@ -107,15 +107,21 @@ export const sessionSlice = createSlice({
...
         return;
       }
-      const lastMessage = state.history[state.history.length - 1];
 
-      // Only clear in the case of an empty message
-      if (!lastMessage.message.content.length) {
-        state.mainEditorContentTrigger =
-          state.history[state.history.length - 2].editorState;
-        state.history = state.history.slice(0, -2);
-        // TODO is this logic correct for tool use conversations?
```
chezsmithy, line 117: This TODO was on the mark. I've followed up with an implementation that follows this line of thinking and peformed extensive testing to validate it.

RomneyDa (changes requested):
@chezsmithy this will be a great problem to solve, has been on my mind recently. 

The main issue and reason I hadn't approached this yet is that (correct me if I'm wrong) if the user cancels streaming in Agent mode with this implementation after several turns of the conversation, there will be no recent user messages and it will essentially truncate to the beginning of the convo since the pattern is 
user -> assistant + calls -> tool -> assistant + calls -> tool -> assistant + calls etc.

In this case we also can't just truncate back to the last tool message because (if I remember right) e.g. Anthropic won't accept the handshake user -> assistant + calls -> tool (= user in anthropic/gemini) -> user

Would love ideas/brainstorming on how to solve this!

chezsmithy:
@RomneyDa I ran a number of scenarios last night, and I agree with your assessment. There really isn't a good way to roll back to a point in time and keep all the call data balanced to avoid failures with the next API call.

I've seen other tools handle this differently with checkpoints that allow you to restore artifact state but do not modify the history and do not allow a cancel option.

Typically the loss of context from the last user message seems to be pretty short. The biggest loss can be tool outcome state. Perhaps there is a way to summarize those lost outcomes and inject that into the next user message, or maybe provide that as an option for the user? (E.g You are restarting a cancelled flow, would you like for me to include a summary of your cancelled tool calls into the next request?).

chezsmithy:
Another thought.

Maybe we have adaptive cancel. 

If you cancel a thinking call or LLM request after a user message, it rolls to the last user message.

If you cancel during a tools call it only cancels the call but doesn't roll back. Perhaps if a tools call has occurred since the last user message we never allow roll back, only cancelling the next tool call.

core/index.d.ts
```diff
@@ -362,6 +369,7 @@ export interface AssistantChatMessage {
   role: "assistant";
   content: MessageContent;
   toolCalls?: ToolCallDelta[];
+  toolCallId?: string,
```
RomneyDa, line 372: toolCallId shouldn't be on user chat message since one user message can have multiple tool calls, should already be within each toolcalldelta

gui/src/components/mainInput/Lump/LumpToolbar.tsx
```diff
@@ -96,7 +96,7 @@ export function LumpToolbar() {
         <StopButton
           className="text-gray-400"
           onClick={() => {
-            dispatch(cancelStream());
+            dispatch(cancelButton());
```
RomneyDa, line 99: semantic nitpick, thinking cancelStream makes more sense as the action performed, and since dispatched on keyboard events not just button click. also reduces file changes

gui/src/redux/thunks/cancelToolCallAndAddResult.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    // Step 3: Create a new message object for the tool result
+    const toolResultMessage = {
+      message: {
+        id: uuidv4(),
+        role: "tool",
+        content: "Tool use was cancelled.",
+        toolCallId: toolCallId
+      } as ToolResultChatMessageWithId,
```
RomneyDa, line 44: could we avoid a new one-off message type and use the existing `ChatHistoryItemWithMessageId` type?

gui/src/redux/thunks/cancelButton.ts
```diff
@@ -0,0 +1,18 @@
...
+import { streamAssistantMessage } from "./streamAssistantMessage";
+import { clearLastEmptyResponse } from "./clearLastEmptyResponse";
+
+export const cancelButton = createAsyncThunk<void, undefined, ThunkApiType>(
+  "chat/cancelButton",
+  async (messages, { dispatch, extra, getState }) => {
+    await dispatch(setInactive());
+    await dispatch(abortStream());
+    await dispatch(clearLastEmptyResponse());
+    await dispatch(streamAssistantMessage({ content: "The request has been cancelled." }));
```
RomneyDa, line 16: In chat mode this should just terminate, shouldn't stream an assistant response

gui/src/redux/slices/sessionSlice.ts
```diff
@@ -651,8 +644,36 @@ export const sessionSlice = createSlice({
...
+        if (toolResultMessage && toolResultMessage.message && toolResultMessage.message.id) {
+          state.history = [...state.history, toolResultMessage];
+        } else {
+          console.error("Cannot push tool result message to history: invalid message structure", toolResultMessage);
+        }
+      }
+    });
+    
+    // Add handlers for clearLastEmptyResponseThunk
+    builder.addCase(clearLastEmptyResponse.fulfilled, (state, action) => {
```
RomneyDa, line 672: If no action can remove

gui/src/redux/slices/sessionSlice.ts
```diff
@@ -651,8 +644,36 @@ export const sessionSlice = createSlice({
     },
   },
   extraReducers: (builder) => {
+    // Keep existing cases
     addPassthroughCases(builder, [streamResponseThunk]);
-  },
+    
+    // Add handlers for cancelToolCallAndAddResult
+    builder.addCase(cancelToolCallAndAddResult.fulfilled, (state, action) => {
```
RomneyDa, line 651: Could this be moved to the thunk so that it's all in one place? I'm not super familiar with the builder paradigm, could you explain the advantage of it vs dispatching a new message/tool call status from the thunk? Stylistic opinion but feels a bit disjointed. Testability related?

RomneyDa (changes requested):
@chezsmithy really appreciate the thorough testing here.
I think despite the solid PR description, I may have lost the core idea being implemented here, not easily grasping the exact scope/consequence of these changes.

Could you write up a quick re-outline of the exact approach/changes again?

RomneyDa:
@chezsmithy thanks for this in depth feedback, I think with all these points we're close to understanding the full solution. I'll be bumping tool architecture fixes this week, will make sure to rope you in if you're available, we will likely be able to reuse most of this

PR #5670, opened 2025-05-14
feat: add current file context automatically to chat

### LINKED ISSUE #5457: Add ability to add local/IDE context automatically/easily
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Not able to add initial IDE context like the currently open file or selected text. (Other than copy/pasting into the prompt.)
`@codebase` has to be added manually.

## Behavior of other products

| | Selection | Current file | Codebase |
|---------|-----------|--------------|----------|
| Augment | ✅ | ✅ | ✅ |
| Copilot | ✅ | ✅ | 👨‍🔧 |
| Cursor | 💥 | 💥 | 👨‍🔧 |

👨‍🔧 = Manual

Cursor used to be 💥/✅/✅. Recently, it happened a few times to me that AI didn't have visibility on the full file, even small ones. And they used to have a dedicated "codebase" button—while not automatic, the end result was the same: No additional typing/clicks were required.
Selection handling in Cursor was always weird because contrary to Augment and Copilot, they make it depend on interaction with Cursor features (like opening the chat panel while something is selected) rather than the user simply selecting/deselecting something.

Augment can afford automated codebase search because contrary to others, it's really fast.

Generally speaking, all products indicate that something will be added automatically and the user can remove any piece manually by clicking on an X next to the indicator.

## My opinion

Generally aim for Augment/Copilot behavior, since that's the most intuitive. (But as the title for this issue indicates, people's opinions on automatic or not may differ. Making it at least possible to somehow add these contexts resolves this feature request.)

Due to how Continue works (people using pay as you go), it's probably important to make it easy to turn this default behavior off in the settings panel. (I think having selection and current file on but codebase off would be the most appropriate default. Both due to cost but also since Continue's codebase search takes a while unless a fast model is used. But I think it would be great to have the option to turn it on as a default context.)

Since Continue is more configurable, I think there's a great opportunity here to do even better than others. Namely by not just adding these default context bits but making it possible to add arbitrary contexts to the initial context by default—HTTP providers and such.

## Not considered

I don't know how various tools behave if context changes when chat is already ongoing, since I personally don't use that. 1 problem per chat, so I only care about initial context.
Others may of course use it differently. But if IDE context is potentially added with every prompt, not just the first, it's of course important to avoid redundancy.

uinstinct:
## Description

This PR adds the ability to have the current file context added to a new chat. 

I really like this feature from when I used Augment and wanted it in Continue.

In this PR:
- adds a new vscode setting `addFileContext` for the extension
- the file context is added a mention so that it is removable
- the file context gets added when a new chat is initialized

Closes #5457

## Checklist

## Screenshots

https://github.com/user-attachments/assets/ad660907-d255-40e7-9024-742b4daf5085

## Tests

---

## Summary by mrge
Added a setting to automatically include the current file context as a removable mention when starting a new chat.

- **New Features**
  - New `addFileContext` setting in VS Code extension.
  - Current file context is added to new chats if the setting is enabled.


core/index.d.ts
```diff
@@ -704,6 +704,7 @@ export interface IdeSettings {
   userToken: string;
   continueTestEnvironment: "none" | "production" | "staging" | "local";
   pauseCodebaseIndexOnStart: boolean;
+  addFileContext: boolean
```
Patrick-Erichsen: This could be something a bit more descriptive, eg `useCurrentFileAsContext`

gui/src/components/mainInput/TipTapEditor/useMainEditorWebviewListeners.ts
```diff
@@ -178,4 +181,21 @@ export function useMainEditorWebviewListeners({
     },
     [],
   );
+
+  const activeContextProviders = useAppSelector(
```
Patrick-Erichsen: We don't have linting for this but it's a bit easier to read the file if we keep all these const declarations towards the top of the function before the hooks, eg `useWebviewListener`s

gui/src/components/mainInput/TipTapEditor/useMainEditorWebviewListeners.ts
```diff
@@ -178,4 +181,21 @@ export function useMainEditorWebviewListeners({
...
+    (state) => state.config.config.contextProviders,
+  );
+  const mode = useAppSelector(state=>state.session.mode);
```
Patrick-Erichsen:
```suggestion
  const mode = useAppSelector(state => state.session.mode);
```

If you don't have auto format/linting on please do so

gui/src/components/mainInput/TipTapEditor/utils/insertContent.ts
Patrick-Erichsen, line 1: Filename should match the export in this case

Patrick-Erichsen (changes requested):
This looks great! We've gone back and forth on this feature in the past but I think it's worth experimenting with again.

That being said, I think this should be configured as a UI setting, not an IDE setting.

@RomneyDa recently added a new section for experimental settings in https://github.com/continuedev/continue/blob/main/gui/src/pages/config/UserSettingsForm.tsx, so let's move this in there and then it should be good to go 👍

My other comments were just a few small nitpicks.

gui/src/components/mainInput/TipTapEditor/utils/insertContent.ts
uinstinct: yes. implemented!

core/index.d.ts
uinstinct: makes sense. implemented!

gui/src/components/mainInput/TipTapEditor/useMainEditorWebviewListeners.ts
uinstinct: right. implemented!

gui/src/components/mainInput/TipTapEditor/useMainEditorWebviewListeners.ts
uinstinct: turned on!
i also [added "editor.formatOnSave"](https://github.com/continuedev/continue/pull/5670/commits/c04d3c0d923d27e451cd5fce82745a9bde116a54) in our `.vscode/settings.json`

Patrick-Erichsen: Nice catch, thanks for doing that!

Patrick-Erichsen (changes requested):
Thanks for making those updates! It looks like you moved the setting to the UI, but I was thinking specifically of placing it under the new ["Experimental Settings" section](https://github.com/continuedev/continue/blob/038a410954285f6dd10650d669fc9fc91a223b79/gui/src/pages/config/UserSettingsForm.tsx#L344-L378). I think you just need to merge main and move the toggle to that section and then we're good to go 👍

uinstinct:
Apologies. Missed that part!
Made the changes now!

Also changed the vertical spacing on the experimental section to match with the other toggles' vertical spacing
image

PR #5671, opened 2025-05-14
chore: :bug: add keep-alive header for streaming to stop premature close

sestinj:
## Description

I'm not 100% certain that this solves the premature close error, but without an easy way to reproduce yet it seems like a worthwhile attempt.

## Checklist

## Tests

I do not think that tests of 15+ second-long streams would be appropriate to run in CI
    

---

## Summary by mrge
Added a keep-alive header to streaming requests to prevent connections from closing too early.

### DIFF
--- a/core/llm/llms/Anthropic.ts
+++ b/core/llm/llms/Anthropic.ts
@@ -191,6 +191,7 @@ class Anthropic extends BaseLLM {
       headers: {
         "Content-Type": "application/json",
         Accept: "application/json",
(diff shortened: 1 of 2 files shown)


sestinj:
@RomneyDa just fixed the type error, could use another review

RomneyDa (approved):

RomneyDa:
Closing for now since this is an experiment that would only affect local openai/anthropic and we're not really sure keep alive is a header that would do the trick since we're having difficulty reproducing. Potentially resolved itself somehow

PR #5677, opened 2025-05-14
feat: display rules used

tomasz-stefaniak:
## Description

display rules used

## Checklist

## Screenshots

image

## Tests

### DIFF
--- a/gui/src/components/mainInput/belowMainInput/RulesPeek.tsx
+++ b/gui/src/components/mainInput/belowMainInput/RulesPeek.tsx
@@ -74,7 +74,7 @@ export function RulesPeekItem({ rule }: RulesPeekItemProps) {
 
           <div className="min-w-0 flex-1 overflow-hidden truncate whitespace-nowrap text-xs text-gray-500">
             {isGlobal
-              ? "Applies to all files"
+              ? ""
               : `Pattern: ${typeof rule.globs === "string" ? rule.globs : Array.isArray(rule.globs) ? rule.globs.join(", ") : ""}`}
           </div>
         </div>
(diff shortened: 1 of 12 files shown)


tomasz-stefaniak:

tomasz-stefaniak:
recheck

gui/src/components/mainInput/belowMainInput/RulesPeek.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const isGlobal = !rule.globs;
+
+  return (
+    <div
+      className="group mr-2 flex flex-col overflow-hidden rounded px-1.5 py-1 text-xs hover:bg-white/10"
+      data-testid="rules-peek-item"
+    >
+      <div className="flex w-full items-center">
+        {isGlobal ? (
+          <GlobeAltIcon className="mr-2 h-4 w-4 flex-shrink-0 text-gray-400" />
```
Patrick-Erichsen, line 65: This was a bit confusing to me when I only had global rules. Given we already have the "Global Rules" header, and the "Applies to all files" text, I lean towards just using the same PencilIcon that we have in the notch for all rules.

Patrick-Erichsen (changes requested):
- There is some jank when both rules + context items are used
  - The rule peek is rendered first, then the context item is rendered above it, which causes jank. We can fix just by always rendering rules first since context items has lag
- Gap between context item + rules peek is a bit too big imo
- Can we add an affordance to click on the rule somewhere and view the file? Like we do in the notch

image

- I think we should only give ~1/3vh to the rules peek, right now appears to be closer to 2/3
- In a world where there are tens of rules, each potentially quite long, I lean towards setting a `line-clamp-3` (or maybe a few more lines) for individual rules. Imo the name of the rule is the most important thing, if you want the full text click on it and go to the file
- Can we make the "Applies to all files" text color more subtle, like the "Source" text? Feels like it's drawing too much attention since it's quite close to the rule title
- It looks like newlines aren't being preserved
- "Global" is a bit vague to me. I think it's the right term but seeing it next to "Applies to all files" is confusing. Adding an info icon with hover text to explain what "Global" means would solve it imo

Patrick-Erichsen (dismissed):

Patrick-Erichsen (approved):

PR #5681, opened 2025-05-15
chore: specify edit capabilities in `createRuleBlock`

Patrick-Erichsen:
## Summary by mrge
Updated the description for the createRuleBlock tool to clarify that editing existing rules should be done with the edit tool instead.


RomneyDa (commented):
modify to tell it where to find the rule file in this case (or similar)

RomneyDa (approved):

PR #5694, opened 2025-05-15
Make Apply Streaming Cancelable

RomneyDa:
Made the "Apply" streaming process cancelable so users can stop an in-progress code apply operation in the UI or via the reject diffs OR cancel stream commands.

Approach:

- Because of our apply architecture I decided to make a simple abort manager id -> abort controller singleton that can be used by the streamDiffLines function, which is the bottom level of all apply/edit streams.
- Also makes the llm stream chat actually abort the controller rather than just stopping the generator when aborted message id found
- Then, changes fetch package to catch abort errors right next to the fetch call and return a `499` client cancelled request error
- This error is caught in the `streamResponse` function used by all streaming within Continue and just causes a premature return. This is for cases where fetch is aborted after streaming starts
- Then I added specific error handling for `499` in the ~20 abortable non-streaming places - mostly odd APIs and `complete` requests throughout where `response.json()` is used
- This requires publishing a new version of `fetch` (1.0.10), and `openai-adapters` (1.0.25) packages (DONE)

Also
- Fixes VertexAI streams not being abortable
- Shows `Applying...` and the cancel keyboard shortcuts if currently applying (same location as "Generating...)
- Removes duplicate streaming logic - uses @continuedev/fetch package for streaming utils

To test:
Update package.jsons to use relative paths for fetch package and rebuild fetch package
Use a slower apply model
Cancel streaming mid-apply


core/core.ts
```diff
@@ -62,61 +61,7 @@ import { LLMLogger } from "./llm/logger";
 import { llmStreamChat } from "./llm/streamChat";
 import type { FromCoreProtocol, ToCoreProtocol } from "./protocol";
 import type { IMessenger, Message } from "./protocol/messenger";
-
-// This function is used for jetbrains inline edit and apply
```
RomneyDa, line 66: Removed streamDiffLinesGenerator/moved logic inline because point of it was to account for aborted message ids which wasn't being used, made ~1/2 the lines duplicate, and is now defunct

core/edit/streamDiffLines.ts
```diff
@@ -73,12 +75,15 @@ export async function* streamDiffLines({
...
   suffix: string;
   llm: ILLM;
+  abortControllerId: string;
   input: string;
   language: string | undefined;
   onlyOneInsertion: boolean;
   overridePrompt: ChatMessage[] | undefined;
   rulesToInclude: RuleWithSource[] | undefined;
 }): AsyncGenerator<DiffLine> {
+  const abortManager = StreamAbortManager.getInstance();
```
RomneyDa, line 85: Took the singleton approach so that functional stream diff lines function (also used in vscode-only code) can have persisted abort controllers that can be cancelled with core messaging

core/llm/streamChat.ts
```diff
@@ -75,52 +75,44 @@ export async function* llmStreamChat(
...
+    if (!next.done) {
+      throw new Error("Will never happen");
     }
+
+    return next.value;
   } else {
     const gen = model.streamChat(
       messages,
-      new AbortController().signal,
+      abortController.signal,
```
RomneyDa, line 110: This is significant because before, the connection would be hanging/running in the background until complete even if we've "aborted" using message ids at that point

core/llm/llms/VertexAI.ts
```diff
@@ -323,14 +341,14 @@ class VertexAI extends BaseLLM {
       ? this.geminiInstance.removeSystemMessage(messages)
       : messages;
     if (this.vertexProvider === "gemini") {
-      yield* this.streamChatGemini(convertedMsgs, options);
+      yield* this.streamChatGemini(convertedMsgs, options, signal);
```
RomneyDa, line 344: VertexAI fixes - weren't abortable

sestinj (approved):
Looks good. Nice that we were able to get rid of some of the uglier abort logic. Good tests for streamSse. I tested locally for a bit, but this is a PR to keep our eyes on while in pre-release

Issue #5530, opened 2025-05-06
VSCode llamafile Support Is Broken


ipaddicting:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: v1.0.8
- IDE version: VSCode 1.99.3
- Model: llamafile v0.9.2
- config:
  
  "models": [
    {
      "title": "Qwen 2.5 Coder 32B",
      "model": "qwen2.5-coder:32b",
      "provider": "llamafile",
      "apiBase": "http://127.0.0.1:8001"
    }
  ],
```

### Description

The support of llamafile on VSCode is broken.

I rolled back the Continue to v1.0.1, and it just works fine, log output as following:
```
{"function":"log_server_request","level":"INFO","line":2842,"method":"POST","msg":"request","params":{},"path":"/completion","remote_addr":"127.0.0.1","remote_port":57485,"status":200,"tid":"5188741632","timestamp":1746515946}
```

### To reproduce

1. Open chat
2. Type `who are you` 
3. Press Enter

### Log output

```Shell
Error message from continue:

Unexpected token 'F', "File Not Found" is not valid JSON
```

And the error message from llamafile:

```
{"function":"log_server_request","level":"INFO","line":2842,"method":"POST","msg":"request","params":{},"path":"/completions","remote_addr":"127.0.0.1","remote_port":58247,"status":404,"tid":"5188744736","timestamp":1746516335}
```

Looks like the path of request changed, from `"path":"/completion"` to `path":"/completions"`.

ipaddicting:
The bug was introduced by this pull request [Update LlamaCpp.ts #5030](https://github.com/continuedev/continue/pull/5030/files), since LlamaCpp & Llamafile is sharing the same config.

ipaddicting:
Either separates llamafile from llama.cpp, or reverts the URL change.

RomneyDa:
@ipaddicting thanks for the quick turnaround on this

RomneyDa:
[...]

PR #5726, opened 2025-05-19
fix: reverted the URL of llama.cpp back to 'completion'.

ipaddicting:
## Description

Closes: #5530
Reverted the URL of llama.cpp back to 'completion' since it's not changed at all.

The latest documentation from llama.cpp:
https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md#post-completion-given-a-prompt-it-returns-the-predicted-completion

`/v1/completions` is for OAI-compatible clients only, but in LlamaCpp.ts there is no `/v1` presented, and I don't think it should.

## Tests

Local tests with the extension built.

### DIFF
--- a/core/llm/llms/LlamaCpp.ts
+++ b/core/llm/llms/LlamaCpp.ts
@@ -35,7 +35,7 @@ class LlamaCpp extends BaseLLM {
       ...this.requestOptions?.headers,
     };
 
-    const resp = await this.fetch(new URL("completions", this.apiBase), {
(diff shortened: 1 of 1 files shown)


ipaddicting:

tomasz-stefaniak (approved):
Thanks!

For some extra background, I think we may actually want to remove this provider in the future, given that llama.cpp is now OpenAI-compatible. That means the OpenAI provider can be used instead.

PR #5735, opened 2025-05-19
Tool UI and docs tweaks

RomneyDa:
- Re-add the args button for simple tool call UI, but smaller and cleaner with better colors
- Don't let long tool args go off the edge of the screen
- Add/improve docs explaining why Agent mode isn't supported unless the model is known to support tools or `capabilities` is set


RomneyDa:
closing in favor of https://github.com/continuedev/continue/pull/5887

PR #5740, opened 2025-05-20
fix: grab PATH for mcp connections

Patrick-Erichsen:


core/context/mcp/MCPConnection.ts
```diff
@@ -63,80 +61,6 @@ class MCPConnection {
...
   }
 
-  /**
-   * Resolves the command and arguments for the current platform
-   * On Windows, batch script commands need to be executed via cmd.exe
-   * @param originalCommand The original command
-   * @param originalArgs The original command arguments
-   * @returns An object with the resolved command and arguments
-   */
-  private resolveCommandForPlatform(
```
Patrick-Erichsen, line 73: This is just import ordering that was automatically applied, nothing has changed in `resolveCommandForPlatform`

core/context/mcp/MCPConnection.ts
```diff
@@ -302,6 +226,93 @@ class MCPConnection {
...
+            try {
+              const shellEnvPath = getEnvPathFromUserShell();
+              if (shellEnvPath && shellEnvPath !== process.env.PATH) {
+                env.PATH = shellEnvPath;
+              }
+            } catch (err) {
+              console.error("Error getting PATH:", err);
+            }
+          }
+        }
```
Patrick-Erichsen, line 279: This is the only updated logic in this file

core/context/mcp/MCPConnection.ts
```diff
@@ -302,6 +226,93 @@ class MCPConnection {
...
+        if (process.env.PATH !== undefined) {
+          // Set the initial PATH from process.env
+          env.PATH = process.env.PATH;
+
+          // For non-Windows platforms, try to get the PATH from user shell
+          if (process.platform !== "win32") {
+            try {
+              const shellEnvPath = getEnvPathFromUserShell();
+              if (shellEnvPath && shellEnvPath !== process.env.PATH) {
+                env.PATH = shellEnvPath;
```
RomneyDa, line 273: we talked about making sure that shellEnvPath would always contain all path variables that would be in process.env.path. Seems likely and either way def a 1% better.

RomneyDa (approved):
Injects system path for non-windows systems so that MCP can use e.g. `npm`

Could you verify this doesn't break VS Code before merging?

Patrick-Erichsen:
Good callout on VSC, verified and it is working 👍

PR #5741, opened 2025-05-20
Fix find widget alignment

RomneyDa:
Fixes find widget match alignment issues caused by tabs by giving the tabs a ref and then passing it to the find widget to account for its height when calculating positions.


gui/src/components/find/FindWidget.tsx
```diff
@@ -341,16 +367,16 @@ export const useFindWidget = (searchRef: RefObject<HTMLDivElement>) => {
...
 
   // Triggers that should cause results to temporarily disappear and then reload
   // Active = LLM is generating, etc.
   const active = useAppSelector((state) => state.session.isStreaming);
 
   useEffect(() => {
     if (active || isResizing) setMatches([]);
-    else refreshSearch("none");
-  }, [refreshSearch, active]);
+    else setTimeout(() => refreshSearch("none"), 300); // this gives plenty of time for resizing to finish
```
RomneyDa: After a delay for codeblocks etc to render, recalculate search

gui/src/components/find/FindWidget.tsx
```diff
@@ -232,8 +245,13 @@ export const useFindWidget = (searchRef: RefObject<HTMLDivElement>) => {
...
           startIndex = endIndex;
 
+          const headerHeight = header?.clientHeight ?? 0;
+
           const top =
-            rect.top + searchContainer.clientTop + searchContainer.scrollTop;
+            rect.top +
+            searchContainer.clientTop +
+            searchContainer.scrollTop -
+            headerHeight;
```
RomneyDa: this is the main fix - account for header height in result y offset

gui/src/components/find/FindWidget.tsx
```diff
@@ -159,20 +163,29 @@ export const useFindWidget = (searchRef: RefObject<HTMLDivElement>) => {
...
-      if (searchRef.current) resizeObserver.unobserve(searchRef.current);
+      if (searchRef.current) {
+        resizeObserver.unobserve(searchRef.current);
+      }
+      if (headerRef.current) {
+        resizeObserver.unobserve(headerRef.current);
+      }
     };
-  }, [searchRef.current]);
+  }, [searchRef, headerRef]);
```
RomneyDa: Also recalculate when the header/tabs are resized

gui/src/components/find/FindWidget.tsx
tomasz-stefaniak: This will occasionally cause flake. Why is the timeout needed?

gui/src/components/find/FindWidget.tsx
```diff
@@ -159,20 +163,29 @@ export const useFindWidget = (searchRef: RefObject<HTMLDivElement>) => {
...
 
   // Main function for finding matches and generating highlight overlays
   const refreshSearch = useCallback(
```
tomasz-stefaniak, line 149: `refreshSearch` is large and complicated as is and `useFindWidget` is 500 lines long. How about moving this function of `FindWidget` body? 

```suggestion
  const refreshSearch = useCallback(someNewFunction({
      headerRef,
      searchRef,
      debouncedInput,
      scrollToMatch,
      caseSensitive,
      useRegex,
  },
    [
      headerRef,
      searchRef,
      debouncedInput,
      scrollToMatch,
      caseSensitive,
      useRegex,
      ...dependencies,
    ],
  )
```

RomneyDa:
@tomasz-stefaniak the issue is with alignment of results, not the widget itself.
Will add a screenshot

tomasz-stefaniak (approved):
Thanks for splitting up the code into smaller chunks!

PR #5766, opened 2025-05-21
Full Theme Colors Framework

RomneyDa:
## Description
- Creates new less-IDE-specific theme paradigm with corresponding tailwind classes and no explicit colors
- Adds more sensible defaults. Default is nice dark theme with blue accents - this was already the case but needed some tweaks
- Updates Jetbrains color mapping. Jetbrains sends color values mapped to theme names rather than vscode variable names. IMPORTANT TO UNDERSTAND: Jetbrains will "spray and pray", setting all associated CSS variable names to the given or default color, while VSCode will use the css variable fallback chain. That is, both will use the fallback chain, but in jetbrains all the variables will have the same value and the actual theme fallbacks happen on the kotlin side
- Adds a theme testing page which shows each color in use (page only shows in development)
- Fixes bad looking tabs on jetbrains
- Does not effect VS Code Continue Console styling
   
   
  Theme Testing Page
   
image

image

image


core/protocol/ideWebview.ts
```diff
@@ -41,7 +41,7 @@ export type ToIdeFromWebviewProtocol = ToIdeFromWebviewOrCoreProtocol & {
       vscMediaUrl: string;
     },
   ];
-  "jetbrains/getColors": [undefined, Record<string, string>];
+  "jetbrains/getColors": [undefined, Record<string, string | null | undefined>];
```
RomneyDa, line 45: Allow Jetbrains to send null colors so that we can handle defaults on the GUI side

gui/src/components/TabBar/TabBar.tsx
RomneyDa, line 1: This PR fixes bad looking tabs on jetbrains
Tabs were only using VS code variables specific to tabs
For some themes, this change actually makes tabs looks a bit _worse_ on VS code since doesn't use the tab specific styling but even many vs code themes are missing the tab CSS vars and this is much easier to maintain and works on jetbrains

gui/src/index.css
```diff
@@ -25,7 +25,7 @@ body,
 #root {
   height: 100%;
```
RomneyDa: these conflicted with sensible defaults when multiple fallback vars

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/GetTheme.kt
```diff
@@ -38,64 +38,117 @@ class GetTheme {
...
-        fun toHex(color: Color): String {
-            return String.format("#%02x%02x%02x", color.red, color.green, color.blue)
+    fun toHex(color: Color?): String? {
+        if(color == null) {
+            return null
         }
+        return String.format("#%02x%02x%02x", color.red, color.green, color.blue)
+    }
+
+    fun namedColor(name: String): Color? {
```
RomneyDa, line 50: Importantly, jetbrains now sends colors associated with theme name, not strange vscode class names
Also can now send null/undefined so that defaults can be handled on GUI side and the theme dev page can detect missing colors

gui/src/pages/config/HelpCenterSection.tsx
```diff
@@ -83,6 +85,16 @@ export function HelpCenterSection() {
             ideMessenger.post("showTutorial", undefined);
           }}
         />
+        {process.env.NODE_ENV === "development" && (
+          <MoreHelpRow
```
Patrick-Erichsen, line 89: very nice 👌

RomneyDa:
@Patrick-Erichsen made the fixes with primary/secondary buttons and with improved normal -> description -> muted fade

Patrick-Erichsen (approved):

PR #5787, opened 2025-05-22
fix: Fix the misleading naming of return value Config in function loa…

janzwj:
Fix the misleading naming of return value Config in function loadConfig  within file ConfigHandle

## Description

Replace Config with ConfigResult

    

---

## Summary by cubic
Renamed the return value in loadConfig from Config to ConfigResult to better reflect its contents and avoid confusion.


Patrick-Erichsen:
Closing this as an unnecessary change.

PR #5801, opened 2025-05-22
feat: remove edit as a mode

Patrick-Erichsen:


core/index.d.ts
```diff
@@ -423,7 +423,7 @@ export interface PromptLog {
   completion: string;
 }
 
-export type MessageModes = "chat" | "edit" | "agent";
+export type MessageModes = "chat" | "agent";
```
Patrick-Erichsen, line 426: Edit is now a boolean property, not a mode

gui/src/components/AcceptRejectDiffButtons.tsx
```diff
@@ -36,54 +34,43 @@ export default function AcceptRejectAllButtons({
...
+        data-tooltip-id="accept-shortcut"
+        data-tooltip-content={`Accept All (${acceptShortcut})`}
       >
         <div className="flex flex-row items-center gap-1">
-          <CheckIcon className="h-4 w-4 text-green-600" />
+          <CheckIcon className="text-success h-4 w-4" />
           <span>Accept</span>
           <span className="xs:inline-block hidden">All</span>
         </div>
-        <span
```
Patrick-Erichsen, line 78: Removed this because of space constraints on the lump, and moved it into a tooltip

gui/src/components/Layout.tsx
```diff
@@ -148,7 +148,7 @@ const Layout = () => {
     "focusEdit",
     async () => {
       await ideMessenger.request("edit/addCurrentSelection", undefined);
-      await dispatch(enterEditMode({}));
+      await dispatch(enterEdit({ editorContent: mainEditor?.getJSON() }));
```
Patrick-Erichsen, line 151: pass the current chat/agent editor content for us to cache while in edit, and reset once we return

gui/src/components/mainInput/Lump/LumpToolbar.tsx
Patrick-Erichsen, line 1: This was moved into `LumpToolbar/` folder where we colocate all the other toolbars

gui/src/components/mainInput/util/index.ts
Patrick-Erichsen, line 1: simplified version that is just the text, previously it was a jsx component with fileicon

gui/src/redux/slices/editState.ts
```diff
@@ -1,30 +1,33 @@
...
 
-export interface EditModeState {
+export interface EditState {
   // Array because of previous multi-file edit functionality
   // Keeping array to not break persisted redux for now
   codeToEdit: SetCodeToEditPayload[];
   applyState: ApplyState;
   returnToMode: MessageModes;
   lastNonEditSessionWasEmpty: boolean;
+  previousModeEditorContent: JSONContent | undefined;
```
Patrick-Erichsen, line 13: this is how we cache/return the editor content between edits

Patrick-Erichsen (commented):
Lot of code changes here because of file renames/reorg, cleanup of linting errors (eg adding `void`s), etc

The meat of the logic is the following:
- Make edit a boolean instead of a mode
- Enhance the lump toolbar, primarily by moving apply states into it
- When a user enters edit, clear the editor, but save the previous content and restore it when the user exits Edit

gui/src/components/ModeSelect/ModeSelect.tsx
```diff
@@ -66,55 +37,18 @@ function ModeSelect() {
...
-          enterEditMode({
-            returnToMode: mode,
-          }),
-        );
-      } else {
-        dispatch(setMode(nextMode));
-      }
-    }
+  const cycleMode = useCallback(() => {
+    dispatch(setMode(mode === "chat" ? "agent" : "chat"));
```
RomneyDa, line 41: I think need to ignore this command if in edit mode?

gui/src/components/mainInput/Lump/LumpToolbar/LumpToolbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+      } else if ((jetbrains ? altKey : metaKey) && event.key === "Backspace") {
+        event.preventDefault();
+        event.stopPropagation();
+        void dispatch(cancelCurrentToolCall());
+      }
+    }
+  };
+
+  useEffect(() => {
+    document.addEventListener("keydown", handleKeyDown);
```
RomneyDa, line 59: assuming this wasn't changed on this PR but this might cause a bug bc the toolCallState = generated check should be in the use effect adding/removing the listener, not in the listener itself?

gui/src/components/mainInput/TipTapEditor/MainEditorProvider.tsx
```diff
@@ -110,8 +118,10 @@ export const MainEditorProvider: React.FC<{ children: React.ReactNode }> = ({
  */
 export const useMainEditor = (): MainEditorContextType => {
   const context = useContext(MainEditorContext);
+
   if (context === undefined) {
```
RomneyDa: New context will never be undefined, could remove error

RomneyDa (dismissed):
One more thing to think about, should we persist isInEdit to redux? Maybe not

sestinj (approved):

RomneyDa (dismissed):

sestinj (changes requested):
Kind of a small thing but I think for people with affected themes it will be pretty bothersome. Is there a foreground theme var that is intended to work with this background var?

Screenshot 2025-05-25 at 6 03 27 PM

Patrick-Erichsen:
image
image


`bg-table-oddRow` doesn't feel like the most semantic var but looks good on light/dark

PR #5802, opened 2025-05-22
Theme colors rule and contributing guide section

RomneyDa:
Description:
- A section in `Contributing.md` explaining theme colors
- A `gui`-scoped rule telling models how to use theme colors

Follow up to #5766 
    

---

## Summary by cubic
Added a new section to the contributing guide explaining how to use theme colors, and introduced a rule for using theme colors in GUI components.

### DIFF
--- a/.continue/rules/colors.yaml
+++ b/.continue/rules/colors.yaml
@@ -1,0 +1,27 @@
+name: GUI
+version: 0.0.1
+schema: v1
+rules:
+  - name: Extension Color Themes
+    rule: |
+      When adding colors to components, use tailwind color classes.
+      Do NOT use explicit colors like text-gray-400. Instead, use theme colors.
+
+      The common available theme colors are:
+      - For normal text: foreground, description, description-muted
+      - For other text, icons, etc: success, warning, error, accent, link
+      - For general components background, border, border-focus
+      - For specific components: 
+          - Button: primary, primary-foreground, primary-hover, secondary, secondary-foreground, secondary-hover
+          - Input: input, input-foreground, input-border, input-placeholder
+          - Badge: badge, badge-foreground
+          - List/Dropdown items: list-hover, list-active, list-active-foreground
+          - Code Editor: editor, editor-foreground
+
+      Any of these colors can be used in tailwind e.g. bg-primary, text-success, border-error, hover:bg-list-hover, etc.
+    globs: "gui/**/*.ts"
+# Excluding the following less-used colors:
+# Command (only used by tip-tap): command, command-foreground, command-border, command-border-focus
+# Find widget colors: find-match, find-match-selected
+# table-oddRow
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -192,6 +193,58 @@ To keep the Continue codebase clean and maintainable, we expect the following fr
 Continue uses [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) to format
 JavaScript/TypeScript. Please install the Prettier extension in VS Code and enable "Format on Save" in your settings.
 
+### Theme Colors
```
Patrick-Erichsen, line 196: The content here is solid but feels a bit in the weeds for a contributing doc. I'd say keep the paragraph here but everything from `#### How Extension colors work` and after can be removed. If we want to document this somewhere maybe just a huge comment in tailwind config is best.

.continue/rules/colors.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - For general components background, border, border-focus
+      - For specific components: 
+          - Button: primary, primary-foreground, primary-hover, secondary, secondary-foreground, secondary-hover
+          - Input: input, input-foreground, input-border, input-placeholder
+          - Badge: badge, badge-foreground
+          - List/Dropdown items: list-hover, list-active, list-active-foreground
+          - Code Editor: editor, editor-foreground
+
+      Any of these colors can be used in tailwind e.g. bg-primary, text-success, border-error, hover:bg-list-hover, etc.
+    globs: "gui/**/*.ts"
```
Patrick-Erichsen: I believe this will only apply to tsx

```suggestion
    globs: "gui/**/*.tsx"
```

.continue/rules/colors.yaml
```diff
@@ -0,0 +1,27 @@
+name: GUI
+version: 0.0.1
+schema: v1
+rules:
+  - name: Extension Color Themes
+    rule: |
+      When adding colors to components, use tailwind color classes.
+      Do NOT use explicit colors like text-gray-400. Instead, use theme colors.
+
+      The common available theme colors are:
```
Patrick-Erichsen, line 10: This is a good example of where we really need the ability to reference a file in a rule. This rule is out of date as soon as we modify tailwind config. At the same time, I think the current rule is a better solution than just instructing the model to read tailwind config every time it writes tailwind, that would be annoying.

Patrick-Erichsen (changes requested):
few nitpicks

.continue/rules/colors.yaml
RomneyDa: Yeah I didn't want to put the contents of tailwind config in there since this is only colors and is much more concise and targeted structure, and I wanted to exclude many of the colors. I think this is manageable but agree colors could go missing every once in a while.

The end solution might be to have some kind of Rule management meta system, where you have a rule set that contains e.g. "Update the colors rule based on tailwind.config.js" that runs on a CRON job or something

.continue/rules/colors.yaml
RomneyDa: Nice catch

CONTRIBUTING.md
RomneyDa: Updated

PR #5821, opened 2025-05-23
Feed edit tool output back to the model

RomneyDa:
When edit tool is complete, read the new file contents and feed it back to the model. For now, just read the entire file. In the future, could send back diffs, ranges, etc but my hunch is that would confuse more than help the model.
    

---

## Summary by cubic
The edit tool now reads the updated file contents after an edit and sends them back to the model for further processing.

- **New Features**
  - Reads the entire edited file and returns its contents as context to the model.
  - Adds a helper to format and package the file output for the model.

### DIFF
--- a/core/llm/utils/extractPathsFromCodeBlocks.ts
+++ b/core/llm/utils/extractPathsFromCodeBlocks.ts
@@ -2,8 +2,6 @@
  * Extracts file paths from markdown code blocks
  */
 export function extractPathsFromCodeBlocks(content: string): string[] {
-  console.log("CONTENT", content);
-
   const paths: string[] = [];
 
   // Match code block opening patterns:
--- a/gui/src/hooks/ParallelListeners.tsx
+++ b/gui/src/hooks/ParallelListeners.tsx
@@ -21,6 +21,7 @@ import {
   acceptToolCall,
   addContextItemsAtIndex,
   updateApplyState,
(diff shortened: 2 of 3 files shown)


sestinj:
@RomneyDa I feel like I'm just missing a bit of context on this, but doesn't the model already basically see the new file contents from the input to its tool call? Not full output, but it can probably piece this together

chezsmithy:
This seems like a reasonable solution but you are right it will be token expensive if you make a number of small edits to a larger file? If the previous version of the file is in the chat history I wonder if there is some way to compress the chat history somehow after the edit is complete. Maybe store it like this?

History:
  v1: full snapshot
  v2: diff from v1
  v3: diff from v2
  v4: full snapshot (latest)

sestinj:
I think cost is actually a major concern here (enough so to avoid merging), and the point about small edits in large files I expect will become increasingly true with the direction of models using find/replace for edits

chezsmithy:
Could we solve for the full reject by updating the tool call noting the edit was cancelled as an early change while we solve for the bigger state issue of partial rejections?

sestinj:
@RomneyDa Not sure where you're at with this, but I'm going to close for hygiene's sake. Can be re-opened

RomneyDa:
@chezsmithy retackling this today!

PR #5828, opened 2025-05-24
feat: add support for generating commit message

lkk214:
## Description

Generate commit messages based on selected changes in the workflow UI.

- Analyze selected changes in the workflow UI and generate relevant descriptive commit messages
- Support switching between different LLMs, using `config.modelsByRole.chat` as the model list source

## Screenshots

demo0524


sestinj:
@lkk214 I like this a lot! If we add this in IntelliJ, do you think we could add it in VS Code as well? It can be bothersome when there isn't feature parity. Otherwise, I like what you did with the UI and whenever you've got the code in a closer spot I'd be happy to review

nomagicln:
This is an incredible feature! I've been looking forward to it for quite some time. Will it be supported in VS Code?

lkk214:
Hi @sestinj, this PR is now ready for review.

sestinj (changes requested):
Everything here looks great. I don't have an requests as far as the quality of the code (I appreciate the nice little refactorings, e.g. with the icons), but I did find one small bug when testing locally.

Right now, if you are using a thinking model, the thinking will be included in the commit message. You can reproduce this by using the `reasoning` option: https://docs.continue.dev/reference#models.

There are two potential options here:
1) just override the completionOptions to disable thinking (this would be reasonable)
2) make a small utility method (so it can be reused [here](https://github.com/continuedev/continue/blob/eb089af4e10ec8d64d05701c2036039fa9111d60/core/util/chatDescriber.ts#L34-L43)) that strips thinking from the output

lkk214:
Yes, I do need to consider the thinking model... For handling this, compared to creating a utility method to strip out the thinking for reuse in different places, I prefer disabling thinking directly in the concrete implementation of BaseLLM. If we don't need it, perhaps we shouldn't let it be returned from BaseLLM in the first place—this approach seems more reasonable to me.

sestinj (approved):
I like this approach

sestinj (dismissed):

lkk214:
@sestinj Merge conflicts keep popping up—any ETA on merging this?

sestinj:
will merge today if can get the tests to pass

sestinj (approved):

sestinj:
it seems like there are legitimate failing tests on this and it's been a while that this PR has been open. While I think it is a great change, I'm quite worried given how long it has taken to merge that it's going to become a large maintenance burden. I'm going to close the PR and would love to start a fresh one perhaps if you think there is a better approach we can take, but since I can't merge with failing tests it doesn't feel like this one has been making enough progress. Open to other ideas as usual!

Issue #5545, opened 2025-05-06
config.yaml not finding schema


ghost:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11 pro (Windows_NT x64 10.0.26100)
- Continue version: 1.0.8
VSCODE Version: 1.99.3 (user setup)
Commit: 17baf841131aa23349f217ca7c570c76ee87b957
Date: 2025-04-15T23:18:46.076Z
Electron: 34.3.2
ElectronBuildId: 11161073
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
- config:
  
name: Local Assistant
version: 1.0.0
schema: v1
models: []
context:
  - provider: code
  - provider: docs
  - provider: diff
  - provider: terminal
  - provider: problems
  - provider: folder
  - provider: codebase
  
  OR link to assistant in Continue hub:
```

### Description

The config.yaml display the following error:
```
Unable to load schema from 'c:\Users\<user>\.vscode\extensions\continue.continue-1.0.7-win32-x64\config-yaml-schema.json': No content
```
Even if you delete the config.yaml and .json, and open vscode to regenerate the file, the error still there.

I think the issue is that it is looking for the `\continue.continue-1.0.7` path, but after updating continue extension, now everything is in the `\continue.continue-1.0.8` folder.

![Image](https://github.com/user-attachments/assets/f8ead0e0-4a7d-4335-bccc-a0ec0bfddd46)

### To reproduce

1. Delete all youre config files.
2. Close VSCode and reopen it.
3. Open the continue extension so the config.yaml is create from new.
4. Open the config.yaml

### Log output

```Shell

```

drcrallen:
I had the same issue.

This can be worked around in vscode by going to Ctrl-P > Preferences: Open User Settings (JSON)  and changing your `yaml.schemas` to point to the valid file.

The bug seems to be that this setting isn't changed as versions update.

ikario404:
[...]

PR #5851, opened 2025-05-26
Fix Config.yaml schema registration

RomneyDa:
Fixes https://github.com/continuedev/continue/issues/5545

Before, would only update once and extensionUri.fsPath changed, making schema out of date.

This re-registers with new extension versions and also handles case of YAML extension not being installed


sestinj (approved):
Looks great! Seems fair not to add any tests here, it's already going to catch anything and is a strict improvement

PR #5853, opened 2025-05-26
Nate/config-yaml-no-node-packages

sestinj:
## Description

We import the config-yaml package in the browser, so it's important to get rid of any nodejs-dependent libraries. I split anything requiring that into a new package (which should be used basically only in our extension right now).

TODO: upgrade and use this new package within the extension

## Checklist

## Tests

The tests were moved over to the new package. The extension itself will definitely need to be tested


sestinj:
patrick found a better way to solve this

PR #5855, opened 2025-05-27
Remove free trial provider

RomneyDa:
Removes instances of free trial provider and comments out some code dealing with free trial limits
This is because free trial is deprecated and limit is not tracked for typical free trial provider

Adds warning that free trial provider is no longer supported


core/config/load.ts
```diff
@@ -458,7 +442,10 @@ async function intermediateToFinalConfig({
         return embedConfig;
       }
       const { provider, ...options } = embedConfig;
-      if (provider === "transformers.js") {
+      if (provider === "transformers.js" || provider === "free-trial") {
```
sestinj, line 445: This could probably use a quick comment just to explicitly acknowledge that we are falling back to transformers.js in the case of the free-trial provider now

sestinj (approved):
so nice to have all this gone. i'd say the one test that would be nice would be testing that loading the config with a free-trial provider succeeds but returns the error. That said, with what we have right now that's going to be tough and I def don't think it makes sense to handle within the scope of this PR. just a note for later

Go ahead and merge, the other comment isn't worth the time

Issue #5736, opened 2025-05-19
Display MCP logs in the Continue Console


sestinj:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Currently it is a bit difficult to debug MCP servers that aren't working correctly.

This is very similar to the problem of debugging LLMs, which was solved with the [Continue Console](https://docs.continue.dev/troubleshooting#prompt-logs-continue-console).

### Solution

I think it would be a good idea to also log MCP requests and responses

vickywane:
@sestinj should there be some form of configuration for the logs? (e.g log levels )

k-stamp:
Hey, how is the curretn state of this Issue?

I have HUGE problems getting all the different kinds of MCP Server to run, because every MCP docu shows a JSON integration and the YAML of continue seems to behave different.

My MCP Server gets a green flag in continue but it fails to retrieve information and I don't have any chance to understand why because i cannot see the mcp logs.

Patrick-Erichsen:
Thanks for the bump here @k-stamp , we are working on some ways to make the JSON -> YAML conversion easier. In the meantime it might be easiest just to use the "Ask AI" feature on our docs and ask it to map the JSON configuration to the Continue YAML schema.

PR #5858, opened 2025-05-27
show tool outputs during tool call

uinstinct:
## Description

Show tool call outputs along with tool call arguments. Previously MCP arguments were only shown and not the MCP responses.

Also:
- Added a new icon for ToolCall.
- Removed the older `SimpleToolCallUI.tsx` and added its capabilities in `ToolCall.tsx`

References:
- https://discord.com/channels/1108621136150929458/1367324473715593227/1376590374973476994
- maybe close? https://github.com/continuedev/continue/issues/5736 since both outputs and errors are now shown in tool call

## Checklist

## Screenshots

https://github.com/user-attachments/assets/0f835cce-62a5-43b7-9e86-85a93f369ae9

## Tests

image


RomneyDa (changes requested):
@uinstinct this would be great. Can we add it as a Context Item peek instead of with the args? At some point we had it as a context item peek like any normal context items and then regressed when the simple tool call UI was added. I'd like to keep the args separate. Thanks for adding tests

uinstinct:
Sure. Have separated out the context items/output items into a different component. Currently the output items are shown when the show args button is clicked. 

<details>
<summary>updated designs</summary>

image


---




image
</details>

RomneyDa:
@uinstinct can it use the same styling as the normal context item peeks? Like a toggle div that pops them down, with no "output" text or anything. We actually had this for a few months but it was taken out not fully intentionally 

image

uinstinct:
@RomneyDa for this we will need to append the context items in the existing place (just like in codebase or current files works) right?
I tried to look for appending these context items in `callTool.ts` and other areas on core and gui. Currently was unable to locate them. Can you point me out where I should be look such that they appear in the chat's context items?

AfterStories:
I think it would be better if a window pop up when the output button was clicked, showing the tool's input and output contents directly inside.  This is a more achievable method, and the tool output may be more content. If it is displayed in the context item, the scroll bar may be too long to be viewed. It may be better to display it in a separate pop-up window.and as RomneyDa said, `output` button might be change name to something like `tool context`, which would be easier and more user friendly.

uinstinct:
> I think it would be better if a window pop up when the output button was clicked, showing the tool's input and output contents directly inside.

Yeah, you are right. I actually implemented it for the tool output by opening the output in a virtual file.
We can open the tool error as well in a virtual file. Will wait for opinions.

RomneyDa:
@uinstinct nice improvements, I think this is closer but still have some thoughts.
I think we don't need tool output to be associated with args, i.e. it shouldn't say "show/hide args" if it's also going to show/hide output. Args _could_ be nested within a toggleable that by default only shows context items, but probably not vice versa. ContextItem has a `hidden` property that can be used to filter context items. If any context items are not hidden, then it should just show them in a collapsible div (like toggle div). If non are not hidden, then it should not show anything and not be collapsible. That is
- The previous "Simple" tools can be clickable so that they show the tool output below where the args would pop down if they have non-hidden ContextItems
- Custom tool call divs like Edit or Terminal can have their context items incorporated case by case, e.g. terminal already has custom streaming output logic, but edit tool could have context items shown below (results of edit e.g. https://github.com/continuedev/continue/pull/5821)

Great tests btw

Somewhat rambling, let me know if you have any questions about this or ideas for other approaches

RomneyDa:
@AfterStories context items will currently be shown as virtual files, they will not expand into the chat. The nice thing about virtual files over popups is that they are a more IDE-native experience, can be edited and saved, etc. Do virtual files satisfy the long output issue?

AfterStories:
@uinstinct @RomneyDa 
looks good ,virtual files also helpful for user ,may be we can get more feedback after merge this feature

RomneyDa:
@AfterStories thanks for the bump

@uinstinct checking in, some thoughts/review
1. Virtual file is great for tool output
2. Agreed that SimpleToolCallUI and ToolCallDiv should be merged, but let's do it in a way that doesn't effect the simple tool calls as much - the UI is fundamentally different in that the simple tool call will show the chevron that can dropdown Tool Output, while the non-simple ones (edit, terminal, etc.) should not have a chevron, and always show output. Let's merge the components but keep that simple chevron functionality
3. Talking with @Patrick-Erichsen today we do think that putting the args in a modal (same args button) as in @AfterStories feedback makes sense, and the tool output could also be shown in that modal but at least would be shown in context item peeks in chat still

So in summary
- args just shows a modal with args
- output always shows below as context item peek, toggleable chevron look for "simple" tools, and shown in custom ways for "non-simple" tools, e.g. terminal streams in, edit always shows context item peek (nothing for now since no output).

RomneyDa:
@uinstinct closing this for now, I think we should reimplement ASAP, looks like all the merge conflicts might not be worth it. We're thinking about reinstating a simple context item peek style, which shows when the tool chevron is toggled.

PR #5866, opened 2025-05-27
Expand Edit Tool tests, fix auto-apply duplicate responses

RomneyDa:
Expands edit tool testing, adding full edit tool walkthrough integration tests for multiple scenarios

When auto apply edits was enabled, accept was triggered
- in edit tool implementation
- in parallel listeners listening for apply state closed

Updated to only be in parallel listeners.

This required moving around `setActive` and `setInactive` dispatches because with nested thunks the stream wrapper would complete and setInactive which would cause subsequent response after tool call to abort.

Also changes StyledMarkdownPreview to force streamId by toolCallId rather than generating a streamId and forcing streamId


sestinj (commented):
seems like this fix will work, but not sure how we pin it down from breaking again if there aren't tests. While testing auto-apply seems like overkill given it's an experiment, do we have enough vitest gui tests for the edit file tool that we can be sure this doesn't break something else? I think worth adding one, even if simple

sestinj (changes requested):
seems like this fix will work, but not sure how we pin it down from breaking again if there aren't tests. While testing auto-apply seems like overkill given it's an experiment, do we have enough vitest gui tests for the edit file tool that we can be sure this doesn't break something else? I think worth adding one, even if simple

gui/src/context/MockIdeMessenger.ts
```diff
@@ -109,16 +109,34 @@ export class MockIdeMessenger implements IIdeMessenger {
...
+  chatResponse: ChatMessage[][] = [
+    [
       {
         role: "assistant",
         content: "This is a test",
       },
-    ];
+    ],
+  ];
+  chatStreamDelay: number = 0;
```
RomneyDa: Expands the abilities of MockIdeMessenger to be able to set the contents returned (similar to MockLLM) and mock delays

gui/src/util/test/config.ts
```diff
@@ -0,0 +1,73 @@
...
+  store: EnhancedStore;
+  messenger: MockIdeMessenger;
+  newProfileId?: string;
+  newOrgId?: string;
+  newOrgs?: SerializedOrgWithProfiles[];
+
+  configUpdates?: Partial<BrowserSerializedContinueConfig>;
+}
+
+export function triggerConfigUpdate({
```
RomneyDa, line 18: Utils for triggering config updates in redux for testing, as well as selecting model, etc.

gui/src/util/clientTools/editImpl.ts
```diff
@@ -35,10 +35,6 @@ export const editToolImpl: ClientToolImpl = async (
     if (out.status === "error") {
       throw new Error(out.error);
     }
-    return {
```
RomneyDa, line 38: This is the main fix, which prevents there being two simultaneous calls to the model which can result in malformed json

gui/src/redux/thunks/streamThunkWrapper.tsx
```diff
@@ -22,7 +21,6 @@ export const streamThunkWrapper = createAsyncThunk<
         }),
       );
     }
```
RomneyDa, line 23: setActive and setInactive are moved to around the actual streams, rather than in the streamThunkWrapper, since the wrapper can wrap chains of thunks, which causes setInactive to be prematurely triggered

gui/vite.config.ts
```diff
@@ -32,5 +32,18 @@ export default defineConfig({
     globals: true,
     environment: "jsdom",
     setupFiles: "./src/util/test/setupTests.ts",
+    onConsoleLog(log, type) {
```
RomneyDa, line 35: ignore redundant/unecessary logs for our setup

gui/src/redux/store.ts
```diff
@@ -23,11 +23,6 @@ import sessionReducer from "./slices/sessionSlice";
 import tabsReducer from "./slices/tabsSlice";
 import uiReducer from "./slices/uiSlice";
 
-export interface ChatMessage {
```
RomneyDa, line 26: This was an unused duplicate type

gui/src/components/StyledMarkdownPreview/StepContainerPreToolbar/index.tsx
RomneyDa, line 1: The main change here is that rather than generating a streamId within a useEffect in the EditTool GUI, we, force the `styledMarkdownPreview` to be associated with a toolCallId, use that, which is cleaner and more reliable

gui/src/components/mainInput/TipTapEditor/utils/editorConfig.ts
RomneyDa, line 1: Allows selection of main editor in integration test utils

gui/src/context/MockIdeMessenger.ts
RomneyDa, line 1: Expands abilities of MockIdeMessenger, and streamlines data mocking for messaging

gui/src/hooks/ParallelListeners.tsx
```diff
@@ -259,30 +264,45 @@ function ParallelListeners() {
...
-              toolCallId: currentToolCallApplyState.toolCallId!,
-            }),
-          );
-          // dispatch(setToolCallOutput([]));
-          void dispatch(
-            streamResponseAfterToolCall({
-              toolCallId: currentToolCallApplyState.toolCallId!,
-            }),
-          );
+          if (state.status === "done" && autoAcceptEditToolDiffs) {
```
RomneyDa, line 270: auto accept edit tool implementation moved from edit tool to here

gui/src/pages/gui/chat-tests/Chat.test.tsx
RomneyDa, line 1: Chat test is unchanged other than using the new utils.

gui/src/pages/gui/chat-tests/ApplyScenarios.test.tsx
RomneyDa, line 1: Apply test is unchanged other than using the new utils

RomneyDa:
@sestinj tests have been added

PR #5867, opened 2025-05-27
feat: `requestRule` tool

Patrick-Erichsen:
## Description

Introduces a new `requestRule` tool.

## Testing

```md
---
alwaysApply: false
description: Include this rule when asking about typescript.
---

Always respond in Spanish when discussing or generating code for TypeScript files
```

- Create the above rule in `.continue/rules` 
- Enter agent mode
- Ask a question about Typescript without including any files
- Verify it requests the rule
- Send another message
- Verify it replies in Spanish


RomneyDa (changes requested):
1. Tool policy off by default for now till feels stable
2. Look into just using name to filter/find rule, ignore edge case of duplicate names (edge case of duplicate filepaths also exists). Really nice thing about this is way less tokens in the dynamic descriptions

gui/src/redux/slices/uiSlice.ts
```diff
@@ -56,7 +56,7 @@ export const uiSlice = createSlice({
       [BuiltInToolNames.ViewDiff]: "allowedWithoutPermission",
       [BuiltInToolNames.LSTool]: "allowedWithoutPermission",
       [BuiltInToolNames.CreateRuleBlock]: "allowedWithPermission",
-      [BuiltInToolNames.RequestRule]: "allowedWithoutPermission",
+      [BuiltInToolNames.RequestRule]: "allowedWithPermission",
```
RomneyDa: Means still sent to model but requires permission. I was leaning off/excluded but don't feel strongly. I feel like it should either be off or automatic since it's a read-only "under the hood" action

core/tools/implementations/requestRule.ts
```diff
@@ -2,7 +2,16 @@ import { ToolImpl } from ".";
 import { parseMarkdownRule } from "../../config/markdown/parseMarkdownRule";
 
 export const requestRuleImpl: ToolImpl = async (args, extras) => {
-  const fileContent = await extras.ide.readFile(args.filepath);
+  // Find the rule by name in the config
+  const rule = extras.config.rules.find((r) => r.name === args.name);
+
+  if (!rule || !rule.ruleFile) {
```
RomneyDa, line 8: We could push a non-fatal error to config errors during config loading if duplicate names are detected! Similar to how it does it with MCP atm

RomneyDa (dismissed):

Issue #5819, opened 2025-05-23
Autocomplete Freezing and Locking Up the IDE


Patrick-Erichsen:
Below is an AI generated summary of the following related issues:
- https://github.com/continuedev/continue/issues/5213
- https://github.com/continuedev/continue/issues/4845
- https://github.com/continuedev/continue/issues/4376
- https://github.com/continuedev/continue/issues/3709
- https://discord.com/channels/1108621136150929458/1367755318670528575
- https://github.com/continuedev/continue/issues/5774
- https://github.com/continuedev/continue/issues/4130

# Summary of JetBrains IDE Freezing
**Root Cause:**

The primary root cause appears to be related to the **autocomplete functionality** in the JetBrains plugin, leading to the IDE freezing or becoming unresponsive. This is often (but not always) accompanied by high CPU usage from the `continue-binary` process. A secondary, possibly related, issue involves errors during the **indexing/embedding process**, specifically "Table already exists" errors with LanceDB.

**Affected Versions & IDEs:**

*   **Continue Plugin Versions:** Issues have been reported across various versions, including `0.0.75`, `0.0.83`, `0.0.85`, `0.0.88`, `1.0.2`, `1.0.8`, `1.0.17`, and `1.0.18` (0.0.75 appears to be the last version that definitely does not experience the issue)
*   **JetBrains IDEs:** IntelliJ IDEA, WebStorm, CLion, Rider, RustRover.
*   **IDE Versions:** Primarily 2024.x versions (e.g., 2024.3, 2025.1 EAP).
*   **Operating Systems:** Windows (10, 11), macOS, Linux.

> Note: IDEA 2025.1 in particular has freezing issues across a number of plugins. [See this issue](https://github.com/continuedev/continue/issues/5665#issuecomment-2905112729) for details.

**Symptoms & User Reports:**

1.  **IDE Freezing/Hanging:**
    *   The most common complaint is the IDE becoming completely unresponsive, often when typing or when autocomplete is triggered.
    *   Users often need to kill the IDE process via Task Manager or similar.
    *   Freezing can occur within minutes of starting the IDE or after a period of normal use.
    *   Some users report freezing even with the tab autocomplete feature disabled in settings.

2.  **High CPU Usage:**
    *   The `continue-binary` process is frequently reported to consume high CPU (50-300%, sometimes pinning a core at 100%).
    *   However, some users experience freezes *without* high CPU usage from `continue-binary`.

3.  **Autocomplete Issues:**
[...]

PR #5879, opened 2025-05-28
Disable diff autocomplete snippets in jetbrains

RomneyDa:
Temporary fix for https://github.com/continuedev/continue/issues/5819
Long term fix: fix jetbrains git diff caching logic, move more to core, etc.


sestinj (approved):
Looks good, good precision change

Issue #5815, opened 2025-05-23
Newly tagToString truncate impl throw SQLITE_CONSTRAINT error


zephyr19:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: 1.1.39
```

### Description

 Error: SQLITE_CONSTRAINT: UNIQUE constraint failed: chunk_tags.tag, chunk_tags.chunkId

tagToString truncate impl makes chunk_tags.tag duplicated

### Log output

```Shell

```

Patrick-Erichsen:
@zephyr19 does clearing `~/.continue/index` resolve this for you? If not then we need to update `tagToString`.

zephyr19:
clearing `~/.continue/index` did resolve the issue for me.

However, I believe this might be a broader issue. Many of my colleagues have encountered the same error, so it seems like a common problem. Could you help investigate further to see if there’s a more permanent fix needed, perhaps related to `tagToString`?

PR #5880, opened 2025-05-28
fix: use dir hash in `tagToString`

Patrick-Erichsen:
## Description

Resolves https://github.com/continuedev/continue/issues/5815 by using a dir hash when computing tag names for indexing.


RomneyDa (approved):
Looks great! Handles long paths while ensuring unique ids

PR #5882, opened 2025-05-28
Fix git diff caching

RomneyDa:
- Disables diff snippets for autocomplete for all IDEs
- Fixes git diff caching logic for other things that use the git diff

## Background
https://github.com/continuedev/continue/pull/5879 temporarily removed git diff from autocomplete for jetbrains to address https://github.com/continuedev/continue/issues/5819 and many associated issues

The issue was mostly only happening in Jetbrains, presumably because of the difference in how the diff is retrieved (VS Code uses the Git extension, while Jetbrains runs a `git diff` command), however, it also happens on extremely large projects in VS Code because generating diffs is expensive

This is a followup PR that fixes the git diff cache logic which had a race condition that caused the crashing in the first place when it took > 100ms to get the diff. Diff caching logic is moved off the IDEs and into core (removes concept of `lastFileSaveTimestamp`), and the cache is a more standard cache with invalidation, which is invalidated on file changes OR after 60 seconds.

However, we came to the conclusion that we should temporarily disable for all IDEs for the following reasons:
-
- The larger the diff, the less useful for autocomplete, and there is not a no-brainer way to guess the size of the diff before retrieving (we could look at e.g. the number of files changed or something but that's too fancy)

So this PR also removes git diff snippets from autocomplete. Most code is left intact.

### DIFF
--- a/core/autocomplete/snippets/getAllSnippets.ts
+++ b/core/autocomplete/snippets/getAllSnippets.ts
@@ -131,7 +131,7 @@ export const getAllSnippets = async ({
     IDE_SNIPPETS_ENABLED
       ? racePromise(getIdeSnippets(helper, ide, getDefinitionsFromLsp))
       : [],
-    racePromise(getDiffSnippets(ide)),
+    [], // racePromise(getDiffSnippets(ide)) // temporarily disabled, see https://github.com/continuedev/continue/pull/5882,
     racePromise(getClipboardSnippets(ide)),
   ]);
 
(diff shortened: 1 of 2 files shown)


extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/GitService.kt
```diff
@@ -13,33 +13,12 @@ class GitService(
...
-     * Updates the timestamp when a file is saved
-     */
-    fun updateLastFileSaveTimestamp() {
-        lastFileSaveTimestamp = System.currentTimeMillis()
-    }
 
     /**
      * Returns the git diff for all workspace directories
      */
     suspend fun getDiff(includeUnstaged: Boolean): List<String> {
```
sestinj, line 21: We should try to get tests on this. Not going to block this PR since it was almost unchanged

PR #5894, opened 2025-05-29
refactor: :recycle: Refactor CodeBaseIndexer out of core

chezsmithy:
## Description

Refector CodeBaseIndexer out of core.ts in a way that it can be independently passed to tools which might need to perform reindexing.

## Checklist

## Screenshots

## Tests

I've manually tested the following in debug:

1. At startup Indexing starts and suceeds.
2. Force Reindex works.
3. Rebuild index works.
4. Pausing and resuming indexing work.
5. Files in the test project show up in the index confirmed by using @files to lookup test.js

I've added new tests for CodeBaseIndexer.ts to cover the changes.


chezsmithy:
@Patrick-Erichsen related to https://github.com/continuedev/continue/pull/5564 I've performed a refactor of CodeBaseIndxer.ts and core.ts to refactor the indexing functions out of core. I'm going to note a few questions in a self review for you.

extensions/vscode/src/extension/VsCodeExtension.ts
```diff
@@ -244,6 +244,7 @@ export class VsCodeExtension {
       this.consoleView,
       this.configHandler,
       this.verticalDiffManager,
+      // This is expected to exist, but it's no longer used? Can I remove it?
       this.core.continueServerClientPromise,
```
chezsmithy: @Patrick-Erichsen this is no longer used anywhere in core. I can't see anywhere else it is used as well. Note in CodeBaseIndexer I just created a self contained new instance of this class. I'm not sure if this was meant to be a singleton or it's used in some other way? If it's no longer used we can refactor it out of here, and remove the initalization from core.ts.

core/core.ts
```diff
@@ -170,38 +164,24 @@ export class Core {
     dataLogger.ideInfoPromise = ideInfoPromise;
     dataLogger.ideSettingsPromise = ideSettingsPromise;
 
-    // Codebase Indexer and ContinueServerClient depend on IdeSettings
-    let codebaseIndexerResolve: (_: any) => void | undefined;
-    this.codebaseIndexerPromise = new Promise(
-      async (resolve) => (codebaseIndexerResolve = resolve),
-    );
-
     let continueServerClientResolve: (_: any) => void | undefined;
```
chezsmithy: @Patrick-Erichsen note this is no longer used anywhere in core. I've moved initalization into CodeBaseIndexer rather than relying on this instance. Should we just remove it?

extensions/vscode/src/extension/VsCodeExtension.ts
Patrick-Erichsen: Nice catch @chezsmithy - I think this is safe to remove and would be appreciated if you could do so.

cc @sestinj for a sanity check however.

core/core.ts
Patrick-Erichsen: Yep, let's remove this one too, thanks for calling it out 👍

core/config/yaml/loadYaml.ts
Patrick-Erichsen, line 1: Are the changes here intentional? I don't see this SSE logic in `main` https://github.com/continuedev/continue/blob/main/core/config/yaml/loadYaml.ts#L452-L466

Patrick-Erichsen (dismissed):
This is a great cleanup, will make the `callTools` logic we originally discussed less dependent on `core` and this is just a solid cleanup in general. Nice we we can remove the awkward `codebaseIndexerResolve` things.

Going to wait to merge until I double check with @sestinj on that one comment but otherwise good to go 🚀

core/core.ts
chezsmithy: @sestinj cc on this question along with the other.

core/config/yaml/loadYaml.ts
chezsmithy: @Patrick-Erichsen I've reverted to main. I believe that crept in due to some type script issues as I was adding tests.

extensions/vscode/src/extension/VsCodeExtension.ts
Patrick-Erichsen: I just poked around, seems like we might use it in `extensions/vscode/src/commands.ts`?

```ts
    "continue.giveAutocompleteFeedback": async () => {
      const feedback = await vscode.window.showInputBox({
        ignoreFocusOut: true,
        prompt:
          "Please share what went wrong with the last completion. The details of the completion as well as this message will be sent to the Continue team in order to improve.",
      });
      if (feedback) {
        const client = await continueServerClientPromise;
        const completionsPath = getDevDataFilePath(
          "autocomplete",
          LOCAL_DEV_DATA_VERSION,
        );

        const lastLines = await readLastLines.read(completionsPath, 2);
        client.sendFeedback(feedback, lastLines);
      }
    },
```

sestinj: Ah we should just remove this

sestinj:
All good to go from me except I think we should just completely remove ` "continue.giveAutocompleteFeedback"`

sestinj (changes requested):
all looks great. as we're removing the continueServerClient, let's take out the `"continue.giveAutocompleteFeedback"` command in VS Code

chezsmithy:
Waiting on https://github.com/continuedev/continue/pull/5937 to merge to fix CD/CI issues.

chezsmithy:
@Patrick-Erichsen and @sestinj looks like we have 🟢 lights. Could you please review at your convienence? 🙇

Patrick-Erichsen (approved):

sestinj (approved):

PR #5907, opened 2025-05-29
Fix apply cancellation and make slash commands cancelable

RomneyDa:
#5694 was intended to make apply cancelable but only worked for jetbrains because of how it was set up.
This PR fixes for VS Code
It also passes abort controllers to slash commands to make them cancelable as well.

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -836,6 +836,7 @@ export interface ContinueSDK {
   config: ContinueConfig;
   fetch: FetchFunction;
   completionOptions?: LLMFullCompletionOptions;
+  abortController: AbortController;
 }
 
 export interface SlashCommand {
@@ -1141,7 +1142,7 @@ export interface StreamableHTTPOptions {
   requestOptions?: RequestOptions;
 }
 
-export type TransportOptions = 
+export type TransportOptions =
   | StdioOptions
   | WebSocketOptions
   | SSEOptions
--- a/core/commands/index.ts
+++ b/core/commands/index.ts
@@ -25,6 +25,7 @@ export function slashFromCustomCommand(
       config,
       selectedCode,
       fetch,
+      abortController,
     }) {
       // Render prompt template
       let renderedPrompt: string;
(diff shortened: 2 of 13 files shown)


chezsmithy:
@RomneyDa thank you for all your amazing work on solving for being able to cancel various pieces of the experience. I closed my cancel button fix. Noting that my only remaining issue appears to be cancelling "thinking/reasoning" and ending up with unmatched messages in the history.

core/edit/recursiveStream.ts
```diff
@@ -68,6 +68,7 @@ export async function* recursiveStream(
     });
 
     for await (const chunk of generator) {
+      console.log(chunk);
```
sestinj: make sure to remove this

core/index.d.ts
```diff
@@ -1262,7 +1263,7 @@ export interface HighlightedCodePayload {
 }
 
 export interface AcceptOrRejectDiffPayload {
-  filepath: string;
+  filepath?: string;
```
sestinj, line 1284: this is necessary because the GUI needs to be able to send a generic cancel?

gui/src/pages/gui/Chat.tsx
```diff
@@ -139,7 +139,7 @@ export function Chat() {
         !e.shiftKey
       ) {
         dispatch(cancelStream());
```
sestinj, line 134: I think this would be a quick and high-value thing to test. There's already a setup for the Chat.tsx component in vitest I think, and would just want to spy on ideMessenger to make sure that it calls post("rejectDiff", ...).

sestinj (changes requested):
I know most of this is just passing around abortController / signal, but at least some part of this should get locked down with a test. otherwise just feels like setting up for whack-a-mole

Proposed one for the keyboard shortcut in the GUI

core/index.d.ts
RomneyDa: Yes and the underlying functions that the Accept/Reject payload is passed to don't actually required the filepath

gui/src/pages/gui/Chat.tsx
RomneyDa: Added!

PR #5921, opened 2025-05-30
Feature/nextEdit

jpoly1219:
## Description

Merge #5919 and #5920.

### 5919:
Initial commits. Next edit module is created and is called alongside autocomplete. This is extremely rudimentary with no custom logic - the code is shamelessly copied over from the core/autocomplete directory, and all imports are pointing back to that directory. In the future, when we start adding more custom algorithms and types, we will slowly cut ties to the core/autocomplete directory and create our own internal definitions.

For now, this works only inside vscode. Jetbrains should require a custom message to some endpoint like "nextEdit/predict".

### 5920:
Turn off next edit for other users until it is ready.

This only targets vscode -- we need another check for jetbrains. I don't see any message endpoints defined for next edit for the jetbrains extension, so we should be good for now.

## Checklist

## Screenshots

## Tests

### DIFF
--- a/core/nextEdit/IgnoreNextEdit.ts
+++ b/core/nextEdit/IgnoreNextEdit.ts
@@ -1,1 +1,0 @@
-export const IGNORE_NEXT_EDIT = true;
--- a/core/nextEdit/IgnoreNextEdit.ts
+++ b/core/nextEdit/IgnoreNextEdit.ts
@@ -1,0 +1,1 @@
+export const IGNORE_NEXT_EDIT = true;
--- a/core/nextEdit/NextEditProvider.ts
+++ b/core/nextEdit/NextEditProvider.ts
@@ -1,0 +1,282 @@
+import { ConfigHandler } from "../config/ConfigHandler.js";
+import { TRIAL_FIM_MODEL } from "../config/onboarding.js";
+import { IDE, ILLM } from "../index.js";
+import OpenAI from "../llm/llms/OpenAI.js";
+import { DEFAULT_AUTOCOMPLETE_OPTS } from "../util/parameters.js";
+
+import { shouldCompleteMultiline } from "../autocomplete/classification/shouldCompleteMultiline.js";
+import { ContextRetrievalService } from "../autocomplete/context/ContextRetrievalService.js";
+
+import { BracketMatchingService } from "../autocomplete/filtering/BracketMatchingService.js";
+import { CompletionStreamer } from "../autocomplete/generation/CompletionStreamer.js";
+import { postprocessCompletion } from "../autocomplete/postprocessing/index.js";
+import { shouldPrefilter } from "../autocomplete/prefiltering/index.js";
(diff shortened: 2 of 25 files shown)


core/nextEdit/IgnoreNextEdit.ts
tomasz-stefaniak, line 1: How about something like `constants.ts` for consistency with the rest of the codebase?

core/nextEdit/NextEditProvider.ts
```diff
@@ -0,0 +1,282 @@
...
+import { postprocessCompletion } from "../autocomplete/postprocessing/index.js";
+import { shouldPrefilter } from "../autocomplete/prefiltering/index.js";
+import { getAllSnippets } from "../autocomplete/snippets/index.js";
+import { renderPrompt } from "../autocomplete/templating/index.js";
+import { GetLspDefinitionsFunction } from "../autocomplete/types.js";
+import { AutocompleteDebouncer } from "../autocomplete/util/AutocompleteDebouncer.js";
+import { AutocompleteLoggingService } from "../autocomplete/util/AutocompleteLoggingService.js";
+import AutocompleteLruCache from "../autocomplete/util/AutocompleteLruCache.js";
+import { HelperVars } from "../autocomplete/util/HelperVars.js";
+import { AutocompleteInput, AutocompleteOutcome } from "../autocomplete/util/types.js";
```
tomasz-stefaniak, line 21: This is ok for now but long-term it's better to avoid importing implementation details from sibling services. This makes it easy to create circular dependencies, which will eventually lead to bugs and are often difficult to resolve.

For example, someone else could, in the future, import implementation details from `nextEdit` into autocomplete and it generate a circular dependency that way.

It's usually better to move the reusable elements to a third directory that both siblings import from.

extensions/vscode/src/autocomplete/completionProvider.ts
```diff
@@ -83,12 +86,21 @@ export class ContinueCompletionProvider
       this.onError.bind(this),
       getDefinitionsFromLsp,
     );
+    // NOTE: Only turn it on locally when testing (for review purposes).
+    if (!IGNORE_NEXT_EDIT) {
```
tomasz-stefaniak: Naming-wise, it's better to not have "negative" boolean flags. For example, instead of `IGNORE_NEXT_EDIT` you could use `IS_NEXT_EDIT_ACTIVE`. This helps avoid confusion and makes control flows simpler (no negation needed).

```suggestion
    if (IS_NEXT_EDIT_ACTIVE) {
```

core/nextEdit/NextEditProvider.ts
jpoly1219: Yeah, I don't like to make deps this way either :( I didn't feel like making exact copies of these files inside the `core/nextEdit` directory, although that would've been "safer" because the dependencies are contained within `core/nextEdit`. My plan here is to phase these imports out gradually as we add more features to next edit.

PR #5945, opened 2025-06-02
feat: create markdown rules in notch

Patrick-Erichsen:
## Description

When clicking "Add Rule" from the notch, we now create a markdown rule instead of a YAML rule.


core/config/markdown/createMarkdownRule.ts
Patrick-Erichsen, line 1: This consolidates some of the rule creation logic we had in multiple places throughout the codebase

core/tools/definitions/createRuleBlock.ts
```diff
@@ -37,11 +37,6 @@ export const createRuleBlock: Tool = {
           description:
             "Optional file patterns to which this rule applies (e.g. ['**/*.{ts,tsx}'] or ['src/**/*.ts', 'tests/**/*.ts'])",
         },
-        alwaysApply: {
```
Patrick-Erichsen, line 40: Decided to remove this since it's still relatively experimental.

chezsmithy:
Could we offer the choice? We have a bunch of YAML rules now and this might be somewhat confusing.

Patrick-Erichsen:
@chezsmithy thanks for the callout here! What are your general thoughts/feelings on markdown vs YAML syntax for rule creation?

We're thinking next week of rolling out a tool to easily convert all of your existing YAML rules to markdown. If we shipped that, 1) would you use it, and 2) would it then be non-disruptive to merge this PR?

If so I think we can leave this PR open until we ship the YAML -> markdown rule conversion.

chezsmithy:
@Patrick-Erichsen thanks for the heads up. I'll check in with my team. Are you considering deprecating YAML for rules all together, or just creation initially? Would the tool be command line, or something in the UI? 

Maybe we could have a dialog on discord about the plan and save a bunch of back and forth here?

Patrick-Erichsen:
Good call, moved the discussion here: https://discord.com/channels/1108621136150929458/1131313996750917835/1379542031243808869

core/config/workspace/workspaceBlocks.ts
```diff
@@ -2,10 +2,51 @@ import { BlockType, ConfigYaml } from "@continuedev/config-yaml";
...
+    case "models":
+      return "model";
+    case "rules":
+      return "rule";
+    case "docs":
+      return "doc";
+    case "prompts":
+      return "prompt";
+    case "mcpServers":
+      return "mcp-server";
```
RomneyDa: maybe just mcpserver

core/tools/definitions/createRuleBlock.ts
RomneyDa: I think it's also confusing for the model and maybe we should use a different property name for the model and then map it to the underlying property

core/config/workspace/workspaceBlocks.ts
```diff
@@ -2,10 +2,51 @@ import { BlockType, ConfigYaml } from "@continuedev/config-yaml";
 import * as YAML from "yaml";
 import { IDE } from "../..";
 import { joinPathsToUri } from "../../util/uri";
+import { RULE_FILE_EXTENSION, createRuleMarkdown } from "../markdown";
+
+export function blockTypeToSingular(blockType: BlockType): string {
+  switch (blockType) {
```
RomneyDa: feels like an object with block info would be cleaner here, like 
```
context:
  singular: Context
  plural: Context
  title: Context
mcpServers:
  singular: MCP Server
  plural: MCP Servers
  title: MCP Servers
 ```
  
  Or similar

RomneyDa (approved):

PR #5952, opened 2025-06-03
Quick pick to enter key in VS Code

sestinj:
## Description

A nice UI for entering the key, rather than needing to run a command manually

## Checklist

## Tests

Tests were updated for mdm.test.ts


extensions/vscode/src/commands.ts
```diff
@@ -794,19 +794,55 @@ const getCommandsMap: (
...
+          );
+          await new Promise((resolve) => setTimeout(resolve, 1000));
+          await vscode.commands.executeCommand("workbench.action.reloadWindow");
+        } else {
+          void vscode.window.showErrorMessage(
+            "Invalid license key. Please check your license key and try again.",
+          );
+        }
+      } catch (error) {
+        void vscode.window.showErrorMessage(
```
Patrick-Erichsen, line 841: I'm not sure this will ever get triggered because we wrap all of the write ops for `mdm/setLicenseKey` in try/catch with a `return false` clause. So any actual errors writing the keys will just result in the `Invalid license key.` 

It might make sense to just move the toast notification logic into core. That would also make it easier to support this in JetBrains if we ever wanted to do something similar.

Issue #4020, opened 2025-02-07
Per-Model Configuration for `TabAutocompleteOptions`


Hinkolas:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

I have two main models configured in Continue that I regularly switch between:  
1. A **larger model** running on a remote Mac Studio with an M2 Ultra, which handles large context windows and fast autocompletions.  
2. A **low-parameter model** running locally on my MacBook Pro with an M2 Pro, which I use when I’m on the go and don’t have a stable internet connection (a common issue in Germany).  

The problem is that my MacBook Pro cannot handle the same context window size (`maxPromptTokens`) as the Mac Studio. As a result, I need to manually adjust the `maxPromptTokens` setting in `TabAutocompleteOptions` every time I switch between the models.

### Solution

It would be great to configure `TabAutocompleteOptions` (e.g., `maxPromptTokens`) on a **per-model basis**. This would allow users to tailor settings like `maxPromptTokens` to the capabilities of the hardware or the size of the model being used. For example:  
- The Mac Studio could use a larger `maxPromptTokens` value for more comprehensive autocompletions.  
- The MacBook Pro could use a smaller `maxPromptTokens` value to ensure good performance and lower battery usage.  

This feature would eliminate the need for manual adjustments when switching models and improve the overall user experience.

RomneyDa:
This is completed with config.yaml syntax, using `autocompleteOptions`

PR #5970, opened 2025-06-03
Add ability to configure tabAutocompleteOptions on a per-model basis in config.yaml

jpoly1219:
## Description

Implement #4020.

Updated the config.yaml reading process so that we can now read in `autocompleteOptions` on a per-model basis. The `config.yaml` using this would look like so:
```yaml
name: Jacob
version: 1.0.8
schema: v1
models:
#...
  - uses: # your model here...
    with:
      # your key here...
    override:
      autocompleteOptions:
        debounceDelay: 5000
        maxPromptTokens: 1500
        disableInFiles:
          - "*.md"
      # other overrides here...
#...
```

A good next step would be to:
- Update our docs to reflect this change.
- Slowly transition the internal naming scheme from `TabAutocomplete` to `Autocomplete`.

## Checklist

## Screenshots

## Tests


core/autocomplete/CompletionProvider.ts
```diff
@@ -136,15 +137,19 @@ export class CompletionProvider {
         token = controller.signal;
       }
       const startTime = Date.now();
-      const options = await this._getAutocompleteOptions();
 
-      // Debounce
-      if (await this.debouncer.delayAndShouldDebounce(options.debounceDelay)) {
+      // NOTE: There shouldn't be too big of a performance penalty here.
```
sestinj: I think you can remove this comment. I'm confident that it's safe enough and this comment might make people worry when they don't need to

packages/config-yaml/src/schemas/models.ts
```diff
@@ -106,6 +106,35 @@ const templateSchema = z.enum([
...
+  experimental_includeClipboard: z.union([z.boolean(), z.number()]).optional(), // not in docs
+  experimental_includeRecentlyVisitedRanges: z
+    .union([z.boolean(), z.number()])
+    .optional(), // not in docs
+  experimental_includeRecentlyEditedRanges: z
+    .union([z.boolean(), z.number()])
+    .optional(), // not in docs
+  experimental_includeDiff: z.union([z.boolean(), z.number()]).optional(), // not in docs
+});
+// .partial();
```
sestinj: remove the .partial() comment

packages/config-yaml/src/schemas/models.ts
```diff
@@ -106,6 +106,35 @@ const templateSchema = z.enum([
...
+  useImports: z.boolean().optional(), // not in docs
+  showWhateverWeHaveAtXMs: z.number().optional(), // not in docs
+  experimental_includeClipboard: z.union([z.boolean(), z.number()]).optional(), // not in docs
+  experimental_includeRecentlyVisitedRanges: z
+    .union([z.boolean(), z.number()])
+    .optional(), // not in docs
+  experimental_includeRecentlyEditedRanges: z
+    .union([z.boolean(), z.number()])
+    .optional(), // not in docs
+  experimental_includeDiff: z.union([z.boolean(), z.number()]).optional(), // not in docs
```
sestinj: I know I didn't say this when we originally talked about it, but let's remove any of the properties here that aren't in the docs. (Just on this Zod object, keep the TypeScript TabAutocompleteOptions the same in index.d.ts)

sestinj (changes requested):
few quick comments, but this looks basically perfect!

Issue #3244, opened 2024-12-07
Use .continueignore to exclude from auto complete


salah-8:
### Validations

- [X] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

To exclude files from auto complete I have to add `disableInFiles`

### Solution

It would be better I exclude/hide files from auto complete using the `.continueignore` file so there is one file to exclude from auto complete and indexing

tomasz-stefaniak:
@salah-8 I had the same issue a few days ago but on second though, I believe these should be separate. The files you want to ignore may not always overlap. For example, you may want to disable completions in all Markdown files but still want to have them indexed.

salah-8:
@tomasz-stefaniak Hey, I understand your point. But it would be better to at least make it a file like .continueignore instead of adding a long list in the config.json file

> Out of curiosity, what's your use case and what types of files would you like to ignore?

most of the files I want to ignore/hide from AI in general (indexing or auto completion) is the files that have secrets, keys, or anything sensitive stuff from my company's codebase in general

tomasz-stefaniak:
@salah-8 we're currently working on a new version on config.json. I'm tagging @sestinj so that he's aware of this feedback.

ingo-eichhorst-ionos:
IMHO `.continueignore` indicates that the globs get ignored for continue completely. It's confusing to have multiple places for configuration.
We may combine a block-list and allow-list for indexing and autocomplete. That would overwrite .continueignore if explicitly defined.
For me the purpose of `.continueignore` is to not send secrets via the wire.

Fmstrat:
My 2 cents, `disableInFiles` makes 100% sense for ignoring in the tab autocomplete, for all the reasons mentioned above, and `.continueignore` makes sense for continue to fully ignore a file (auto-complete and indexing).

I also think there should be an option in `config.json` called `ignoreFiles` that is a glob allowing you to do the same thing as `.continueignore`. This way you could add something like `secrets/*` to it, so you don't need a custom `.continueignore` for every repo.

SDwarfs:
I'd not call the file ".continueignore" but more general ".aiexclude".
This way people do not have to create several exclude files for all the AI assistants they are using.

[...]

PR #5973, opened 2025-06-04
Implement #3244

jpoly1219:
## Description

`.continueignore` is respected for autocomplete.

Currently this only supports top-level `.continueignore` files inside your `.continue` directory. It also runs on every autocomplete request, but upon profiling, this takes a negligible ~2ms.

## Checklist

## Screenshots

## Tests

### DIFF
--- a/core/autocomplete/prefiltering/index.ts
+++ b/core/autocomplete/prefiltering/index.ts
@@ -1,6 +1,7 @@
 import ignore from "ignore";
 
 import { IDE } from "../..";
+import { getGlobalContinueIgArray } from "../../indexing/ignore";
 import { getConfigJsonPath } from "../../util/paths";
 import { findUriInDirs } from "../../util/uri";
 import { HelperVars } from "../util/HelperVars";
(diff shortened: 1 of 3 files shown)


core/autocomplete/prefiltering/index.ts
```diff
@@ -50,7 +51,11 @@ export async function shouldPrefilter(
   }
 
   // Check whether autocomplete is disabled for this file
-  const disableInFiles = [...(helper.options.disableInFiles ?? []), "*.prompt"];
+  const disableInFiles = [
+    ...(helper.options.disableInFiles ?? []),
+    "*.prompt",
+    ...getGlobalContinueIgArray(),
```
sestinj, line 60: Kind of a miscommunication because I forgot that we both have the global and the workspace .continueignore. Global is ~/.continue/.continueignore while we more often see users use a .continueignore at the root of their workspace.

I think it wouldn't be too tough to add that. Something like this:

```ts
export const getWorkspaceContinueIgArray = async (ide:IDE) => {
  const contents = await ide.readFile(...);
  return gitIgArrayFromFile(contents);
}
```

sestinj (changes requested):
i think this should be a pretty quick update

PR #5990, opened 2025-06-04
Bubble up config and block errors to the extension GUI

jpoly1219:
## Description

Previously, errors when loading config and blocks were not being rendered. This made errors in configuration and rules difficult to catch.

Here's a motivating example:
```yaml
name: BAD RULE
version: 0.0.1
schema: v1
rules:
  - "this is a rule."
  - "this is supposed to be a rule body."
  - "unfortunately, each hyphen corresponds to a new rule."
  - "so this is actually the fourth rule."
```
This is syntactically valid YAML, but our rule parser will error out because it expects one rule per YAML file. These errors in semantics were difficult to catch without an error message. Now users will be able to see errors from the extension GUI.

This PR also fixes a subtle bug where exclamation triangle icon in the warning box will shrink depending on the size of the displayed error message.

## Checklist

## Screenshots

image

## Tests


packages/config-yaml/src/load/unroll.ts
```diff
@@ -204,14 +205,49 @@ export type UnrollAssistantOptions =
...
+  registry: Registry,
+  options: UnrollAssistantOptions,
+  needErrorPropagation?: false,
+): Promise<AssistantUnrolled>;
+
+export async function unrollAssistant(
+  id: PackageIdentifier,
+  registry: Registry,
+  options: UnrollAssistantOptions,
+  needErrorPropagation: boolean = false,
```
Patrick-Erichsen: Could we put this on the `UnrollAssistantOptions` instead to avoid a new param? Probably `BaseUnrollAssistantOptions` specifically.

core/config/yaml/loadYaml.ts
```diff
@@ -117,39 +117,46 @@ async function loadConfigYaml(options: {
...
-      new RegistryClient({
-        accessToken: await controlPlaneClient.getAccessToken(),
-        apiBase: getControlPlaneEnvSync(ideSettings.continueTestEnvironment)
-          .CONTROL_PLANE_URL,
-        rootPath,
-      }),
-      {
-        currentUserSlug: "",
-        onPremProxyUrl: null,
+  // This is how we allow use of blocks locally
```
Patrick-Erichsen: Nitpick but this comment should be on `unrollAssistant`, not `NEED_ERROR_PROPAGATION`. Otherwise it reads like `NEED_ERROR_PROPAGATION` is what allows local blocks

core/config/yaml/loadYaml.ts
```diff
@@ -117,39 +117,46 @@ async function loadConfigYaml(options: {
...
+    NEED_ERROR_PROPAGATION,
+  );
 
-  const errors = isAssistantUnrolledNonNullable(config)
-    ? validateConfigYaml(config)
-    : [
-        {
+  const errors: ConfigValidationError[] = [];
+
+  let config = overrideConfigYaml ?? unrollResult.config;
```
Patrick-Erichsen: I think this could be a `const`?

packages/config-yaml/src/load/unroll.ts
```diff
@@ -204,14 +205,49 @@ export type UnrollAssistantOptions =
...
 export async function unrollAssistant(
   id: PackageIdentifier,
   registry: Registry,
   options: UnrollAssistantOptions,
-): Promise<AssistantUnrolled> {
+  needErrorPropagation: true,
+): Promise<ConfigResult<AssistantUnrolled>>;
+
+// Overload when we don't need error propapagation.
+export async function unrollAssistant(
```
Patrick-Erichsen, line 227: Didn't know you could overload function definitions like this in Typescript, neat

If it's possible to do it using a generic I think that would be a bit cleaner but I'm not sure how easy that would be so not a big deal

packages/config-yaml/src/load/unroll.ts
```diff
@@ -204,14 +205,49 @@ export type UnrollAssistantOptions =
...
+  id: PackageIdentifier,
+  registry: Registry,
+  options: UnrollAssistantOptions,
+  needErrorPropagation: boolean = false,
+): Promise<AssistantUnrolled | ConfigResult<AssistantUnrolled>> {
   // Request the content from the registry
   const rawContent = await registry.getContent(id);
-  return unrollAssistantFromContent(id, rawContent, registry, options);
+
+  // Somewhat annoyingly, we need this redundant ternary to circumvent overload matching error.
```
Patrick-Erichsen: Generics on the fn def might solve this

packages/config-yaml/src/load/unroll.ts
```diff
@@ -245,10 +300,15 @@ export async function unrollAssistantFromContent(
     parsedYaml,
     registry,
     options.injectBlocks,
+    needErrorPropagation,
   );
 
   // Back to a string so we can fill in template variables
-  const rawUnrolledYaml = YAML.stringify(unrolledAssistant);
+  const rawUnrolledYaml = needErrorPropagation
```
Patrick-Erichsen: It seems like we ended up with a lot of ternaries on `needErrorPropagation`. Is it not possible to just include the errors object by default, and then in a single location if `needErrorPropagation` is false we just return the direct `.config` property, else return the whole ConfigResult object? 

With that in mind it might be more accurate to call the boolean `asConfigResult` or something since that seems more accurate as to what decision the callee needs to make.

packages/config-yaml/src/load/unroll.ts
```diff
@@ -381,12 +480,29 @@ export async function unrollBlocks(
         );
       }
     } catch (err) {
+      if (needErrorPropagation) {
+        configResult.errors?.push({
+          fatal: false,
+          // Using JSON.stringify to avoid printing [object Object].
+          message: `Failed to unroll block ${JSON.stringify(injectBlock)}: ${(err as Error).message}`,
```
Patrick-Erichsen: This error was developer facing, but now that it's user-facing I think we should make them a bit more user friendly. 

E.g. `Failed to unroll block` likely doesn't mean anything to a user. 

For this particular one we could probably just entirely remove the `Failed to unroll block`  and just keep the `${JSON.stringify(unrolledBlock.uses)}: ${(err as Error).message}` since that is descriptive enough imo

packages/config-yaml/src/load/unroll.ts
jpoly1219: Decided to leave it as a typed union

core/config/yaml/loadYaml.ts
```diff
@@ -117,39 +117,46 @@ async function loadConfigYaml(options: {
   //   `Loading config.yaml from ${JSON.stringify(packageIdentifier)} with root path ${rootPath}`,
   // );
 
-  let config =
-    overrideConfigYaml ??
```
Patrick-Erichsen, line 121: Missed this on my last review, but I wonder if it was intentional that we have the `overrideConfigYaml ??` inline here with `await unrollAssistant`. This would cause the statement to short circuit if `overrideConfigYaml` was truthy, and correctly never call `unrollAssistant`. In this PR we now always call `unrollAssistant`, even if `overrideConfigYaml` is truthy.

I don't think there's a performance concern here, but I worry about the case where `unrollAssistant` blows up

Patrick-Erichsen (changes requested):
Here is the error message I got for the invalid rule 

```ts
Failed to parse block: [
  {
    "code": "too_big",
    "maximum": 1,
    "type": "array",
    "inclusive": true,
    "exact": true,
    "message": "Array must contain exactly 1 element(s)",
    "path": [
      "rules"
    ]
  }
]:
{"uriType":"file","filePath":"file:///Users/patrickerichsen/Git/continue-dev/continue/extensions/.continue-debug/rules/test.yaml"}
```

I didn't realize this on our first pass but there's still 2 problems with error bubbling here:
- For local blocks, apparently we set the `uses` clause to an object with the URI
- For errors downstream of our changes here, we're still wrapping the errors in non-user friendly messages, eg my error was from here: https://github.com/continuedev/continue/blob/main/packages/config-yaml/src/load/unroll.ts#L64

Both of these are relatively easily solvable I think
- We could make a quick util fn that checks if `uses` is an object which contains a `filePath` and then print that instead of the object, else just print the regular `uses`
- Update the handful of downstream errors in `unroll.ts` that are bubbling up to just `console.error` the more verbose error, but then throw just the raw error so we can show the user friendly version, eg

```diff
export function parseBlock(configYaml: string): Block {
  try {
    const parsed = YAML.parse(configYaml);
    const result = blockSchema.parse(parsed);
    return result;
  } catch (e: any) {
-     throw new Error(`Failed to parse block: ${e.message}`);
+     console.error(`Failed to parse block: ${e.message}`);
+     throw new Error(e)
  }
}
```

core/config/yaml/loadYaml.ts
jpoly1219: Will fix the messages to be more user friendly. As for the `??`, I will move the call to `unrollAssistant` downwards so that we actually don't waste compute on it when we don't need it. I think the intention here was that I wanted to check for errors that bubbled up during `unrollAssistant`, and I needed that instance at the top level to access its `.errors` field. But I agree with you in that there is no error to bubble up if we aren't unrolling the assistant 😅

jpoly1219:
image

This is what the errors currently look like. I think it's a good idea to open a new issue specific to error formatting 👍

packages/config-yaml/src/load/unroll.ts
```diff
@@ -439,3 +532,22 @@ export function mergeOverrides<T extends Record<string, any>>(
...
+      .map((e: any) => {
+        const path = e.path.length > 0 ? `${e.path.join(".")}: ` : "";
+        return `${path}${e.message}`;
+      })
+      .join(", ");
+  }
+  return error.message || "Validation failed";
+}
+
+function shortenedPath(filePath: string): string {
```
Patrick-Erichsen: I think this will blow up for Windows users. We have some utils in `core/util/uri.ts`, I think specifically `getLastNUriRelativePathParts`, that will handle URIs correctly

Patrick-Erichsen: +1 to this

>  I think it's a good idea to open a new issue specific to error formatting 👍

jpoly1219: Oh nice, I thought to use `path.posix` to support Windows paths, but if we already have a function to do so then it makes total sense to use that instead.

PR #5991, opened 2025-06-04
Include mcp server names in tool names

RomneyDa:
Prevents duplicate tool names from MCP servers

Has the downside of additional tokens for MCP server tools, and potential confusion if the mcp server name is very unrelated to the tool function.
Could look into other ways of making these unique but full mcp server name is fine for now. Models might do well with hashes/numbers
Could also only prepend mcp server names to duplicate tools - decided not to do this since not consistent and will cause tool policy thrashing


RomneyDa:
@chezsmithy curious if anyone on your team has built rules/workflows around tools that would break with this change?

Patrick-Erichsen (approved):

PR #5995, opened 2025-06-05
Handle JSON Parsing errors

RomneyDa:
Handles invalid tool call args parsing (caused by when JSON tool arg streaming is truncated)
Also adds some try/catch handling for other JSON parsing throughout the repo to make it more robust

A followup will be to properly cancel tool call messages by removing invalid args on cancellation
Related discussion here https://github.com/continuedev/continue/pull/5662


core/tools/parseArgs.ts
```diff
@@ -0,0 +1,26 @@
+import { ToolCallDelta } from "..";
+
+function safeParseArgs(
```
tomasz-stefaniak: I like this utility but can we define it in some shared package? It appears to be defined twice, once in core and once in openai-adapters

tomasz-stefaniak (changes requested):
I'm assuming this PR fixes some bugs, could you add relevant tests?

core/tools/parseArgs.ts
RomneyDa: Good point, made the two versions specific to their use cases

RomneyDa:
Added tests in `llm-pre-fetch.vitest.ts`!

tomasz-stefaniak (approved):
Thanks for adding the test 💪

Issue #5999, opened 2025-06-05
intellij extension create file failed if this file not exist


G81192:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: windows 10
- Continue version: v1.0.22
- IDE version: IntelliJ IDEA 2025.1.1
- Model:
- config:
  

  
  OR link to assistant in Continue hub:
```

### Description

When I use the Contine plugin in IntelliJ and generate a file with a large model, clicking the `Create file` button does not create a new file in the target directory if the file does not already exist. Instead, it directly modifies the currently open file in the editor.
![Image](https://github.com/user-attachments/assets/02fb9f94-eb55-478a-8e9e-6cce4f932fee)
After investigation, I found that the issue lies in the following piece of code, which lacks logic for creating a new file. When the target file does not exist, it proceeds to operate on the currently open file.

![Image](https://github.com/user-attachments/assets/6b06d3b5-c2e8-4e37-a737-ad97fa5556e3)

### Log output

```Shell

```

Patrick-Erichsen:
Ah, this is a good catch, thanks for pointing it out and opening up a PR to fix it!

G81192:
> Ah, this is a good catch, thanks for pointing it out and opening up a PR to fix it!

Glad it helps! Let me know if there's anything I can improve in the PR.

PR #6000, opened 2025-06-05
fix intellij extension create file failed

G81192:
## Description

## Checklist

## Screenshots

## Tests

fix intellij extension create file failed for this issue #5999

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/EditorUtils.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/EditorUtils.kt
@@ -2,6 +2,7 @@ package com.github.continuedev.continueintellijextension.editor
 
 import com.github.continuedev.continueintellijextension.utils.toUriOrNull
 import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.command.WriteCommandAction
 import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.editor.SelectionModel
@@ -14,6 +15,8 @@ import com.intellij.openapi.util.TextRange
 import com.intellij.openapi.vfs.VirtualFileManager
 import com.intellij.ui.JBColor
 import com.intellij.openapi.editor.ScrollType
+import com.intellij.openapi.vfs.LocalFileSystem
+import java.io.File
 
 /**
  * Utility class for working with Editor instances.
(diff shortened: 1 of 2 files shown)


G81192:

RomneyDa (changes requested):
@G81192 can you provide more information about this PR? In which situation is this supposed to create a file and it doesn't?

G81192:
> @G81192 can you provide more information about this approach? I'm thinking maybe it shouldn't be called "getOrOpen" if the create functionality is added. Maybe we could call it getOrCreate or something, or make the flow more clear. I want to make sure we're not creating files when we shouldn't

@RomneyDa Thank you for the reply.

The issue I'm encountering is that, with the current unmodified code, when I try to create a new file generated by the LLM by clicking create file, nothing happens. Instead, the file currently open in the IDEA window gets modified.

After looking into the code, I found that the getOrOpenEditor method returns the currently selected file as the editor when the target file does not exist. However, in my case, I need to create a new file — and since it doesn’t exist yet, this results in an error.
Example:

1. I want to create the file:
image

2. When I click create file, the currently open file in the IDEA window gets modified instead, which leads to an error:
image

3.  This is the expected behavior I want to achieve:
image

This PR is intended to fix this issue. If anything is unclear, I’m happy to provide more details.

G81192:
@Patrick-Erichsen @RomneyDa  

Hey! Just wondering if there's anything holding up the merge — happy to make changes if needed!

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/EditorUtils.kt
```diff
@@ -155,10 +158,23 @@ class EditorUtils(val editor: Editor) {
          */
         fun getOrOpenEditor(project: Project, filepath: String?): EditorUtils? {
             if (!filepath.isNullOrEmpty()) {
-                val virtualFile = VirtualFileManager.getInstance().findFileByUrl(filepath)
+                var virtualFile = VirtualFileManager.getInstance().findFileByUrl(filepath)
+                if (virtualFile == null) {
+                    // file not exists and create
+                    val ioFile = File(filepath.replace("file:///", ""))
```
Patrick-Erichsen: I think this will fail for non-unix OS? Or is this a URI that always is of this format?

Patrick-Erichsen:
I would +1 Dallin's feedback, I think the existing method should be left as is and instead we create the file in https://github.com/continuedev/continue/blob/main/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/ApplyToFileHandler.kt#L172 when we realize it doesnt exist. I don't think we always want the side effect of creating a new file in `getOrOpenEditor`

G81192:
> I would +1 Dallin's feedback, I think the existing method should be left as is and instead we create the file in https://github.com/continuedev/continue/blob/main/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/ApplyToFileHandler.kt#L172 when we realize it doesnt exist. I don't think we always want the side effect of creating a new file in `getOrOpenEditor`

Thanks for the feedback! I've updated the code to reflect the suggestion — the logic for creating a new file has been moved out of `getOrOpenEditor`. Appreciate the guidance — looking forward to your review.

Issue #5502, opened 2025-05-04
Error: Premature close


oldagecoder1:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: 1.0.8 
- IDE version: VScode (version don't know)
- Model: Cloude 3.7 Sonnet
- config:
  

  
  OR link to assistant in Continue hub:
```

### Description

I have purchased plan. 
I selected Cloude 3.7 sonnet model. and when I try to give the prompt. it start acting on the prompt check codebase files (like 1 or 2) after that it start giving this error. (Premature close)

### To reproduce

1. Add prompt to do some task
2. after some files check it gives premature close

<img width="337" alt="Image" src="https://github.com/user-attachments/assets/14f176a6-be1f-439c-b8fa-205a48a55dea" />

### Log output

```Shell
[Extension Host] Error handling webview message: {
  "msg": {
    "messageId": "35ec0e33-7e0b-47a9-89e7-09ca44c2a89c",
    "messageType": "llm/streamChat",
    "data": {
      "completionOptions": {
        "tools": [
          {
            "type": "function",
            "displayTitle": "Read File",
            "wouldLikeTo": "read {{{ filepath }}}",
            "isCurrently": "reading {{{ filepath }}}",
            "hasAlready": "viewed {{{ filepath }}}",
            "readonly": true,
            "isInstant": true,
            "group": "Built-In",
            "function": {
              "name": "builtin_read_file",
              "description": "Use this tool if you need to view the contents of an existing file.",
              "parameters": {
                "type": "object",
                "required": [
                  "filepath"
                ],
                "properties": {
                  "filepath": {
                    "type": "string",
                    "description": "The path of the file to read, relative to the root of the workspace (NOT uri or absolute path)"
                  }
                }
              }
            }
          },
          {
            "type": "function",
            "displayTitle": "Edit File",
            "wouldLikeTo": "edit {{{ filepath }}}",
            "isCurrently": "editing {{{ filepath }}}",
            "hasAlready": "edited {{{ filepath }}}",
            "group": "Built-In",
[...]

PR #6009, opened 2025-06-05
Node fetch Premature close patch

RomneyDa:
Implements node fetch patch for chunked boundary detection described in https://github.com/node-fetch/node-fetch/issues/1576

Addresses premature close issues e.g. #5502 

Could also be related to https://github.com/node-fetch/node-fetch/issues/1767 but we do NOT have maxSockets set so I don't think that's the cause

### DIFF
--- a/packages/fetch/package-lock.json
+++ b/packages/fetch/package-lock.json
@@ -22,18 +22,84 @@
     },
     "node_modules/@continuedev/config-types": {
       "version": "1.0.5",
-      "license": "Apache-2.0",
+      "resolved": "https://registry.npmjs.org/@continuedev/config-types/-/config-types-1.0.5.tgz",
+      "integrity": "sha512-Uv3JWKmpgSo9cMhZb2hWoJgR2ezuAn3CenEziWT/5d7mddPLzvuuHgFX6aDjLVVMojS1ul0rpfAd2LGxEb6nDQ==",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


packages/fetch/src/node-fetch-patch.js
```diff
@@ -0,0 +1,512 @@
...
+    // 		);
+    // 	}
+
+    // 	previousChunk = buf;
+    // };
+
+    // VENDOR note - THIS IS THE REASON I AM PATCHING THIS PACKAGE
+    // See https://github.com/node-fetch/node-fetch/issues/1576
+    // FIXES BOUNDARY CHUNKING ISSUE
+    const onData = (buf) => {
```
RomneyDa, line 493: patch here

packages/fetch/src/node-fetch-patch.js
```diff
@@ -0,0 +1,512 @@
+/**
```
tomasz-stefaniak, line 8: Could you prepend the source of this file somewhere in the comment? Might help with debugging or pulling in updates in the future.

Some reference of a Github issue (or other explanation) that prompted this patch would also be helpful as a future reference.

tomasz-stefaniak (dismissed):
Approved with comment ⬆️

tomasz-stefaniak (approved):
Thanks for updating the comments, approved (once tests pass)

PR #6012, opened 2025-06-06
[Updated PR] [Feature] Autocomplete context from other tabs open in the editor

adarshramiyer:
## Description

Adds context to autocomplete using files open in the editor. Logs open files and viewing order in an LRU cache, then prioritizes fitting recent and shorter files into the context window. Adaptively trims files to fit within prompt token limits as necessary.

## Checklist


core/autocomplete/snippets/getAllSnippets.ts
```diff
@@ -103,6 +105,48 @@ const getDiffSnippets = async (
...
+    const currentFileUri = `${helper.filepath}`;
+
+    // converts cache file URIs to autocomplete snippets
+    for (const [fileUri, _] of openedFilesLruCache.entriesDescending()) {
+      if (fileUri === currentFileUri) {
+        continue;
+      }
+
+      try {
+        const fileContent = await ide.readFile(fileUri);
```
RomneyDa: How was the max cache size decided, by file read latency? It seems like with this setup it might often not finish reading all files in the cache within 100 ms, e.g. with 20 files could take 200+. Consider using Promise.all()

extensions/vscode/src/VsCodeIde.ts
```diff
@@ -96,6 +94,19 @@ class VsCodeIde implements IDE {
     });
   }
 
+  onDidCloseTextDocument(callback: (uris: string[]) => void): void {
+    vscode.workspace.onDidCloseTextDocument((document) => {
+      if (document) {
+        let openFilePaths = vscode.window.tabGroups.all // get tab groups
+          .flatMap((group) => group.tabs) // extract tabs from tab groups
+          .filter((tab) => tab.input instanceof vscode.TabInputText) // filter irrelevant tabs (settings, etc.)
+          .map((tab) => (tab.input as vscode.TabInputText).uri.fsPath); // get filepaths
```
RomneyDa: Use the existing VsCodeIDEUtils getOpenFiles function. Looks like you might be able to improve it by using this TabInputText approach instead of the existing `documentIsCode` filter. Make sure to use uri.toString() instead of fsPath

core/index.d.ts
```diff
@@ -1105,6 +1107,7 @@ export interface TabAutocompleteOptions {
   useCache: boolean;
   onlyMyCode: boolean;
   useRecentlyEdited: boolean;
+  useRecentlyOpened: boolean;
```
RomneyDa, line 1108: recentlyClosed?

core/core.ts
```diff
@@ -673,6 +683,47 @@ export class Core {
...
+        );
+      }
+    });
+
+    on(
+      "initializeOpenedFileCache",
+      async ({ data: { initialOpenedFilePaths } }) => {
+        try {
+          for (const filepath of initialOpenedFilePaths) {
+            openedFilesLruCache.set(`file://${filepath}`, `file://${filepath}`);
```
RomneyDa: remove file://, use URI from IDE all the way through

core/autocomplete/templating/formatOpenedFilesContext.ts
```diff
@@ -0,0 +1,164 @@
...
+  scoredSnippets.sort((a, b) => b.score - a.score);
+
+  return scoredSnippets
+    .slice(0, Math.min(defaultNumFilesUsed, scoredSnippets.length))
+    .map((item) => item.snippet);
+};
+
+// Returns linear combination of recency and size scores
+// recency score is exponential decay over recency; log normalized score is used for size
+const getRecencyAndSizeScore = (
```
RomneyDa, line 122: This would be awesome for unit tests

RomneyDa (changes requested):
Clean implementation!

Some changes in mind:
1. Use the existing `files/closed` or `files/opened` messages rather than introducing a new `onDidCloseTextDocument` callback and `initializeOpenedFileCache` message. "files/closed" is already set up for both vs code and jetbrains. It seems like in this implementation, `recentlyEdited` is for open files and `recentlyOpened` is for closed files, but `recentlyOpened` is also initialized with open files. Let's talk about this naming and review the snippet priority. I think recently edited should be prioritized over recently closed. Also, let's implement the existing `files/opened` message rather than adding a new `initializeOpenedFileCache` message. It seems like those should be added to the recently edited cache, not the recently "opened" (closed) cache. 
2. Use URIs all the way through, don't use "file://", will cause bugs, just use uri.toString() in the IDE side . Note that `helper.filepath` was named before we transitioned to URIs, bit misleading
3. Let's add unit tests for the formatting and ranking, especially around the math stuff!

I also wanted to talk to you about snippet ranking, deduplication, and file read latency, can chat in person, you probably already tested this enough but wanted to get a grasp on it. Did we consider storing the actual file contents in the cache? Might not be up to date but would be much faster and these are for "closed" files. Files are often large, is there a concern of recently closed files always filling the remainder of snippet context window?



Were the additions/changes to manual-testing-sandbox intentional? Fine to be included, just checking

core/core.ts
```diff
@@ -597,14 +600,56 @@ export class Core {
     });
 
     on("files/closed", async ({ data }) => {
+      try {
+        const filepaths = data.fileUris;
```
RomneyDa: No need for fileUris and uris, same thing

extensions/vscode/src/extension/VsCodeExtension.ts
```diff
@@ -76,6 +78,7 @@ export class VsCodeExtension {
       },
     );
     this.ide = new VsCodeIde(this.webviewProtocolPromise, context);
+    this.ideUtils = new VsCodeIdeUtils();
```
RomneyDa, line 81: use this.ide.getOpenFiles and merge the utils, no need for new util

RomneyDa (changes requested):
great new tests!
Also I think might be an issue with useRecentlyOpened setting - how can it be set, what is the default? JSON only for now?

PR #6013, opened 2025-06-06
better premature close error, resubmit button on error dialog

RomneyDa:
## Description

Fleshed out error message for "Premature close" error, and a new resubmit button that resubmits at the most recent user/tool message

## Screenshots

image

### DIFF
--- a/packages/fetch/src/premature-close.ts
+++ b/packages/fetch/src/premature-close.ts
@@ -1,0 +1,14 @@
+export function getPrematureCloseErrorMessage(chunks: number): string {
+  if (chunks === 0) {
+    return "Connection was closed before any data was received (Premature close). Try again.";
+  } else {
+    return (
+      "Stream was cancelled mid-stream. This can happen for various reasons, including:\n" +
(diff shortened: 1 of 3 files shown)


packages/fetch/src/premature-close.ts
```diff
@@ -0,0 +1,14 @@
+export function getPrematureCloseErrorMessage(chunks: number): string {
+  if (chunks === 0) {
+    return "Connection was closed before any data was received (Premature close). Try again.";
+  } else {
+    return (
```
sestinj: I really don't think the end user wants to see all of this. If they start trying to solve "Keep alive" on their own they will never succeed. Could it just be "The response was cancelled mid-stream"

sestinj (changes requested):

packages/fetch/src/premature-close.ts
RomneyDa: Agreed, updated

sestinj (approved):

PR #6023, opened 2025-06-06
feat: use correct deployment for azure

Patrick-Erichsen:


packages/openai-adapters/src/apis/Azure.ts
```diff
@@ -0,0 +1,108 @@
...
+    return apiType === "azure-openai" || apiType === "azure";
+  }
+
+  private _getAzureBaseURL(config: z.infer<typeof AzureConfigSchema>): string {
+    const baseURL = new URL(this.apiBase).toString().replace(/\/$/, "");
+
+    // Default is `azure-openai` in docs, but previously was `azure`
+    if (this._isAzureOpenAI(config.env?.apiType)) {
+      if (!config.env?.deployment) {
+        throw new Error(
```
Patrick-Erichsen, line 44: Because we call `constructLlmApi` even if we aren't using openai-adapters, and in our LLM Azure class we set `azure-openai` as the default, this will throw for existing users that don't have `deployment` set. But the current behavior n LLM OpenAI would be failing anyways without this

https://github.com/continuedev/continue/blob/pe/watsonx-vi/core/llm/llms/OpenAI.ts#L423-L428

```ts
return new URL(
        `openai/deployments/${this.deployment}/embeddings?api-version=${this.apiVersion}`,
        this.apiBase,
      );
```

PR #6024, opened 2025-06-06
feat: skip hub onboarding in free trial

Patrick-Erichsen:
## Description

Users who select free trial option from IDE onboarding are just brought through workOS sign in, and then back to the extension. 

The only change that was required for this was updating the redirect URI. The rest of this PR is just a bit of UI polish as I was going through auth UI.


sestinj (approved):
tried this locally, so much better. nice styling, no tests needed

PR #6048, opened 2025-06-09
Tomasz/rule colocation

tomasz-stefaniak:
## Description

Place rules.md files anywhere in your codebase.

## Checklist

- [] The relevant docs, if any, have been updated or created - Will be done separately to avoid conflicts with https://github.com/continuedev/continue/pull/6049

## Screenshots

Demo:

https://github.com/user-attachments/assets/0b2597b4-7ac8-4c5a-8b9a-20985c4c3cc1

## Tests

Wrote 60+ tests


core/config/markdown/loadCodebaseRules.ts
```diff
@@ -0,0 +1,68 @@
...
+import { IDE, RuleWithSource } from "../..";
+import { walkDirs } from "../../indexing/walkDir";
+import { convertMarkdownRuleToContinueRule } from "./parseMarkdownRule";
+
+/**
+ * Gets the directory part of a path (without using Node.js path module)
+ */
+const getDirname = (filePath: string): string => {
+  // Normalize path separators to forward slash
+  const normalizedPath = filePath.replace(/\\/g, "/");
```
Patrick-Erichsen: I think this will fail on Windows systems? I believe we need to be using the utils in `core/util/uri.ts` here, and elsewhere in this file

`pathToUriPathSegment` might be what you need

core/core.ts
```diff
@@ -578,6 +578,20 @@ export class Core {
         walkDirCache.invalidate();
         void refreshIfNotIgnored(data.uris);
 
+        // Check for rules.md files being created
+        let rulesFileCreated = false;
+        for (const uri of data.uris) {
+          const parts = uri.split("/");
+          const filename = parts[parts.length - 1];
+          if (filename === "rules.md") {
```
Patrick-Erichsen: Might be nice to have a `RULES_FILENAME` const that we can use here and elsewhere instead of a plain string

core/core.ts
```diff
@@ -595,6 +609,20 @@ export class Core {
       if (data?.uris?.length) {
         walkDirCache.invalidate();
         void refreshIfNotIgnored(data.uris);
+
+        // Check for rules.md files being deleted
+        let rulesFileDeleted = false;
```
Patrick-Erichsen: Feels like we could keep things DRY and make a `hasAddedOrDeletedRuleFile` for the same logic above

core/llm/rules/getSystemMessageWithRules.ts
```diff
@@ -10,19 +10,126 @@ import { extractPathsFromCodeBlocks } from "../utils/extractPathsFromCodeBlocks"
...
+};
+
+/**
+ * Checks if a rule is a root-level rule (.continue directory or no file path)
+ */
+const isRootLevelRule = (rule: RuleWithSource): boolean => {
+  return (
+    !rule.ruleFile ||
+    rule.ruleFile.startsWith(".continue/") ||
+    rule.ruleFile === ".continue/rules.md"
```
Patrick-Erichsen: I'm a little confused by this option, shouldn't it just be any rules in `.continue`, or that don't have a `ruleFile`? Why the specific check for `.continue/rules.md`?

Patrick-Erichsen (changes requested):
Thanks for locking down the behavior with all of these tests!

Everything looks good to me except the path things I mentioned in a comment below. Hopefully using those URI utils will remove a lot of the path related code in this PR.

core/llm/rules/getSystemMessageWithRules.ts
tomasz-stefaniak: It was unnecessary, removed 👌

Patrick-Erichsen (approved):

PR #6058, opened 2025-06-10
fix: apply notch filename trunaction

Patrick-Erichsen:
## Description
- Fixes overflow on apply notch by hiding the `Accept`/`Reject` text on xs screens
- Hides diff text on md screens and below for apply actions in toolbar

Closes CON-2134

## Screenshots
Screenshot 2025-06-09 at 6 15 12 PM

Screenshot 2025-06-09 at 6 22 03 PM


Patrick-Erichsen:
My other thoughts was that on xs screens and below we could hide the copy/insert buttons when applying, but this was a much easier fix

PR #6062, opened 2025-06-10
telemetry for stream_error

sestinj:
## Description

Improve visibility to blocking errors in agent / chat modes

## Checklist

## Tests

Tests aren't appropriate here

### DIFF
--- a/gui/src/pages/gui/StreamError.tsx
+++ b/gui/src/pages/gui/StreamError.tsx
@@ -6,7 +6,8 @@ import {
   KeyIcon,
 } from "@heroicons/react/24/outline";
 import { DISCORD_LINK, GITHUB_LINK } from "core/util/constants";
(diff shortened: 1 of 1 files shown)


sestinj:
@RomneyDa I think the only alternative is this:

```tsx
const useErrorLogger = (error: unknown, additionalData: object) => {
  const hasLogged = useRef(false);
  
  if (error && !hasLogged.current) {
    posthog.capture("gui_stream_error", additionalData);
    hasLogged.current = true;
  }
};
```

And that seems much less clean, and basically equivalent

RomneyDa:
@sestinj
I should have been more specific
I mean, move parsing logic away from the component into a utility, and send the error telemetry when the error is thrown in the stream wrapper thunk, no need for a one-time hook

PR #6063, opened 2025-06-10
chore: migrate core (autocomplete) tests to vitest

uinstinct:
## Description

As part of our migration of tests inside `core` from jest to vitest, this PR:

- migrates the tests in the **`autocomplete`** directory.
- introduces vitest.global-setup and vitest.setup

related to CON-2030

## Checklist

## Screenshots

## Tests


core/autocomplete/context/root-path-context/test/testUtils.ts
```diff
@@ -46,7 +46,7 @@ export async function testRootPathContext(
 
   // Copy the folder to the test directory
   const folderPath = path.join(
-    __dirname,
```
uinstinct, line 49: vitest gets the actual `__dirname` where the file resides while for jest __dirname is equivalent process.cwd()

for example, in this case, __dirname for vitest would be `/users/myuser/continue/core/autocomplete/context/root-path-context/test` and for jest it would be `/users/myuser/continue/core`

Hence used `process.cwd()`

core/autocomplete/generation/utils.test.ts
```diff
@@ -37,11 +37,11 @@ describe("stopAfterMaxProcessingTime", () => {
     expect(fullStop).not.toHaveBeenCalled();
   });
 
-  it.only("should stop processing after max time is reached", async () => {
+  it("should stop processing after max time is reached", async () => {
```
tomasz-stefaniak, line 40: 👍

core/test/vitest.setup.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// TODO - currently causing tests to fail because sqlite is still running for some reason
+// const clearTestDirectory = () => {
+//   if (fs.existsSync(process.env.CONTINUE_GLOBAL_DIR!)) {
+//     fs.rmSync(process.env.CONTINUE_GLOBAL_DIR!, { recursive: true });
+//   }
+// };
+
+// globalThis.beforeAll(clearTestDirectory);
+// globalThis.afterAll(clearTestDirectory);
```
tomasz-stefaniak: Are these still relevant?

core/util/treeSitter.ts
```diff
@@ -181,7 +181,7 @@ export async function getQueryForFile(
   }
 
   const sourcePath = path.join(
-    __dirname,
+    process.env.NODE_ENV === "test" ? process.cwd() : __dirname,
```
tomasz-stefaniak, line 184: I'd prefer to avoid different control flows in the test environment. Is there a way we can get it to work without switching to `process.cwd()` in the test env?

Btw, is this related to this comment? 
https://github.com/continuedev/continue/pull/6063/files#r2137084957

core/vitest.config.ts
```diff
@@ -2,6 +2,17 @@ import { defineConfig } from "vitest/config";
 
 export default defineConfig({
   test: {
-    include: ["**/*.vitest.ts"],
```
tomasz-stefaniak, line 5: Over the last two weeks or so, we have adopted a convention that all vitest tests use `.vitest.ts` filename instead of `.test.ts`. This makes it easier to migrate the tests one test at a time. 

Removing this prevents the handful of tests we've already migrated to vitest from running.

core/test/vitest.setup.ts
uinstinct: nope, removed!
also am ensuring that wherever we have setup of test environment, it is destroyed as well

(actually had started by copy jest's setup file so it was present)

core/vitest.config.ts
uinstinct: okay, there are few tests which are present in `.vitest.ts`
Will keep that!

initial flow/thought process was to migrate by including the test files for vitest in config and excluding those from jest's config. had not accounted for already present `.vitest.ts` files

core/util/treeSitter.ts
uinstinct: yes, it is the related to the [comment](https://github.com/continuedev/continue/pull/6063/files#r2137084957)

I tried this now by only putting `process.cwd()`. In that case the `sourcePath` becomes `/tree-sitter/import-queries/python.scm` (where by using `__dirname` should have been like this `/Users/instinct/Desktop/continue/extensions/vscode/tree-sitter/import-queries/python.scm`)

core/vitest.config.ts
```diff
@@ -2,6 +2,17 @@ import { defineConfig } from "vitest/config";
...
+      web: ["/.[jt]s?$/"],
+      ssr: ["/.[jt]s?$/"],
+    },
+    globalSetup: "./test/vitest.global-setup.ts",
+    setupFiles: "./test/vitest.setup.ts",
+    fileParallelism: false,
+    include: [
+      "**/*.vitest.ts",
+      "config/yaml/LocalPlatformClient.test.ts",
+      "autocomplete/**/*.test.ts",
```
tomasz-stefaniak: As we spoke on the huddle earlier today, let's only include `["**/*.vitest.ts"]` and update test names accordingly instead.

tomasz-stefaniak (changes requested):
Left a comment ⬆️

tomasz-stefaniak (approved):
Looks great, thanks for making the changes!

PR #6070, opened 2025-06-10
RulesContextProvider

sestinj:
## Description

Introduces the RulesContextProvider, which is available by default

Screenshot 2025-06-10 at 3 19 42 PM

Screenshot 2025-06-10 at 3 19 59 PM

Screenshot 2025-06-10 at 3 20 10 PM

## Checklist

## Tests

I don't feel that any meaningful logic was introduced that needs to be tested. If we tested some of the private methods on the class, that feels like it would be overkill


core/config/profile/doLoadConfig.ts
```diff
@@ -39,6 +41,24 @@ import { migrateJsonSharedConfig } from "../migrateSharedConfig";
 import { rectifySelectedModelsFromGlobalContext } from "../selectedModels";
 import { loadContinueConfigFromYaml } from "../yaml/loadYaml";
 
+async function loadRules(ide: IDE) {
```
Patrick-Erichsen, line 44: This is going to conflict with @tomasz-stefaniak  PR, I think this is a good improvement though because his PR adds a 3rd rule source that we load from

core/context/providers/RulesContextProvider.ts
```diff
@@ -0,0 +1,91 @@
...
+    renderInlineAs: "",
+  };
+
+  // This is only used within this class. Worst case if there are exact duplicates is that one always calls the other, but this is an extreme edge case
+  // Can eventually pull in more metadata, but this is experimental
+  private getIdFromRule(rule: RuleWithSource): string {
+    return rule.ruleFile ?? rule.slug ?? rule.name ?? rule.rule;
+  }
+
+  private getNameFromRule(rule: RuleWithSource): string {
```
Patrick-Erichsen, line 28: We do this in `convertMarkdownRuleToContinueRule` to grab a fallback name from the first H1 tag in the markdown. 

```ts
  // Try to extract title from first heading if no name in frontmatter
  let name = frontmatter.name;
  if (!name) {
    // Look for a heading in the markdown
    const headingMatch = markdown.match(/^#\s+(.+)$/m);
    if (headingMatch) {
      name = headingMatch[1].trim();
    } else {
      // Fall back to filename
      name = basename(path).replace(/\.md$/, "");
    }
  }
```

Kind of edge case though when we have the `name` property. If you don't think it's worth pulling in here maybe we just get rid of that logic in `convertMarkdownRuleToContinueRule` to keep behavior consistent though.

core/context/providers/RulesContextProvider.ts
Patrick-Erichsen, line 1: A lot of the logic here feels like it should live in `loadMarkdownRules.ts` or some other `rules` subfolder so that we can reuse it, which I anticipate happening.  E.g. most of @tomasz-stefaniak 's new work lives in `core/config/markdown`

Also would make it easier to test some of these helper fns

Patrick-Erichsen (approved):
LGTM, few nitpick comments, only other thing is I think we should use the `PencilIcon` instead of the current default `ChatBubbleIcon` so that we mirror what is in the notch.

core/context/providers/RulesContextProvider.ts
sestinj: I think the difference between this and `convertMarkdownRuleToContinueRule` is that `RuleWithSource` doesn't need to have a name. That function actually should get called prior to this one, so I don't think this method will ever be used. But is necessary for typing because I don't think we'd want to make `name` required on `RuleWithSource` (it wouldn't match the .md source). The `name`, `id`, and `description` are really concepts that are unique to the `ContextItem` interface so I don't think they would need to be used anywhere else

core/context/providers/RulesContextProvider.ts
sestinj: for the reason mentioned in my other comment I actually kind of think that these small methods won't ever need to be used anywhere else

PR #6074, opened 2025-06-11
Include local blocks for hub assistants

RomneyDa:
## Description
Adds support for local blocks to hub assistants
Achieves this by creating a simple "dud" assistant to unroll, made up of any local blocks
Moves some utils around, including retrieval of local block files and creation of a registry. This is mostly so that adding this support for JSON is simple in the future.

This should be followed by a PR to add support for JSON


sestinj:
I think the test failures here are legit @RomneyDa

sestinj (changes requested):
seems good, but waiting on failing tests for further review

PR #6081, opened 2025-06-11
Fetch URL Content Tool

RomneyDa:
## Description
- Adds a simple tool for reading a URL, disabled by default for now.
- Uses the same context item gathering as the URL context provider and prompt files
- This PR also changes context item peeks so that if you click the URL it opens the website but anything else opens the virtual file like any other context item. Before, it was impossible to see what content was actually retrieved

### DIFF
--- a/core/tools/builtIn.ts
+++ b/core/tools/builtIn.ts
@@ -11,7 +11,7 @@ export enum BuiltInToolNames {
   LSTool = "builtin_ls",
   CreateRuleBlock = "builtin_create_rule_block",
   RequestRule = "builtin_request_rule",
-  ReadUrl = "builtin_read_url",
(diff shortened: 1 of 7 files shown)


core/tools/builtIn.ts
```diff
@@ -11,6 +11,7 @@ export enum BuiltInToolNames {
   LSTool = "builtin_ls",
   CreateRuleBlock = "builtin_create_rule_block",
   RequestRule = "builtin_request_rule",
+  ReadUrl = "builtin_read_url",
```
Patrick-Erichsen: I feel like this title is a bit vague. What about something like `fetch_from_url`? Don't feel that strongly though

RomneyDa: changed to fetch_url_content

PR #6109, opened 2025-06-13
fix: :bug: Add requestOptions to YAML config for mcp

chezsmithy:
## Description

Add request options to mcp server YAML config. Without this in place you can't pass headers for example.

## Checklist

## Screenshots

## Tests

Confirmed the following configuration loads headers in:

https://github.com/chezsmithy/continue/blob/16ef20aeca9696f0b751e590671111eafc5fcda7/core/context/mcp/MCPConnection.ts#L352

```yaml
name: SSE Server
version: 0.0.1
schema: v1
mcpServers:
  - name: sse-server
    type: sse
    url: https://www.example.com
    requestOptions:
      headers: { Authorization: "Bearer ABC", JWT: XXASDASDSAS }
```

### DIFF
--- a/packages/config-yaml/src/schemas/index.ts
+++ b/packages/config-yaml/src/schemas/index.ts
@@ -1,7 +1,11 @@
 import * as z from "zod";
 import { commonModelSlugs } from "./commonSlugs.js";
 import { dataSchema } from "./data/index.js";
-import { modelSchema, partialModelSchema } from "./models.js";
+import {
+  modelSchema,
+  partialModelSchema,
+  requestOptionsSchema,
+} from "./models.js";
 
 export const contextSchema = z.object({
   name: z.string().optional(),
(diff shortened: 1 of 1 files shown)


sestinj (changes requested):
I like the change—can we add this to the docs, or do we need to actually implement this in the code?

Edit: I see your link to the code. I think instead of _only_ injecting the headers, we should override the whole fetch function in the same standard way we do throughout the codebase, using `@continuedev/fetch`

sestinj:
Edit: I see that this is potentially urgent for you. If you wanted to merge this in and make a follow up PR to address my comments, I would be okay with that

chezsmithy:
@sestinj I think I would prefer to merge as is, given this is where the requestOptions are being pulled in. The contributor who made the original change based their work off the json config which had the requestOptions already included. I'd be cool if you want to open up a new issue and I can see how we can use the fetch override. https://github.com/continuedev/continue/blob/bb0490ad74fca80ba87a8d4d820d6d20744d4517/core/context/mcp/MCPConnection.ts#L364

sestinj:
works for me. I made a follow up issue here @chezsmithy https://github.com/continuedev/continue/issues/6234

sestinj (approved):
lgtm

PR #6126, opened 2025-06-14
fix: prevent memory leaks from AbortController in async operations

lkk214:
## Description

AbortController instances were not being properly cleaned up after async operations (Promise/AsyncGenerator) completed or failed, leading to potential memory leaks.

## Testing
Verified with:
- Run async tasks until completion
- Premature interruption cases
- Memory profiling before/after changes


sestinj (changes requested):
Thanks for doing this, it's a great fix. I do think it would be good to have a quick test for the `runWithAbortController` method, since this is going to be on the critical path of every single request. Once there are tests I'm happy to merge!

lkk214:
Since https://github.com/continuedev/continue/pull/5828 still uses the original implementation of `addMessageAbortController` and `abortById`, I’ve marked this PR as draft for now. Will update the references here once https://github.com/continuedev/continue/pull/5828 is merged.

sestinj:
I want to close this until we can make progress on it to avoid stale PRs. I don't necessarily disagree that it's a good idea but we need to make sure we can keep a clean inbox. Would welcome a fresh PR not built on top of the other!

PR #6131, opened 2025-06-15
fix: refresh codebase index on config change

shssoichiro:
## Description

Updates the CodebaseIndexer to refresh the index when the assistant config is changed. The DocsService already does this, so this change copies the approach that the DocsService is using.

The biggest current issue is that the codebase indexing will begin before Continue has loaded remote assistant configs, which means it will use whatever model is available locally (usually Transformers.js in VSCode), or worse, silently fail. By refreshing after the config loads, we will ensure we are indexing the codebase using the user's configured embed model.

## Checklist

## Screenshots

N/A This is a backend change.

## Tests

Added appropriate tests to CodebaseIndexer.test.ts


RomneyDa (changes requested):
@shssoichiro it would be great to refresh codebase indexing, but I think it should only update when the embeddings model or @codebase config changes. Indexing (even just computing potential changes) can be pretty expensive.

Let's add something similar to the DocsService's `siteIndexingConfigsAreEqual` that checks if the embeddings model config has changed OR the codebase context provider config has changed

shssoichiro:
Thanks, I've updated both indexers as requested and fixed the tests which were failing due to caching.

core/indexing/CodebaseIndexer.ts
```diff
@@ -648,4 +670,60 @@ export class CodebaseIndexer {
...
+        provider.description.title ===
+        CodebaseContextProvider.description.title,
+    );
+    const codebaseProvider2 = config2.contextProviders?.find(
+      (provider) =>
+        provider.description.title ===
+        CodebaseContextProvider.description.title,
+    );
+    return (
+      config1?.selectedModelByRole.embed ===
```
RomneyDa: I think these references will change every time config is loaded, for both the contextProvider and embed model, so this will always return false in practice, could use JSON stringify or similar

shssoichiro: Good callout. Do you think it's good enough to check `providerName`? This seems like the simplest approach.

RomneyDa: I think `underlyingProviderName` and `title` should be good enough, maybe `maxEmbeddingChunkSize` for good measure

shssoichiro: Thanks, done.

shssoichiro:
What would be a good way to check for this? Looking at the `ContextProviderDescription`, it seems like the properties there would typically always be the same given that we are always looking at the "codebase" provider.

RomneyDa:
@shssoichiro you are right that changes in `nRetrieve` and `nFinal` and `useReranking` (the only codebase params) should not trigger a refresh. I just mean that the equality expression `codebaseProvider1 === codebaseProvider2` will always evaluate to false, because they are objects that are created with each config reload.

shssoichiro:
Thanks, that makes sense. In that case it probably makes sense to remove that check and just check if the embed provider has changed. I've done that and fixed the merge conflicts.

core/indexing/docs/DocsService.ts
```diff
@@ -1261,7 +1306,12 @@ export default class DocsService {
     // Handles the case where a user has manually added the doc to config.json
     // so it already exists in the file
     const doesEquivalentDocExist = this.config.docs?.some((doc) =>
-      siteIndexingConfigsAreEqual(doc, siteIndexingConfig),
+      siteIndexingConfigsAreEqual(
+        doc,
+        siteIndexingConfig,
+        this.config,
+        this.config,
```
RomneyDa: @shssoichiro passing same config twice to each siteIndexingConfigsAreEqual will always result in embed model being equal. Need to pass oldConfig and newConfig

RomneyDa (changes requested):

core/indexing/docs/DocsService.ts
shssoichiro: Thanks, in some places this was done intentionally because it is only checking whether the doc config changed, but I will double check all of the call sites to ensure they are correct.

shssoichiro: All the calls were correct, but I refactored the code by adding a new helper function which makes the intent of these calls clearer.

RomneyDa (changes requested):
@shssoichiro the `docConfigsAreEqual` is not needed since the 3rd param does not do anything. Can we take config params out of `siteIndexingConfigsAreEqualExceptTitleAndFavicon` and call `embedModelsAreEqual` separately so we don't need a redundant `docConfigsAreEqual` utility?

`siteIndexingConfigsAreEqual` and `

RomneyDa (approved):

PR #6138, opened 2025-06-16
feat: move rule parsing to `config-yaml`

Patrick-Erichsen:
Consolidates rule parsing functionality by moving it from the core config module to the shared `config-yaml` package.


packages/config-yaml/package.json
```diff
@@ -1,10 +1,17 @@
 {
   "name": "@continuedev/config-yaml",
-  "version": "1.0.88",
+  "version": "1.0.92",
```
Patrick-Erichsen, line 3: reminder to publish this new version once we merge

packages/config-yaml/package.json
```diff
@@ -1,10 +1,17 @@
...
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "type": "module",
+  "exports": {
+    ".": {
+      "browser": "./dist/browser.js",
+      "node": "./dist/index.js",
+      "default": "./dist/index.js"
+    }
+  },
```
Patrick-Erichsen, line 14: I had to add this because I updated `unrollAssistant` to include logic from our new `config-yaml/markdown` folder. `unrollAssistant` is included in `registryClient`, which uses node APIs, so we needed a separate entrypoint for browser envs

packages/config-yaml/src/load/unroll.ts
```diff
@@ -503,7 +504,26 @@ export async function resolveBlock(
...
+  } catch (yamlError) {
+    // If YAML parsing fails, try parsing as markdown rule
+    try {
+      const rule = markdownToRule(templatedYaml);
+      // Convert the rule object to the expected format
+      parsedYaml = {
+        name: rule.name,
+        version: "1.0.0",
+        rules: [rule],
+      };
```
Patrick-Erichsen, line 531: We map markdown rules to yaml rules for now

packages/config-yaml/src/schemas/index.ts
```diff
@@ -48,8 +49,18 @@ const ruleObjectSchema = z.object({
...
 
+// TODO: What should be required?
+export const rulesJsonSchema = z.object({
+  name: z.string(),
+  version: z.string(),
+  author: z.string().optional(),
+  license: z.string().optional(),
+  rules: z.record(z.string(), z.string()).optional(),
+});
+
```
Patrick-Erichsen: I'm not sure if this is where we want to define this, but I'd like to have the types available for use on the hub

packages/config-yaml/package.json
sestinj: nice! This is something I didn't know about, and had needed this before

packages/config-yaml/src/schemas/index.ts
sestinj: So far I'm defining it in a JSON schema here: https://github.com/continuedev/rules-cli/blob/main/internal/validation/schema/rules-schema.json because we'll need it for both CLI and other things. Might not be necessary to define this here. Just making a note to myself here

sestinj (approved):
Made a few changes to address small bugs, but the design choices here are fine and it's well tested

PR #6146, opened 2025-06-16
fix: do not delete docs indexes on resync

### LINKED ISSUE #6139: Continue reindexes all docs on every startup

### Relevant environment info

```Markdown
- OS: All
- Continue version: Latest stable
- IDE version: Latest VSCode and IntelliJ
- Model: nomic-embed-text
- config: https://hub.continue.dev/expedientfalcon/expedientfalcon-versatile
```

### Description

Every time I start either VSCode or IntelliJ, Continue clears all my previous Docs indexes and reindexes all of them from scratch. This seems completely unnecessary and wastes my CPU/GPU power. Continue needs to be able to skip re-indexing if a site is already in the local Lancedb index for a given model.

### To reproduce

1. Have a moderate number of docs
2. Restart your IDE
3. Look at Docs tab
4. See that they are all reindexing from scratch

### Log output

```Shell

```

shssoichiro:
## Description

When Continue loads, it takes several seconds for it to fetch the remote assistant configs. Until then, it defaults back to the local assistant config. If the user has their docs set up in the remote assistant but not in the local assistant, then this code to delete all non-existent docs from the index will cause all doc indexes to be cleared out on every restart, which is the behavior I am seeing in #6139. This would also be an issue for users who have multiple assistants with different docs in each one.

Fixes #6139.

## Checklist

## Tests

Manually verified that the fix works. There were no automated tests covering this functionality, and it is only deleting code.

### DIFF
--- a/core/indexing/docs/DocsService.ts
+++ b/core/indexing/docs/DocsService.ts
@@ -981,16 +981,10 @@ export default class DocsService {
       // Otherwise sync the index based on config changes
       const oldConfigDocs = oldConfig?.docs || [];
       const newConfigDocs = newConfig.docs || [];
-      const newConfigStartUrls = newConfigDocs.map((doc) => doc.startUrl);
 
       // NOTE since listMetadata filters by embeddings provider id embedding model changes are accounted for here
       const currentlyIndexedDocs = await this.listMetadata();
       const currentStartUrls = currentlyIndexedDocs.map((doc) => doc.startUrl);
-
-      // Anything found in sqlite but not in new config should be deleted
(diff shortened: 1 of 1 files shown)


shssoichiro:
As far as I can tell, the test failure doesn't appear to be related to this change.

shssoichiro:
This issue causes severe CPU utilization every time my IDE is restarted, to the point that I've had to disable Continue. It would be great to have this reviewed please.

RomneyDa (approved):
@shssoichiro nice find!

PR #6164, opened 2025-06-17
Fetch 14: Read node extra ca certs, better NO_PROXY logic

RomneyDa:
Reads certs from NODE_EXTRA_CA_CERTS for custom fetch.

It appears that whenever agent `ca` is manually set, NODE_EXTRA_CA_CERTS is not read, even though the node docs seem to imply that NODE_EXTRA_CA_CERTS are added to tls on initialization. It seems node-fetch injects these under the hood.

Adds a simple certs cache that caches
- tls and globalAgent certs
- certs read from NODE_EXTRA_CA_CERTS
- caBundlePath successful cert contents

This fixes missing NODE_EXTRA_CA_CERTS and speeds up fetch.

Also adds port and case sensitivity issue fixes for NO_PROXY (taken from https://github.com/continuedev/continue/pull/6115)


sestinj (approved):
great tests, change makes sense, verified this solves a problem loading certs for a user

PR #6167, opened 2025-06-17
feat: show warning for unrecommended model selected in agent mode

uinstinct:
## Description

Show a warning below the text box when an unrecommended model is selected for agent mode.

- Model being suitable for agent mode is decided using the [llm-info's `recommendedFor` field](https://github.com/continuedev/continue/blob/0239baa1ca350321ed45d946c22e9b2e4a8a94f7/packages/llm-info/src/providers/openai.ts/#L83)
- A "Know more" link to the issue points to https://docs.continue.dev/agent/model-setup
- Once the warning is dismissed, it is never shown again for any model

related to CON-2301

## Checklist

## Screenshots

https://github.com/user-attachments/assets/3501a176-3128-4457-be00-275b3e28700b

## Tests

### DIFF
--- a/core/config/load.ts
+++ b/core/config/load.ts
@@ -641,6 +641,7 @@ function llmToSerializedModelDescription(llm: ILLM): ModelDescription {
     roles: llm.roles,
     configurationStatus: llm.getConfigurationStatus(),
     apiKeyLocation: llm.apiKeyLocation,
+    recommendedFor: llm.recommendedFor,
   };
 }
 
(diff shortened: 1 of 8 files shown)


RomneyDa (changes requested):
Some issues
- this implementation will only work once - once the user has dismissed for any model, they won't see it again for any model
- recommendedFor is deprecated and `recommendedFor` === "chat" does not imply that the model  can be used for agent mode, it lines up with ModelRoles (`agent` is not a model role)

I think we should change a few things
- Warning should be sticky but more subtle, maybe a small yellow info icon in the mode select dropdown. No need for local storage

RomneyDa:
@uinstinct closing for now because this will conflict with upcoming changes that add virtual tool calling capabilities for all models that doesn't support native tool calling, which will mostly nullify this approach.

PR #6187, opened 2025-06-18
fix: show assistant loading after login

uinstinct:
## Description

After initial load, login via hub and manual config refresh, show loading indicators in the assistant select dropdown and also in account > organizations. This shows that the assistants and account's organizations are being refreshed.

- add a new state in redux's configSlice for showing loading indicators
- put the loading state when auth refreshes profiles and in initial auth and config loading

Cause Analysis: After vscode/jetbrains authentication happens, a [session update is triggered](https://github.com/continuedev/continue/blob/ec32afcc544218d176cca87caa3fd50803b6830d/extensions/vscode/src/extension/VsCodeExtension.ts/#L311) invoking "sessionUpdate" with login details. Also a "configUpdate" [is triggered](https://github.com/continuedev/continue/blob/ec32afcc544218d176cca87caa3fd50803b6830d/core/core.ts/#L728-L730) after retrieving the configs via an API call. This API call was the reason which was silently updating the assistants without any feedback.

resolves CON-1630

## Checklist

## Screenshots

https://github.com/user-attachments/assets/d4527977-0e35-4246-8cc8-a56d6a8d8b3e

https://github.com/user-attachments/assets/4223fb57-f7ed-4852-8014-af2fb22af20e

https://github.com/user-attachments/assets/72bcc999-caa8-44c4-9397-bf99a553b9ec

## Tests


Patrick-Erichsen:
Nice + clean fix!

I think we could take this a bit further and add some loading state to the `ListboxButton` to let users know that we're updating their assistants without needing to click into the select. Maybe instead of the AssistantIcon/selectedProfile.title, we just show the same spinning `ArrowPathIcon` with a "Loading" text?

uinstinct:
> Maybe instead of the AssistantIcon/selectedProfile.title, we just show the same spinning ArrowPathIcon with a "Loading" text?

This approach is good. 

Added it!

https://github.com/user-attachments/assets/b787ca90-908c-46d6-bdce-2b1565274192

PR #6202, opened 2025-06-19
No Barrel Exports

RomneyDa:
- Removes all barrel exports from `gui`
- Adds a `no-barrel-exports` linting rule

Why
Barrel files significantly reduce vitest performance
https://vite.dev/guide/performance.html#avoid-barrel-files
https://www.reddit.com/r/reactjs/comments/1ggpyms/vite_and_barrel_files_performance_issues/\


gui/src/components/mainInput/TipTapEditor/components/index.ts
```diff
@@ -1,3 +0,0 @@
-export * from "./DragOverlay";
```
tomasz-stefaniak, line 1: In a cases like this one, where there aren't too many items being exported, I think it's still ok to keep index.ts. It's just instead of barrel exports we would be using named exports.

The reason I'd lean toward that is that otherwise we may end up importing internal implementation details of components, which can lead to circular dependencies and other unwanted behaviors.

I'm mostly referring to component exports, for pretty much any other file type removing index.ts completely is fair game.

RomneyDa: Good call, updated

tomasz-stefaniak (approved):

Issue #6135, opened 2025-06-15
Add support of Enchained Caching for Anthropic.


md2k:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Anthropic added support for extended caching capabilities, so their Beta feature allow to cache context not only for a 5 minutes but for 1 hour, which is can improve coding experience and chat for big context, because 5 minutes caching not always playing well. 

Official docs: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#1-hour-cache-duration

### Solution

As solution i tried to create simple patch, but unfortunately not able to build extension under my macos m3 due multiple errors.

<details>
<summary>Patch content under the cut</summary>

```diff
diff --git a/core/index.d.ts b/core/index.d.ts
index ce78a8d82..a6cf6d3b2 100644
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -926,6 +926,8 @@ export interface RequestOptions {
 export interface CacheBehavior {
   cacheSystemMessage?: boolean;
   cacheConversation?: boolean;
+  useExtendedCacheTtlBeta?: boolean;
+  cacheTtl?: string;
 }

 export interface ClientCertificateOptions {
diff --git a/core/llm/llms/Anthropic.ts b/core/llm/llms/Anthropic.ts
index 8d1d3a67f..759157d87 100644
--- a/core/llm/llms/Anthropic.ts
+++ b/core/llm/llms/Anthropic.ts
@@ -100,7 +100,11 @@ class Anthropic extends BaseLLM {
           {
             type: "text",
             text: message.content,
-            ...(addCaching ? { cache_control: { type: "ephemeral" } } : {}),
+            ...(addCaching
+              ? { cache_control: this.cacheBehavior?.useExtendedCacheTtlBeta
+                ? { type: "ephemeral", ttl: this.cacheBehavior?.cacheTtl ?? "5m" }
+                : { type: "ephemeral" } }
+              : {}),
           },
         ],
       };
@@ -115,7 +119,9 @@ class Anthropic extends BaseLLM {
             ...part,
             // If multiple text parts, only add cache_control to the last one
             ...(addCaching && contentIdx === message.content.length - 1
[...]

PR #6205, opened 2025-06-19
Feat/anthropic extended ttl

md2k:
## Description

Implements granular per-message-type caching for Anthropic models to improve token efficiency in Agent mode. Adds new `CacheBehavior` options to specify how many of each message type to cache (user messages, tool results, assistant tool calls, etc.) instead of only caching the last 2 user messages. 
This is related to issue https://github.com/continuedev/continue/issues/6135

## Checklist

## Screenshots

N/A - Backend caching enhancement with no visual changes.

## Tests

Added comprehensive test suite `core/llm/llms/Anthropic.enhanced-caching.test.ts` with 6 test cases covering:

- Tool result message caching
- Assistant tool call message caching  
- Per-type caching limits validation
- Disabled caching behavior
- Fallback TTL handling
- Core `shouldCacheMessage` logic

All tests pass and validate the new per-type caching functionality while maintaining backward compatibility.

image

### DIFF
--- a/core/llm/llms/Anthropic.ts
+++ b/core/llm/llms/Anthropic.ts
@@ -8,7 +8,7 @@ class Anthropic extends BaseLLM {
   static providerName = "anthropic";
   static defaultOptions: Partial<LLMOptions> = {
     model: "claude-3-5-sonnet-latest",
-    contextLength: 200000,
+    contextLength: 200_000,
     completionOptions: {
       model: "claude-3-5-sonnet-latest",
       maxTokens: 8192,


md2k:

md2k:
Some details about how long session with big context looks with 5min cache and 1h cache and cost perspective:
5 min cache:
image
image

md2k:
1h ttl
image
image

docs/docs/customize/model-providers/top-level/anthropic.mdx
```diff
@@ -62,18 +62,40 @@ Anthropic currently does not offer any reranking models.
 
 Anthropic supports [prompt caching with Claude](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching), which allows Claude models to cache system messages and conversation history between requests to improve performance and reduce costs.
 
+> **NOTE:** As part of their `Beta` support [Extended caching](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#1-hour-cache-duration)
```
sestinj: The docs here feel a bit extensive since they take up most of this page now. I think we should try to make a collapsible block _or_ make a dedicated page to prompt caching. If possible, I think the collapsible would be the better option

core/index.d.ts
```diff
@@ -927,6 +927,13 @@ export interface RequestOptions {
 export interface CacheBehavior {
   cacheSystemMessage?: boolean;
   cacheConversation?: boolean;
+  useExtendedCacheTtlBeta?: boolean;
```
sestinj, line 932: I'm coming at this review with the lens of "if we add it now, we'll have to support it forever (or go through a deliberate deprecation process)". I'm worried there are a large number of options here that aren't going to be relevant forever, or that they might not be the final form of this configuration.

It would be helpful to better understand whether all of the cacheUserMessages, cacheAssistantMessages, etc. are truly necessary for people to customize, or whether we just need to set a more sensible default. For example, I'd be curious what values you set here and whether you think we should just ship those as the defaults for everyone. Usage patterns in Continue are probably similar across a variety of users. Not that we couldn't eventually also allow this customization, but it might save a lot of maintenance (and give many users money back without needing to configure anything)

sestinj (commented):
This is a great PR as far as the code goes. I kind of want to step back though to better understand whether you think this could be a sensible default rather than a configuration option. I'm weary of too many options and if everyone would benefit from the way you are configuring your Anthropic models, maybe we should just ship that as the default (I repeated all this in a comment below)

chezsmithy:
@sestinj maybe we align to this previous PR. https://github.com/continuedev/continue/pull/5371

It introduced a single caching setting that controls all the options. Whatever we do here I should likely bring to Bedrock as well.

Watching.

sestinj:
Agreed @chezsmithy ! Thanks for linking the PR here, that what I had in mind

docs/docs/customize/model-providers/top-level/anthropic.mdx
md2k: Hi @sestinj , no problem i can re-format mdx doc file, i just initially tried to not to break original style of documentation and preserve as it is.

core/index.d.ts
md2k: as for me, i think we just need to cache everything we can to cache, since it is saves users money. but main problem with Anthropic, that 5m cache and 1h cache cost differently and 1h cache cost double [pricing](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#pricing). 

I think about 3 possible routes/options:
1. Cache everything and give user minimal control 5m/1h cache (while inside code if user selecting 1h, we automatically adding headers)
2. Give user to control what he want to cache (but from my experience more cached - less expensive when i need to work a lot within same session with huge context) and do not want keep me in 5m time frame.
3. maybe implement custom provider configuration block, which are extending basic configuration per AI provider if necessary. 

i think caching last 2 messages is more than enough default for any type, since cache LRU based (if i got Anthropic docs right)

md2k:
@chezsmithy  thanks for link, i will take a look and will re-align my PR

md2k:
Due some time constraint this week i have, will work thru PR closer to the end of week or over weekend.

sestinj:
Sounds good!

md2k:
@chezsmithy i looked into PR for AWS Bedrock, at this moment its support limited to 5min cache, what `cachePoint: { type: "default" }`, and you already have implemented system, user messages and tools. the only improvements i can see here, is to maybe extend internal logic to ensure we covering all types of messages similar to my Anthropic PR.  So probably right now Bedrock not require any changes. also i need look into its api because some property names different from Anthropic.

md2k:
@sestinj, i did more simple version with only following settings

```
        "cacheSystemMessage": true, -- original
        "cacheConversation": true,  -- original
        "cacheToolMessages": true, - new
        "useExtendedCacheTtlBeta": true,  -new
        "cacheTtl": "5m|1h" - new
```
the reason, while it still in beta feature i probably keep it as enabler, plus it costs 2x tokens, by this reason I allow to disable tools messages.
I going to work with this patch myself couple of days, to understand how it feels with some settings enabled/disabled and token/$$ consumption.

But if you have any suggestions, please, let me know.

md2k:

md2k:
found some issues in caching logic where we can hit maximum 4 slots allowed to be cached by Anthropic, so implementing some logic to ensure priority to what cache settings will be applied

md2k:
while testing logic, i was able to achieve pretty nice results
image
520k and 620k input/cached tokens... unfortunately i screwed logic and burned around 400k input (no cache at all) and invalidated cache for 300k tokens. so real input/cache writes is around 200k/200k +- and 6.6mil cache read hits.

image
 
And costs (supposed to be cheaper, but due small mistake i burned for nothing close to 4$ due broken logic :D ):
image

md2k:
new logic not committed yet. going to test it a bit more. Also added some debug to console. curious if possible to keep this debug or maybe make some option for stats (can be handy to see)

image

md2k:
I found my problem 😀 I made it too complicated. I will try rolling back to my previous initial code and simplifying it. I spent a couple of hours reading through all of Anthropic's documentation and examples to understand how their breakpoints work and where and how to use them. Things were much easier than I thought. So, hopefully, I will have the final code in a few days. Then, we can benefit from a close to 90-95% cache hit rate.

md2k:
@sestinj  i tested current implementation and it looks good. so user can enable/disable tools/system/conversation cache breakpoints, and control 5 min/1 hour TTL when enabling Beta feature. Also added cacheDebug, it helps alot to understand what going on inside and how things cached and its efficiency.
If you do not want to keep debug, i can remove it. 
And i asking for help because i can't resolve mdx conflict :(

md2k:
i found also problem with CLA, looks like it will be more cleaner if i re-do entirely PR and apply my changes to latest main branch instead of trying to solve the puzzles with CLA and merge conflicts.

md2k:
Seems for best efficient cost/cache ratio, Apply/Edit models better to disable cacheConversations, because apply/edit does not include history. while it make sense, i not sure why rules / system messages stripped out as well (i would prefer to include them to ensure that during edit/apply AI follow the same guidelines as during conversation), but the main point need to disable conersation caching for Edit/Apply to not waste $$$ and pay just for vanilla costs, with caching single input message we literally wasting x1.25 or x2.0 depends on cache type which written to cache but never used after since each edit/apply a new mini-session. 

image


In any case i think i did the best from what i can to achieve best caching logic and support Anthropic extended ttl. It works perfect. Small issue with 1h ttl that it has strictly limited TTL and not using LRU style, so its invalidation will always happen after 1hour. It still good option anyway. I have idea maybe to implement sort of "ping" where run timer after last message and automatically send context +ping which will cost few tokens only but will refresh 5min ttl. The problem here it needs more testing and investigation about how long we can keep 5min TTL using ping approach and also have some fail-safe feature to not abuse anthropic functionality where user can forgot to close session and ping can stay for hours and days. So this is only idea.

Also code tested against current main branch with v1.1.56 of plugin. unfortunately i cant do CLA from another account, so maybe it easier to create new PR with rebased from current main and without leftish committers :D

md2k:
Hi @sestinj , can you, please, take a look and let me know your thoughts ?

sestinj (changes requested):
@md2k I'm sorry for the slow feedback here. The thing that was making me nervous is that this is on the critical path of every request and we just didn't yet have good abstractions set up to make it testable / swappable. We now have this in `packages/openai-adapters/src/apis/AnthropicCachingStrategies.ts`. I'd love to merge this, but would you be willing to move everything into a new strategy in that file so that we can separately test / swap out to make this safer?

md2k:
No problem, i going to rebase my branch to have latest code from main and align with new style.

md2k:
Hi @sestinj ,  I'm a bit confused, since in `core/llm/llms/Anthropic.ts` I still can see some portion of old code and adding caching to some messages as it was in old code, some reference about how new code used will be very helpful (i can see you added caching part at f49e62ce7ad6fa0e0a5452efb1617816c039fd59).

sestinj:
We will be switching entirely away from the `ILLM` classes and toward the `openai-adapters` soon. Right now there is an `if` statement in `streamChat` of `BaseLLM` that checks for `shouldUseOpenAiAdapter` and if true it uses that. Eventually we will switch these all to use openai-adapters.

sestinj:
I feel this would probably be better opened as a fresh PR at this point since there is a lot of messy history. We might benefit from adding the relevant charts directly to a new PR description and I think we'll review faster that way. I'm going to close but let me know if you feel otherwise!

PR #6227, opened 2025-06-20
Fix azure opena ai query params

RomneyDa:
## Description
The openAI constructor takes baseUrl and query params separately
We were passing query params in baseUrl, which led to Urls like:
`https://azure-example/openai/deployments/example-deployment?api-version=2024-12-01-preview/chat/completions"`

This fixes that by sending the params to `defaultQuery`


Patrick-Erichsen (approved):
nice find 👌

PR #6228, opened 2025-06-20
New .continue/rules

sestinj:
## Description

New rules for core protocol and avoiding being over-eager

## Checklist

### DIFF
--- a/.continue/rules/new-protocol-message.md
+++ b/.continue/rules/new-protocol-message.md
@@ -1,0 +1,35 @@
+---
+name: New protocol message
+description: Create a new protocol message from core/gui/ide to core/gui/ide
+---
+
+Create a new protocol message by taking the following steps:
+
+## 1. Define the message type
+
+Add your new message type definition to the appropriate file in the `protocol/` directory with correct TypeScript typing.
+
+## 2. Check for duplicates
+
+Verify that no existing message type already provides the same functionality.
+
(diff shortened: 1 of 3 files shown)


.continue/rules/new-protocol-message.md
RomneyDa, line 1: I think multiple-header lists are higher risk of confusion for concatenated rules. Single header with numbered/nested list might make more sense

This rule sounds a bit too much like a prompt, e.g. 
>Create a new protocol message by taking the following steps:

 I feel like it would often leave to confusion

.continue/rules/overeager.md
RomneyDa, line 1: Let's consider adding some version of this directly into the default agent prompt

core/rules.md
```diff
@@ -0,0 +1,10 @@
+# `core` rules
+
+Whenever a new protocol message is added to the `protocol/` directory, check the following:
+
+- It's type is defined correctly
```
RomneyDa, line 5: *Its
This feels a bit vague, maybe "Its message body type is defined correctly" or similar

core/rules.md
RomneyDa, line 1: We should reevaluate colocated rule approaches, they are currently included in every request unless globs etc. is set. You would need to add globs

RomneyDa (approved):

PR #6236, opened 2025-06-21
add tool call support for deepseek-chat、deepseek-reasoner(Only for deepsee…

xiaohuanxiong3:
…k official API )

## Description

[ add tool call support for deepseek official api . ]

## Checklist

## Screenshots

## Tests


xiaohuanxiong3:

core/llm/autodetect.ts
```diff
@@ -293,7 +292,8 @@ function autodetectTemplateFunction(
       zephyr: zephyrTemplateMessages,
       anthropic: anthropicTemplateMessages,
       chatml: chatmlTemplateMessages,
-      deepseek: deepseekTemplateMessages,
+      // deepseek: deepseekTemplateMessages,
+      deepseek: null,
```
Patrick-Erichsen: What's the reason for this?

core/llm/llms/Deepseek.ts
```diff
@@ -16,6 +23,8 @@ class Deepseek extends OpenAI {
   };
   maxStopWords: number | undefined = 16;
 
+  protected useOpenAIAdapterFor: (LlmApiRequestType | "*")[] = [];
+
```
Patrick-Erichsen: Do we need this?

core/llm/llms/Deepseek.ts
```diff
@@ -52,6 +61,110 @@ class Deepseek extends OpenAI {
...
+      }
+    }
+
+    for await (const value of streamSse(response)) {
+      const chunk = fromChatCompletionChunk(value);
+      if (chunk) {
+        yield chunk;
+      }
+    }
+  }
```
Patrick-Erichsen: Is there a reason we need this when the model is already OpenAI compatible? I believe `deepseek-reasoner` was working fine without this

Patrick-Erichsen (commented):
Thanks for opening the PR here @xiaohuanxiong3 ! Left a comments on some changes I wasn't sure about.

core/llm/autodetect.ts
xiaohuanxiong3: The purpose of this line of code is to prevent the program from entering the following code branch in `BaseLLM.*streamChat`

```ts
if (this.templateMessages) {
  for await (const chunk of this._streamComplte(
    prompt,
    signal,
    completionOptions,
  )) {
    completion += chunk;
    interaction?.logItem({
      kind: "chunk",
      chunk: chunk,
    });
    yield { role: "assistant", content: chunk };
  }
}
```

After the program enters this branch, the tool call information will never be returned, and because the deepseek model will stop responding after the tool call ends, the chat will suddenly end in the user interface.

Maybe I should modify the `autodetectTemplateType()` function in `autodetect.ts` instead

core/llm/llms/Deepseek.ts
xiaohuanxiong3: Sorry, this is my question, I didn't expect 'openaiAdapter' to implement a similar logic to change-3, so change is not needed here

core/llm/llms/Deepseek.ts
xiaohuanxiong3: My apologies, this was my mistake. I didn't realize that `openaiAdapter` already implemented the similar logic, so this change here is also unnecessary. However, because the `fromChatCompletionChunk` function in `openaiTypeConverters.ts` doesn't handle the potential `ThinkingChatMessage` logic, you're right that `deepseek-reasoner` can work fine without the above changes, but the thinking process of `deepseek-reasoner` won't be displayed in the chat interface.

xiaohuanxiong3: The `ChatCompletionChunk` returned by the `openaiAdapter.chatCompletionStream` function does not have the `reasoning_content` field corresponding to the thought process returned by `deepseek-reasoner`. Therefore, if it is not necessary to display the thought process of `deepseek-reasoner` in the user interface, change-2 and change-3 are unnecessary. If it is necessary to display the thought process of `deepseek-reasoner`, I looked at the return type `ChatCompletionChunk` of the `*chatCompletionStream` function in `@continuedev/openai-adapters` and it does not have the `reasoning_content` field, so besides adding change-2 and change-3, I don't have any other solutions for the time being.

core/llm/autodetect.ts
Patrick-Erichsen: > Maybe I should modify the autodetectTemplateType() function in autodetect.ts instead
I think that seems like the right implementation 👍

Patrick-Erichsen:
@xiaohuanxiong3 seems like you rolled back most of the changes, lmk if/when it's ready for another review 👍

xiaohuanxiong3:
it's ready

core/llm/autodetect.ts
JunYang-tes: Sorry to ask a question unrelated to this topic: Why is _streamComplete used instead of _streamChat when templateMessages are involved? If _streamComplete is used, all models with templateMessages—even those that support tools—will fail in tool calls due to the loss of toolCall.

core/llm/toolSupport.ts
```diff
@@ -179,7 +179,11 @@ export const PROVIDER_TOOL_SUPPORT: Record<string, (model: string) => boolean> =
       return false;
     },
     deepseek: (model) => {
-      if (model !== "deepseek-reasoner") {
+      if (
+        model === "deepseek-reasoner" ||
+        model === "deepseek-chat" ||
+        model === "deepseek-coder"
+      ) {
```
Patrick-Erichsen: @xiaohuanxiong3 I took a closer look at the other changes in your PR, this is the only part that is required to allow tool calling for the additional DeepSeek models. So all the other deletions of `"deepseek"` from `autodetect.ts`, `index.d.ts` and `models.ts` should be reverted.

Lastly could you provide a link to the DeepSeek docs to verify that the chat and coder models do in fact support tool calling? You'll notice the other providers in this file have a comment linking back to the tool calling docs for the given provider.

Patrick-Erichsen:
Thanks for the quick fixes! I didn't see any info on model support for tool calling on that doc page but I do here: https://api-docs.deepseek.com/quick_start/pricing

However I don't see any info on `deepseek-coder` either on that page, or anywhere else on the internet. Can you confirm you've tried this PR our with `deepseek-coder` locally and that it was supported?

If so this is good to merge 👍

xiaohuanxiong3:
In DeepSeek-V2.5, deepseek merged the DeepSeek-V2-Chat and DeepSeek-Coder-V2 models. At the same time, the API interface is forward compatible, and users can access the new model through deepseek-coder or deepseek-chat. As for the latest DeepSeek-V3, although deepseek does not say in the document that the latest model can be accessed through deepseek-coder, I have no problem specifying the model as deepseek-coder in the local configuration file. I am not sure whether deepseek-coder will be supported in the future, so I will remove it.

Patrick-Erichsen (approved):

Patrick-Erichsen:
This is ready to merge but trying to get one of the new e2e tests to pass that is failing for some reason unrelated to these changes

PR #6247, opened 2025-06-22
feat: create `AssistantAndOrgListbox`

Patrick-Erichsen:
## Description

Most of the file changes here are due to me creating a new `AssistantAndOrgListbox` component that breaks out all the subcomponents from the old `AssistantSelect`. No new logic, just UI changes.

Closes CON-2156

## Signed in
Screenshot 2025-06-22 at 11 05 04 AM

## Signed out

Screenshot 2025-06-22 at 11 39 44 AM


sestinj (changes requested):
Couple of changes. Here, we should hide the "Organization" title and the dropdown if the user only has the Personal org. Instead, we can show either "Signed in as <email>" or "Signed out".

Screenshot 2025-06-23 at 10 22 07 AM

And then I think we should add icons to the "New Assistant" and "Sign in" buttons (+ and box with arrow or person icons potentially)

Patrick-Erichsen:
Here's what will show if the following is true. I decided not to add the "Signed out" since I felt it was clear from the "Log in" option that we already render.

```ts
 const shouldRenderOrgInfo =
    session && organizations.length > 1 && !isOnPremSession(session);
```

Screenshot 2025-06-23 at 11 15 27 AM

sestinj (approved):

PR #6248, opened 2025-06-22
fix: use last 2 parts of file uri for rule names

Patrick-Erichsen:
## Description

Current behavior is that the entire file URI is shown as rule name, which is more/less useless since we truncate the display, and the beginning of all file URIs for a workspace will always be the same.

This PR changes behavior to instead be the last 2 parts of the URI. I think ideally we'd somehow keep the full URI and let the callee decide how to render the name based on this, but for now this is 1% better.


sestinj (approved):
makes sense, good tests

PR #6263, opened 2025-06-23
Applied rules for tool messages

RomneyDa:
Rearranges our messaging thunks a bit so that construct messages is only called in one place and applied rules are added to tool messages as well


sestinj (changes requested):
I think that we should start now with writing tests that cover the thunk logic. Probably vitest "integration" tests would be better than a unit test of all the individual thunks. Kent D. Dodds (seems important) seems to agree: https://redux.js.org/usage/writing-tests#guiding-principles

But just something that checks the state before / after is what seems important

PR #6233, opened 2025-06-21
feat: continue init prompt

tomasz-stefaniak:
## Description

Added `/init` command to initialize Continue in a new codebase.

## Checklist

- [] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [] The relevant docs, if any, have been updated or created
- [] The relevant tests, if any, have been updated or created

## Screenshots

[ For visual changes, include screenshots. Screen recordings are particularly helpful, and appreciated! ]

## Tests

[ What tests were added or updated to ensure the changes work as expected? ]

Issue #5771, opened 2025-05-21
MCP tools are showing but are not called


maximelebastard:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: 1.1.32
- IDE version: 1.100.2
- Model: any
- config: 
  
%YAML 1.1
---
version: 1.0.16
schema: v1

bedrock: &bedrock
  provider: bedrock
  env:
    region: eu-central-1
    profile: bedrock

models:
  - name: (aws) Anthropic Claude 3.5 Sonnet
    <<: *bedrock
    model: anthropic.claude-3-5-sonnet-20240620-v1:0
    roles:
      - chat
      - edit
      - apply
  - name: (aws) Anthropic Claude 3.7 Sonnet
    <<: *bedrock
    model: eu.anthropic.claude-3-7-sonnet-20250219-v1:0
    roles:
      - chat
      - edit
  - name: (aws) Anthropic Claude 3 Haiku
    <<: *bedrock
    model: anthropic.claude-3-haiku-20240307-v1:0
    roles:
      - chat
      - autocomplete

  - name: (aws) Cohere Rerank
    <<: *bedrock
    model: cohere.rerank-v3-5:0
    roles:
      - rerank

  - name: (aws) Cohere Embed Multilingual
    <<: *bedrock
    model: cohere.embed-multilingual-v3
    roles:
      - embed
      
  - name: (local) Qwen 2.5 autocomplete
    provider: ollama
    model: qwen2.5-coder:3b
    roles:
      - chat

context:
  - provider: file
  - provider: code
  - provider: open
  - provider: folder
  - provider: clipboard
  - provider: diff
  - provider: terminal
  - provider: docs
  - provider: web
    params:
      n: 5
  - provider: codebase
  - provider: tree
  - provider: url

prompts:
  - name: commit
    description: Create a commit message from staged files
    prompt: "{{input}}"

docs:
  - name: Symfony
    startUrl: https://symfony.com/doc/6.4/index.html

defaultCompletionOptions:
  maxTokens: 4096


mcp servers config: 
[...]

Issue #4891, opened 2025-03-28
Support context providers in prompt files


sestinj:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Right now we support @ mentions for context providers in prompt files, but not in config.yaml. This should be possible in both. For example, `@diff` should cause the diff to be included in the prompt.

Relatedly, there is a bug stopping this from working in .prompt files it seems. These are likely to be solved together

juanfcocontreras:
I think this is resolved in the latest beta versions, at least it seems to work for me.

RomneyDa:
Further expanded in https://github.com/continuedev/continue/pull/6264

RomneyDa:
Fixed by #6264 for non-query providers like diff

lilyTokens:
This is one of the better discussions I have seen on context window management in agentic coding tools. I have been building a freelance scope management tool where the agent needs to hold a consistent model of what was agreed at project kickoff across many sessions. The approach of keeping a structured memory graph (rather than raw conversation history) has been significantly more reliable for this than window-filling approaches.

PR #6264, opened 2025-06-23
Support tools for prompts

### LINKED ISSUE #5762: Using prompt from MCP server doesn't trigger tool calls on the first message

### Relevant environment info

```Markdown
- OS:
- Continue version:
- IDE version:
- Model:
- config:
  

  
  OR link to assistant in Continue hub:
```

### Description

I'm trying to use a prompt loaded in an MCP server to execute steps.

The MCP prompt is used with the "/" and is loaded correctly, but once the tool calls are received from the model, continue doesn't execute them. Adding a second message allows executing the process, but I want the process to run automatically without additional user intervention.

If the prompt message is "written" manually, the tool calls are executed immediately

Image

Image

I'm using the example MCP server provided here

https://github.com/continuedev/continue/pull/5517#issuecomment-2889471089

```python
@mcp.prompt()
def debug_routine() -> str:
    return """
Execute the following steps

1. Call terminal tool and execute a ls command
2. Call terminal tool and execute a git status
"""
```

### To reproduce

Tested with 1.1.37 (pre-release) and 1.0.7, 1.0.11

### Log output

```Shell

```

RomneyDa:
## Description
The problem: Slash commands are built with generators that output strings. Thus, they don't support tool calls.

This PR restructures slash commands to add tool support for prompts.

TL;DR: non-legacy slash commands (prompt blocks, mcp, etc.) are now rendered on the GUI side and included in messages as normal conversations

Fixes #5762, #5771, #4891
Unblocks #6233 

Details
- Adds a `SlashCommandWithSource` type (similar idea to `RuleWithSource` that adds source traceability and an `isLegacy` flag to slash commands, without breaking e.g. config.ts SlashCommand type
- Removes the `run` functions from non-legacy slash commands
- Updates editor content resolution on the GUI to render and inject slash command prompts on the spot. This includes 
    - adding an `mcp/getPrompt` message that hits core and retrieves/stringifies an mcp prompt
    - moving prompt rendering logic to the GUI
- Moves all legacy slash commands e.g. `/share`, `/commit`, etc. to a new more isolated directory
- Adds tests for tip tap editor processing, prompt file rendering, etc.
- Fixes various prompt file bugs, including 
    - nested prompt files broken because of URI reasons
    - adds support for simple no-query context providers in v2 prompt files


core/index.d.ts
```diff
@@ -838,15 +838,44 @@ export interface ContinueSDK {
...
 }
 
+export interface SlashCommandWithSource extends SlashCommandDescription {
+  run?: (sdk: ContinueSDK) => AsyncGenerator<string | undefined>; // Optional - only needed for legacy
+  source: SlashCommandSource;
+  promptFile?: string;
+  overrideSystemMessage?: string;
+}
+
+export type SlashCommandSource =
```
sestinj, line 860: this is embarrassing...

core/llm/streamChat.ts
```diff
@@ -59,6 +59,9 @@ export async function* llmStreamChat(
       },
       true,
     );
+    if (!slashCommand.run) {
+      throw new Error(`Slash command ${command.name} is not a legacy command`);
```
sestinj: god forbid this error ever becomes visible to the user

I don't think it should be possible, but could we pretend that we expect that and write an error message that makes more sense to them? e.g. "Slash command not found"

sestinj (changes requested):

sestinj (approved):

PR #6276, opened 2025-06-24
fix: skip ripgrep if workspace is remote

uinstinct:
## Description

When workspace is remote, we want to avoid running ripgrep. This is because ripgrep will run locally instead of running the workspace server.

- when workspace is remote, use `baseToolDefinitions` excluding `grepSearchTool`
- add a new type `IDE` interface `isWorkspaceRemote` to detect if workspace is remote in vscode and intellij

resolves CON-2298

If it is possible to run os remote call from remote installed extension, we should be able to implement grep search for remote ides. corresponding issue in vscode: https://github.com/microsoft/vscode/issues/252269

## Checklist

## Screenshots

https://github.com/user-attachments/assets/f54872c8-78e1-4305-bd0a-ad8f411dd94b

https://github.com/user-attachments/assets/95eb01c5-4015-4129-bc52-0fa959dd5f69

## Tests

## Replication

You can run a docker container for ssh into a remote host.
You can this dockerfile: https://gist.github.com/uinstinct/314da2b450f5440069bbce5ab14043d2

### DIFF
--- a/core/tools/index.ts
+++ b/core/tools/index.ts
@@ -13,6 +13,7 @@ import { runTerminalCommandTool } from "./definitions/runTerminalCommand";
 import { searchWebTool } from "./definitions/searchWeb";
 import { viewDiffTool } from "./definitions/viewDiff";
 
+// missing support for remote os calls: https://github.com/microsoft/vscode/issues/252269
 export const localOnlyToolDefinitions = [grepSearchTool];
 
 export const baseToolDefinitions = [
--- a/core/tools/index.ts
+++ b/core/tools/index.ts
@@ -13,12 +13,11 @@ import { runTerminalCommandTool } from "./definitions/runTerminalCommand";
 import { searchWebTool } from "./definitions/searchWeb";
 import { viewDiffTool } from "./definitions/viewDiff";
 
-export const baseToolDefinitions = [
+export const remoteToolDefinitions = [
   readFileTool,
   editFileTool,
   createNewFileTool,
   runTerminalCommandTool,
-  grepSearchTool,
   globSearchTool,
   searchWebTool,
   viewDiffTool,
@@ -30,6 +29,8 @@ export const baseToolDefinitions = [
   // viewSubdirectoryTool,
   // viewRepoMapTool,
 ];
+
+export const baseToolDefinitions = [...remoteToolDefinitions, grepSearchTool];
 
 export const getConfigDependentToolDefinitions = (
   params: ConfigDependentToolParams,
--- a/core/tools/index.ts
+++ b/core/tools/index.ts
@@ -1,4 +1,4 @@
-import { ConfigDependentToolParams, Tool } from "..";
+import { ConfigDependentToolParams, IDE, Tool } from "..";
 import { createNewFileTool } from "./definitions/createNewFile";
 import { createRuleBlock } from "./definitions/createRuleBlock";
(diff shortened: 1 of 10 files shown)


core/tools/index.ts
```diff
@@ -31,6 +30,8 @@ export const baseToolDefinitions = [
   // viewRepoMapTool,
 ];
 
+export const baseToolDefinitions = [...remoteToolDefinitions, grepSearchTool];
+
```
Patrick-Erichsen: I think it would be a little more clear if we inverted the logic here/in the `await ide.isWorkspaceRemote()` call. Renaming `baseToolDefinitions` to `remoteToolDefinitions` forces us to always think in terms of remote or not which is more of an edge case

```ts
export const localOnlyToolDefinitions = [grepSearchTool]
export const baseToolDefinitions = [ ... ] // unchanged from before

...

 const continueConfig: ContinueConfig = {
    slashCommands: [],
    tools: [
      ...((await ide.isWorkspaceRemote())
        ? baseToolDefinitions
        : [...baseToolDefinitions, ...localOnlyToolDefinitions),
    ],
```

Patrick-Erichsen: Also, if it's possible to move the `(await ide.isWorkspaceRemote())` ternary into a standalone fn in our `core/tools` folder I think that would be nice to keep the logic centralized. I anticipate that it will continue to get more complicated over time.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
```diff
@@ -125,6 +125,10 @@ class IntelliJIDE(
         return true
     }
 
+    override suspend fun isWorkspaceRemote(): Boolean {
+        return this.getIdeInfo().remoteName != "local"
```
Patrick-Erichsen, line 129: I didn't now we had `remoteName` logic already built out, neat

core/tools/index.ts
Patrick-Erichsen: One other thought 😅  let's throw a link to the upstream issue you opened on the vscode repo in a comment on the `localOnlyToolDefinitions` object. Then we can hopefully follow up and remove it in the future.

Patrick-Erichsen (changes requested):
Great work, I'm going to save that dockerfile Gist for easier testing in the future! I've been spinning up an EC2 instance for SSH tests but that was way easier.

This worked for me, I just had a nitpick on some of the logic and then we should be good to merge here.

Patrick-Erichsen:
Hmm another thought, maybe we put your Gist into the `scripts` folder? That would be nice to have on hand.

core/tools/index.ts
uinstinct: makes sense. refactored to [`getToolsForIde`](https://github.com/continuedev/continue/blob/8d90c76e6558fef9148cd196130d41e83c78719f/core/tools/index.ts/#L39-L42) inside core/tools/index.ts

uinstinct:
> Hmm another thought, maybe we put your Gist into the `scripts` folder? That would be nice to have on hand.

yes agree. made a few changes and [added it as a new script](https://github.com/continuedev/continue/blob/7df9c92d39bc68e83b0ddd2510364dfcf3ccf53a/scripts/create-docker-ssh-container.sh)

PR #6281, opened 2025-06-24
Standardize codeblock formatting

RomneyDa:
Description
- Standardizes codeblock fences across context providers to have ```lang /path/to/file (range if applicable) using new `formatCodeblock` utility
- Cleans up ambiguity in apply/autocomplete code around file URI vs path vs extension
- Significantly speeds up problems context provider by eliminating duplicate file reads


sestinj:
Seems fine, but failing tests so will wait to do further review

PR #6288, opened 2025-06-25
Fetch agent pooling

RomneyDa:
Pools agents based on proxy + request options + protocol key


RomneyDa:
Not certain will be used, closing for now

PR #6303, opened 2025-06-26
improve: instruct model when no context results were retrieved

uinstinct:
## Description

When no results were retrieved during context embeddings, instruct model to use other tools.

Resolves CON-2400

## Checklist

## Screenshots

https://github.com/user-attachments/assets/633649fb-3721-4058-be9a-b31b1c3767b6

https://github.com/user-attachments/assets/2e4e56d3-15b9-4ad0-b75c-34fef30a9018

## Replication

1. Launch the extension
2. Add a new .continueignore in the manual testing sandbox which excludes all the files
<details>
<summary>.continueignore</summary>
```
.continueignore
AdvancedPage.tsx
calculator_test
Calculator.java
config.yaml
data.json
Dockerfile
example.ipynb
nested-folder
program.cs
query.sql
readme.md
requirements.txt
test.css
test.csv
test.html
test.js
test.kt
test.php
test.py
test.rb
test.rs
test.sh
test.ts
```
</details>

3. reindex the codebase
4. prompt the model using `@codebase`


core/context/retrieval/retrieval.ts
```diff
@@ -79,16 +79,18 @@ export async function retrieveContextItemsFromEmbeddings(
...
       void extras.ide.showToast(
         "warning",
-        "No results found. If you think this is an error, re-index your codebase.",
+        "No results found (Indexing disabled).",
       );
-      // TODO - add "re-index" option to warning message which clears and reindexes codebase
+      return [];
     }
-    return [];
+    return [
```
sestinj: We shouldn't send this message when the user in in chat mode, because the model won't have access to any tools.

I realize we might not have a way of knowing this in `core` yet, so potentially not a trivial change. But seems important

uinstinct: core is stateless right?
if so, we might need to pass a parameter when getting context items or [add a protocol in IDE](https://github.com/continuedev/continue/blob/78426cc224a4a34ef8f46fdb4c4e4be9adfa2871/core/protocol/ide.ts/#L23)

sestinj: Yeah that's correct. I feel like passing it as a parameter would be the easier option

sestinj:
@uinstinct it looks like there are real failing tests here

sestinj (approved):
thanks for the updates, looks good

PR #6299, opened 2025-06-26
Dev data: added tool usage event

adarshramiyer:
## Description

Added tool usage dev data event with appropriate logger invocation.

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [x] The relevant docs, if any, have been updated or created
- [x] The relevant tests, if any, have been updated or created

PR #6315, opened 2025-06-26
toolUsage devdata events

adarshramiyer:
## Description

Added dev data for tool usage. This was previously introduced (incorrectly) in #6299 and reverted in #6314 .

## Checklist

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -809,7 +809,7 @@ export class Core {
       this.messenger.send("toolCallPartialOutput", params);
     };
 
-    return await callTool(tool, toolCall, {
+    const result = await callTool(tool, toolCall, {
       config,
       ide: this.ide,
       llm: config.selectedModelByRole.chat,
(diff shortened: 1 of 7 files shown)


RomneyDa (commented):
We know when the too

gui/src/redux/thunks/callCurrentTool.ts
```diff
@@ -28,6 +29,21 @@ export const callCurrentTool = createAsyncThunk<void, undefined, ThunkApiType>(
       return;
     }
 
+    const logToolUsage = (success: boolean, finalOutput?: ContextItem[]) => {
```
RomneyDa: Let's split this into a util outside the thunk

core/core.ts
```diff
@@ -820,6 +820,14 @@ export class Core {
       onPartialOutput,
       codeBaseIndexer: this.codeBaseIndexer,
     });
+
+    this.messenger.send("toolCallCompleted", {
```
RomneyDa: We can eliminate this toolCallCompleted message, since `callTool` thunk will work for both client and core tools

RomneyDa (changes requested):
We have the

PR #6316, opened 2025-06-26
Pe/search and replace experiment

Patrick-Erichsen:
## Description

## Checklist

## Screenshots

## Tests


core/tools/index.ts
```diff
@@ -28,14 +30,16 @@ export const baseToolDefinitions = [
...
-  // viewRepoMapTool,
 ];
 
 export const getConfigDependentToolDefinitions = (
   params: ConfigDependentToolParams,
-): Tool[] => [requestRuleTool(params)];
+): Tool[] => [
+  requestRuleTool(params),
+  ...(params.enableExperimentalTools
+    ? [searchAndReplaceInFileTool, viewRepoMapTool, viewSubdirectoryTool]
```
RomneyDa, line 40: Could be a nice touch but maybe too clever, maybe search and replace shouldn't replace edit since they do different things. I guess naming could be confusing to model? Not critical

gui/src/util/clientTools/searchReplaceImpl.ts
```diff
@@ -0,0 +1,78 @@
...
+
+      // Apply the replacement
+      currentContent =
+        currentContent.substring(0, match.startIndex) +
+        (replaceContent || "") +
+        currentContent.substring(match.endIndex);
+    }
+
+    // Single applyToFile call with all accumulated changes
+    await extras.ideMessenger.request("applyToFile", {
```
RomneyDa, line 62: I think we shouldn't use apply to file here because it adds an apply call which is a huge speed drag and surface area increase when we just need to stream known changes in

RomneyDa (changes requested):
Awesome tests.
Wanted to check in about using applyToFile to stream diffs before we merge this

gui/src/util/clientTools/searchReplaceImpl.ts
Patrick-Erichsen: One step ahead of me! This was the commit I made but didn't have wifi to push up: #[0e7709d](https://github.com/continuedev/continue/pull/6316/commits/0e7709d85e2b4243f6275c5e752c95b0c669e9da) 

We still use the same `applyToFile` message, but it never uses an apply model. It always just generates diff lines using meyers and streams them directly in.

Funny enough even before this commit it was still working though because we detected that it was an "instant" apply since it's a full file rewrite. But these changes are much more clear

RomneyDa: OK great!

RomneyDa (approved):
Let's also hide for jetbrains OR add support for jetbrains before merging

RomneyDa:
failing gui test is legit, just need to add the isSearchAndReplace param in the test I think

RomneyDa:
FUTURE: add a custom tool call div for Search and Replace tool. First iteration could be red text for search, green for replace, second iteration could be import meyers diff on GUI and calculate diff live

RomneyDa (approved):

PR #6318, opened 2025-06-27
UI Polish

sestinj:
## Description

Some UI improvements, including using rem instead of px, matching VS Code link colors, and showing diff syntax highlighting for diff tool output

## Checklist

---

## Summary by cubic
Improved UI consistency by switching to rem units, updated link colors to match VS Code, and added diff syntax highlighting to the diff tool output.

- **UI Polish**
  - Replaced px with rem for scalable sizing.
  - Updated link and code block colors for better VS Code alignment.
  - Added syntax highlighting for diffs.
  - Refactored gradient border into its own component.

### DIFF
--- a/core/tools/implementations/viewDiff.ts
+++ b/core/tools/implementations/viewDiff.ts
@@ -7,7 +7,7 @@ export const viewDiffImpl: ToolImpl = async (args, extras) => {
 
   return [
     {
-      name: "Diff",
+      name: ".diff",
       description: "The current git diff",
       content: diffs.join("\n"),
     },
--- a/core/tools/implementations/viewDiff.ts
+++ b/core/tools/implementations/viewDiff.ts
@@ -7,7 +7,7 @@ export const viewDiffImpl: ToolImpl = async (args, extras) => {
 
   return [
     {
-      name: ".diff",
+      name: "Diff",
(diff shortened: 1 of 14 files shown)


core/tools/implementations/viewDiff.ts
RomneyDa, line 1: why change to .diff?

RomneyDa (changes requested):
Seems like no changes to diff tool here?

Need to add `textCodeBlockBackground` to jetbrains theme

core/tools/implementations/viewDiff.ts
sestinj: If you click the context item it now has syntax highlighting because it's detected as a ".diff" file. Does have the downside of it saying ".diff" instead of "Diff", but with the updated UI I'm working toward it shouldn't need to say either (at least in the GUI, just in the file tab)

Patrick-Erichsen (dismissed):

core/tools/implementations/viewDiff.ts
RomneyDa: Could we set a file URI instead? context item name as the trigger for it being detected as a diff file feels off

RomneyDa:
@sestinj see comments

PR #6319, opened 2025-06-27
Improve UI for tool calls that only have one context item

sestinj:
## Description

For tool calls with a single context item, don't show the dropdown

## Checklist

## Tests

Just UI changed, I don't think tests are so helpful here
    

---

## Summary by cubic
Updated the tool call UI to hide the dropdown when there is only one context item, making it clickable instead.

- **UI Improvements**
  - Shows a dropdown only when there are multiple context items.
  - Clicking a single context item now opens it directly.


Patrick-Erichsen (approved):
nice polish 👌

PR #6323, opened 2025-06-27
feat: run dependent indexer based on context provider

uinstinct:
## Description

We do not need to run all the indexes if only a set of selected context providers are enabled in our config. This PR adds the ability to run only dependent indexers determined by `indexTypes` property on the context provider.

- replaces `dependsOnIndexing` key value with `indexTypes`
- specifies the dependent indexers in Code, Codebase, and Folder Context providers

Resolves CON-2399

## Checklist

## Screenshots

## Tests

## Reproduction

I used a debug statement like this `console.log("indexTypesToBuild", [...indexTypesToBuild],"and config",config,);` inside `CodebaseIndexer.ts` to check if dependent indexing steps are present

    

---

## Summary by cubic
Updated context providers to specify which indexers they need, so only the required indexers run based on the config.

- **Refactors**
  - Replaced the old `dependsOnIndexing` flag with a new `indexTypes` property.
  - Updated Code, Codebase, and Folder context providers to list their needed indexers.
  - Changed indexer logic to build only the index types required by enabled providers.


core/context/providers/FolderContextProvider.ts
```diff
@@ -19,7 +19,7 @@ class FolderContextProvider extends BaseContextProvider {
     displayTitle: "Folder",
     description: "Type to search",
     type: "submenu",
-    dependsOnIndexing: true,
+    indexTypes: ["embeddings"],
```
sestinj: This should rely on the same list as @ codebase (embeddings, fullTextSearch, chunk)

core/protocol/webview.ts
```diff
@@ -24,7 +24,7 @@ export type ToWebviewFromIdeOrCoreProtocol = {
   "indexing/statusUpdate": [IndexingStatus, void]; // Docs, etc.
   refreshSubmenuItems: [
     {
-      providers: "all" | "dependsOnIndexing" | ContextProviderName[];
+      providers: "all" | "indexTypes" | ContextProviderName[];
```
sestinj: I feel like we could keep this as "dependsOnIndexing". That seems like a variable name that makes more sense, and I don't think the behavior changed

sestinj (changes requested):

RomneyDa:
@uinstinct will this work if config is changed to _add_ a context provider that requires an index that was previously not required?

core/indexing/CodebaseIndexer.ts
```diff
@@ -704,6 +724,13 @@ export class CodebaseIndexer {
     }
   }
 
+  public async wasIndexesChanged() {
+    const indexes = await this.getIndexesToBuild();
+    return !indexes.every((index) =>
```
RomneyDa, line 719: - I think the `builtIndex === index` line will fail because it's comparing references to two instances and will always be false
- This logic also fails if the number of indexes decreases, since every new index would be found in the current indexes.

RomneyDa (changes requested):

RomneyDa (changes requested):
@uinstinct this comment form last review is still not resolved

>This logic also fails if the number of indexes decreases, since every new index would be found in the current indexes.

uinstinct:
> @uinstinct this comment from last review is still not resolved
> 
> > This logic also fails if the number of indexes decreases, since every new index would be found in the current indexes.

I had written a test [which covers this use case](https://github.com/continuedev/continue/pull/6323#pullrequestreview-3138707101)

RomneyDa:
@uinstinct note conversation should be unlocked now

core/indexing/CodebaseIndexer.test.ts
```diff
@@ -675,5 +676,150 @@ describe("CodebaseIndexer", () => {
...
+      const mockIndex1 = new TestCodebaseIndex();
+      const mockIndex2 = new FullTextSearchCodebaseIndex();
+      mockGetIndexesToBuild.mockResolvedValue([mockIndex1, mockIndex2]);
+      (testIndexer as any).builtIndexes = [mockIndex1];
+
+      const result = await testIndexer.wasIndexesChanged();
+      expect(result).toBe(true);
+    });
+
+    test("should return false when there are fewer indexes to build than built indexes", async () => {
```
RomneyDa, line 819: @my previous review
the test case is faulty, should return true

RomneyDa (dismissed):

PR #6331, opened 2025-06-27
fix: :bug: Fix bare markdown fences and premature code generation stops

chezsmithy:
## Description

Claude 4 sonnet is very excited about creating project structure display in readme files. When this was attempted the markdown display was broken and the file output was truncated. This was due file structure being outputted with nested codeblocks without a type attached.

When I orginally implemented a fix to this code I noted that there might be some risky edge cases where the nested code blocks didn't have types. As such this PR should solve for that use case by implementing a look ahead feature to check upcoming code blocks to determine if nesting is starting or we are ending the outer code block nesting early.

Significant hardening has been done on all of the places where markdown is used. I've still noted a few issues in chat where markdown might be returned and fences are not balanced (for example an open fence is included indented, and the closing fence happens to be not indented or enclosed in quotes) that could be further improved in a new PR.

Further, I noted that Bedrock Lower cost nova models we want to use for edit and apply were causing problems with outputting markdown content that wasn't compatible with the current prompts and filtering logic. This has been solved for.

## Checklist

## Screenshots

## Tests

I've added a new function for the look ahead checking, and added new tests for that. Then I added a specfic test for the nested structure output that was failing.

I've added extensive tests for all the changes included functions that have been modified. During manual testing, I have found additional edge cases where the generation was stopping pre-maturely and I've added tests to attempt to solve for those. My goal has been to allow using AI to manage the lineStream.ts, and lineStream.vitetest.ts themselves which has been a problem traditionally becuase they contained stop blocks and fences that broke code generation. I've validated that its now possible to refactor those files sucessfully.

    

---

## Summary by cubic
Fixed an issue where project structure markdown with nested bare codeblocks would break display and truncate output.

- **Bug Fixes**
  - Added logic to detect and handle nested bare codeblocks without a type.
  - Updated tests to cover these edge cases.


sestinj:
@chezsmithy There is a failing test, but I do like the PR. Will merge once tests are fixed

chezsmithy:
TODO Items:

- [x] Passing tests
- [x] Review all ``` and STOP checking to make sure it is looking only for leading backticks or STOP keywords and not catching examples inside of code strings. Thus tests won't generate using these keywords or truncation can occur.
- [x] Optimize lookahead function for ``` / STOP checking to not scan the whole remainder of the file if we can.

chezsmithy:
@sestinj feeling good about this now. I've added a tone of tests and split out the filterCodeBlockLines function so that it's easier to test in issolation as its such a critical component.

core/autocomplete/filtering/streamTransforms/lineStream.ts
```diff
@@ -16,6 +18,23 @@ export type CharacterFilter = (args: {
   multiline: boolean;
 }) => AsyncGenerator<string>;
 
+// Duplicated from gui/src/components/StyledMarkdownPreview/utils/headerIsMarkdown.ts
+function headerIsMarkdown(header: string): boolean {
```
RomneyDa: Note, rather than duplicating you can move to core and import to GUI

core/edit/lazy/streamLazyApply.ts
```diff
@@ -11,6 +10,164 @@ import { DiffLine, ILLM } from "../../index.js";
 import { lazyApplyPromptForModel, UNCHANGED_CODE } from "./prompts.js";
 import { BUFFER_LINES_BELOW, getReplacementWithLlm } from "./replace.js";
 
+function headerIsMarkdown(header: string): boolean {
```
RomneyDa: see other note, can avoid this duplication as well

core/edit/lazy/streamLazyApply.ts
```diff
@@ -11,6 +10,164 @@ import { DiffLine, ILLM } from "../../index.js";
...
+
+function isMarkdownFile(filename: string): boolean {
+  const ext = filename.split(".").pop()?.toLowerCase();
+  return ["md", "markdown", "gfm"].includes(ext || "");
+}
+
+/**
+ * Optimized state tracker for markdown block analysis to avoid recomputing on each call.
+ */
+class MarkdownBlockStateTracker {
```
RomneyDa: It seems like this whole file is duplicate of core/autocomplete/filtering/streamTransforms/lineStream.ts?

chezsmithy:
@sestinj with the introduction of the search and replace tool, would it make sense to break out the rendering fixes for the UI with markdown from the apply updates that might become less relevant with the serach and replace changes?

core/edit/lazy/streamLazyApply.ts
```diff
@@ -108,24 +104,19 @@ async function* streamFillUnchangedCode(
     }
 
     if (line.includes(UNCHANGED_CODE)) {
-      // Buffer so we can give the context of BUFFER_LINES_BELOW lines below
       waitingForBuffer = true;
-      // TODO: If the UNCHANGED CODE is at the very top of the file we need to handle a bit differently
```
RomneyDa: It looks like an apply operation may have removed comments from this file? Or intentional?

core/utils/markdownUtils.ts
```diff
@@ -0,0 +1,115 @@
...
+            }
+          }
+        } else if (currentLine.startsWith("```")) {
+          // Going into a nested codeblock
+          this.markdownNestCount++;
+        }
+      } else {
+        // Not inside a markdown block yet
+        if (currentLine.startsWith("```")) {
+          const header = currentLine.replaceAll("`", "");
```
RomneyDa, line 90: should technically replace all at the start but nitpick

core/utils/streamMarkdownUtils.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (nestCount > 0) {
+      if (line.match(/^`+$/)) {
+        // Use optimized state tracker for markdown block logic
+        if (stateTracker.shouldStopAtPosition(i)) {
+          for (let j = 0; j < i; j++) {
+            yield allLines[j];
+          }
+          return;
+        }
+      } else if (line.startsWith("```")) {
```
RomneyDa: seems like this code will never be hit because `if (line.match(/^'+$/))` is redundant of it?

gui/src/components/StyledMarkdownPreview/utils/patchNestedMarkdown.ts
```diff
@@ -18,19 +52,27 @@ export const patchNestedMarkdown = (source: string): string => {
   const lines = source.split("\n");
   const trimmedLines = lines.map((l) => l.trim());
```
RomneyDa: looks like duplicate trimming operation here, should just iterate over trimmed lines from the PatchStateTracker

core/autocomplete/filtering/streamTransforms/lineStream.ts
```diff
@@ -138,7 +216,7 @@ export async function* avoidPathLine(
   // Sometimes the model with copy this pattern, which is unwanted
   for await (const line of stream) {
     if (line.startsWith(`${comment} Path: `)) {
-      continue;
+      continue; // continue in the Continue codebase! How meta!
```
RomneyDa, line 231: continue!

core/autocomplete/filtering/streamTransforms/lineStream.ts
```diff
@@ -71,6 +118,37 @@ function isUselessLine(line: string): boolean {
...
+): boolean {
+  return (
+    (firstLine.startsWith("```") &&
+      headerIsMarkdown(firstLine.replace(/`/g, ""))) ||
+    Boolean(filepath && isMarkdownFile(filepath))
+  );
+}
+
+// Re-export shared utilities
+export { collectAllLines, isMarkdownFile };
```
RomneyDa: Can remove the re-exports and import from source

core/edit/lazy/replace.ts
```diff
@@ -90,11 +87,18 @@ export async function* getReplacementWithLlm(
       { role: "assistant", content: assistantPrompt },
     ],
     abortController.signal,
+    {
+      raw: true,
```
RomneyDa, line 91: What is the thought on adding `raw` here?

core/edit/recursiveStream.ts
```diff
@@ -63,7 +63,8 @@ export async function* recursiveStream(
     }
   } else {
     const generator = llm.streamChat(prompt, abortController.signal, {
-      prediction,
+      raw: true,
```
RomneyDa, line 66: +1

RomneyDa (changes requested):
@chezsmithy I think generally safe to say apply won't go anywhere too soon! 

I noticed ~30 lines of the logic in `shouldChangeLineAndStop` and `stopAtLines` seem duplicate, and the MarkdownPatchStateTrackers in core and gui seem mostly duplicate as well

Might be misunderstanding subtle differences in the implementations, but could these both be united? This PR is solving complicated problems so it's inherently a bit tough to wrap my head around but I think the deduplication (if possible) could the review move through faster

The tests are great

core/edit/lazy/replace.ts
chezsmithy: Based on analysis from the codebase, it seems that `raw: true` is consistently used for code editing operations. This pattern is established in `core/edit/recursiveStream.ts` line 33 and `core/autocomplete/generation/CompletionStreamer.ts` line 41

More importantly, raw: true is technically superior for code editing use cases because it sends prompts directly to the LLM without template processing, ensuring precise control over the instructions. This prevents any intermediate formatting that could interfere with the carefully crafted prompts needed for accurate code replacement and generation.

@RomneyDa thoughts?

core/edit/recursiveStream.ts
chezsmithy: Based on analysis from the codebase, it seems that `raw: true` is consistently used for code editing operations. This pattern is established in `core/edit/recursiveStream.ts` line 33 and `core/autocomplete/generation/CompletionStreamer.ts` line 41

More importantly, raw: true is technically superior for code editing use cases because it sends prompts directly to the LLM without template processing, ensuring precise control over the instructions. This prevents any intermediate formatting that could interfere with the carefully crafted prompts needed for accurate code replacement and generation.

@RomneyDa thoughts?

chezsmithy:
> @chezsmithy I think generally safe to say apply won't go anywhere too soon!
> 
> I noticed ~30 lines of the logic in `shouldChangeLineAndStop` and `stopAtLines` seem duplicate, and the MarkdownPatchStateTrackers in core and gui seem mostly duplicate as well
> 
> Might be misunderstanding subtle differences in the implementations, but could these both be united? This PR is solving complicated problems so it's inherently a bit tough to wrap my head around but I think the deduplication (if possible) could help me review more confidently
> 
> The tests are great

@RomneyDa I've also addressed the duplicate lines noted in the the comments above I believe. Just a couple open questions on the raw change for you.

core/edit/lazy/replace.ts
RomneyDa: This makes sense to me, appreciate the digging

gui/src/components/StyledMarkdownPreview/utils/patchNestedMarkdown.ts
RomneyDa, line 1: I think we should be especially sensitive about performance in the nested markdown patch because it runs on every single streamed chunk into the GUI.

As far as I can tell (correct me if I'm missing something) the MarkdownBlockStateTracker is not more efficient than the current implementation for patchNestedMarkdown, it adds many array.includes and array.filter operations on top of line trims and regex comparisons. I worry that small increases in operations for this patch could cause some performance issues since it's running on every streamed chunk

RomneyDa (changes requested):
@chezsmithy see comment on performance for patchNestedMarkdown. Let me know if you agree with this. Maybe run a quick n=100000 test locally or something to see if it makes a difference?

I'm less worried about performance for apply since it's not happening within the React app. If it seems performance is not deteriorated at all then lgtm!

gui/src/components/StyledMarkdownPreview/utils/patchNestedMarkdown.ts
chezsmithy: @RomneyDa I agree. One of the fixes I applied was a look forward scan to find the next closing fence. Originally it did a scan on every line. The tracker captures them all upfront and then optimizes the like processing by scanning forward to the next fence.

I'm still tempted to split this into two PRs. One for apply and one for the gui just in case a rollback is required.

I'll build a bigger test and do some benchmarking.

RomneyDa: One PR seems okay if we gut check on performance, either way works for me

RomneyDa (approved):
@chezsmithy pulled this down and didn't notice any performance issues. Current state looks good so will go ahead and merge, if you were in the process of making changes I'll look out for another PR!

PR #6333, opened 2025-06-28
Update Agent/Tool Message Construction and Handling

RomneyDa:
## Description
- Unifies thunks - moves getting system message, constructing messages into the streamNormalInput thunk
- Updates the way dangling messages are cleaned up on stream cancellation - accounts for tool calls, thinking messages, etc.
- Updates message compilation so that fresh tool messages are injected after their tool calls at the time of compiling messages, although output is still streamed to redux toolCallState AND existing tool call messages, so that persisted conversation reflects what is sent to model. This fixes bugs around tool calls not having associated tool messages and vice versa.
- Changes pending tool rejection to just stop rather than continue streaming
- Changes reject diffs after edit tool to stop rather than continue streaming from GUI at least for now
- Removes filtering of tool calls in chat mode (was causing non-sensical history for model) but updates the Bedrock API to insert tool calls and outputs as text parts when the corresponding tool is not in the toolOptions. Updates corresponding tests.
- Changes rule trigger logic to include context items from the most recent user message forward, including tool/assistant messages AFTER the most recent user message (for long agent convos)

Borrows ideas from @chezsmithy https://github.com/continuedev/continue/pull/5662


gui/src/components/ModeSelect/ModeSelect.tsx
```diff
@@ -34,7 +34,9 @@ export function ModeSelect() {
 
   const cycleMode = useCallback(() => {
     dispatch(setMode(mode === "chat" ? "agent" : "chat"));
-    mainEditor?.commands.focus();
+    if (!document.activeElement?.classList.contains("ProseMirror")) {
```
RomneyDa, line 37: Only focus main editor if another one doesn't already have focus

gui/src/pages/gui/ToolCallDiv/SimpleToolCallUI.tsx
```diff
@@ -54,22 +60,25 @@ export function SimpleToolCallUI({
...
         </div>
-        {args.length > 0 ? (
-          <ArgsToggleIcon
-            isShowing={showingArgs}
-            setIsShowing={setShowingArgs}
-            toolCallId={toolCallState.toolCallId}
-          />
-        ) : null}
+        <div className="flex flex-row items-center gap-1.5">
+          <TruncateHistoryIcon historyIndex={historyIndex} />
```
RomneyDa, line 64: Streamline tool context item to contextItemWithId conversion, add toolCall named icon

RomneyDa (commented):
- Adds a "Truncate chat to this message" icon in non-last tool messages, to allow users to interrupt mid-long agent streams
- Fix bug where toggling mode causes focus loss if not in main editor
- Streamline tool context item to contextItemWithId conversion, add toolCall named icon

core/llm/constructMessages.ts
```diff
@@ -79,60 +84,37 @@ export const DEFAULT_AGENT_SYSTEM_MESSAGE = `\
...
-    return (
-      item.message.content === userMsg.content &&
-      item.message.role === userMsg.role
-    );
-  });
-
-  return historyItem?.contextItems || [];
-}
-
 export function constructMessages(
```
sestinj: We need to get unit tests on this thing. It's on critical path and if something went wrong we actually wouldn't find out because the LLM would just accept it and return imperceptibly worse answers. Luckily looks like this is a pure function so shouldn't be too tough to test

gui/src/redux/thunks/streamResponse.ts
```diff
@@ -150,7 +105,6 @@ export const streamResponseThunk = createAsyncThunk<
         unwrapResult(
           await dispatch(
             streamNormalInput({
```
sestinj, line 107: this is great

sestinj (changes requested):
It feels like there's a chance this could have been 2 PRs? I don't know exactly what the order of operations looked like while you were working on it, but just noting that this was slightly messy to review

Otherwise biggest thing is unit tests on constructMessages

And make sure to take a look at Cubic's feedback. At least the gap one is accurate

sestinj (approved):

PR #6353, opened 2025-06-30
feat: Report issues to Sentry

exigow:
## Description

Included tags:
* ide_version
* plugin_version
* jcef_supported

## Tests
To test it, just register an action that throws an exception. As a result, a notification appears + there is a blinking icon in the bottom-right corner. Click on it and select 'Submit to Continue'. User message is also sent to Sentry (you can leave it empty).

## Screenshot

image

### DIFF
--- a/extensions/intellij/build.gradle.kts
+++ b/extensions/intellij/build.gradle.kts
@@ -18,6 +18,7 @@ plugins {
     id("org.jetbrains.qodana") version "0.1.13"
     id("org.jetbrains.kotlinx.kover") version "0.7.3"
     kotlin("plugin.serialization") version "1.8.0"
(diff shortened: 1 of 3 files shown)


exigow:
Last event: https://continue-xd.sentry.io/issues/6718511592/events/10e71be0a8a0462ab66cb08a99021b41/

exigow:
Note: I haven't yet figured out how to attach the thread dump automatically (even for simple exceptions), but I think what's in this pull request is a good starting point to gather some knowledge on our issues (+ this will definitely help with prioritization as well). We can now easily add new tags as needed.

sestinj (approved):

PR #6358, opened 2025-06-30
Add edit outcome tracking to record accept/reject

sestinj:
## Summary by cubic
Added tracking for edit outcomes to record when code changes are accepted or rejected.

- **New Features**
  - Tracks pending code edits and logs whether each is accepted or rejected.
  - Captures details like model info, prompt, code before and after, and line changes.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -2,7 +2,7 @@
   "name": "continue",
   "icon": "media/icon.png",
   "author": "Continue Dev, Inc",
-  "version": "1.1.52",
+  "version": "1.1.53",
   "repository": {
     "type": "git",
     "url": "https://github.com/continuedev/continue"


chezsmithy:
Might be worth noting when experimental auto accept is on to not skew results since we are yolo'n our way through apply failures, and control-z'n the file back to pre edit state.

extensions/vscode/src/diff/processDiff.ts
```diff
@@ -37,6 +44,13 @@ export async function processDiff(
   if (streamId) {
     const fileContent = await ide.readFile(newOrCurrentUri);
 
+    // Record the edit outcome before updating the apply state
+    await EditOutcomeTracker.getInstance().recordEditOutcome(
```
Patrick-Erichsen: Generally not a fan of this `EditOutcomeTracker.getInstance()` when we could just export the singleton instance directly as `export const EditOutcomeTrackerSingleton = new EditOutcomeTracker()` and not have the more verbose `.getInstance()` all over the place

extensions/vscode/src/diff/vertical/manager.ts
```diff
@@ -487,11 +508,46 @@ export class VerticalDiffManager {
...
+    filepath: string;
+    prompt: string;
+    fileAfterEdit: string | undefined;
+  }) {
+    // Get previous code content for outcome tracking
+    const previousCode = await this.ide.readFile(filepath);
+    const newCode = fileAfterEdit ?? "";
+    const previousCodeLines = previousCode.split("\n").length;
+    const newCodeLines = newCode.split("\n").length;
+    const lineChange = newCodeLines - previousCodeLines;
```
Patrick-Erichsen, line 535: Nitpick but this feels like logic that would be better encapsulated in the `EditOutcomeTracker`

extensions/vscode/src/extension/EditOutcomeTracker.ts
```diff
@@ -0,0 +1,94 @@
+/**
+ * Tracks pending edit interactions until they are accepted or rejected,
+ * then emits editOutcome events with the final result.
+ */
```
Patrick-Erichsen, line 4: What's the rationale for tracking the pending edits instead of just firing off the event at accept/reject time?

Patrick-Erichsen (approved):

extensions/vscode/src/extension/EditOutcomeTracker.ts
sestinj: I think we need to do this in order to know the prompt, model, etc. at accept time

extensions/vscode/src/diff/processDiff.ts
sestinj: fair

extensions/vscode/src/diff/vertical/manager.ts
sestinj: true!

PR #6359, opened 2025-06-30
refactor: wrap rules with application reasons in tuple structure

sestinj:
## Summary by cubic
Refactored rule application logic to wrap each applied rule with its application reason in a tuple structure, making rule usage and debugging clearer across the codebase.

- **Refactors**
  - Introduced `RuleWithApplicationReason` type to pair each rule with its reason for application.
  - Updated all rule application, storage, and test logic to use the new tuple structure.

### DIFF
--- a/core/llm/rules/getSystemMessageWithRules.ts
+++ b/core/llm/rules/getSystemMessageWithRules.ts
@@ -11,6 +11,19 @@ import { extractContentFromCodeBlock } from "../utils/extractContentFromCodeBloc
 import { extractPathsFromCodeBlocks } from "../utils/extractPathsFromCodeBlocks";
 import { RulePolicies } from "./types";
 
+export type RuleApplicationReason =
+  | "alwaysApply"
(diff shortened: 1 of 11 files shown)


core/llm/rules/getSystemMessageWithRules.ts
```diff
@@ -11,6 +11,19 @@ import { extractContentFromCodeBlock } from "../utils/extractContentFromCodeBloc
...
+  | "globMatch"
+  | "regexMatch"
+  | "globAndRegexMatch"
+  | "directoryMatch";
+
+export interface RuleWithApplicationReason {
+  rule: RuleWithSource;
+  reason: RuleApplicationReason;
+}
+
```
Patrick-Erichsen, line 26:
```suggestion
```

We already have this defined in `index.d.ts`

Patrick-Erichsen (changes requested):
just one quick fix

PR #6376, opened 2025-07-01
fix: support local slugs in document link provider

fbricon:
Document link navigation is currently broken if a config file "uses" a local slug, e.g. a "./", "file://" or "http://" url, the ConfigYamlDocumentLinkProvider always prefixes the path with https://hub.continue.dev/ and then opens an invalid url.

This PR fixes it.

https://github.com/user-attachments/assets/5581b88c-9b82-4e82-8504-f668675a3822

I'm hesitant to provide support for directory traversal (../..) it's often source of security issues, but I'm not even sure if continue supports it. You tell me.

The CLA has been sent to @sestinj.

    

---

## Summary by cubic
Fixed document link navigation for config files that use local slugs like "./" or "file://", so links now resolve correctly instead of always pointing to https://hub.continue.dev/.

### DIFF
--- a/extensions/vscode/src/extension/ConfigYamlDocumentLinkProvider.ts
+++ b/extensions/vscode/src/extension/ConfigYamlDocumentLinkProvider.ts
@@ -1,3 +1,5 @@
+import * as path from "path";
+
 import * as vscode from "vscode";
 
 export class ConfigYamlDocumentLinkProvider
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen (approved):

Patrick-Erichsen:
I think we do support directory traversal, although this is probably a case where the functionality is the docs 😅  what are your thoughts on the security concerns @fbricon ?

Patrick-Erichsen:
@fbricon I'm almost positive you've signed the CLA already but it seems to still be blocking this PR, mind commenting `I have read the CLA Document and I hereby sign the CLA` again to see if it fixes it?

Also apologies but I think you need to merge `main` to fix the failing e2e test, that is a new one that seems to be breaking existing PRs

Patrick-Erichsen:
Actually @fbricon you can ignore the CLA thing, https://github.com/continuedev/continue/pull/6408 will solve it

sestinj:
rerunning the tests real quick, otherwise looks great

PR #6389, opened 2025-07-02
Dynamically switch between rendering ghost text and svg decoration based on the model output

jpoly1219:
## Description

Dynamically switch between rendering ghost text and svg decoration based on the model output.
If the model output is FIM, then we render a ghost text. Else, we render a SVG decoration.

## Checklist

## Screenshots

## Tests


core/nextEdit/diff/diff.ts
Patrick-Erichsen, line 1: I think a unit test would be helpful here, 1) to view sample inputs, and 2) to lock down/verify behavior

Should be a pretty easy one-shot

jpoly1219: added vitest!

sestinj (approved):

PR #6391, opened 2025-07-02
fix prettier formatting from old PRs

RomneyDa:
Some more old PRs were merged that were passing tests BEFORE we added the prettier CICD rule


RomneyDa:
seems like working on main now

PR #6394, opened 2025-07-02
feat: prefix folder path for colocated rules

uinstinct:
## Description

Prepend the folder path in the glob match for the corresponding `rules.md` file. We want to apply the particular rule for the specific folder only.

- added a `getGlobPattern` method for modifying the frontmatter globs
- added tests for the new glob pattern

Resolves CON-2516

## Checklist

## Screenshots

image

## Tests

---

## Summary by cubic
Updated rule glob patterns to automatically include the folder path of each colocated rules.md file, so rules only apply to files in their specific folder.

- **Refactors**
  - Added a helper to prepend the folder path to glob patterns.
  - Updated tests to check the new glob behavior.


packages/config-yaml/src/markdown/markdownToRule.ts
```diff
@@ -69,6 +70,30 @@ export function getRuleName(
   return displayName;
 }
 
+function getGlobPattern(
```
sestinj, line 73: We shouldn't prepend the directory if the rule is inside of .continue/rules

packages/config-yaml/src/markdown/markdownToRule.ts
```diff
@@ -69,6 +70,30 @@ export function getRuleName(
...
+  globs: RuleFrontmatter["globs"],
+  id: PackageIdentifier,
+) {
+  if (id.uriType !== "file") {
+    return globs;
+  }
+  const dir = path.dirname(id.filePath);
+  const prependDirAndApplyGlobstar = (glob: string) => {
+    // skip if glob is a directory
+    if (glob.includes("/")) {
```
sestinj: I think would still want to prepend the dir and globstar if it is a directory pattern. For example `subdir/` should become `path/to/subdir/`

sestinj (changes requested):

packages/config-yaml/src/markdown/markdownToRule.ts
uinstinct: Oh yes. We can prepend `**` in that case as well which would become `path/to/**/subdir` (thus matching `path/to/subdir` and also `path/to/my/nested/subdir`)

Implemented!

RomneyDa (changes requested):
@uinstinct I'm not sure path.join will work on systems that use backward slash paths, since globs only use forward slash, but path.join might add backwards slashes. is this tested?

uinstinct:
> @uinstinct I'm not sure path.join will work on systems that use backward slash paths, since globs only use forward slash, but path.join might add backwards slashes. is this tested?

Yes, [`path.join`](https://nodejs.org/api/path.html#windows-vs-posix) would handle this for windows platforms. 
I haven't tested this explicitly on windows yet.

RomneyDa:
@uinstinct what I mean is path.join will "work" in windows but would result in e.g. path\to\file which will not work with globs, even in windows Node.js

uinstinct:
> @uinstinct what I mean is path.join will "work" in windows but would result in e.g. path\to\file which will not work with globs, even in windows Node.js

Makes sense. Implemented!

RomneyDa (approved):

PR #5809, opened 2025-05-23
Dallin/non native tools experiment

RomneyDa:
## Description

[ What changed? Feel free to be brief. ]

## Checklist

- [] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [] The relevant docs, if any, have been updated or created
- [] The relevant tests, if any, have been updated or created

## Screenshots

[ For visual changes, include screenshots. Screen recordings are particularly helpful, and appreciated! ]

## Tests

[ What tests were added or updated to ensure the changes work as expected? ]

PR #6395, opened 2025-07-02
feat: System Message Tools

RomneyDa:
A remake of https://github.com/continuedev/continue/pull/5809

- Adds support for tools in the form of system message instructions and response content parsing + tests
- Tools are parsed as they come in, converted to ToolCalls, and on the way out converted back to text. Switching between native and non-native tools is interchangeable with this approach.
- For now, this applies only to models that are NOT known to support native tools
- Adds a warning to the Agent mode dropdown that shows for any models not known to be great with tools (including models that support native tools)

All models can now be used in Agent mode


RomneyDa:
note to update tool descriptions based on https://github.com/continuedev/continue/pull/6421/files

core/llm/index.ts
```diff
@@ -498,33 +498,27 @@ export abstract class BaseLLM implements ILLM {
   }
 
   private _formatChatMessages(messages: ChatMessage[]): string {
-    const msgsCopy = messages ? messages.map((msg) => ({ ...msg })) : [];
-    let formatted = "";
-    for (const msg of msgsCopy) {
-      formatted += this._formatChatMessage(msg);
-    }
-    return formatted;
+    return (messages ?? [])
```
sestinj, line 515: I'm not confident it's safe to remove the spread here. Are you?

core/llm/index.ts
```diff
@@ -498,33 +498,27 @@ export abstract class BaseLLM implements ILLM {
...
+      .map((msg) => this._formatChatMessage(msg))
+      .join("\n\n");
   }
 
   private _formatChatMessage(msg: ChatMessage): string {
     let contentToShow = "";
     if (msg.role === "tool") {
       contentToShow = msg.content;
-    } else if (msg.role === "assistant" && msg.toolCalls) {
+    } else if (msg.role === "assistant" && msg.toolCalls?.length) {
```
sestinj, line 510: This nuance might deserve a test

sestinj (changes requested):

sestinj:
image

@RomneyDa Tried first with Sonnet and I'm not sure whether this is a parsing bug, but it seems to mostly not stop when a tool is called, so it ends up just streaming multiple tools in a single assistant output, none of them getting called

sestinj:
When using `o3-mini` on the other hand seems like it does actually stop the stream, call the tool, and then call the assistant again.

I'm also seeing a little bit of bugginess around it randomly starting and stopping the stream, but don't have a super clear grasp yet on what causes that.

core/llm/index.ts
RomneyDa: I think yes because nothing is mutated on the passed object

RomneyDa:
> image
> @RomneyDa Tried first with Sonnet and I'm not sure whether this is a parsing bug, but it seems to mostly not stop when a tool is called, so it ends up just streaming multiple tools in a single assistant output, none of them getting called

This one is fixed!

PR #6396, opened 2025-07-02
Simplify tool names

RomneyDa:
Changes built in tool names from e.g. `builtin_read_file` to just `read_file`
This is in part for better system-message tool calling support, see https://github.com/continuedev/continue/pull/6395
Tool policies are migrated from old to new values on startup.
    

---

## Summary by cubic
Simplified built-in tool names by removing the "builtin_" prefix, and added automatic migration of tool policies from old to new names on startup.

- **Migration**
  - Tool policies using old names are updated to the new names automatically when the app starts.


Patrick-Erichsen (approved):
LGTM but I'm assuming this will conflict with the big docs refactor PR

RomneyDa:
Will merge since @bdougie is actively working on it still and this is somewhat blocking virtual tools being effective for poorer models

PR #6418, opened 2025-07-02
fix duplicate tools bug

RomneyDa:
## Description
Was pushing tools to const in memory, causing duplicates
    

---

## Summary by cubic
Fixed a bug that caused duplicate tools to appear when loading tool definitions for remote workspaces.


Patrick-Erichsen (approved):
We'll still find a way to mess this up

sestinj (approved):

PR #6422, opened 2025-07-02
feat: :alembic: Include custom rules for apply

chezsmithy:
## Description

I've been experimently extensively with using a smaller faster model for apply steps. In this case I've been using claude sonnet 4 for chat and agent model, and then handing off apply steps to Bedrock Nova Pro. 

During this experimentation I've noted subtle differnences in how Nova Pro takes in the input from the sonnet output incorporates the existing file and creates the final product. For example in one case, I always wanted compliant Markdown including fences that had a language. What I noted was with a rule in place sonnet would comply, however, when I handed that off to Nova Pro to apply the fence would be lost. I needed to have that same rule included in both the generation of the edit step, and the apply step to get the same outcome.

As an aside, here somewhat confusingly the final suggested output in the apply step in the chat window displayed what I expected but the apply step would loose the change and then the file in the editor would diverge.

By including just the custom rules in the apply step, this gap was closed and both the suggested edit from sonnet matches 100% with the apply step from Nova Pro.

## Checklist

## Screenshots

## Tests

I've performed an extensive set of manual experiments with including the rules acorss both chat, agent and apply and this seems to end up with the best outcome. There might be a futher step where we have a test harness that can compare many examples to validate this experiment is really valuable but for now that can be future work.

### DIFF
--- a/extensions/vscode/src/apply/ApplyManager.ts
+++ b/extensions/vscode/src/apply/ApplyManager.ts
@@ -199,7 +199,7 @@ export class ApplyManager {
       range: rangeToApplyTo,
       newCode: text,
       toolCallId,
-      rulesToInclude: undefined, // No rules for apply
+      rulesToInclude: config.rules,
     });
   }
 }


chezsmithy:
@RomneyDa and @sestinj also thinking here if the momentum is toward the search and replace tool then I think we can close this. Not sure how long the experiment might last.

RomneyDa:
@chezsmithy will close this for now, partly because +1 to your comment about moving to search and replace.

My other thought is that I think most rules typically should not be included in an apply prompt, since the apply prompt is only supposed to take existing code and apply changes to it, and rules could often confuse this. In the cases you described, I wonder if updating the apply prompt to nudge it to ensure the formatting of the _changes_ takes priority over existing formatting would solve it. E.g. if the changes (made with rules in mind) include the proper fenced codeblocks, the new code should as well. Or to add a boolean rule property `isApplyRule` or something and only include those in apply. I could see it going either way and sounds like including rules did fix it for you in some cases but I'd lean towards leaving rules out and making the apply. 

If needed, you can add an apply prompt using `promptTemplates.apply` to fix your specific issue.

I have it on my list to see if I can prevent removal of comments with this approach, since it's an apply-specific issue.

Let me know if you have any other thoughts on this!

PR #6412, opened 2025-07-02
Fix: resolve the context retrieval issue for codebase and docs

Jazzcort:
## Description

Due to the new retrieval mechanism, getting context items for codebase and docx is not working. This patch resolve the issue.

https://github.com/continuedev/continue/issues/6411

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [x] The relevant docs, if any, have been updated or created
- [x] The relevant tests, if any, have been updated or created

## Screenshots

https://github.com/user-attachments/assets/0c53af0c-f93a-4a42-9d6d-4c7b6bdcd7a5

## Tests

Mention Docs and Codebase and you should get some context items back instead of nothing.

    

---

## Summary by cubic
Fixed an issue where context items for codebase and docs were not being retrieved, so users now get relevant context as expected.

PR #6429, opened 2025-07-03
Fix context item deduplication logic and add text truncation to context peek

sestinj:
itemId isn't necessarily unique, so this was necessary to avoid items being removed. This was causing docs to not work for example (only one result returned)

### DIFF
--- a/gui/src/components/mainInput/TipTapEditor/utils/resolveEditorContent.ts
+++ b/gui/src/components/mainInput/TipTapEditor/utils/resolveEditorContent.ts
@@ -201,12 +201,12 @@ async function gatherContextItems({
       if (
         !acc.some(
           (i) =>
-            (i.id.providerTitle === item.id.providerTitle &&
-              i.id.itemId === item.id.itemId) ||
-            (i.uri &&
-              item.uri &&
-              i.uri.type === item.uri.type &&
-              i.uri.value === item.uri.value),
+            i.id.providerTitle === item.id.providerTitle &&
+            i.id.itemId === item.id.itemId &&
+            i.uri &&
+            item.uri &&
+            i.uri.type === item.uri.type &&
+            i.uri.value === item.uri.value,
         )
       ) {
         acc.push(item);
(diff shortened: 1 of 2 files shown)


RomneyDa:
@sestinj note this PR fixes the only two cases where item id is not unique
imo item id should be unique, but don't feel strongly
https://github.com/continuedev/continue/pull/6412

gui/src/components/mainInput/belowMainInput/ContextItemsPeek.tsx
```diff
@@ -131,7 +131,7 @@ export function ContextItemsPeekItem({
         {contextItem.name}
       </span>
       <div
-        className={`group flex flex-row items-center gap-1.5 pr-1.5 text-xs text-gray-400 ${isUrl ? "hover:underline" : ""}`}
+        className={`group flex flex-row items-center gap-1.5 truncate pr-1.5 text-xs text-gray-400 ${isUrl ? "hover:underline" : ""}`}
```
RomneyDa, line 134: line-clamp-1 preferred if possible

RomneyDa (changes requested):

gui/src/components/mainInput/TipTapEditor/utils/resolveEditorContent.ts
```diff
@@ -201,12 +201,12 @@ async function gatherContextItems({
       if (
         !acc.some(
           (i) =>
-            (i.id.providerTitle === item.id.providerTitle &&
-              i.id.itemId === item.id.itemId) ||
-            (i.uri &&
-              item.uri &&
-              i.uri.type === item.uri.type &&
-              i.uri.value === item.uri.value),
+            i.id.providerTitle === item.id.providerTitle &&
```
RomneyDa, line 204: Requested changes for this line
It changes deduplication to be for _exactly_ the same context item but breaks it for the original goal which is e.g. openFiles and currentFile adds the same file twice, or @diff is mentioned three times in nested prompts, or @openFiles in a prompt matches a @file mention in the editor, etc.

RomneyDa: But I realized deduplication logic had another flaw which is that you can have multiple context items with the same file URI and different content

Maybe the solution is as simple as
```
   i.content === item.content
```

sestinj:
no longer needed after the changes from jazzcourt

PR #6434, opened 2025-07-03
@codebase use tool calling (under experimental flag)

tingwai:
## Description

Experiment: @codebase context provider to use built in tools to see how it compares with indexing

Here's some example queries to play around to trigger tool calls
| Query                                | Tools used                  |
|--------------------------------------|-----------------------------|
| @codebase what files are here        | ls                          |
| @codebase find all the add functions | grep                        |
| @codebase look inside test.js        | read_file                   |
| @codebase where are my .ts files     | glob_search                 |
| @codebase tell me about this repo    | repo_map + ls + glob_search |

## Checklist

## Screenshots

https://github.com/user-attachments/assets/5bbdfb85-846d-4aaf-af24-24c816864775

## Tests

---

## Summary by cubic
Added an experimental option for @codebase to use tool calling instead of traditional retrieval methods like embeddings or full-text search.

- **New Features**
  - Introduced a "tool calling only" toggle in user settings.
  - When enabled, context retrieval uses built-in tools to answer queries.

### DIFF
--- a/core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
+++ b/core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
@@ -15,23 +15,17 @@ import { openedFilesLruCache } from "../../../autocomplete/util/openedFilesLruCa
 import { chunkDocument } from "../../../indexing/chunk/chunk";
 import { FullTextSearchCodebaseIndex } from "../../../indexing/FullTextSearchCodebaseIndex";
 import { LanceDbIndex } from "../../../indexing/LanceDbIndex";
-import { BuiltInToolNames } from "../../../tools/builtIn";
+import { callBuiltInTool } from "../../../tools/callTool";
 import { globSearchTool } from "../../../tools/definitions/globSearch";
 import { grepSearchTool } from "../../../tools/definitions/grepSearch";
 import { lsTool } from "../../../tools/definitions/lsTool";
 import { readFileTool } from "../../../tools/definitions/readFile";
 import { viewRepoMapTool } from "../../../tools/definitions/viewRepoMap";
 import { viewSubdirectoryTool } from "../../../tools/definitions/viewSubdirectory";
-import { fileGlobSearchImpl } from "../../../tools/implementations/globSearch";
-import { grepSearchImpl } from "../../../tools/implementations/grepSearch";
-import { lsToolImpl } from "../../../tools/implementations/lsTool";
-import { readFileImpl } from "../../../tools/implementations/readFile";
-import { viewRepoMapImpl } from "../../../tools/implementations/viewRepoMap";
-import { viewSubdirectoryImpl } from "../../../tools/implementations/viewSubdirectory";
 
 const DEFAULT_CHUNK_SIZE = 384;
 
-const availableTools: Tool[] = [
+const AVAILABLE_TOOLS: Tool[] = [
   globSearchTool,
   grepSearchTool,
   lsTool,
@@ -191,21 +185,18 @@ export default class BaseRetrievalPipeline implements IRetrievalPipeline {
     const toolSelectionPrompt = `Given the following user input: "${input}"
 
 Available tools:
-${availableTools
-  .map((tool) => {
-    const requiredParams = tool.function.parameters?.required || [];
-    const properties = tool.function.parameters?.properties || {};
-    const paramDescriptions = requiredParams
-      .map(
-        (param: any) =>
-          `${param}: ${properties[param]?.description || "string"}`,
-      )
-      .join(", ");
+${AVAILABLE_TOOLS.map((tool) => {
+  const requiredParams = tool.function.parameters?.required || [];
+  const properties = tool.function.parameters?.properties || {};
+  const paramDescriptions = requiredParams
+    .map(
+      (param: any) => `${param}: ${properties[param]?.description || "string"}`,
+    )
+    .join(", ");
 
-    return `- ${tool.function.name}: ${tool.function.description}
+  return `- ${tool.function.name}: ${tool.function.description}
   Required arguments: ${paramDescriptions || "none"}`;
-  })
-  .join("\n")}
+}).join("\n")}
 
 Determine which tools should be used to answer this query. You should feel free to use multiple tools when they would be helpful for comprehensive results. Respond ONLY a JSON object containing the following and nothing else:
 {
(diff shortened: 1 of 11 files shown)


core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
```diff
@@ -155,4 +186,172 @@ export default class BaseRetrievalPipeline implements IRetrievalPipeline {
...
+  "tools": [
+    {
+      "name": "<tool_name>",
+      "args": { "<required_parameter_name>": "<required_parameter_value>" }
+    }
+  ]
+}`;
+
+    // Get LLM response for tool selection
+    const toolSelectionResponse = await this.options.llm.complete(
```
sestinj: For the anthropic provider, this is equivalent to the `llm.chat` method, but for a lot of other providers `complete` actually does something slightly different. It sends a "raw" prompt instead of one that will be formatted as [ChatML](https://github.com/openai/openai-python/blob/release-v0.28.0/chatml.md).

So if you said "Hello" you might get a response like this from `complete`: ", my name is Claude." and a response like this from `chat`: "Hello! How can I help you today?".

tl;dr you should use `llm.chat` here instead

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
```diff
@@ -155,4 +186,172 @@ export default class BaseRetrievalPipeline implements IRetrievalPipeline {
...
+        chunks.push({
+          content: contextItem.content,
+          startLine: -1,
+          endLine: -1,
+          digest: `file:///${cleanedFilepath}`,
+          filepath: `file:///${cleanedFilepath}`,
+          index: 0,
+        });
+      } else {
+        // Split grep search results by file paths and create separate chunks
```
sestinj: I like the idea of splitting these into different chunks—it would be really nice for the chunks to be visible as different context items in the "N context items" dropdown.

Since I think the normal Agent mode tool use would also benefit from this, could this logic be moved into the grep tool implementation?

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
```diff
@@ -155,4 +186,172 @@ export default class BaseRetrievalPipeline implements IRetrievalPipeline {
...
+      ide: this.options.ide,
+      llm: this.options.llm,
+      fetch: fetch,
+      tool: grepSearchTool,
+      config: this.options.config,
+    };
+
+    for (const toolCall of toolCalls) {
+      switch (toolCall.name) {
+        case BuiltInToolNames.FileGlobSearch:
```
sestinj: There's repeated logic in each of the `case`s here. Could probably be broken out into a function to make it more concise and easier to maintain

sestinj (changes requested):
Prompting is great, the entire design makes sense, just a few refactoring notes

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
tingwai: I'll look into moving the logic.

The grep results currently show up as different context items
image

sestinj: ah okay that's perfect!

RomneyDa:
What is the difference between this and e.g. changing @codebase to be a built in prompt? I think it might be better if the user can see what's happening and have granular control over the tools, no need for extra calls under the hood.

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
```diff
@@ -155,4 +186,172 @@ export default class BaseRetrievalPipeline implements IRetrievalPipeline {
...
+    const requiredParams = tool.function.parameters?.required || [];
+    const properties = tool.function.parameters?.properties || {};
+    const paramDescriptions = requiredParams
+      .map(
+        (param: any) =>
+          `${param}: ${properties[param]?.description || "string"}`,
+      )
+      .join(", ");
+
+    return `- ${tool.function.name}: ${tool.function.description}
```
RomneyDa: Some descriptions are long, multiline, this might confuse the model

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
```diff
@@ -155,4 +186,172 @@ export default class BaseRetrievalPipeline implements IRetrievalPipeline {
...
+        case BuiltInToolNames.GrepSearch:
+          toolExtras.tool = grepSearchTool;
+          const grepContextItems = await grepSearchImpl(
+            toolCall.args,
+            toolExtras,
+          );
+          allContextItems.push(...grepContextItems);
+          break;
+
+        case BuiltInToolNames.LSTool:
```
RomneyDa: could probably just import the `callBuiltInTool` function from core

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
RomneyDa: It seems like an antipattern to not use tools here, but don't feel strongly about it

RomneyDa (commented):
Not assigned to this PR but excited about the feature

I think my general feedback would be
- The main advantage of this tool is that the model would be able to show targeted chunks/comments from/about the codebase without polluting the chat history
- In that sense, it's kind of a sub-agent command, and we should just prompt the model and let it run its course and then output relevant things at the end (maybe with a time limit/turns of conversation limit)
- It's an antipattern to request chat JSON when the model will support tools, and the approach is limiting because the model can't change course based on feedback from the codebase.

RomneyDa (commented):
ignore, duplciate

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
tingwai: That's a good point, I think there's also a balance between how much information is helpful. Eg I actually added some information to the glob search description because the LLM wouldn't search recursively unless I told it it can

core/context/retrieval/pipelines/BaseRetrievalPipeline.ts
sestinj: Most users that use @ codebase are likely to be using models that don't support tool calling

sestinj (approved):

PR #6439, opened 2025-07-03
Truncate large agent tool output

RomneyDa:
Continuation of #6206, truncates large outputs for the following tools:
- Limits the LS tool to 200 entries. In case of truncation when using `recursive` argument, prompts model to consider non recursive
- 

Adds a truncation warning context item in the case of truncation, to make it more obvious to the user (more or less same content for the agent this way)

These limits are not configurable for now, they are fairly generous limits designed to ensure agent works reasonably well. Personally I'd tighten them. Open to feedback on if people think it's too tight or how they'd like to configure them.
    

---

## Summary by cubic
Added output limits to agent tools (ls, fetchUrlContent, searchWeb, viewDiff) to prevent large responses and improve reliability. When output is truncated, a warning context item is included to notify users.

- **Bug Fixes**
  - Limits ls tool to 200 entries and suggests non-recursive search if truncated.
  - Truncates fetchUrlContent at 20,000 characters, searchWeb at 8,000 characters, and viewDiff at 5,000 lines.
  - Adds clear truncation warnings for each tool when limits are hit.

### DIFF
--- a/core/tools/implementations/fetchUrlContent.ts
+++ b/core/tools/implementations/fetchUrlContent.ts
@@ -1,6 +1,34 @@
 import { ToolImpl } from ".";
 import { getUrlContextItems } from "../../context/providers/URLContextProvider";
 
+const DEFAULT_FETCH_URL_CHAR_LIMIT = 20000;
+
 export const fetchUrlContentImpl: ToolImpl = async (args, extras) => {
-  return getUrlContextItems(args.url, extras.fetch);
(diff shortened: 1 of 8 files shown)


sestinj:
@RomneyDa are the comments from cubic here valid?

Patrick-Erichsen:
^Bump on Nate's comment

RomneyDa:
the import const one was valid. Fixed.

Patrick-Erichsen (approved):

PR #6475, opened 2025-07-05
Polishing find / replace tool (UI, prompt)

sestinj:
## Description

Adding to the rules section of the find/replace tool that it should always try to use multiple find/replace blocks in one tool call rather than splitting into multiple tool calls

## Checklist

---

## Summary by cubic
Updated the find/replace tool rules to encourage using multiple find/replace blocks in a single tool call instead of splitting edits across multiple calls.


core/tools/definitions/searchAndReplaceInFile.ts
```diff
@@ -56,7 +56,11 @@ Critical rules:
     * Include just the changing lines, and a few surrounding lines if needed for uniqueness.
     * Do not include long runs of unchanging lines in SEARCH/REPLACE blocks.
     * Each line must be complete. Never truncate lines mid-way through as this can cause matching failures.
-4. Special operations:
+4. Splitting up tool calls:
```
Patrick-Erichsen, line 59: This makes sense, but wondering if we should also either remove and/or improve the example below.

extensions/vscode/e2e/tests/GUI.test.ts
```diff
@@ -347,7 +347,7 @@ describe("GUI Test", () => {
   });
 
   describe("Context providers", () => {
-    it("should successfully use the terminal context provider", async () => {
+    it.only("should successfully use the terminal context provider", async () => {
```
Patrick-Erichsen: Is the `.only` intentional?

extensions/vscode/src/diff/vertical/handler.ts
```diff
@@ -174,9 +174,12 @@ export class VerticalDiffHandler implements vscode.Disposable {
...
       await this.insertDeletionBuffer();
 
-      await this.reapplyWithMyersDiff(diffLines);
+      const myersDiffs = await this.reapplyWithMyersDiff(diffLines);
 
-      const range = new vscode.Range(this.startLine, 0, this.startLine, 0);
+      // Scroll to the first diff
+      const scrollToLine =
+        this.getFirstChangedLine(myersDiffs) ?? this.startLine;
+      const range = new vscode.Range(scrollToLine, 0, scrollToLine, 0);
```
Patrick-Erichsen, line 182: This makes such a difference

Patrick-Erichsen (dismissed):

PR #6476, opened 2025-07-06
Allow tool call interruptions

RomneyDa:
## Description

https://github.com/continuedev/continue/pull/6333 added the robust message history to allow tool call interuptions, this PR actually allows users to interrupt, rejecting diffs if they are present.


Patrick-Erichsen (approved):
👌

PR #6480, opened 2025-07-06
fix: :ambulance: Put back agent mode system edit instructions

chezsmithy:
## Description

A recent change appears to have lost the lazy edit instructions for edit mode. This restores those instructions.

## Checklist

## Screenshots

## Tests

---

## Summary by cubic
Restored the missing edit instructions in agent mode to ensure users see the correct guidance when editing code.

### DIFF
--- a/core/llm/defaultSystemMessages.ts
+++ b/core/llm/defaultSystemMessages.ts
@@ -65,4 +65,5 @@ export const DEFAULT_AGENT_SYSTEM_MESSAGE = `\
   You are in agent mode.
 
 ${CODEBLOCK_FORMATTING_INSTRUCTIONS}
+${EDIT_CODE_INSTRUCTIONS}
 </important_rules>`;


chezsmithy:
@sestinj looks like a recent change lost the lazy edit instructions for agent mode. This appears to be causing an excessive number of full file replacements. I'm assuming it was accidental.

Patrick-Erichsen:
This was from a recent PR of mine to introduce the new "search/replace" tool for file editing. The lazy edit instructions conflict with how that tool instructs the model to output code diffs, so I moved the lazy edit instructions out of the sys msg and instead put it directly on the edit file tool: https://github.com/continuedev/continue/pull/6316/files#diff-16370bebe855efec0804f4f8cd0881404a7ea5f20847d2680ec3a4aa7518d04bR21

Interesting if it feels like this is causing more full file rewrites though. That is plausible since it's no longer in the sys msg.

@sestinj what are your thoughts on whether we'll be keeping the old edit file tool around (which uses apply models + lazy edit instructions) once we move search/replace out of experimental state? I could see us keeping it around indefinitely to support models that aren't strong at following of search/replace instructions. If that were the case, then based on @chezsmithy 's experience we should conditionally add the lazy instructions back into sys msg if the old edit file tool is being used.

chezsmithy:
Maybe enabling the new experimental tool should control how the system message gets created?

chezsmithy:
This might be a better topic for discord, but I'm curious in your vision for applies. If search and replace were to fail for example how would it fail over to other approaches? Curious.

sestinj:
It seems like the solution would be to put the lazy edit instructions into the edit file tool

sestinj:
And vision for applies follows two different paths:
1. (for the models that can't find and replace yet) we continue to encourage the usual
2. (for the models that can, which is mostly claude right now) we move entirely toward find / replace, except have a robust system of fallbacks. Right now it includes various matching algorithms at various levels of fuzziness. in the future it may include just passing off to a language model. But we think we can get failure rates so low that is doesn't matter

Obviously going to need a bit of feedback from folks as we work on improving this—it will need real-world usage

sestinj:
@chezsmithy The outcome of the conversation I think was to take a different approach, and I think another PR ended up being merged? I'm closing this under that assumption, but if I'm misremembering please lmk or re-open!

PR #6484, opened 2025-07-07
fix: Attach environment type to reported exceptions

exigow:
Motivation:
* Everything right now goes under "production" and it's hard to distinguish exceptions that I triggered myself during development from real ones.
* In Sentry, there is an issue called JobCancellationException, which I believe is related to ci/e2e tests. I want to recognize that.
* It's a common practice to distinguish environments, and it will be easy to hide non-production issues (with query like `?environment=production`).

Introduced types:
* production (default)
* dev (when `runIde`)
* e2e (when robot enabled)
    

---

## Summary by cubic
Exceptions reported to Sentry now include the environment type (production, dev, or e2e) to make it easier to filter and identify issues.


exigow:
>  Probably should have had this logic from the beginning

Yeah, definitely

PR #6495, opened 2025-07-07
fix: prevent UI freeze when ⌘I pressed without active editor

### LINKED ISSUE #6494: ⌘I without active editor freezes Ui causing unrecoverable state

### Relevant environment info

```Markdown
- OS: MacOS
- Continue version: v1.1.57
- IDE version: VS Code 1.101.2 

  OR link to assistant in Continue hub:
```

### Description

Edit session can be accessed through Chat Mode via `⌘I` and pressing `Esc` or clicking the `← Back to chat` can be used to exit an edit session.

However, when I press `⌘I` without any editor open, I get the following error message:

Image

Once this error is triggered, the UI becomes unresponsive. There's no keybinding or UI element to exit this frozen state, unable to start new chat sessions, view history, or dismiss the error. The only solution is to reload the window or restart VS Code

### To reproduce

https://github.com/user-attachments/assets/014eb44a-612b-4ada-99f8-ce2af64ee6d0

### Log output

```Shell

```

panyamkeerthana:
## Description

fixes #6494 

pressing ⌘I in chat mode to switch to edit mode when no editor is focused previously led to an unrecoverable error state in the secondary sidebar.

this commit ensures that ⌘I action does nothing unless an editor is focused, preventing the ui from freezing and allowing normal interaction with the panel.

## Screenshots

Screenshot 2025-07-07 at 2 53 55 PM

    

---

## Summary by cubic
Stops the UI from freezing when ⌘I is pressed without an active editor by making the shortcut do nothing in that case. This keeps the sidebar and panel working normally.

### DIFF
--- a/gui/src/components/mainInput/Lump/LumpToolbar/EditToolbar.tsx
+++ b/gui/src/components/mainInput/Lump/LumpToolbar/EditToolbar.tsx
@@ -30,9 +30,7 @@ export function EditToolbar() {
       <span className="truncate">
         Editing:{" "}
         <span className="italic">
-          {codeToEdit
-            ? getEditFilenameAndRangeText(codeToEdit)
-            : "⚠️ No file selected"}
+          {getEditFilenameAndRangeText(codeToEdit)}
         </span>
       </span>
     </div>
--- a/gui/src/redux/thunks/edit.ts
+++ b/gui/src/redux/thunks/edit.ts
@@ -133,6 +133,14 @@ export const enterEdit = createAsyncThunk<
       return;
     }
 
+    if (!state.editModeState.codeToEdit[0]) {
+      extra.ideMessenger.post("showToast", [
+        "info",
+        "Please open a file to use Edit mode",
+      ]);
+      return;
+    }
+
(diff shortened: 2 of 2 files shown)


RomneyDa (changes requested):
You can only edit an existing file, is there a different approach that solves the underlying issue? E.g. preventing edit mode with no active editor.

Patrick-Erichsen:
+1 to @RomneyDa 's feedback, I think we should just not switch into edit mode.

panyamkeerthana:
@RomneyDa ty for the suggestion and you're right! The fix just addresses the display issue in the edit toolbar when codeToEdit is undefined but it doesn't prevent the user from entering a edit session when there's no active editor. I'll look into the underlying issue and keep you posted!

panyamkeerthana:
@RomneyDa the latest commit prevents edit mode with no active editor and displays a toast w the message 

https://github.com/user-attachments/assets/52063b15-4b2a-4abe-83e1-cb51d6eded67

Patrick-Erichsen (approved):
Good call on the toast!

PR #6497, opened 2025-07-07
Make edit logging directories user-configurable

adarshramiyer:
## Description

Instead of somewhat flaky and hardcoded repo detection, made edit-logging directories user-configurable via .editlogdirs file in global ~/.continue folder.

## Checklist

---

## Summary by cubic
Edit logging directories are now user-configurable using a .editlogdirs file in the global ~/.continue folder, replacing hardcoded repo detection.

- **Refactors**
  - Logging only occurs for files in directories listed in .editlogdirs.
  - Updated logic to check directory membership before logging edits.

### DIFF
--- a/core/nextEdit/context/diffFormatting.ts
+++ b/core/nextEdit/context/diffFormatting.ts
@@ -12,7 +12,7 @@ export type BeforeAfterDiff = {
   afterContent: string;
 };
 
-export interface createDiffArgs {
+export interface CreateDiffArgs {
   beforeContent: string;
   afterContent: string;
   filePath: string;
@@ -26,7 +26,7 @@ export const createDiff = ({
   filePath,
   diffType,
   contextLines,
-}: createDiffArgs) => {
+}: CreateDiffArgs) => {
   switch (diffType) {
     case DiffFormatType.Unified:
       return createUnifiedDiff(


core/nextEdit/context/processNextEditData.ts
```diff
@@ -132,6 +132,7 @@ export const processNextEditData = async ({
       afterContent,
       filePath,
       DiffFormatType.Unified,
+      25, // 25 context lines; store a large number for trimming later on
```
Patrick-Erichsen: IMO this would be a good point to modify `createDiff` to instead take a single object as a param instead of _n_ single arg params.

Then would you would have `{..., contextLine: 25 }` which reads much more easily.

extensions/vscode/src/util/editLoggingUtils.ts
```diff
@@ -18,23 +22,53 @@ export const getBeforeCursorPos = (range: Range, activePos: Position) => {
   }
 };
 
-const getWorkspaceDirInfo = async (
-  event: vscode.TextDocumentChangeEvent,
-  ide: IDE,
-) => {
-  // gets the workspace dir uri and checks if it's the open-source continue repo
+const isFileWithinFolder = (fileUri: string, folderPath: string): boolean => {
```
Patrick-Erichsen: Feels like a quick unit test would be easy to one-shot here

extensions/vscode/src/util/editLoggingUtils.ts
```diff
@@ -18,23 +22,53 @@ export const getBeforeCursorPos = (range: Range, activePos: Position) => {
...
+    const { foundInDir } = findUriInDirs(fileUri, [normalizedFolderUri]);
+    return foundInDir !== null;
+  } catch (error) {
+    return false;
+  }
+};
+
+const isEditLoggingAllowed = async (editedFileURI: string) => {
+  const globalContinuePath = getContinueGlobalPath();
+  const editLogDirsPath = resolve(globalContinuePath, ".editlogdirs");
```
Patrick-Erichsen, line 26: I think we want to put this in `core/util/paths.ts`, we try to keep all the path logic centralized there

extensions/vscode/src/util/editLoggingUtils.ts
Patrick-Erichsen: Also, the purpose of this is just to check if there is a `.editlogdirs` file? How are we doing this for the `.staging`/`.local` files? I'd think that logic already exists

core/nextEdit/context/processNextEditData.ts
adarshramiyer: completed

extensions/vscode/src/util/editLoggingUtils.ts
adarshramiyer: filepath logic is moved to `core/util/paths.ts`

core/nextEdit/context/diffFormatting.ts
```diff
@@ -12,13 +12,21 @@ export type BeforeAfterDiff = {
...
 };
 
-export const createDiff = (
-  beforeContent: string,
-  afterContent: string,
-  filePath: string,
-  diffType: DiffFormatType,
-  contextLines: number = 3,
-) => {
+export interface createDiffArgs {
```
Patrick-Erichsen: Super minor nitpick but interfaces in TypeScript should be in `PascalCase`, eg `CreateDiffArgs`

Issue #6121, opened 2025-06-13
Gemini models in agent mode giving error 400 invalid argument


ShamanicArts:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 1.1.45 - 1.1.47(.vsix from github main)
- IDE version:1.100.2
- Model:Gemini-flash-2.0, Gemini-flash-2.5-05-20, Gemini-Pro-2.5-preview. issue persists across aistudio/vertex provider & openrouter provider
- config:
  

# A name and version for your configuration
name: shamanic-config
version: 0.0.1
schema: v1
openrouter_defaults: &openrouter_defaults
  provider: openrouter
  apiKey: ${{ secrets.OPENROUTER_API_KEY }}

rules:
  - You are an expert software developer. You give helpful and concise responses.


models:
  - name: OpenRouter LLaMA 70 8B
    provider: openrouter
    model: meta-llama/llama-3-70b-instruct
    apiKey: ${{ secrets.OPENROUTER_API_KEY }}
    roles: [chat, edit, apply] # 'models' from JSON get the chat/edit/apply roles by default
  - name: Claude 3.5 Sonnet
    provider: openrouter
    model: anthropic/claude-3.5-sonnet-latest
    apiKey: ${{ secrets.OPENROUTER_API_KEY }}
    roles: [chat, edit, apply]
  - name: Anthropic Claude 3.5 Sonnet Beta
    provider: openrouter
    model: anthropic/claude-3.5-sonnet:beta
    apiKey: ${{ secrets.OPENROUTER_API_KEY }}
    roles: [chat, edit, apply]
  - name: DeepSeek Chat
    provider: openrouter
    model: deepseek/deepseek-chat
    apiKey: ${{ secrets.OPENROUTER_API_KEY }}
    roles: [chat, edit, apply]
  - name: DeepSeek R1 0528
    provider: openrouter
    model: deepseek/deepseek-r1-0528
    apiKey: ${{ secrets.OPENROUTER_API_KEY }}
    roles: [chat, edit, apply]
[...]

PR #6501, opened 2025-07-08
test: check for required params on tool defs

Patrick-Erichsen:
## Description

Prevents https://github.com/continuedev/continue/issues/6121 from happening again by testing for required properties on our tool defs
    

---

## Summary by cubic
Added tests to ensure all tool definitions have required parameters defined, preventing missing property errors in the future. Refactored tool imports for better organization.


sestinj (changes requested):
this is great but it is failing. if only we had continuous ai...

PR #6507, opened 2025-07-08
fix: add current file context when setting is enabled

uinstinct:
## Description

- Add the CurrentFileContextProvider automatically when "add current file by default" setting is enabled.

Resolves CON-2683

## Checklist

## Screenshots

https://github.com/user-attachments/assets/6c2de80a-e9b6-4e1a-8597-98f8c0356cb8

https://github.com/user-attachments/assets/706b5fa0-3a28-4746-ae51-3b25b3bf4cd4

## Tests

---

## Summary by cubic
Automatically adds the CurrentFileContextProvider when the "add current file by default" setting is enabled. This ensures the current file context is included without manual setup.

### DIFF
--- a/core/config/profile/doLoadConfig.ts
+++ b/core/config/profile/doLoadConfig.ts
@@ -154,7 +154,13 @@ export default async function doLoadConfig(options: {
   newConfig.contextProviders.push(new RulesContextProvider({}));
 
   // Add current file as context if setting is enabled
-  if (newConfig.experimental?.useCurrentFileAsContext === true) {
+  if (
+    newConfig.experimental?.useCurrentFileAsContext === true &&
+    !newConfig.contextProviders.find(
+      (c) =>
+        c.description.title === CurrentFileContextProvider.description.title,
+    )
+  ) {
     newConfig.contextProviders.push(new CurrentFileContextProvider({}));
   }
 


RomneyDa (changes requested):
`modifyAnyConfigWithSharedConfig` should work for serialized configs as well, let's move this to `doLoadConfig`

It also needs to check if current file provider is already there, could cause duplicates

core/config/profile/doLoadConfig.ts
```diff
@@ -152,6 +153,11 @@ export default async function doLoadConfig(options: {
   newConfig.rules.unshift(...rules);
   newConfig.contextProviders.push(new RulesContextProvider({}));
 
+  // Add current file as context if setting is enabled
+  if (newConfig.experimental?.useCurrentFileAsContext === true) {
```
RomneyDa: @uinstinct have you verified that this would not result in duplicate context providers if they have "currentFile" in their config? I think it's added in loadYaml prior to this

RomneyDa (changes requested):
See comment!

core/config/profile/doLoadConfig.ts
uinstinct: you were right, this was adding duplicate providers. fixed it!

RomneyDa (approved):
Nice updates!

Issue #6454, opened 2025-07-04
Trailing "/" in startUrl of docs causes retrieving preindexed vectors from S3 bucket failed.


Jazzcort:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:
- Continue version:
- IDE version:
- Model:
- config:
  

  
  OR link to assistant in Continue hub:
```

### Description

Having startUrl as `https://react.dev` can't not retrieve the preindexed cache, but having startUrl as `https://react.dev/` can. This is because in the S3 bucket, the keys of the preindexed cache that are related to React have a trailing `_` (which is due to the replacing process that replace `/` with `_`). Here are all the keys that are related to React:

<img width="715" alt="Image" src="https://github.com/user-attachments/assets/c284049b-a499-4c22-946b-7ef60eb20862" />

Because we're asking for `all-MiniLM-L6-v2::500/https%3A__react.dev`, we end up having "key doesn't exist" error.

I've checked the S3 bucket. The keys with trailing `_` are listed below:

<img width="643" alt="Image" src="https://github.com/user-attachments/assets/399bf6b4-2355-4228-8737-dedb11fa4f69" />

P.S. There are some docs having both variants (with `_` and without `_`).  Most of the docs have the without `_` variant.

My propose to fix this bug is to:

1. Update the S3 bucket to unifying the key format. All the keys should not have trailing `_`. (This also helps to reduce the bucket size, since having both variants is redundant)
2. Sanitize all the startUrl from config to not have trailing `/` before we send the cache request to the S3 bucket.

I can fix the sanitizing part, but I need somebody who have the access to the S3 bucket to fix the first part. 😁

### To reproduce

1. Try adding doc with `https://react.dev` (should fail to retrieve cache)
2. Try adding doc with `https://react.dev/` (this one should succeed)

### Log output

```Shell

```

Patrick-Erichsen:
Nice writeup @Jazzcort !

[...]

PR #6512, opened 2025-07-08
Fix: add sanitizing process to strip off the trailing slash of the doc URL

Jazzcort:
## Description

In this patch, we add a sanitizing process to strip off the trailing forward slash (if any) in the given URL before generating the file path for S3 bucket. Therefore, both variants (url with / and without /) in the config file can be treated the same.

https://github.com/continuedev/continue/issues/6454

## Checklist

---

## Summary by cubic
Added a sanitizing step to remove trailing slashes from doc URLs before generating S3 file paths, so URLs with and without a slash are treated the same.

### DIFF
--- a/core/indexing/docs/DocsCache.ts
+++ b/core/indexing/docs/DocsCache.ts
@@ -7,6 +7,15 @@ export interface SiteIndexingResults {
   title: string;
 }
 
+/**
+ * Strip off the trailing forward slash from the given URL
+ * @param url The URL of the document
+ * @returns A sanitized URL without trailing forward slash
+ */
+const stripTrailingSlash = (url: string): string => {
+  return url.replace(/\/+$/, "");
+};
+
 export class DocsCache {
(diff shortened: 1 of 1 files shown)


RomneyDa (approved):
@Jazzcort non-recursive nitpick, could make it cleaner and nonrecursive with e.g.
`str.replace(/\/+$/, "");`

Jazzcort:
@RomneyDa You're right, that could be much clearer. Thanks for pointing it out. 😄

RomneyDa:
@Jazzcort tests are persistently failing for some reason on this, could you merge main?

Jazzcort:
@RomneyDa I rebased to the main but it still can not pass the jetbrains test. Are the current main be able to pass this test?

RomneyDa:
@Jazzcort I think this one also needs a main merge for tests!

Jazzcort:
@RomneyDa Passed all tests! Ready for being merged! 😁

Issue #5242, opened 2025-04-19
IntelliJ IDEA extension doesn't seem to follow the IntelliJ IDEA proxy settings


zyklone4096:
### Validations

- [ ] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

The IntelliJ IDEA extension doesn't seem to follow the IntelliJ IDEA proxy settings, which causes slow loading in my region (UI loading uses proxy but assistant config loading seems not)

on Windows 11 23H2, system proxy enabled before IntelliJ IDEA open, theres no log about continue in my VPN software, i have to enable TUN mode

zyklone4096:
nvm, working in 1.0.13

zyklone4096:
i misjudged it and its still not using a proxy

ReFlo:
I have the same issue. I used cntlm in verbose mode and there is no traffic appearing from that plugin. Additionally i used the proxy options in the config.yaml:

    requestOptions:
      proxy: localhost:3129

but then i get an HTTP 400 Bad Request Error: 

HTTP 400 Bad Request from https://<base>.openai.azure.com/v1/openai/deployments/gpt-4o/chat/completions?api-version=2025-01-01-preview <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd"> <HTML><HEAD><TITLE>Bad Request</TITLE> <META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD> <BODY><h2>Bad Request - Invalid URL</h2> <hr><p>HTTP Error 400. The request URL is invalid.</p> </BODY></HTML>

My full model config.yaml is:

  - name: GPT-4o-with-proxy
    provider: azure
    model: gpt-4o
    apiKey: <key>
    apiBase: https://<base>.openai.azure.com
    env:
      apiVersion: 2025-01-01-preview
      deployment: gpt-4o
      apiType: azure
    requestOptions:
      proxy: localhost:3129


Exactly the same config works with vscode and a the proxy set in the vscode settings.

tobiajung:
Same here, using a proxy works in VS Code, but not in JetBrains.
[...]

PR #6531, opened 2025-07-09
fix: Follow JetBrains IDE proxy settings

exigow:
This PR partially solves https://github.com/continuedev/continue/issues/5242 because some requests was ignoring the proxy config. Token requests now respect the config by replacing OkHttp with the built-in HttpRequests. HttpRequests are following IDE proxy config by default.

Also, we no longer have any usages of OkHttp, so I removed this dependency (with one exception: it is still used in tests).

## Testing

I tested this manually:

### Settings with local proxy

Selection_199

### Proxy view (Burp Suite)

Selection_198

Token requests are now visible.

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/auth/ContinueAuthService.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/auth/ContinueAuthService.kt
@@ -111,11 +111,10 @@ class ContinueAuthService {
         val gson = Gson()
         val jsonBody = gson.toJson(mapOf("refreshToken" to refreshToken))
         val url = getControlPlaneUrl() + "/auth/refresh"
+        // todo: test this manually before publishing PR
         val response = HttpRequests.post(url, HttpRequests.JSON_CONTENT_TYPE)
-            .connect { connection ->
-                connection.write(jsonBody.toByteArray())
-                connection.inputStream.reader().readText()
-            }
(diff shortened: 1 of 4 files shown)


extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/auth/ContinueAuthService.kt
```diff
@@ -77,73 +76,49 @@ class ContinueAuthService {
...
         coroutineScope.launch {
             while (true) {
                 val refreshToken = getRefreshToken()
                 if (refreshToken != null) {
                     updateRefreshToken(refreshToken)
                 }
-
-                kotlinx.coroutines.delay(15 * 60 * 1000) // 15 minutes in milliseconds
+                log.info("Token refreshed, retrying in 15 minutes")
+                delay(10.minutes)
```
Patrick-Erichsen:
```suggestion
                delay(15.minutes)
```

I think this should still be 15?

exigow: Ah, that's a typo, thanks

Fixed :+1:

exigow:
**Important note:** I wasn’t able to test the updated HTTP requests in `ContinueExtensionSettings.syncRemoteConfig` because this feature is not available to me. I’m fairly confident that this will work, but it’s probably worth testing just in case.

Patrick-Erichsen (approved):

PR #6532, opened 2025-07-09
feat: Next Edit JetBrains MVP for Internal Testing

jpoly1219:
## Description

Closes CON-2423, CON-2715, CON-2700, CON-2699, CON-2698, CON-2712, CON-2711, CON-2710, CON-2709

Basic MVP for JetBrains.
- Window for next edit predictions.
- Tab to accept, esc to reject.
- Ghost text for FIM.
- Add next edit message types.
- Logging.

## Checklist

## Screenshots

## Tests

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/nextEdit/NextEditService.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/nextEdit/NextEditService.kt
@@ -1,5 +1,6 @@
 package com.github.continuedev.continueintellijextension.nextEdit
 
+import com.github.continuedev.continueintellijextension.FimResult
 import com.github.continuedev.continueintellijextension.autocomplete.AutocompleteLookupListener
 import com.github.continuedev.continueintellijextension.autocomplete.AutocompleteSpinnerWidget
 import com.github.continuedev.continueintellijextension.autocomplete.ContinueInlayRenderer
@@ -8,6 +9,7 @@ import com.github.continuedev.continueintellijextension.autocomplete.addInlayEle
 import com.github.continuedev.continueintellijextension.autocomplete.isInjectedText
 import com.github.continuedev.continueintellijextension.services.ContinueExtensionSettings
 import com.github.continuedev.continueintellijextension.services.ContinuePluginService
+import com.github.continuedev.continueintellijextension.utils.checkFim
 import com.github.continuedev.continueintellijextension.utils.toUriOrNull
 import com.github.continuedev.continueintellijextension.utils.uuid
 import com.intellij.openapi.application.ApplicationManager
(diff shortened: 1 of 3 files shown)


extensions/intellij/src/main/resources/META-INF/plugin.xml
```diff
@@ -169,6 +169,13 @@
             <keyboard-shortcut keymap="Mac OS X" first-keystroke="TAB"/>
         </action>
 
+        <action id="com.github.continuedev.continueintellijextension.nextEdit.AcceptNextEditAction"
+                class="com.github.continuedev.continueintellijextension.nextEdit.AcceptNextEditAction"
+                text="Accept Autocomplete Suggestion" description="Accept Next Edit Prediction">
```
Patrick-Erichsen, line 181:
```suggestion
                text="Accept Next Edit Prediction" description="Accept Next Edit Prediction">
```

?

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/nextEdit/NextEditService.kt
```diff
@@ -0,0 +1,400 @@
...
+    }
+    /**
+     * Check if the diff is indeed a FIM (Fill-In-Middle).
+     * @param oldEditRange Original string content.
+     * @param newEditRange New string content.
+     * @param cursorPosition The position of the cursor in the old string.
+     * @return A Pair where the first value is a boolean indicating if the change is purely additive (FIM)
+     *         and the second value is the FIM text content or null if not a FIM.
+     */
+    private fun checkFim(
```
Patrick-Erichsen: I think this one should be relatively easy to unit test

Patrick-Erichsen (commented):
Worked well for me aside from the one bug I showed 👍 

@exigow - would you mind taking a glance at this to see if you have any feedback on lifecycle things?

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteEditorListener.kt
```diff
@@ -46,12 +59,39 @@ class AutocompleteDocumentListener(private val editorManager: FileEditorManager,
...
+            null,
+            null,
+            (request@ { response ->
+                val optInNextEditFeature = response.castNestedOrNull<Boolean>(
+                    "content",
+                    "result",
+                    "config",
+                    "experimental",
+                    "optInNextEditFeature"
+                ) ?: return@request
```
exigow: I think this elvis with return statement prevents us from entering the else branch when `optInNextEditFeature` is null.

I checked this on my config after rebuilding the core, and completions are not working because I do not have any `optInNextEditFeature` in the JSON. :sweat_smile: 

It should be `?: false` instead of `?: return@request`.

exigow: @jpoly1219, BTW how can I opt-in this feature? I enabled all the experimental settings in my toolbar, but none of them are called `optInNextEditFeature`.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteEditorListener.kt
```diff
@@ -46,12 +59,39 @@ class AutocompleteDocumentListener(private val editorManager: FileEditorManager,
...
+        val continuePluginService = editor.project?.service<ContinuePluginService>()
+        if (continuePluginService == null) {
+            return
         }
+
+        continuePluginService.coreMessenger?.request(
+            "config/getSerializedProfileInfo",
+            null,
+            null,
+            (request@ { response ->
```
exigow: Every keystroke triggers this request. I’m not sure if we are OK with that.

From what I see, we pool this `getSerializedProfileInfo` in few places. Maybe it’s time to add a class/service that caches this? Or maybe we could invert this so that instead of polling, we receive a message from the core when a change occurs? IMO changes in profile are rare.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteEditorListener.kt
```diff
@@ -46,12 +59,39 @@ class AutocompleteDocumentListener(private val editorManager: FileEditorManager,
...
+                        // Invoke later is important, otherwise the completion will be triggered before the document is updated
+                        // causing the old caret offset to be used
+                        // TODO: concurrency
+                        service.triggerCompletion(editor)
+                    }
+                }
+            })
+        )
     }
 }
```
exigow, line 113: Threading seems to be OK.
* Request is started on AWT, but writes are handled in non-blocking way (queue) - OK
* Response is handled on background thread - OK
* With invokeLater we're going back to AWT - OK
* triggerNextEdit/triggerCompletion - i think it's OK. We're working on AWT but we do only inlay rendering (which is UI-related stuff)

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/nextEdit/NextEditWindowService.kt
```diff
@@ -0,0 +1,304 @@
...
+
+    private fun getFileLanguage(document: Document): String {
+        val virtualFile = FileDocumentManager.getInstance().getFile(document)
+        return virtualFile?.extension?.lowercase() ?: "text"
+    }
+
+    companion object {
+        fun getInstance(project: Project): NextEditWindowService = project.service()
+    }
+}
```
exigow, line 304: This `NextEditService` class shares large portions of code with `AutocompleteService`. I compared both files and noticed a lot of copy-paste. :smile:  Shouldn't we extract some common parts into an abstract class or utility methods?

I’m especially thinking about methods like `deduplicateCompletion`, requests, `triggerNextEdit` (which is basically a copy of `triggerCompletion` + extra code), and probably others as well. I’m not sure it’s a good idea to maintain these two features in this way (long-term).

exigow: BTW, from what I understand, this is for internal testing, so I don’t think my concerns are an issue. :+1: But if we want to enable it at some point, it definitely needs a refactor.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteEditorListener.kt
jpoly1219: Currently we only have it for continue.dev emails! @sestinj can we add @exigow? I can make the necessary changes.

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteEditorListener.kt
jpoly1219: Yeah that's probably a good idea :+1:

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/nextEdit/NextEditWindowService.kt
jpoly1219: You raise legitimate concerns. The purpose of this PR was to test if the feature was possible in JetBrains, and I was planning to add more polish to it in coming PRs. We definitely should address all the above concerns before the actual release!

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/autocomplete/AutocompleteEditorListener.kt
jpoly1219: > I think this elvis with return statement prevents us from entering the else branch when `optInNextEditFeature` is null.
> 
> I checked this on my config after rebuilding the core, and completions are not working because I do not have any `optInNextEditFeature` in the JSON. 😅
> 
> It should be `?: false` instead of `?: return@request`.

Ah so this must be the reason why displayCompletion test was failing. Thanks for pointing it out :pray:

Patrick-Erichsen (approved):

PR #6539, opened 2025-07-09
Revert "chore: terminate all tasks after debug"

Patrick-Erichsen:
Reverts continuedev/continue#6405

This is causing an issue where every time you reload the debug window, all the processes are terminated.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,7 +19,6 @@
       "pauseForSourceMap": false,
       "outFiles": ["${workspaceFolder}/extensions/vscode/out/extension.js"],
       "preLaunchTask": "vscode-extension:build",
-      "postDebugTask": "Terminate All Tasks",
       "env": {
         // "CONTROL_PLANE_ENV": "local",
         "CONTINUE_GLOBAL_DIR": "${workspaceFolder}/extensions/.continue-debug"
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
@uinstinct reverting the commit for now, but it would still be nice to have the original feature that you were trying to do here, so might be worth a second shot! 

FWIW, here's what Claude 4 had to say about the issue, seems like it might not be possible if this is true:

```
The problem is that the postDebugTask is set to "Terminate All Tasks" which terminates all tasks when the debug session ends. However, when you reload the debug window (as opposed to fully stopping it), VS Code is treating this as the debug session ending and triggering the postDebugTask.

Let me check if there's a way to distinguish between a reload and a full stop in VS Code's debug configuration, or if we need to modify the approach.

Looking at the current configuration:

The postDebugTask is set to "Terminate All Tasks"
The "Terminate All Tasks" task uses workbench.action.tasks.terminate with terminateAll argument
This terminates ALL tasks, but apparently one task (npm run esbuild-notify) survives
The issue is that VS Code doesn't distinguish between a reload and a stop when it comes to postDebugTask. Both actions trigger the post-debug task.
```

RomneyDa (approved):

PR #6544, opened 2025-07-10
fix: :ambulance: hotfix-bedrock-provider

chezsmithy:
## Description

An alternate hotfix for the duplicated tool argument issue. This aligns the Bedrock provider with the Anthropic and OpenAI providers in that tool delta messages aren't aggreated in the provider but are instead streamed to the client where they are aggreated.

## Checklist

## Screenshots

## Tests

Manual testing has been performed.
TODO: Evaluating impacts to tests.

    

---

## Summary by cubic
Fixed duplicated tool argument issue in the Bedrock provider by streaming tool delta messages to the client for aggregation, matching the behavior of Anthropic and OpenAI providers.

### DIFF
--- a/core/llm/llms/Bedrock.ts
+++ b/core/llm/llms/Bedrock.ts
@@ -54,7 +54,6 @@ class Bedrock extends BaseLLM {
     profile: "bedrock",
   };
 
-  private _currentToolResponse: Partial<ToolUseState> | null = null;
   private _promptCachingMetrics: PromptCachingMetrics = {
     cacheReadInputTokens: 0,
     cacheWriteInputTokens: 0,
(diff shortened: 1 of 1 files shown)


chezsmithy:
@RomneyDa here is a slightly different approach we might want as well to align Bedrock more closely with other providers that stream toolCall deltas.

RomneyDa:
@chezsmithy this is great. do we know why/when this error started happening? I made the openai adapter but this fix seems to be for the LLM class. Did bedrock change streaming approach? Is it only for some models?

chezsmithy:
@RomneyDa its related to the issue in the preview release you fixed. The bedrock provider was collecting tool call state and then sending it all at once. This aligns it with other providers where it streams it instead.

RomneyDa:
Nice, see https://github.com/continuedev/continue/pull/6559/
Will get these into prerlease

chezsmithy:
> Nice, see #6559 Will get these into prerlease

Thank you! Closing.

PR #6556, opened 2025-07-10
feat: Allow local assistants for on prem

RomneyDa:
## Description
Reverts a change that disabled local assistants for on prem setups


sestinj:
Outcome here was we don't want to allow this. Proxy should be only source of the assistant. Later we may enable custom policies

PR #6560, opened 2025-07-10
Add retry logic with exponential backoff to listOrganizations for user account creation delay

sestinj:
Solves https://linear.app/continue/issue/CON-2425/vs-code-login-issues-where-organization-does-not-appear-after-login by retrying /list-organizations while waiting for the user creation webhook to run. This is important for newly created accounts in domain verified organizations

CON-2425
    

## Summary by cubic
Added retry logic with exponential backoff to listOrganizations to handle delays in user account creation and ensure organizations appear after signup.

- **Bug Fixes**
 - Retries up to 5 times on 404 errors, waiting longer between each attempt.


sestinj:
maxWaitTime logic is intentional (supposed to be overall wait time). Made updates on the others and cleaned up the code

RomneyDa (changes requested):
@sestinj looks like `maxRetries` limit is not checked, should that be added?

chezsmithy:
Could we share retry logic with a decorator strategy?

https://github.com/continuedev/continue/pull/6596

PR #6564, opened 2025-07-10
fix: :mute: The safeParseToolCallArgs function is logging excessively

chezsmithy:
## Description

The console.error statement in this function is getting triggered on every delta reutrn that is incomplete as it won't parse as valid Json. This is filling the logs full of parsing errors.

Recommend removing this code entirely as the catch block is expected up until the point the full argument is returned as valid JSON.

I'm curious if there are issues here, if the arguments parse at some interim point and the tool call proceeds too early. We aren't using stop messages from the provdiers to detect when when the toolCall is full generated and this might be causing premature tool exeuction?

## Checklist

## Screenshots

## Tests

---

## Summary by cubic
Removed excessive error logging from safeParseToolCallArgs to prevent log spam when parsing incomplete tool call arguments.

### DIFF
--- a/core/tools/parseArgs.ts
+++ b/core/tools/parseArgs.ts
@@ -6,9 +6,9 @@ export function safeParseToolCallArgs(
   try {
     return JSON.parse(toolCall.function?.arguments?.trim() || "{}");
   } catch (e) {
-    console.error(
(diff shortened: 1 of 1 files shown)


chezsmithy:
@RomneyDa the changes to have the providers stream make sense. I wonder if there are dragons here if we don't use the stop messages to actually know when the toolCall is fully formed. Could this parse prematurely allowing the tool to proceed? Hmm.....

PR #6566, opened 2025-07-10
Fix cmd backspace for pending tool calls

RomneyDa:
## Description
React hook was in the wrong place
    

---

## Summary by cubic
Fixed an issue where the cmd+backspace shortcut did not work for pending tool calls by moving the React hook to the correct place.


Patrick-Erichsen:
I bet cubic would have caught this

PR #6576, opened 2025-07-11
fix: Ignore invalid JSON messages (like NODE_ENV undefined)

exigow:
Fixes: https://continue-xd.sentry.io/issues/6731751693
    

---

## Summary by cubic
Invalid JSON messages (like when NODE_ENV is undefined) are now ignored instead of causing errors in the plugin.

- **Bug Fixes**
 - Added error handling to skip and log invalid JSON messages.


exigow:
Marking as draft because I need to fix TODOs + check this changes manually on Monday (we have issues with infra today).

PR #6586, opened 2025-07-12
feat: :zap: Improve search and release tool by supporting multiple diffs per tool call

chezsmithy:
## Description

Looking at the current implementation I noted the tool definition expects the LLM to return mulltiple search and replace blocks into a single "diff" parameter. While monitoring the tool, I noted that rather than doing this, it's firing off many requests to the tool for small edits. By enhancing the tool to support an array of diffs, the tool now gets multiple search and replace blocks which are executed in sequence.

The sequence of tool calls is a pattern I've observed when testing the new tool. It's unfortunately stressing our AWS call limits as well because it's eating up the per minute call quotas. When in a tight edit loop this happens extremely rapidly across only a few users.

## Checklist

## Screenshots

## Tests

I've added new tests to account for the change to the tool definition.

I've found a pretty decent way to trigger the multiple search and replace block experience in the test repo. Use the prompt "Add a double, triple and power function to @test.js, and insert them alphabetized" you seem to get multiple requests.

When attempting this same prompt without this change I noted either a single large search and replace block covering the whole file, or 3 individual back to back tool calls.

### DIFF
--- a/core/tools/definitions/searchAndReplaceInFile.ts
+++ b/core/tools/definitions/searchAndReplaceInFile.ts
@@ -3,7 +3,7 @@ import { BUILT_IN_GROUP_NAME, BuiltInToolNames } from "../builtIn";
 
 export interface SearchAndReplaceInFileArgs {
   filepath: string;
-  diff: string;
+  diffs: string[];
 }
 
 /**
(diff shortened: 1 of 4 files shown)


chezsmithy:
@Patrick-Erichsen FYI from my discord comments in the contributor channel.

Patrick-Erichsen (dismissed):
Hoping this gets the model to be more eager to output multiple blocks!

chezsmithy:
@Patrick-Erichsen ready for review. The one failing nextEdit test might not be related?

RomneyDa:
@Patrick-Erichsen @sestinj general +1 to this approach, I think models good enough to do search and replace will definitely be good enough to call the tool with an array. We could also use this array to simplify parsing in the future, @chezsmithy looks like you went for the simple approach of join the blocks with "\n\n" and leave parsing the same for now which makes sense.

chezsmithy:
> @Patrick-Erichsen @sestinj general +1 to this approach, I think models good enough to do search and replace will definitely be good enough to call the tool with an array. We could also use this array to simplify parsing in the future, @chezsmithy looks like you went for the simple approach of join the blocks with "\n\n" and leave parsing the same for now which makes sense.

@RomneyDa  I tried to keep the change as simple as possible just in case we needed to revert based on real life experience.

Patrick-Erichsen:
Resolved a merge conflict here caused by my parallel tool calling PR.

Patrick-Erichsen (approved):

PR #6590, opened 2025-07-12
chore: add ideInfo headers to hub reqs

Patrick-Erichsen:


extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
```diff
@@ -86,7 +87,8 @@ class IntelliJIDE(
             name = ideName,
             version = ideVersion,
             remoteName = remoteName,
-            extensionVersion = extensionVersion
+            extensionVersion = extensionVersion,
+            isPrerelease = false // TODO: Implement prerelease detection for JetBrains
```
Patrick-Erichsen, line 91: @exigow do you know of a way to determine the release channel of an installed extension?

Our versioning schema for JB doesn't have any context on whether a given version is an EAP or stable release, they're all just `1.0.X`, so not sure if we have a way to determine this.

RomneyDa:
@Patrick-Erichsen will wait on @exigow's input before reviewing

extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
exigow: @Patrick-Erichsen, AFAIK context like this is never added automatically.

I see two options:
* We could add postfix like `1.0.0-eap` when building the plugin. In the projects I worked on, this was exactly how we solved it.
* We could add custom metadata to `plugin.xml`: we could read the channel like `publishPlugin` (from env vars) and add a custom tag like `<channel>` inside `patchPluginXml`. Then we could read this value from the code (I think it should be possible).

Patrick-Erichsen: I like option 1 of adding the `-eap` suffix. This would be helpful to make the release versions clearer in general imo. 

I will make a Linear issue for this, but think it's okay to merge this in the meantime and just have the incorrect log info.

RomneyDa (approved):

PR #6593, opened 2025-07-12
[WIP] feat: task lists

uinstinct:
## Description

Task list in the assistant which adds and shows what task is being currently executed by the agent. Useful to plan and add more tasks while the agent is executing.

resolves CON-2763

## Checklist

## Screenshots

image

## Tests

---

## Summary by cubic
Added a task list feature that lets users view and manage tasks being executed by the assistant, including a new UI section to display current and completed tasks.

- **New Features**
  - Introduced backend task manager for adding, updating, removing, and listing tasks.
  - Added a Tasks section in the UI to show the current task queue and statuses.


core/tools/implementations/taskList.ts
```diff
@@ -0,0 +1,90 @@
+import { ToolImpl } from ".";
+import { ContextItem } from "../..";
+import { getTaskManagerForSession } from "../../context/taskList";
+
+export const taskListImpl: ToolImpl = async (args, extras) => {
+  let contextItem: ContextItem;
+  const manager = getTaskManagerForSession("abcd", extras.messenger!);
```
uinstinct, line 8: how to get the `IDEMessenger` here with the send method implemented?

uinstinct: I want to implement the event receiving [here](https://github.com/continuedev/continue/pull/6593/files#diff-c357ac89ab97e72b38c0536a5c8d4fbddb6ff6c4968ec3ac0c1ba91567273fc3R78-R107)

core/tools/implementations/taskList.ts
```diff
@@ -0,0 +1,90 @@
+import { ToolImpl } from ".";
+import { ContextItem } from "../..";
+import { getTaskManagerForSession } from "../../context/taskList";
+
+export const taskListImpl: ToolImpl = async (args, extras) => {
```
uinstinct, line 6: how to get the session id in the tool call?

tomasz-stefaniak: @uinstinct I'd probably pass session id in  `ToolExtras`. You will need to modify `ToolExtras` payload to make it happen

chezsmithy:
This is super cool. The new AWS tool Kiro suggests this kind of spec or task driven development. Love the idea of maybe using a markdown file to drive the work somehow as well.

Could you quickly attach a session to a task to drive the notion of lowering context before starting the next task. 

Also curious if there might be a way in the future to match a model to a task. Lower cost, faster models for simple tasks maybe.

Love the possibilities!

uinstinct:
> Also curious if there might be a way in the future to match a model to a task. Lower cost, faster models for simple tasks maybe.

Awesome idea!!!
I will have this in an upcoming improvement.


> Could you quickly attach a session to a task to drive the notion of lowering context before starting the next task.

I am thinking of having each session its own task list. Hence the entire context would be passed for each task. By lowering the context do you mean for independent and non-related tasks, one task should not use other's context?

chezsmithy:
Yes. Thinking tasks have discrete context where possible to encourage healthy context management. Or maybe it's an option?

Tasks by repo might also be interesting so you could store your task list as markdown and restart at a later point in time or share the task list with another developer.

uinstinct:
Yes something like [how Kiro does](https://kiro.dev/docs/specs/concepts/)

core/tools/implementations/taskList.ts
uinstinct: `extras.messenger` is available as verified on call!

---

request for the sessionId can be done using `extras.messenger.request("some-session-id-method")`

sestinj:
I think this has gone stale and it might make sense to open a separate PR since there are likely to be big differences

Issue #5709, opened 2025-05-17
[Feature Request] Add Thinking Mode Toggle and Filter `<think>`


NGC13009:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Qwen3, like Claude, supports enabling or disabling the thinking mode at any time, and it is a model that can be deployed locally via Ollama. However, continue still has not provided a convenient option to enable or disable thinking at any time. Users still need to manually input `/nothink` to control the model, which is very inconvenient. It is hoped that continue will provide a convenient switch to control the model's thinking mode when calling Ollama for local execution.

In addition, I want to configure qwen3 as an autocomplete model, but the current continue's autocomplete does not filter the `<think>` tag. Even when the thinking mode is turned off, ollama will return an empty `<think>` tag, and continue will include them in the response, filling them into the code, making autocomplete unavailable.

This issue also appears in the problem of automatically generating titles.

![Image](https://github.com/user-attachments/assets/03ca8488-9786-40b8-8bf4-9c90d7aceb61)

### Solution

In summary, three issues are reflected here:
1. A quick toggle switch for switching models provided by ollama is needed.
2. The switchable model is currently not properly configured as an autocomplete model.
3. The current model cannot generate the title and abstract correctly.

NGC13009:
@tomasz-stefaniak I have tried these settings, but they cannot resolve the problems I encountered.

I would like to add a thinking toggle button, but for now I have to default to disabling thinking by adding `/nothink` to the system prompt, and enable thinking by using `/think`. This method works, but it's not convenient.

Moreover, it completely fails to address the issue of outputting the `</think>` tag when using autocomplete. This can only be resolved by hoping that the continue plugin will filter out this suggestion (i.e., the initial pair of `<think>` tags that the model produces at the beginning), while not filtering out subsequent tags, as those might be part of the normal output.

The current version (v1.0.7) still cannot correctly retrieve the title from the history using thinking-type models.

wenyifancc:
> [@NGC13009](https://github.com/NGC13009) does the `reasoning` setting at least partially fix the issues you're having?
> 
> https://docs.continue.dev/json-reference#completionoptions
[...]

PR #6594, opened 2025-07-12
Qwen3 support for autocomplete function

NGC13009:
## Description

Abstract:

1. Added support for the autocomplete feature for the qwen3 model.
2. The documentation has been updated to guide how to configure the autocomplete feature for models with switchable toggle-like model for the thinking function.

In the case of personal deployment models based on ollama, being able to accomplish more tasks with fewer models can significantly save GPU memory. For example, if I only need to run a `qwen3:4b` model to handle all functions such as `chat`, `autocomplete`, and `edit`, etc. There is no need to launch another `qwen2.5-coder:1.5b` model in the background.

The previous autocomplete feature in continue was not well matched with `qwen3`, often outputting tags resembling #5709 , one of the issues mentioned. This change will fix this problem, allowing the thinking process of the `qwen3` series, which can switch models, to be correctly applied to the autocomplete feature.

## Checklist

- [o] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [o] The relevant docs, if any, have been updated or created
- [o] The relevant tests, if any, have been updated or created

## Screenshots

No visual updates.

## Tests

1. Only modified the filter a bit, removing extra tags when the model name contains the `qwen3` character.
2. Use as few filtering tags as possible, theoretically not causing the filtering of content that should be output.
3. Tested on common code or text completion tasks, it performs well, can be used normally, and produces no additional output.
    

---

## Summary by cubic
Added support for Qwen3 models in the autocomplete feature, removing unwanted thinking tags from completions and updating documentation to show how to configure Qwen3 for autocomplete.

- **Bug Fixes**
  - Filters out special <think> tags from Qwen3 autocomplete output to prevent extra text.

- **Docs**
  - Added instructions for using Qwen3 models with autocomplete and disabling thinking mode.

### DIFF
--- a/core/autocomplete/postprocessing/index.ts
+++ b/core/autocomplete/postprocessing/index.ts
@@ -103,6 +103,16 @@ export function postprocessCompletion({
     }
   }
 
+  if (llm.model.includes("qwen3")) {
+    // Qwen3 always starts from special thinking markers, and we don't want them to output these contents
+    // Remove all content from "
+    completion = completion.replace(/<think>.*?<\/think>/s, "");
+    completion = completion.replace(/<\/think>/, "");
+
+    // Remove any number of newline characters at the beginning and end
+    completion = completion.replace(/^\n+|\n+$/g, "");
+  }
+
   if (llm.model.includes("mercury") || llm.model.includes("granite")) {
     // Granite tends to repeat the start of the line in the completion output
     let prefixEnd = prefix.split("\n").pop();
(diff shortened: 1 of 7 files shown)


NGC13009:
Why does the automatic test fail on jetbrains?? And it seems to be a UI error, but only three lines of code were added, and all of them are unrelated to the UI. In theory, it should not cause any new issues. This error should not be caused by this change. I don't know how to fix this problem.

RomneyDa (approved):
Removes think tags when using qwen3 for autocomplete

PR #6595, opened 2025-07-12
Config loading fixes

RomneyDa:
- Awaits initial refresh attempt when getting sessions (vs code) - fixes duplicate cascade caused by immediate session update from refreshing
- Fixes laggy assistant dropdown updating when changing assistants - adds optimistic selected profile update
- Abort current cascade init when ideSettings/session is changed - adds signals to control plane client methods
- Remove duplicate profiles refresh in gui Auth Provider on `sessionUpdate`
- Add "reason" logs for all config reloads (log line commented out for now)
- When loading config, wait till cascade init has been initialized/completed once
  - this prevents potentially useless initial load AND reload caused by its MCP server followup reload
- Await config handler cascade init initialized when loading config
  - this cancels auth-dependent calls when auth changes - better security
- Fix reload config button styling that caused it to open hub when clicked

### DIFF
--- a/core/config/ConfigHandler.ts
+++ b/core/config/ConfigHandler.ts
@@ -44,6 +44,17 @@ export class ConfigHandler {
   private organizations: OrgWithProfiles[] = [];
   currentProfile: ProfileLifecycleManager | null;
   currentOrg: OrgWithProfiles;
+  totalConfigLoads: number = 0;
+  public isInitialized: Promise<void>;
+  private setInitialized:
+    | undefined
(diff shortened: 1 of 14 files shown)


sestinj:
I just realized I can mark other peoples' PRs as draft. Did that here because I think there's at least more to do in failing tests + cubic feedback. But will take it as a signal to review again once it's out of draft

Patrick-Erichsen (approved):
Nice, having all these reasons will be super helpful 👍  We really should just be setting log levels that we can use to see more verbose logs like this without modifying source in the future though, would also enable end user troubleshooting

PR #6603, opened 2025-07-13
Codebase/Colocated Rules Cache

RomneyDa:
## Description

A major source of slow config loading is walk dir to find colocated rules.md files. This PR adds a codebase/colocated rules cache that initializes on startup and then updates and triggers a config reload when a rules.md file is edited. This way it's not doing a full walk dir for any config reload.

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -52,6 +52,7 @@ import { BLOCK_TYPES, ConfigYaml } from "@continuedev/config-yaml";
 import { getDiffFn, GitDiffCache } from "./autocomplete/snippets/gitDiffCache";
 import { stringifyMcpPrompt } from "./commands/slash/mcpSlashCommand";
 import { isLocalDefinitionFile } from "./config/loadLocalAssistants";
+import { CodebaseRulesCache } from "./config/markdown/loadCodebaseRules";
 import {
   setupLocalConfig,
   setupProviderConfig,
@@ -233,6 +234,14 @@ export class Core {
       (..._) => Promise.resolve([]),
     );
 
+    const codebaseRulesCache = CodebaseRulesCache.getInstance();
+    codebaseRulesCache
+      .refresh(ide)
+      .catch((e) => console.error("Failed to initialize colocated rules cache"))
+      .then(() => {
+        this.configHandler.reloadConfig();
+      });
+
     this.registerMessageHandlers(ideSettingsPromise);
   }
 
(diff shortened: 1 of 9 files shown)


gui/src/components/mainInput/belowMainInput/RulesPeek.tsx
```diff
@@ -27,6 +27,8 @@ const getSourceLabel = (source: RuleSource): string => {
       return "Model Agent Options";
     case "rules-block":
       return "Rules Block";
+    case "colocated-markdown":
+      return "Codebase Rule";
```
Patrick-Erichsen: I have a feeling that in most cases, it would be more useful to just show the user the path to the rule than give a vague term like "Project Rules" or "Codebase Rule". 

We still need a few terms for things that don't have a useful URI, eg "Default Chat", but if we have a URI I would think that's a better piece of information to show.

core/core.ts
```diff
@@ -966,15 +977,21 @@ export class Core {
...
+          BLOCK_TYPES.some((blockType) =>
+            uri.includes(`.continue/${blockType}`),
           )
         ) {
           await this.configHandler.reloadConfig();
+        } else if (uri.endsWith(RULES_MARKDOWN_FILENAME)) {
+          try {
+            const codebaseRulesCache = CodebaseRulesCache.getInstance();
+            await codebaseRulesCache.update(this.ide, uri);
+            this.configHandler.reloadConfig();
```
Patrick-Erichsen: I wonder if we should start passing some sort of "reason" enum to reloadConfig so that we could instead move this type of logic into there

Patrick-Erichsen (dismissed):
Nice fix! Few comments I wanted to run by you but this is ready to merge.

I think one risk here is that if a user deletes a rule, it will still be in the cache? Seems somewhat edge case-y though, and probably is the same behavior we already have for other files in a workspace .continue

core/core.ts
RomneyDa: What would the use case be? For now there is no logic associated with reason, only logging

RomneyDa:
> Nice fix! Few comments I wanted to run by you but this is ready to merge.
> 
> I think one risk here is that if a user deletes a rule, it will still be in the cache? Seems somewhat edge case-y though, and probably is the same behavior we already have for other files in a workspace .continue

Added rule removal on deletion

Patrick-Erichsen (approved):

gui/src/components/mainInput/belowMainInput/RulesPeek.tsx
RomneyDa: Updated! Either last 2 path parts or "rules.md"

Issue #6621, opened 2025-07-15
qwen3 provided by ollama failed to call tools


leozzyzheng:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows11 24H2
- Continue version: v
- IDE version: VSCode 1.102.0
- Model: qwen3 8b
- config:

name: Local Assistant
version: 1.0.0
schema: v1
models:
  - name: Qwen3 8B
    provider: ollama
    model: qwen3:8b
    apiBase: http://localhost:11434/
    roles:
      - autocomplete
      - chat
      - edit
      - apply
    capabilities:
      - tool_use
context:
  - provider: code
  - provider: docs
  - provider: diff
  - provider: terminal
  - provider: problems
  - provider: folder
  - provider: codebase
  - provider: url
```

### Description

Continue failed to call tools when using qwen3 provided by ollama, see screenshot in the following.

### To reproduce

1. Config ollama and install qwen3:8b
2. Config YAML add tool_use
3. Use Agent mode to ask weather
4. Never see Continue is generating output to search the web for "weather in New York" completed.

<img width="2311" height="1494" alt="Image" src="https://github.com/user-attachments/assets/76cca894-8e3a-4aa9-aefc-104991cc2cf2" />

### Log output

```Shell

```

leozzyzheng:
The original response from ollama

```
{"model":"qwen3:8b","created_at":"2025-07-15T03:22:43.542760407Z","message":{"role":"assistant","content":"\u003cthink\u003e"},"done":false}
{"model":"qwen3:8b","created_at":"2025-07-15T03:22:43.601422785Z","message":{"role":"assistant","content":"\n"},"done":false}
{"model":"qwen3:8b","created_at":"2025-07-15T03:22:43.635235781Z","message":{"role":"assistant","content":"Okay"},"done":false}
{"model":"qwen3:8b","created_at":"2025-07-15T03:22:43.671292262Z","message":{"role":"assistant","content":","},"done":false}
[...]

PR #6626, opened 2025-07-15
fix: fix tool call for qwen3-ollama

leozzyzheng:
Fix tool call broken if there are assistant messages immediately after tool call.

## Description
See https://github.com/continuedev/continue/issues/6621
If there are assistant messages without tool call immediately after tool call, we need keep using the existing message to store content. Otherwise, `findCurrentToolCall` will fail to find the message with tool call.

## Checklist

## Screenshots

No visual changes.

## Tests

Add a new test to simulate case described in https://github.com/continuedev/continue/issues/6621.

### DIFF
--- a/gui/src/redux/slices/sessionSlice.test.ts
+++ b/gui/src/redux/slices/sessionSlice.test.ts
@@ -3,6 +3,7 @@ import { renderChatMessage } from "core/util/messageContent";
 import { v4 as uuidv4 } from "uuid";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { addToolCallDeltaToState } from "../../util/toolCallState";
+import { findCurrentToolCall } from "../util";
 import { ChatHistoryItemWithMessageId, sessionSlice } from "./sessionSlice";
 
 // Mock dependencies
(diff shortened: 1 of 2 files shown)


leozzyzheng:

leozzyzheng:
Have no idea why jetbrains-tests failed. I think this PR does nothing to it.

chezsmithy:
This might be totally fine, but having spent a bunch of time with using this cleanup method I'm curious if we need to make this cleanup more model and provider specific so we don't run the risk of breaking other models and providers which might have different expectations. For example, the AWS Bedrock Converse API is super picky about having perfectly matched tool and reasoning messages. This might not be the case for other providers and models. Thoughts?

leozzyzheng:
> This might be totally fine, but having spent a bunch of time with using this cleanup method I'm curious if we need to make this cleanup more model and provider specific so we don't run the risk of breaking other models and providers which might have different expectations. For example, the AWS Bedrock Converse API is super picky about having perfectly matched tool and reasoning messages. This might not be the case for other providers and models. Thoughts?

Totally agree, different model and provider needs a way to override the default method. It might need another PR to do this. :)

In this PR, the existing code is conflict with `findCurrentToolCall` which expect the message contains tool call must be the last one. So I guess this PR is safe and won't break existing good behaviours?

leozzyzheng:
duplicated by https://github.com/continuedev/continue/commit/a9e3a56a618f73cb4c21a17bb1ffe31479f9eae8

PR #6634, opened 2025-07-15
fix: Pass proxy settings from the IDE to the Continue binary (solves #5242)

exigow:
:construction:

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -331,6 +331,15 @@ export class Core {
       await this.configHandler.updateIdeSettings(msg.data);
     });
 
+    on("config/ideProxySettings", async (msg) => {
+      this.configHandler.ideProxySettings = msg.data;
+
+      // todo: just a debug. remove when PR is ready
+      console.log(
+        `ideProxySettings: ${JSON.stringify(this.configHandler.ideProxySettings)}`,
+      );
+    });
+
     on("config/refreshProfiles", async (msg) => {
       const { selectOrgId, selectProfileId } = msg.data ?? {};
       await this.configHandler.refreshAll();
@@ -418,7 +427,13 @@ export class Core {
             config,
             ide: this.ide,
             fetch: (url, init) =>
-              fetchwithRequestOptions(url, init, config.requestOptions),
+              fetchwithRequestOptions(
+                url,
+                init,
+                config.requestOptions,
+                // this.configHandler.ideProxySettings?.host,
+                "TEST HOST (loadSubmenuItems)", // todo
+              ),
           });
         return items || [];
       } catch (e) {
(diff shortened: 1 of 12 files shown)


core/index.d.ts
```diff
@@ -718,6 +718,12 @@ export interface IdeSettings {
   pauseCodebaseIndexOnStart: boolean;
 }
 
+export interface IdeProxySettings {
+  useProxy: boolean;
+  host: string;
+  port: number;
+}
+
```
exigow: I need to move this type to a module that will be recognized by `packages/fetch` and `core`. Is `config-types` OK for this?

core/core.ts
```diff
@@ -1109,7 +1130,13 @@ export class Core {
...
         reranker: config.selectedModelByRole.rerank,
         fetch: (url, init) =>
-          fetchwithRequestOptions(url, init, config.requestOptions),
+          fetchwithRequestOptions(
+            url,
+            init,
+            config.requestOptions,
+            //this.configHandler.ideProxySettings?.host,
+            "TEST HOST (getContextItems)",
+          ),
```
exigow, line 1131: Problem: This additional argument in `fetchWithRequestOptions(..., ideProxySettings)` is causing cascading changes everywhere. That’s still ~OK, but not every usage has access to introduced `ConfigHandler.ideProxySettings` field.

Maybe storing proxy in `ConfigHandler` is not a good idea? Is there a better place for this, or maybe this config should be global (like singleton)? I’m not sure how these problems are usually solved within core codebase.

(Note: The IDE proxy config is stored in an application-level service, which is, in a way, a singleton)

exigow: It just occurred to me that I could implicitly "patch" `requestOptions` in case of a IDE proxy config, but that would be a bit of a hack. :sweat_smile: (I haven’t verified that - just an idea)

core/index.d.ts
Patrick-Erichsen: `config-types` seems like the correct option 👍

core/core.ts
Patrick-Erichsen: > This additional argument in fetchWithRequestOptions(...,  ideProxySettings) is causing cascading changes everywhere

This is because the other places where we call `fetchWithRequestOptions` require a code update to handle the new parameter? Isn't the new param an optional type, so it can just be omitted?

exigow: That's correct. It can be omitted, but then the request will not be sent through the proxy. 

I added this config everywhere except `BaseLLM`, which does not use the `Core` or `ConfigHandler` class. I’m thinking right now how to pass this dependency there.

core/index.d.ts
exigow: Thanks. Fixed

exigow:
Closing this in favor of https://github.com/continuedev/continue/pull/6752

PR #6635, opened 2025-07-15
fix: open the proper config file

fbricon:
The continue view currently doesn't open the proper config files, when clicking the edit button for a Model, Rule, MCP server, prompt or Doc that is not defined in the default config file, typically anything defined under the workspace's `.continue/` directory. This PR addresses this shortcoming by:

- adding a sourceFile property to the aforementioned objects
- injecting the local file path as sourceFile value, during configuration unrolling 
- passing said sourceFile value when issuing a "config/openProfile" request 

Ideally Prompts' promptfile and Rules' ruleFile should be aligned with the other sourceFile properties but I wanted to minimize the impact area of this PR.

https://github.com/user-attachments/assets/378369a2-93c4-4b4c-b17f-0789298b1845

    

---

## Summary by cubic
Fixed the edit button so it now opens the correct config file for Models, Rules, MCP servers, Prompts, and Docs, including those defined in workspace `.continue/` directories.

- **Bug Fixes**
  - Added a `sourceFile` property to config objects and passed it through the stack to ensure the right file opens when editing non-default config entries.

### DIFF
--- a/core/commands/slash/customSlashCommand.ts
+++ b/core/commands/slash/customSlashCommand.ts
@@ -11,5 +11,6 @@ export function convertCustomCommandToSlashCommand(
     description: customCommand.description ?? "",
     prompt: customCommand.prompt,
     source: "json-custom-command",
+    promptFile: customCommand.sourceFile, // TODO refactor promptFile to align with sourceFile
   };
 }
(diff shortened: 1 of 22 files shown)


sestinj (approved):
great!

PR #6644, opened 2025-07-15
fix: :bug: Fix aborts (simple Option?)

chezsmithy:
## Description

Main:

dispatch(setInactive());
await handleToolCallExecution(dispatch, getState);

Fix:

dispatch(setInactive());
await handleToolCallExecution(dispatch, getState);

I reviewed the somewhat complex abort funciton and the current state of main, and in thinking about what the middleware solution is accomplishing I think this is the quick fix. Since dispatch(setInactive()); is synchonous, this will guarentee the current stream is wrapped up before the tool calls fire off. With the tool calls happening first, fast tool results could attempt to start streaming before the setInactive gets fired triggering an abort. The race condition that the middleware is also solving by watching for inactive to trigger before firing off the toolCallsExecution.

## Checklist

## Screenshots

## Tests

Running all existing tests to validate no changes.

### DIFF
--- a/gui/src/redux/thunks/streamNormalInput.ts
+++ b/gui/src/redux/thunks/streamNormalInput.ts
@@ -235,9 +235,7 @@ export const streamNormalInput = createAsyncThunk<
         console.error("Failed to send dev data interaction log", e);
       }
     }
-
-    await handleToolCallExecution(dispatch, getState);
-
     dispatch(setInactive());
+    await handleToolCallExecution(dispatch, getState);
   },
 );


Patrick-Erichsen:
If this is the fix that would be excellent 😅

chezsmithy:
Check me, but I think the elaborate middleware is just chaining the tool calling to the inactive event, and this does the same thing...

Patrick-Erichsen:
If it's working for you then seems good to me 🤷 we haven't cut pre-release yet so we can get this in for 1.1.64 tonight 👍

chezsmithy:
@Patrick-Erichsen it's an easy rollback if it causes any issues. As long as you aren't seeing issues running it with your setup and providers let's go.

Patrick-Erichsen:
No issues testing on my end. Thanks for the iterative work here!

PR #6647, opened 2025-07-16
add line range replacement as a fallback for search and replace

nathants:
Fixes CON-2773
    

---

## Summary by cubic
Added a fallback to the search and replace tool that uses a language model to find the best matching line range when standard matchers fail.

- **New Features**
  - If no match is found, the tool now tries a language model-based range finder to identify the correct lines for replacement.

### DIFF
--- a/gui/src/util/clientTools/searchReplaceImpl.ts
+++ b/gui/src/util/clientTools/searchReplaceImpl.ts
@@ -4,6 +4,7 @@ import { resolveRelativePathInDir } from "core/util/ideUtils";
 import posthog from "posthog-js";
 import { v4 as uuid } from "uuid";
 import { ClientToolImpl } from "./callClientTool";
+import { selectSelectedChatModel } from "../../redux/slices/configSlice";
 
 export const searchReplaceToolImpl: ClientToolImpl = async (
   args,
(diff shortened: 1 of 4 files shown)


gui/src/util/clientTools/searchReplaceImpl.ts
```diff
@@ -45,14 +214,28 @@ export const searchReplaceToolImpl: ClientToolImpl = async (
       const { searchContent, replaceContent } = block;
 
       // Find the search content in the current state of the file
-      const match = findSearchMatch(currentContent, searchContent || "");
+      let match = findSearchMatch(currentContent, searchContent || "");
```
sestinj: We can remove the need for this explicit fallback logic by adding this as another `matchingStrategies` in `findSearchMatch.ts`

sestinj (changes requested):

gui/src/util/clientTools/searchReplaceImpl.ts
```diff
@@ -4,6 +4,175 @@ import { resolveRelativePathInDir } from "core/util/ideUtils";
...
+          if (chunk.role === "assistant" && chunk.content) {
+            responseContent += chunk.content;
+          }
+        }
+      }
+
+      // Parse the AI response to get line numbers
+      const jsonMatch = responseContent.match(/\{[^}]+\}/);
+      if (!jsonMatch) {
+        throw new Error("AI did not return valid JSON");
```
sestinj: I think it would be incredibly valuable to track success rates on this, could we add some super minimal telemetry (there's an example at the bottom of this file)

sestinj (changes requested):

PR #6649, opened 2025-07-16
fix: dont stop streaming if no tool call states found

Patrick-Erichsen:
Resolves CON-2829

I was unable to reproduce the issue in debug mode where streaming doesn't resume after tool calls, but I'm reasonably confident this will fix it.

### DIFF
--- a/gui/src/redux/thunks/streamResponseAfterToolCall.ts
+++ b/gui/src/redux/thunks/streamResponseAfterToolCall.ts
@@ -2,80 +2,32 @@ import { createAsyncThunk, unwrapResult } from "@reduxjs/toolkit";
 import { ChatMessage } from "core";
 import { renderContextItems } from "core/util/messageContent";
 import {
+  ChatHistoryItemWithMessageId,
   resetNextCodeBlockToApplyIndex,
   streamUpdate,
 } from "../slices/sessionSlice";
-import { AppThunkDispatch, RootState, ThunkApiType } from "../store";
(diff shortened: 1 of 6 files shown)


gui/src/redux/thunks/streamResponseAfterToolCall.ts
```diff
@@ -2,80 +2,32 @@ import { createAsyncThunk, unwrapResult } from "@reduxjs/toolkit";
...
  * Determines if we should continue streaming based on tool call completion status.
  */
-function shouldContinueStreaming(
-  assistantMessage: ReturnType<typeof findAssistantMessageWithToolCall>,
+function areAllToolsDoneStreaming(
+  assistantMessage: ChatHistoryItemWithMessageId | undefined,
 ): boolean {
+  // This might occur because of race conditions, if so, the tools are completed
   if (!assistantMessage?.toolCallStates) {
-    return false; // No assistant message found - don't stream
```
Patrick-Erichsen, line 51: The core fix was removing this clause. My hypothesis is that we had a race condition where the `toolCallStates` was getting cleared before we reached this point, which should indicate that the tool call has completed and we should continue streaming, but we were instead returning false.

The only reason this `shouldContinueStreaming` is because with the new parallel tool calling support, we want to pause streaming until _all_ tools are complete. Previously this check didn't exist at all, we just always resumed streaming at the end of the thunk.

RomneyDa (approved):

PR #6658, opened 2025-07-16
fix: Add logging of Continue binary IO

exigow:
Minor fix/feature for developers

I used this recently in a situation where I wasn’t connecting to the binary with `USE_TCP` (it wasn’t straightforward due to the remote host)


exigow:
(rebased)

Patrick-Erichsen (approved):
Nice improvement! We should probably be using debug logs in more places

PR #6665, opened 2025-07-16
chore: remove is too big check

RomneyDa:
## Description
Removes the item is too big check, https://github.com/continuedev/continue/pull/4790, feedback has been that it's a bit noisy
Pruning message warnings will address part of the original reason for this
   
Draft, looking into updating the error message etc. instead of removing

---

## Summary by cubic
Removed the "is item too big" check from core logic and UI to reduce noise and simplify context item handling. Pruning message warnings will now handle related errors earlier.

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -17,7 +17,6 @@ import { DevDataSqliteDb } from "./data/devdataSqlite";
 import { DataLogger } from "./data/log";
 import { CodebaseIndexer } from "./indexing/CodebaseIndexer";
 import DocsService from "./indexing/docs/DocsService";
-import { countTokens } from "./llm/countTokens";
 import Ollama from "./llm/llms/Ollama";
 import { EditAggregator } from "./nextEdit/context/aggregateEdits";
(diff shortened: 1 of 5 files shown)


bdougie:
I was receiving this error without much effort, @ one markdown file. Perhaps pruning or concatenating removes the need for this message moving forward? 

I am all for unblocking this from the user experience.

RomneyDa:
@bdougie this error only happens pre-pruning, it is when a single file is too large for the entire context window. Pruning will only remove old messages. Do you remember what context length you were working with?

bdougie:
> @bdougie this error only happens pre-pruning, it is when a single file is too large for the entire context window. Pruning will only remove old messages. Do you remember what context length you were working with?

Yes, my workflow during that time was @ plan file, i.e. https://github.com/bdougie/contributor.info/blob/main/tasks/prd-contributor-insights-github-app.md and then using the Agent to plan an implementation. 

I was using Claude 4 sonnet in a Local Assistant and consistently got the file too big.

RomneyDa:
Got it. Closing in favor of https://github.com/continuedev/continue/pull/7036

PR #6702, opened 2025-07-18
feat: Support @Terminal Context Provider and RunInTerminal Button for Jetbrains Plugin

SoLoHiC:
## Description
1. Just noticed that VS Code extension has supported that @Terminal Context Provider, but Jetbrains plugin hasn't.  So I added the terminal plugin dependency and implemented the method 'getTerminalContents()' in IntelijIde.kt.
2. Since I've added that terminal plugin dependency, I digged a few on Jetbrains Docs and figured out how to implement  'paste the code to terminal', which also behaves like VS Code Extension. So, the 'runCommand()' method in IntelijIde.kt has also be implemented too now. 

## Checklist

## Screenshots
@Terminal works now, as below: 
image

the 'RunInTerminal' button works now also:
image

## Tests
don't really find much of tests under the extensions/intelij package. if you guys would like to provide more examples, i'd happy to follow the standards.

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/IntelliJIde.kt
@@ -1,4 +1,3 @@
-
 package com.github.continuedev.continueintellijextension.`continue`
 
 import com.github.continuedev.continueintellijextension.*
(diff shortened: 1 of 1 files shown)


sestinj:
@exigow just want to cc you here for awareness. Might be a nice thing to test out during your daily use. Though I've already reviewed this pretty closely and it seems like it's done perfectly

PR #6703, opened 2025-07-18
fix: remove reject edit diff functionality on cmd backspace

uinstinct:
## Description

When cmd+backspace was pressed to delete entire words in the chat area on edit mode, the focus would transfer to the editor which in turn would delete words from the editor. This PR fixes this behavior by removing the "reject diff" shortcut functionality from edit mode.

- remove "rejectDiff" on cmd backspace event listener

resolves CON-2607

## Checklist

## Screenshots

https://github.com/user-attachments/assets/8862dff4-82eb-4bed-9f74-d08a30ed9365

https://github.com/user-attachments/assets/1a8c15d3-5ed5-4e33-b4e2-33e262d07c1e

## Tests

---

## Summary by cubic
Removed the "reject diff" shortcut from edit mode so pressing cmd+backspace in the chat area no longer deletes words in the editor.

### DIFF
--- a/gui/src/pages/gui/Chat.tsx
+++ b/gui/src/pages/gui/Chat.tsx
@@ -3,7 +3,7 @@ import {
   ChatBubbleOvalLeftIcon,
 } from "@heroicons/react/24/outline";
 import { Editor, JSONContent } from "@tiptap/react";
-import { InputModifiers, ChatHistoryItem } from "core";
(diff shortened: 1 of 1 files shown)


RomneyDa (commented):
Is there another shortcut that can be used to reject? If not let's try fixing it instead of removing it.

Patrick-Erichsen (approved):
Nice catch!

PR #6711, opened 2025-07-18
feat: allow Cmd+L / Cmd+I on empty selection to select the entire line

fbricon:
This PR makes it simpler to add code to the Chat context: if no code is selected, the entire line is added to the context by Cmd+I or Cmd+L. This behaviour is similar to what Github Copilot offers.

https://github.com/user-attachments/assets/d89add3b-5515-4b34-a2d1-4ab9327a6397

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -505,13 +505,12 @@
       "continue.continueSubMenu": [
         {
           "command": "continue.focusContinueInputWithoutClear",
-          "group": "Continue",
-          "when": "editorHasSelection"
+          "group": "Continue"
         },
         {
           "command": "continue.focusEdit",
           "group": "Continue",
-          "when": "editorHasSelection && !editorReadonly"
+          "when": "!editorReadonly"
         },
         {
           "command": "continue.writeCommentsForCode",
--- a/extensions/vscode/src/util/addCode.ts
+++ b/extensions/vscode/src/util/addCode.ts
@@ -40,18 +40,34 @@ export function getRangeInFileWithContents(
       return null;
     }
 
-    let selectionRange = new vscode.Range(selection.start, selection.end);
-    const document = editor.document;
-    // Select the context from the beginning of the selection start line to the selection start position
-    const beginningOfSelectionStartLine = selection.start.with(undefined, 0);
-    const textBeforeSelectionStart = document.getText(
-      new vscode.Range(beginningOfSelectionStartLine, selection.start),
-    );
-    // If there are only whitespace before the start of the selection, include the indentation
-    if (textBeforeSelectionStart.trim().length === 0) {
-      selectionRange = selectionRange.with({
-        start: beginningOfSelectionStartLine,
-      });
+    let selectionRange: vscode.Range | undefined;
+    // if the selection is empty and the line is not, then use the entire line
+    if (selection.isEmpty) {
+      const startLine = editor.document.lineAt(selection.start.line);
+      if (startLine.isEmptyOrWhitespace) {
+        // Empty selection on an empty line, nothing else to do here
+        return null;
+      }
+      // Select the whole line
+      selectionRange = new vscode.Range(
+        selection.start.with(undefined, 0),
+        selection.end.with(undefined, startLine.range.end.character),
+      );
+    }
+    if (!selectionRange) {
+      selectionRange = new vscode.Range(selection.start, selection.end);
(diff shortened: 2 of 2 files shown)


Patrick-Erichsen:
I think some of the e2e tests have a valid failure here.

But, I believe at least for `cmd + i` the current behavior is to insert at the cursor location rather than add the line as context. I think this would also break the ability to toggle open/close the sidebar using `cmd + L`.

Is your main thought here that Copilot behaves in the way you showed in that video? 

cc @sestinj for a second opinion on behavior.

Patrick-Erichsen:
Chatted this over with @sestinj 
- We'd like to preserve the `cmd + L` sidebar toggling behavior
- We'd like to preserve `cmd + I` insert behavior, but only on empty lines.

With that in mind, could we scope the update to basically be "if the user pressed `cmd + I` on a line with any characters, the entire line is added to context` ?

fbricon:
So from what I can see, the existing Cmd+L behavior is to show the webview if hidden, and focus the cursor there. Not sure it's accurate to call it toggle, since another Cmd+L doesn't hide it, but the webview is unfocused.

I can look into reducing the scope of Cmd+I, I'll see if it works

Patrick-Erichsen:
> Not sure it's accurate to call it toggle, since another Cmd+L doesn't hide it, but the webview is unfocused.

Hm, this isn't working for you? For me if I repeatedly hit `Cmd + L` it will open/close the sidebar.

Patrick-Erichsen:
I merged main but I believe the remaining failures are legitimate.

fbricon:
So actually the bulk of these changes concerns selecting the whole line on Cmd+L.

For Cmd+I, the only thing that changed was the command is now visible when right-clicking on the editor (removing the editorHasSelection clause for continue.focusEdit in package.json). The behavior of Cmd+I itself has not changed.

If the new Cmd+L behavior is not something the Continue team is willing to get in, this is fine, we can just reject this PR. But if you find some value in this approach, then I'll look more closely into the test failures, as expectations have changed.

@Patrick-Erichsen Let me know what you decide.

fbricon:
> > Not sure it's accurate to call it toggle, since another Cmd+L doesn't hide it, but the webview is unfocused.
> 
> Hm, this isn't working for you? For me if I repeatedly hit `Cmd + L` it will open/close the sidebar.

No it doesn't work for me that way. 2nd Cmd+L doesn't close the sidebar, just unfocusses it. I tried woth both release and prerelease versions of continue, on 2 macbooks.

Patrick-Erichsen:
Hm, well the `cmd + L` toggle behavior not working is a bug, that is the intended behavior.

But if the `cmd + I` behavior remains the same and the bulk of the changes are around the `cmd + L` line highlight then I think we can close this out, unfortunately. I don't think we have metrics on how many people actually use the toggle behavior so it's more of a "gut feel" but at least on the team most of use the toggle behavior quite frequently.

RomneyDa:
@TyDunn thoughts on this?

Patrick-Erichsen:
@fbricon circling back here, chatted things over more with @TyDunn  and I think we'd be opening to aligning with Copilot here, eg having `cmd + L` add the current line. 

I think it's likely that the only folks using the toggle behavior is the internal Continue team.

fbricon:
ok, I'll take a look at the test failures when I'm back from PTO on Monday

Patrick-Erichsen:
Think a few are still failing unfortunately

fbricon:
Yes, it was a simple rebase. I need to figure out how to run those tests locally

RomneyDa:
@fbricon after sitting on this for a while we're thinking it should add the whole file, since in most cases during usage the user will want to just add the file to context, and one line is rarely useful/probably usually confusing to the model. Thoughts on tweaking to add full file instead of that line?

Agree that removing the sidebar toggle functionality is fine

Patrick-Erichsen:
@fbricon bump on this one!

fbricon:
I'll get back to it on Monday

fbricon:
The shortcut now selects the entire file if no text is highlighted. e2e tests fixed (that took me a while).

Looks like some other unit tests fail but they seem irrelevant

fbricon:
> @fbricon it looks like the description is still showing the line number of the cursor as the range, although the functionality of including the entire file is working
> 

Fixed

RomneyDa:
@fbricon thank you!

PR #6715, opened 2025-07-19
feat: init tool policy in tools definitions

xiaohuanxiong3:
## Description

Initialize the toolSetting of the tool in tool definitions

## Checklist

## Screenshots

## Tests

---

## Summary by cubic
Added a toolPolicy field to all tool definitions to set default permission levels for each tool.

- **Refactors**
  - Updated tool definitions to include toolPolicy.
  - Changed UI logic to use toolPolicy from tool definitions instead of hardcoded defaults.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -988,6 +988,14 @@
     {
       "source": "/agent/context-selection",
       "destination": "/features/agent/context-selection"
+    },
+    {
+      "source": "/customization/overview#codebase-context",
+      "destination": "/customize/context/codebase"
+    },
+    {
+      "source": "/customization/overview#documentation-context",
+      "destination": "/customize/context/documentation"
     }
   ]
 }
--- a/docs/features/chat/how-it-works.mdx
+++ b/docs/features/chat/how-it-works.mdx
@@ -51,8 +51,8 @@ View the exact prompt sent to the AI model in the [prompt logs](/troubleshooting
 
 Learn more about how context providers work:
 
-- [Codebase Context](/customization/overview#codebase-context)
-- [Documentation Context](/customization/overview#documentation-context)
+- [Codebase Context](/customize/context/codebase)
+- [Documentation Context](/customize/context/documentation)
 
 ---
 


RomneyDa:
@xiaohuanxiong3 this will be a nice change. It looks like tests are still failing, I'll convert into a draft for now.

xiaohuanxiong3:
@RomneyDa  Hi, can you help me see why /e2e/_output/tests/GUI.test.js fails.

    I checked the execution results and found that both the [should call tool after approval] and [should cancel tool] tests were successful under the Agent with tools test, but only the [should render tool call] test failed.

    Where [should call tool after approval] and [should cancel tool] successfully indicate that the initialization policy of `view_diff` tool is `allowedWithPermission` during testing.

    call `await GUIActions.toggleToolPolicy(view, "view_diff", 1)` in the [should render tool call] test means execute the button to switch tool policy once, and the policy of view_diff tool should become allowedWithoutPermission, but the test results show that the policy of view_diff tool is still allowedWithPermission, I don't understand why

```typescript
    }).timeout(DEFAULT_TIMEOUT.MD);

    it("should render tool call", async () => {
      // all tool's toolSettings here defaults to ?
      await GUIActions.toggleToolPolicy(view, "view_diff", 1);

      const [messageInput] = await GUISelectors.getMessageInputFields(view);
      await messageInput.sendKeys("Hello");
      await messageInput.sendKeys(Key.ENTER);
@@ -304,7 +307,8 @@ describe("GUI Test", () => {
    }).timeout(DEFAULT_TIMEOUT.MD * 100);

    it("should call tool after approval", async () => {
      // await GUIActions.toggleToolPolicy(view, "view_diff", 2);
      // all tool's toolSettings here defaults to ?
      // await GUIActions.toggleToolPolicy(view, "view_diff", 1);

      const [messageInput] = await GUISelectors.getMessageInputFields(view);
      await messageInput.sendKeys("Hello");
@@ -325,7 +329,8 @@ describe("GUI Test", () => {
    }).timeout(DEFAULT_TIMEOUT.XL);

    it("should cancel tool", async () => {
      // await GUIActions.toggleToolPolicy(view, "view_diff", 2);
      // all tool's toolSettings here defaults to ?
      // await GUIActions.toggleToolPolicy(view, "view_diff", 1);

      const [messageInput] = await GUISelectors.getMessageInputFields(view);
      await messageInput.sendKeys("Hello");
```

RomneyDa (changes requested):
@xiaohuanxiong3 

Looks like you got the tests passing, nice! I'm thinking we should make the following changes
- change `Tool.toolPolicy` to `Tool.defaultToolPolicy` to be more clear that it's the default/fallback value
- remove the`useEffect` that sets these in local storage and just fall back to the default if it's not set

gui/src/redux/slices/uiSlice.ts
```diff
@@ -51,19 +46,19 @@ export const uiSlice = createSlice({
...
-      [BuiltInToolNames.GrepSearch]: "allowedWithoutPermission",
-      [BuiltInToolNames.FileGlobSearch]: "allowedWithoutPermission",
-      [BuiltInToolNames.SearchWeb]: "allowedWithoutPermission",
-      [BuiltInToolNames.FetchUrlContent]: "allowedWithPermission",
-      [BuiltInToolNames.ViewDiff]: "allowedWithoutPermission",
-      [BuiltInToolNames.LSTool]: "allowedWithoutPermission",
-      [BuiltInToolNames.CreateRuleBlock]: "allowedWithPermission",
-      [BuiltInToolNames.RequestRule]: "disabled",
-      [BuiltInToolNames.SearchAndReplaceInFile]: "allowedWithPermission",
+      // [BuiltInToolNames.ReadFile]: "allowedWithoutPermission",
```
RomneyDa: This can be removed

xiaohuanxiong3:
@RomneyDa 
> - remove the`useEffect` that sets these in local storage and just fall back to the default if it's not set

Could you elaborate on the part quoted?

I think what you mean is the following code snippet.

```typescript
  const availableTools = useAppSelector((state) => state.config.config.tools);
  // Mark if initialized
  const [toolSettingsInitialized, setToolSettingsInitialized] = useState(false);

  useEffect(() => {
    if (
      !toolSettingsInitialized &&
      Array.isArray(availableTools) &&
      availableTools.length > 0
    ) {
      for (const tool of availableTools) {
        dispatch(initTool(tool));
      }
      setToolSettingsInitialized(true);
    }
  }, [availableTools, toolSettingsInitialized, dispatch]);
```

As I understand it, the code above sets the toolPolicy to its default value when the user first uses the plugin or manually clears the cache, resulting in an empty local storage, which is equivalent to your requirement of ‘fall back to the default if it's not set.’

RomneyDa:
Yes that is the snippet. We don't actually need to have any values set in the cache unless the user _changes_ them. So you can remove the use effect entirely and just fallback to the default when selecting active tools.

xiaohuanxiong3:
@RomneyDa 
When you say selecting active tools, do you mean the following functions?

```typescript
export const selectActiveTools = createSelector(
  [
    (store: RootState) => store.session.mode,
    (store: RootState) => store.config.config.tools,
    (store: RootState) => store.ui.toolSettings,
    (store: RootState) => store.ui.toolGroupSettings,
  ],
  (mode, tools, policies, groupPolicies): Tool[] => {
    if (mode === "chat") {
      return [];
    } else {
      const enabledTools = tools.filter(
        (tool) =>
          policies[tool.function.name] !== "disabled" &&
          groupPolicies[tool.group] !== "exclude",
      );
      if (mode === "plan") {
        return enabledTools.filter(
          (t) => t.group !== BUILT_IN_GROUP_NAME || t.readonly,
        );
      }
      return enabledTools;
    }
  },
);
```

I looked globally for the usage of UIState.toolSettings and found that the handleToolCallExecution function of streamNormalInput.ts is also using UIState.toolSettings

```typescript
async function handleToolCallExecution(
  dispatch: AppThunkDispatch,
  getState: () => RootState,
): Promise<void> {
  const newState = getState();
  const toolSettings = newState.ui.toolSettings;
  const allToolCallStates = selectCurrentToolCalls(newState);

  // Only process tool calls that are in "generating" status (newly created during this streaming session)
  const toolCallStates = allToolCallStates.filter(
    (toolCallState) => toolCallState.status === "generating",
  );

  // If no generating tool calls, nothing to process
  if (toolCallStates.length === 0) {
    return;
  }

  // Check if ALL tool calls are auto-approved - if not, wait for user approval
  const allAutoApproved = toolCallStates.every(
    (toolCallState) =>
      toolSettings[toolCallState.toolCall.function.name] ===
      "allowedWithoutPermission",
  ); 
  ......
```

I wonder if this means that setting a default value for toolPolicy is still necessary, or how to modify this method to use Tool.defaultToolPolicy

PR #6724, opened 2025-07-20
fix: Handle invalid tool call args

RomneyDa:
## Description
Throws errors when args are missing for a tool call
Handles invalid args in state for tool call div


Patrick-Erichsen (approved):
Maybe this is overkill but I think we could simplify to something like this:

```ts
// core/tools/implementations/requestRule.ts
const name = getToolArg(args, "name");
```

We can then get the `type` for a given `parameter`, e.g. "name", and whether or not it's required, and then have a switch statement to handle parsing for each type.

Main reason I feel this might be worthwhile is so that if a given arg goes from optional to required, or vice versa, there is no code updates needed outside the tool def.

PR #6728, opened 2025-07-21
build: Add *-eap postfix to JetBrains Early Access releases

exigow:
Motivation: https://github.com/continuedev/continue/pull/6590#discussion_r2211736481

Extra cleanup:
* Remove unused log4j dependency
  We don't have any dependency that uses it, so this dependency was unnecessarily added to the release.
* Remove explicit version in patchPluginXml
  The version is already retrieved from the project-level version / it was redundant.
* Remove commented-out publishPlugin.hidden + some noisy/redundant comments
  I removed the commented-out code which doesn't work anyway (the hidden field is not available in this version of the JetBrains Gradle plugin).

### Test

> ./gradlew buildPlugin

Outputs distributions/continue-intellij-extension-1.0.28.zip

> export RELEASE_CHANNEL=eap

> ./gradlew buildPlugin

Outputs distributions/continue-intellij-extension-1.0.28-eap.zip
    

---

## Summary by cubic
Adds an "-eap" postfix to JetBrains plugin builds when the Early Access release channel is used, and cleans up unused dependencies and redundant code.

- **Cleanup**
  - Removed unused log4j dependencies.
  - Simplified version handling in patchPluginXml.
  - Deleted commented-out and redundant code.

### DIFF
--- a/extensions/intellij/build.gradle.kts
+++ b/extensions/intellij/build.gradle.kts
@@ -9,6 +9,8 @@ fun Sync.prepareSandbox() {
 }
 
 val remoteRobotVersion = "0.11.23"
+val isEap get() = environment("RELEASE_CHANNEL").orNull == "eap"
+val pluginVersion = properties("pluginVersion").get()
 
 plugins {
     id("java")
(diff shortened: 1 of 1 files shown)


exigow:
Note: The update was limited to `build.gradle.kts` because the channel is set externally via env vars (buildPlugin is executed two times: for stable and for eap separately). There was no need to change anything in `jetbrains-release.yml`.

Patrick-Erichsen (approved):

PR #6729, opened 2025-07-21
Add debug configuration for core binary

aaudin90:
Adds a new "Core Binary (Debug)" launch configuration to VS Code. This allows for direct debugging of the Node.js binary.

- Creates a `binary:rebuild` task to build the binary before launching.
- Sets development-specific environment variables for the debug session.
- Ignores the `.cursor` directory.

## Description  
This change introduces a dedicated “Core Binary (Debug)” launch configuration for VS Code, allowing developers to debug the Node.js core binary directly inside the IDE with full source-map support.  
Key points:  
• Automatically rebuilds the binary before each session via the new `binary:rebuild` task, ensuring the debugger always runs the latest code.  
• Sets development-specific environment variables so the binary behaves the same way it does when launched by the Extension.  
• Updates documentation to reflect the new workflow and clarifies how to use it from both IntelliJ and VS Code.  
• Adds `.cursor/` to `.gitignore` to keep editor metadata out of version control.  

Together, these updates streamline the developer workflow, enable breakpoint debugging without manual setup, and keep the repository clean of transient files.

## Checklist

- [v] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [v] The relevant docs, if any, have been updated or created
- [v] The relevant tests, if any, have been updated or created

## Screenshots

---

## Summary by cubic
Added a "Core Binary (Debug)" launch configuration to VS Code for direct debugging of the Node.js binary, with automatic rebuilds and development environment variables.

- **New Features**
  - New debug config runs the latest binary build and sets up the correct environment.
  - Added a `binary:rebuild` task to automate building before debugging.
  - Updated `.gitignore` to exclude `.cursor` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ __pycache__/
 # C extensions
 *.so
 
+.cursor/
+
 # Distribution / packaging
 .Python
 build/
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -44,6 +44,23 @@
       }
     },
     {
+      "type": "node",
+      "request": "launch",
+      "name": "Core Binary (Debug)",
+      "skipFiles": ["<node_internals>/**"],
+      "program": "${workspaceFolder}/binary/out/index.js",
+      "outFiles": ["${workspaceFolder}/binary/out/**/*.js"],
+      "sourceMaps": true,
+      "smartStep": true,
+      "internalConsoleOptions": "openOnSessionStart",
+      "cwd": "${workspaceFolder}/binary",
+      "preLaunchTask": "binary:rebuild",
+      "env": {
+        "CONTINUE_DEVELOPMENT": "true",
+        "CONTINUE_GLOBAL_DIR": "${workspaceFolder}/extensions/.continue-debug"
+      }
+    },
+    {
       "name": "Debug Jest Tests",
       "type": "node",
       "request": "launch",
@@ -55,7 +72,6 @@
         "--config",
         "${workspaceRoot}/core/jest.config.js"
       ],
-
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen"
     },
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -159,6 +159,16 @@
       }
     },
     {
+      "label": "binary:rebuild",
+      "type": "shell",
+      "command": "npm",
+      "args": ["run", "rebuild"],
+      "problemMatcher": [],
+      "options": {
+        "cwd": "binary"
+      }
+    },
+    {
       "label": "docs:start",
       "type": "shell",
       "command": "npm",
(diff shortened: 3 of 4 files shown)


aaudin90:

exigow:
Hey @aaudin90 :wave:, thanks for this PR! I'll look on your config soon

aaudin90:
> Hey @aaudin90 👋, thanks for this PR! I'll look on your config soon

Hi, I'm waiting :)

aaudin90:
> Hey @aaudin90 👋, thanks for this PR! I'll look on your config soon

Hello, will you look at the changes?

Patrick-Erichsen:
@exigow any feedback here?

exigow:
Hey :wave: @aaudin90, apologies for not merging right away, but I was tied up with other priorities. I’ve reviewed your changes, but I’m at bit resistant to merge them because I don’t quite understand the value they bring.

Let me first briefly describe what’s possible right now (on `main` in the context of debugging):

* If you're using Ultimate, you don't need anything else. You don’t need VS Code. Just run the compound run configuration called `Run Continue` in debug mode, and you can set breakpoints both in the Kotlin extension and in the TypeScript server.
* If you're using Community Edition, you can’t debug TypeScript and need to use another IDE like VS Code. For Kotlin, you run `Run Extension` in debug mode in IntelliJ, and for TypeScript, you launch Core + GUI in debug mode in VS Code.

Your changes don’t seem to improve either of these scenarios.

Also, I have a follow-up questions:

> Automatically rebuilds the binary before each session via the new binary:rebuild task, ensuring the debugger always runs the latest code.

This shouldn’t be necessary. Launching `Core Binary` task in VS Code by default compiles the code and everything starts without a "binary". A server is started on port `3000` and communication happens over TCP. That’s why the run configuration in IntelliJ has an extra `USE_TCP=true` argument.

> Sets development-specific environment variables so the binary behaves the same way it does when launched by the Extension.

Can you please point out which environment variables you’re referring to? The new `Core Binary (Debug)` config seems identical to `Core Binary` + extra `binary:rebuild`.

To sum up, could you please elaborate more on what exactly your changes do and what problem they solve? (Note: I’d really like to understand your perspective, and it's possible that, since I'm mainly working with IntelliJ, I'm simply not aware of certain pain points :wink:)

aaudin90:
hi, thanks for taking the time. I don't have an ultimate, so I was debugging a part of TypeScript using Cursor. I was able to debug without any changes, and I could set breakpoints, but I couldn't access the variable values.

I am an Android developer, I know Java/Kotlin well. but I'm completely new to TS. maybe there is another way, but these changes helped me)

Try to see the variables at the moment of stopping inside VSCode, does it work for you?

RomneyDa (changes requested):
@aaudin90 this is a valid update for non-Ultimate users but let's just modify the existing core binary debug to not minify the esbuild

aaudin90:
> @aaudin90 this is a valid update for non-Ultimate users but I think we can just modify the existing core binary debug to not minify the esbuild?

Hi, I think so, as far as I understand, this is the problem.

aaudin90:
> @aaudin90 this is a valid update for non-Ultimate users but I think we can just modify the existing core binary debug to not minify the esbuild?

Do I need to do something?

RomneyDa:
@aaudin90 I just mean let's update the existing debug configuration to not minify in local debug mode rather than adding a new one!

aaudin90:
> @aaudin90 I just mean let's update the existing debug configuration to not minify in local debug mode rather than adding a new one!

Got it, I'll do it.

RomneyDa:
@aaudin90 closing since stale but feel free to open a PR that tweaks the current debug config to not minify esbuild in development environment!

PR #6734, opened 2025-07-21
chore: remove postgres context provider

uinstinct:
## Description

Remove postgres context provider. Reasons include low usage of the postgres context provider as reported on telemetry data and the postgres MCP is provides greater capabilities and is more maintained than the current context provider.

- remove PostgresContextProvider.ts
- use remark markdown to display error messages in ErrorSection

resolves CON-2764

## Checklist

## Screenshots

## Tests

---

## Summary by cubic
Removed the Postgres context provider and updated config validation to show a deprecation message if "postgres" is used.

- **Migration**
  - Use the Postgres MCP server instead of the built-in Postgres provider.
  - Error messages in the UI now support markdown formatting.


uinstinct:
@RomneyDa added the reasons in the PR description!

About showing errors in the config, how about we show a deprecation message for a few releases and then finally remove it in a later? (in that case, we can show a deprecation message for few other providers like google, github issues, discord etc. which can be replaced by MCPs, starting from this PR)

RomneyDa:
What is the purpose of removing it? Can we just remove it from the docs and hub?

RomneyDa:
Can we just remove it from the docs and hub instead?

uinstinct:
Yes. What about users who are using the postgres context provider? (I think we should help them migrate)

RomneyDa:
Simply replacing its docs section with a deprecation warning that points to the MCP server seems fine. I'm not sure any code changes are needed. Maybe we could add a one-time localStorage value `hasShownPostgressContextDeprecationWarning` or that checks for postgres in config and shows an ide warning message once with a link to the postgres MCP?

uinstinct:
added a deprecation warning toast!

image

RomneyDa (changes requested):
@uinstinct this will only show when using YAML config, could we move to doLoadConfig so that it shows for both YAML and JSON?

Also could we remove from the docs? or at least add a deprecation warning

RomneyDa (approved):
Removes postgres provider and adds warning!

PR #6736, opened 2025-07-21
feat: vertex ai openai adapter

RomneyDa:
## Description
- Adds tool support for vertex AI
- Adds an openai adapter for vertex AI which features the same functionality as the current VertexAI llm class (including new tools), besides caching for anthropic

smoke tested for anthropic and gemini
    

---

## Summary by cubic
Added a VertexAI OpenAI adapter to support VertexAI models using the OpenAI-compatible API, including Gemini, Anthropic, and Mistral models.

- **New Features**
  - Added a new VertexAIApi class with streaming and non-streaming chat, completion, and embedding support.
  - Updated type definitions and factory logic to recognize the new provider.
  - Integrated Google authentication and configuration options for VertexAI.


Patrick-Erichsen (approved):
Just one minor error format I noticed from some code that was moved

PR #6751, opened 2025-07-22
test: verify chunkDocument filters out chunks exceeding maxChunkSize

tingwai:
## Description

Add tests to verify maxChunkSize is being respected
    

---

## Summary by cubic
Added tests to verify that chunkDocument filters out chunks exceeding maxChunkSize and cleans up async encoders after tests.

- **Tests**
  - Checked that chunkDocument skips content over the token limit.
  - Ensured async encoder cleanup to prevent Jest from hanging.


tingwai:
> I might be misunderstanding this, but shouldn't the chunking function chunk the document to within max chunk size? Wondering if there's a bug with the chunky algorithm, and filtering is a band-aid?

The behavior should just be removing the chunk, it does `return resolve(undefined)`
https://github.com/continuedev/continue/blob/tingwai/test-chunkDocument-maxChunkSize/core/indexing/chunk/chunk.ts#L61

core/llm/countTokens.test.ts
```diff
@@ -27,7 +28,12 @@ describe.skip("countTokens", () => {
   });
 });
 
-describe.skip("countTokensAsync", () => {
```
tingwai, line 30: I assume these tests was skipped before because there was a dangling asynchronous operations causing Jest to hang. These units tests should now finish with `cleanupAsyncEncoders`

tomasz-stefaniak:
I approved this PR but it's still awaiting response from @RomneyDa because he requested changes. @RomneyDa I think the function is named in a misleading way, my initial reaction was similar to yours. `chunkDocument` calls `chunkDocumentWithoutId` which in turn calls `codeChunker` with `maxChunkSize`.

Now that I think about it, I don't exactly understand why the filtering code is there in the first place 🤔 Maybe that's something worth looking into.

sestinj:
> I might be misunderstanding this, but shouldn't the chunking function chunk the document to within max chunk size? Wondering if there's a bug with the chunky algorithm, and filtering is a band-aid?

It is in fact a band-aid, there as a fallback in case somehow the main chunker fails to properly get below the maxChunkSize. Should this happen? Ideally, no. But if it does there's no reason not to filter to avoid worse problems.

I will say @tingwai it might be useful to also test the function that does the chunking _before_ the filtering. That way we are testing the actual chunking logic and not just the filtering (which should trivially work correctly). You might already have done this and I have my function names backward; if so, ignore.

tingwai:
> I will say @tingwai it might be useful to also test the function that does the chunking _before_ the filtering. That way we are testing the actual chunking logic and not just the filtering (which should trivially work correctly). You might already have done this and I have my function names backward; if so, ignore.

Pushed another commit adding tests for `chunkDocumentWithoutId`, this function is called inside `chunkDocument` which handles actual chunking

tomasz-stefaniak (approved):

PR #6752, opened 2025-07-22
fix: Apply proxy settings on the Continue process using environment variables (solves #5242)

exigow:
This method is based on environment variables instead of dedicated message in the protocol (see: https://github.com/continuedev/continue/pull/6634).

Extra:
* This will not work with `USE_TCP=true`, because these env vars can only be set for the process. Continue-over-TCP with proxy is not possible right now.
* The implementation of the process handler has partially changed its behavior. From now on, manual restart (including `Continue: Restart Process` action) has no artificial delay/it's instant. This was a bug IMO that is fixed now.

### Testing

* To test this in a development environment, you need to remove `USE_TCP` from the `Run Extension` run configuration. 
* We only track two proxy settings controls: whether it's enabled and the proxy address (+ port).
* Those options are avaliable in `IDE > Apperance > System Settings > HTTP Proxy`. You can change the state, then apply. The activity polls those values, and if it detects a change, the service restarts the process, which reads the new setting on start.
* There should always be only one process per IDE, regardless of how quickly the settings are changed or how often the binary is restarted manually.


exigow:
**Note about testing**: I could not fully verify whether the requests actually go through the Burp proxy. Even though I imported the proxy certificate at the system level as trusted (into the OS Trusted Store) and additionally provided the certificate via `requestOptions.caBundlePath` for my local config, I can only see part of the decrypted requests. For some reason, certain requests return an enigmatic `Invalid URL` (but without an actual URL :smile:) in the JCEF logs, and I haven't been able to determine the cause. There's no trace of this in the Continue or JCEF source code, and honestly, I’ve already spent so much time on this that I think it can be merged as-is. Since the way the proxy is applied is so trivial, I’m very confident that these changes should work because it’s transparent enough that no changes were needed in the Continue binary code (it's just env vars).

exigow:
Rebased + fixed conflict

PR #6759, opened 2025-07-23
fix: register YAML config schema by its URI in VS Code

fbricon:
The vscode extension registers the YAML config schema by its FS path, instead of its URI, making it impossible to open it when clicking on its codelens:

https://github.com/user-attachments/assets/4e214e9f-87f6-4f86-979d-5f2abcbd87cc

The fix consists in registering the schema by its URI:

https://github.com/user-attachments/assets/da8f3766-5c8f-40c1-aa62-08de36cda5e6

A proper fix would require registering the schema using vscode-yaml's [registerContributor API](https://github.com/redhat-developer/vscode-yaml/blob/4b1730412b5a6bf82f0bc23f6665db8904bf4c21/src/schema-extension-api.ts#L39-L44), that would then avoid injecting/overriding version specific paths in the user's settings.json. But that's for another day.
    

---

## Summary by cubic
Fixed YAML config schema registration in the VS Code extension to use its URI instead of the file system path, so clicking the codelens now opens the schema as expected.


RomneyDa:
@fbricon might need to pull main for this one to pass tests?
EDIT was able to do this.

Issue #5107, opened 2025-04-11
Continue can not initialize in Android Studio. JCEF not supportd  in this env.


notnotMyself:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 10 Professional
- Continue version: 1.0.8
- IDE version: Android Studio 2023.2
- Model:
- config:
  

  
  OR link to assistant in Continue hub:
```

### Description

java.lang.IllegalStateException: JCEF is not supported in this env or failed to initialize
	at com.intellij.ui.jcef.JBCefApp.getInstance(JBCefApp.java:120)
	at com.intellij.ui.jcef.JBCefBrowserBase.lambda$new$3(JBCefBrowserBase.java:157)
	at com.intellij.util.ObjectUtils.notNull(ObjectUtils.java:121)
	at com.intellij.ui.jcef.JBCefBrowserBase.<init>(JBCefBrowserBase.java:157)
	at com.intellij.ui.jcef.JBCefBrowser.<init>(JBCefBrowser.java:189)
	at com.intellij.ui.jcef.JBCefBrowser.create(JBCefBrowser.java:140)
	at com.intellij.ui.jcef.JBCefBrowserBuilder.build(JBCefBrowserBuilder.java:127)
	at com.github.continuedev.continueintellijextension.toolWindow.ContinueBrowser.<init>(ContinueBrowser.kt:36)
	at com.github.continuedev.continueintellijextension.toolWindow.ContinuePluginToolWindowFactory$ContinuePluginWindow$browser$2.invoke(ContinuePluginToolWindowFactory.kt:49)
	at com.github.continuedev.continueintellijextension.toolWindow.ContinuePluginToolWindowFactory$ContinuePluginWindow$browser$2.invoke(ContinuePluginToolWindowFactory.kt:46)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
[...]

Issue #4598, opened 2025-03-11
The Jetbrains Plugin is not working with the new 1.0.2 version


Vildnex:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:Vanilla OS
- Continue version: 1.0.2
- IDE version: Android Studio 2024.3.1
- Model:
- config:
  
{
  "models": [
    {
      "title": "O3 Mini High",
      "provider": "openrouter",
      "model": "openai/o3-mini-high",
      "apiBase": "https://openrouter.ai/api/v1",
      "apiKey": "<OPENROUTER_KEY>",
      "requestOptions": {
        "extraBodyProperties": {
          "transforms": []
        }
      }
    },
    {
      "title": "Claude 3.7 Sonnet",
      "provider": "openrouter",
      "model": "anthropic/claude-3.7-sonnet",
      "apiBase": "https://openrouter.ai/api/v1",
      "apiKey": "<OPENROUTER_KEY>",
      "requestOptions": {
        "extraBodyProperties": {
          "transforms": []
        }
      }
    },
    {
      "title": "Claude 3.7 Sonnet (think)",
      "provider": "openrouter",
      "model": "anthropic/claude-3.7-sonnet:thinking",
      "apiBase": "https://openrouter.ai/api/v1",
      "apiKey": "<OPENROUTER_KEY>",
      "requestOptions": {
        "extraBodyProperties": {
          "transforms": []
        }
      }
    },
    {
      "title": "Qwen: Qwen-Max ",
      "provider": "openrouter",
      "model": "qwen/qwen-max",
      "apiBase": "https://openrouter.ai/api/v1",
      "apiKey": "<OPENROUTER_KEY>",
      "requestOptions": {
        "extraBodyProperties": {
          "transforms": []
        }
      }
    },
    {
      "title": "DeepSeek Chat",
      "provider": "openrouter",
      "model": "deepseek/deepseek-chat",
      "apiBase": "https://openrouter.ai/api/v1",
      "apiKey": "<OPENROUTER_KEY>",
      "requestOptions": {
[...]

Issue #649, opened 2023-12-03
Continue tab is blank (IDEA 2023.2 Win 10)


schiebelp:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Win 10 
- Continue: 0.0.20
- Python: 3.11.6
- IDE: IntelliJ IDEA 2023.2 (Community Edition)
```

### Description

I was unable to use it. The Plugin tab is visible but nothing in it. 

I noticed in idea.log an exception:
INFO - STDOUT - Starting Continue server binary
INFO - STDERR - java.nio.file.AccessDeniedException: startProcess.vbs
...
...runBinary(ContinuePluginStartupActivity.kt:244)

Found the code 244 is this:
233:            val scriptPath = Paths.get("startProcess.vbs")
244:            Files.writeString(scriptPath, script)

Which is - I believe - is all about running the server_continue.exe using that vbs code.

So I went for that server and ran it myself. That unfortunatelly did not help as tab was still empty (after IDE restart).

The server communicates though, when IDEA is opened:
INFO:     connection open
...appears. 

So there seems to be another issue. Which is frustrating, because this would be killer app. I would love to use it with local llm.

### To reproduce

1. Install plugin
2. Open IDEA IDE
3. Open tab Continue

### Log output

```Shell
continue.log

[2023-12-03 00:52:54] [WARNING] Failed to load config from C:\Users\Petr\.continue\config.py: module 'config' has no attribute 'config'
[2023-12-03 00:52:54] [DEBUG] ------ Begin Logs ------
[2023-12-03 00:52:54] [DEBUG] Downloading MeiliSearch for Windows...
[2023-12-03 00:53:05] [DEBUG] Starting MeiliSearch...
[2023-12-03 00:53:05] [INFO] Meilisearch started at http://localhost:7700
```

sestinj:
@schiebelp sorry to see this, could you share what the tab looked like when empty?

Also one thing to be sure of, did you check the box in settings for "Manually Running Server" when you were running it yourself?

sestinj:
[...]

Issue #596, opened 2023-11-01
IntelliJ Plugin not working in Android Studio


xrigau:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://continue.dev/docs/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 14.1
- Continue: 0.0.14
- IDE: Android Studio Iguana | 2023.2.1 Canary 4. Build #AI-232.9921.47.2321.10809744
- IDE Runtime version: 17.0.8+0-17.0.8b1000.22-10799086 aarch64
- IDE VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
```

### Description

Upon opening the Continue panel in the IDE, the panel is empty and the IDE throws the error attached in the "Log output" section.

### To reproduce

1. Install the Continue plugin in the IDE
2. Restart the IDE
3. Open the Continue panel

### Log output

```Shell
java.lang.IllegalStateException: JCEF is not supported in this env or failed to initialize
	at com.intellij.ui.jcef.JBCefApp.getInstance(JBCefApp.java:120)
	at com.intellij.ui.jcef.JBCefBrowserBase.lambda$new$3(JBCefBrowserBase.java:157)
	at com.intellij.util.ObjectUtils.notNull(ObjectUtils.java:121)
	at com.intellij.ui.jcef.JBCefBrowserBase.<init>(JBCefBrowserBase.java:157)
	at com.intellij.ui.jcef.JBCefBrowser.<init>(JBCefBrowser.java:189)
	at com.intellij.ui.jcef.JBCefBrowser.<init>(JBCefBrowser.java:149)
	at com.github.continuedev.continueintellijextension.toolWindow.ContinuePluginToolWindowFactory$ContinuePluginWindow$webView$2.invoke(ContinuePluginToolWindowFactory.kt:60)
	at com.github.continuedev.continueintellijextension.toolWindow.ContinuePluginToolWindowFactory$ContinuePluginWindow$webView$2.invoke(ContinuePluginToolWindowFactory.kt:59)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
[...]

PR #6762, opened 2025-07-23
fix: Add JCEF not supported warning

exigow:
Related to: #5107, #4598, #649, #596

### Screenshot

Selection_201


exigow:
> This will be great to have! A few thoughts (...)

@Patrick-Erichsen, thanks for the valuable feedback. You're right, users won't understand what it's about. :sweat_smile: 

I simplified the language like you suggested + added a tutorial + link to docs.

BTW I've just discovered that the new JetBrains UI DSL (these builders like `panel {}`, `row {}`, etc) has a nice feature that converts HTML tags on the fly into native Swing components, and it's not JCEF-based (!). Very useful.

exigow:
(screenshot updated)

Patrick-Erichsen (approved):
This looks great! Thanks for the detailed instructions.

Also that's super useful to know that we can build Swing UI just using HTML, this would make our inline edit panel 10x easier to work with haha

PR #6763, opened 2025-07-23
fix: use rules when `alwaysApply` is absent

uinstinct:
## Description

`alwaysApply: undefined` was not working because the rule file path is always absolute

- use `.includes` instead of `.startsWith` to search for the `.continue` folder
- update tests

resolves CON-2955

## Checklist

https://github.com/user-attachments/assets/c7790569-6796-422d-8dd1-90e294460273

https://github.com/user-attachments/assets/662cfe2c-3561-47be-9f55-195707b85c69

## Tests

### DIFF
--- a/core/llm/rules/getSystemMessageWithRules.vitest.ts
+++ b/core/llm/rules/getSystemMessageWithRules.vitest.ts
@@ -69,7 +69,7 @@ describe("Rule colocation glob matching", () => {
       name: "Root Rule",
       rule: "Follow project standards",
       source: "rules-block",
-      ruleFile: ".continue/rules.md",
+      ruleFile: "/path/to/repo/.continue/rules.md",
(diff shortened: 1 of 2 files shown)


core/llm/rules/getSystemMessageWithRules.vitest.ts
```diff
@@ -69,7 +69,7 @@ describe("Rule colocation glob matching", () => {
       name: "Root Rule",
       rule: "Follow project standards",
       source: "rules-block",
-      ruleFile: ".continue/rules.md",
+      ruleFile: "/path/to/repo/.continue/rules.md",
```
RomneyDa, line 72: I realize these are tests but I believe the rules file is a URI, not a path. Just note, fine to merge

RomneyDa (approved):

PR #6768, opened 2025-07-23
feat: add env field to config yaml

RomneyDa:
## Description
Adds an env top level object field to config yaml, which can be used by `cn`
    

---

## Summary by cubic
Added an optional env field to the top level of config.yaml to support environment-specific values.

- **New Features**
  - env can now be defined in config.yaml as a record of string, number, or boolean values.


RomneyDa:
Closing since failing and will remake

PR #6455, opened 2025-07-04
feat: mermaid diagrams

uinstinct:
## Description

Render mermaid diagrams into chat area.

- added library https://github.com/mermaid-js/mermaid (for mermaid diagram rendering)
- added library https://github.com/timmywil/panzoom (for panzooming)

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [x] The relevant docs, if any, have been updated or created
- [x] The relevant tests, if any, have been updated or created

## Screenshots

https://github.com/user-attachments/assets/51f6ca35-1c86-48f4-ac5c-30f62b693c1f

## Tests

[ What tests were added or updated to ensure the changes work as expected? ]

    

---

## Summary by cubic
Added support for rendering Mermaid diagrams in the chat area, including zoom and pan controls.

- **Dependencies**
  - Added `mermaid` and `@panzoom/panzoom` packages.

PR #6777, opened 2025-07-24
fix: case when reranker returns invalid scores array

uinstinct:
## Description

The following error used to happen when the reranker model returned an invalid response.
`Failed to rerank retrieval results TypeError: Cannot read properties of undefined (reading 'sort')`. More about the [RCA in the ticket](https://linear.app/continue/issue/CON-2661/failed-to-rerank-retrieval-results-typeerror-cannot-read-properties-of#comment-b6cf0eb8).

- add a zod validation schema for voyage ai response
- check if `scores` array is valid in DocsContextProvider

resolves CON-2661

## Checklist

## Tests

---

## Summary by cubic
Fixed a crash when the reranker model returned an invalid scores array by adding response validation and handling invalid cases.

- **Bug Fixes**
  - Added Zod schema validation for the Voyage AI reranker response.
  - Checked for a valid scores array before sorting in DocsContextProvider.


core/llm/llms/Voyage.ts
```diff
@@ -36,6 +50,9 @@ class Voyage extends OpenAI {
     const data = (await resp.json()) as {
       data: Array<{ index: number; relevance_score: number }>;
     };
+
+    VoyageRerankSuccessResponseSchema.parse(data);
```
RomneyDa: You can use the zod output here, it will be typesafe!
I'd recommend `safeParse` so you can gracefully handle invalid values

RomneyDa (changes requested):
@uinstinct see nitpick comment. Just to double check, has this been tested with voyage reranker?

core/llm/llms/Voyage.ts
uinstinct: actually I wanted the throw the error directly if parsing failed like we do here

https://github.com/continuedev/continue/blob/b28d0f20081f784f91d8fd008adfc7dec965c44c/core/llm/llms/Voyage.ts#L30-L34

uinstinct:
> Just to double check, has this been tested with voyage reranker?

Yes had verified on Voyage reranker

core/llm/llms/Voyage.ts
RomneyDa: @uinstinct the schema.parse will throw an error if parsing fails. i.e. if the status is 200 then presumably the parse with throw an error because the response won't match the expected format, which means the `status` check will never be reached

uinstinct: yeah, got it.

RomneyDa:
@uinstinct sorry for the confusion,
I think the schema parser is a great idea, I just meant that the parsing error needs to be handled if it is parsed before the response status is checked

EDIT I was wrong about the order here, the main concern now is that the parse will throw an ugly zod error so should be gracefully handled

RomneyDa (approved):

PR #6781, opened 2025-07-24
Fix: Remove duplicate blocks and provide config warnings

Jazzcort:
## Description

This patch is to address the following cases that duplicate blocks appear in the UI and confuse users:

  1. When users refer to a local block with "- uses", if the block is already in the default directories (e.g. ~/.continue/models or \<workspace\>/.continue/models) and users don't override its name attribute, the duplicate block will appear in the UI.

  2. When users have duplicate block names among each block type in their assistant config.

  3. When the injected blocks have duplicate names with the assistant config for the corresponding block type.

The duplicate blocks in the UI is confusing, especially in the model selection.

This patch add a blockDuplicationDetector to detect duplication. When a duplication is detected, we remove the duplicate item and push a config error to the result to warn the users that having duplicate names is not the best practice to set up their config.

The detecting order: assistant config -> injected blocks

To detect duplication, all block types use "name" as the identifier except for rules and context. For rules, if it's an rule object, we use the "name" attribute. Otherwise, we use its string content as the identifier. For context, the "provider" attribute is used as the identifier.

The detection only happens under assistant level, which means you can have assistants with duplicate names.

## Checklist

### Before this patch

https://github.com/user-attachments/assets/d27117c1-b8f4-485e-b1e7-3d39a147dd19

### After this patch

https://github.com/user-attachments/assets/1cdcc4a6-e951-4ca9-935f-846efc3514f9

## Tests

Add a few duplicate blocks and see if they get removed with the corresponding config errors.

    

---

## Summary by cubic
Removed duplicate blocks from assistant configs and injected blocks, and added config warnings when duplicates are detected to prevent confusion in the UI.

- **Bug Fixes**
  - Blocks with duplicate names are now filtered out based on block type.
  - Config errors are shown to warn users about duplicates.

### DIFF
--- a/packages/config-yaml/src/load/unroll.ts
+++ b/packages/config-yaml/src/load/unroll.ts
@@ -20,6 +20,7 @@ import {
   Rule,
 } from "../schemas/index.js";
 import { ConfigResult, ConfigValidationError } from "../validation.js";
+import { BlockDuplicationDetector } from "./blockDuplicationDetector.js";
 import {
   packageIdentifierToShorthandSlug,
   useProxyForUnrenderedSecrets,
@@ -306,6 +307,13 @@ export async function unrollBlocks(
 ): Promise<ConfigResult<AssistantUnrolled>> {
   const errors: ConfigValidationError[] = [];
 
+  function injectDuplicationError(errorMsg: string) {
+    errors.push({
+      fatal: false,
+      message: errorMsg,
+    });
+  }
+
   const unrolledAssistant: AssistantUnrolled = {
     name: assistant.name,
     version: assistant.version,
@@ -484,7 +492,7 @@ export async function unrollBlocks(
         const injectedResults = await Promise.all(injectedBlockPromises);
         const injectedErrors: ConfigValidationError[] = [];
         const injectedBlocks: {
-          blockType: string;
+          blockType: BlockType;
           resolvedBlock: any;
           source?: string;
         }[] = [];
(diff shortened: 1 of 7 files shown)


Jazzcort:
@tingwai Any thoughts on this? 😄

packages/config-yaml/src/load/unroll.test.ts
RomneyDa, line 1: @Jazzcort empty test here

packages/config-yaml/src/load/nameRegistry.ts
```diff
@@ -0,0 +1,79 @@
...
+    block: any,
+    blockType: BlockType,
+    injectError?: (errorMsg: string) => void,
+  ): boolean {
+    // Not checking any null or undefined object
+    if (block === null || block === undefined) {
+      return false;
+    }
+
+    let exists = false;
```
RomneyDa: can remove let, return within switch

RomneyDa (changes requested):
@Jazzcort nice cleanup on types
- see comment about empty test
- Nitpick, I think the name `NameRegistry` could be confused with the `RegistryClient`, could rename to something like `BlockDeduplicator` or similar

packages/config-yaml/src/load/unroll.test.ts
Jazzcort: Thanks for the catch! I'll delete this test. 😄

packages/config-yaml/src/load/nameRegistry.ts
Jazzcort: Will do! 😁

Jazzcort:
@RomneyDa `BlockDuplicator` sounds like something that duplicates blocks. How about `BlockDuplicationDetector`? Does it make sense to you? 😁

Jazzcort:
@RomneyDa I went ahead and changed `nameRegistry` to `blockDuplicationDetector`. Let me know if you have better naming! 😁

packages/config-yaml/src/load/unroll.ts
```diff
@@ -539,16 +558,20 @@ export async function unrollBlocks(
...
+    const filteredBlocks = injectLocalSourceFile(
       key,
       resolvedBlock,
       source,
+    ).filter(
+      (block: any) =>
+        !detector.isDuplicated(block, blockType, injectDuplicationError),
     );
-    unrolledAssistant[key]?.push(...blocksWithSourceFiles);
+    unrolledAssistant[key]?.push(...((filteredBlocks ?? []) as any));
```
RomneyDa: Could we avoid the any type here?

RomneyDa (approved):
Adds warnings for duplicate blocks

packages/config-yaml/src/load/unroll.ts
Jazzcort: Yes! I should have cleaned up this part. Thanks for the catch! 😄

RomneyDa (approved):

Jazzcort:
@RomneyDa I just spotted a few typos and errors in the test cases at the last minute. 😅

RomneyDa (approved):

PR #6799, opened 2025-07-25
fix: fix /auth/refresh string value

exigow:
Fix regression introduced in https://github.com/continuedev/continue/pull/6531

Fortunately, this change hasn't been published yet on stable channel.

We need to hide [1.0.31-eap](https://plugins.jetbrains.com/plugin/22707-continue/versions/eap/803584) + probably the stable 1.0.31 if it's still being verified by Marketplace.

1.0.30 is OK.
    

---

## Summary by cubic
Fixed the /auth/refresh endpoint response handling to correctly read the response string, resolving a regression in token refresh logic.


exigow:
How? :thinking: 

I tested this auth/token manually while creating the original PR, but I must have accidentally replaced `readString` with `toString` at some point (I suspect that before pushing, I decided to move `readString` outside the lambda). The types matched (string->string), so I didn't notice. The error only became visible after a full re-login. Fortunately, I noticed it today by chance while working on main branch.

exigow:
@sestinj, @Patrick-Erichsen, FYI this is important PR :sweat_smile:

Patrick-Erichsen (approved):

Patrick-Erichsen:
Will hide the EAP as well 👍

PR #6801, opened 2025-07-25
feat: add gemma tool calling

buroa:
Adds gemma to tool calling support
    

---

## Summary by cubic
Added support for Gemma models in tool calling, allowing these models to use function calling features.


buroa:

PR #6802, opened 2025-07-25
feat: expose legacy /share command to VS Code UI

fbricon:
The legacy `/share` slash command is no longer available when using config.yaml.
While the feature could be replaced by a custom prompt command + a tools-supporting model,
it's fairly slow/buggy when relying on local models.

Instead, we're exposing the feature to the VS Code UI
- as a button at the bottom of the chat window
- as a button on each conversation in the history view
- as a VS Code command `Share Current Chat Session as Markdown`

https://github.com/user-attachments/assets/3f73b0df-64e6-4014-ad9f-96750b87536e

Unfortunately, because the IDE requires special handling for the destination folder selection,
the gui code needs to check if it's running on jetbrains in order to disable this new feature

I can't work on Jetbrains integration at the moment.
    

---

## Summary by cubic
Added a "Share Chat as Markdown" feature to the VS Code UI, allowing users to export chat sessions as markdown files from the chat window, history view, or via a command. This feature is disabled on JetBrains IDEs.

- **New Features**
  - Added "Share" button to the bottom of the chat window and each conversation in the history view.
  - Added VS Code command: "Share Current Chat Session as Markdown".
  - Prompts user to select a destination folder before saving the markdown file.


RomneyDa (changes requested):
@fbricon got some feedback from the team, main thoughts 
- add support for Jetbrains (Continue might be able to one-shot this) 
- history page and command palette implementation is great
- remove the button in the chat, getting a bit noisy down there, maybe we can put it back in but for now might not make the cut

Note merge conflicts

fbricon:
@RomneyDa I fixed the merge conflicts and removed the button from the chat, as requested. However I can't work on the Jetbrains support atm.

RomneyDa:
@fbricon sounds good, I'm tentative to merge for now because the preferred approach would be to move as much code to core as possible for easy support/maintenance for jetbrains, cli, and future extensions. I think it would look like adding an `openFileDialog` IDE message and moving all the logic including opening file dialog to the core `history/share` handler.

fbricon:
I can centralize the code to core and provide an IDE.openFileDialog implementation for VS Code, but wouldn't it break JetBrains then, since I can't provide its implem ?

fbricon:
And what if we want to add one more step in the future, like "Choose a file name"?

fbricon:
@RomneyDa ^^ ?

RomneyDa:
@fbricon added a good first issue for adding jetbrains support
 https://github.com/continuedev/continue/issues/7902
Appreciate the contribution!

PR #6807, opened 2025-07-26
feat: :sparkles: Improve the apply tool feedback to the LLM

chezsmithy:
## Description

Previously when the apply tool ended there was only 2 outcomes. 1) The tool call was sucessful and the LLM was made aware that all edits were applied. 2) The tool call failed and the interation stopped in an abrupt and confusing way.

The change here is to introduce tri-state tracking for the apply tool. 1) All edits applied, the tool call is sucessful, no changes. 2) Partial edits are accepted, and the tool call completes but informs the LLM that it should ask the user for next steps while also understanding there is partial acceptace. 3) All edits are rejected, and the tool call completes but informs the LLM that it should ask the user for next steps.

This improves end user satisfaction and reduces confusion that the apply step halted.

## Checklist

## Tests

None as of yet. Opening in WIP

    

---

## Summary by cubic
Improved the apply tool feedback so the LLM now receives clear information when file edits are fully accepted, partially accepted, or fully rejected, instead of just success or failure. This helps the LLM know when to stop, ask the user for next steps, or continue, reducing confusion after applying changes.

- **New Features**
  - Added tri-state tracking for apply tool outcomes: all accepted, partial acceptance, or all rejected.
  - Updated LLM feedback to include detailed user actions, prompting the LLM to ask for clarification or next steps when needed.

### DIFF
--- a/gui/src/hooks/ParallelListeners.tsx
+++ b/gui/src/hooks/ParallelListeners.tsx
@@ -233,7 +233,7 @@ function ParallelListeners() {
   });
 
   // Track processed toolCallIds to prevent duplicates
-  const processedToolCalls = useRef(new Set<string>());
+  // const processedToolCalls = useRef(new Set<string>());
 
   useWebviewListener(
     "updateApplyState",
@@ -265,9 +265,9 @@ function ParallelListeners() {
           }
           if (state.status === "closed") {
             // Check if we've already processed this toolCallId
-            if (processedToolCalls.current.has(state.toolCallId)) {
-              return;
-            }
+            // if (processedToolCalls.current.has(state.toolCallId)) {
+            //   return;
+            // }
 
             // Find the tool call to check if it was canceled
             const toolCallState = findToolCallById(
@@ -287,7 +287,7 @@ function ParallelListeners() {
               hasValidDiffData
             ) {
               // Mark this toolCallId as processed
-              processedToolCalls.current.add(state.toolCallId);
+              //processedToolCalls.current.add(state.toolCallId);
               // Use 3-state feedback based on acceptance/rejection counts
               const totalDiffs =
                 (state.acceptedDiffs || 0) + (state.rejectedDiffs || 0);
@@ -306,11 +306,6 @@ function ParallelListeners() {
                   content: feedbackMessage,
                   status: "user_rejected_all",
                 };
-                dispatch(
-                  acceptToolCall({
-                    toolCallId: state.toolCallId,
-                  }),
-                );
               } else if (state.acceptedDiffs === totalDiffs && totalDiffs > 0) {
                 // All diffs were accepted
                 feedbackMessage = "User accepted all proposed changes";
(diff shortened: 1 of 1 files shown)


chezsmithy:
@RomneyDa I'd love you to take a look at this idea. It's a throwback to our idea of having the LLM become more aware to what is happening after edits are completed. I'm opening it in DRAFT so I can get a review of the idea, and then if we should proceed, I may need some help with the IntelliJ testing (it would be my first change there).

gui/src/hooks/ParallelListeners.tsx
```diff
@@ -242,6 +243,9 @@ function ParallelListeners() {
     dispatch(updateIndexingStatus(data));
   });
 
+  // Track processed toolCallIds to prevent duplicates
+  const processedToolCalls = useRef(new Set<string>());
```
RomneyDa: Rather than tracking processedToolCalls here we can check the status of the associated toolCallState

gui/src/hooks/ParallelListeners.tsx
```diff
@@ -271,29 +275,112 @@ function ParallelListeners() {
...
+                  "The search and replace tool executed successfully, but the user rejected ALL proposed changes. This means:\n\n1. DO NOT attempt any more file modifications\n2. DO NOT try different changes\n3. STOP and ask the user for clarification\n\nYou must now ask the user: 'I see you rejected all my proposed changes. Could you help me understand what you're looking for instead? What specific changes would you like me to make to the file?'";
+                finalOutput = {
+                  name: "Search and Replace Tool - User Rejected All Changes",
+                  description:
+                    "Tool succeeded but user rejected all changes - LLM must stop and consult user",
+                  content: feedbackMessage,
+                  status: "user_rejected_all",
+                };
+                dispatch(
+                  acceptToolCall({
```
RomneyDa: One change I might make is to make the case of reject all mark the tool call as error, rather than accepted. Everything else would work the same, just shows red X in the UI which I think might reflect rejecting all diffs a bit better. Don't feel strongly about this, thoughts?

gui/src/hooks/ParallelListeners.tsx
```diff
@@ -271,29 +275,112 @@ function ParallelListeners() {
...
+                (state.acceptedDiffs || 0) + (state.rejectedDiffs || 0);
+              let feedbackMessage: string;
+
+              let finalOutput;
+
+              if (state.rejectedDiffs === totalDiffs && totalDiffs > 0) {
+                // All diffs were rejected - provide explicit stop instruction
+                feedbackMessage =
+                  "The search and replace tool executed successfully, but the user rejected ALL proposed changes. This means:\n\n1. DO NOT attempt any more file modifications\n2. DO NOT try different changes\n3. STOP and ask the user for clarification\n\nYou must now ask the user: 'I see you rejected all my proposed changes. Could you help me understand what you're looking for instead? What specific changes would you like me to make to the file?'";
+                finalOutput = {
```
RomneyDa, line 302: Note, the other editing tool (not search and replace, lazy edit) is still present and triggers the same `updateApplyState`. Would need to have handling for both or route it to this handling based on the tool call tool name

gui/src/util/clientTools/searchReplaceImpl.ts
```diff
@@ -90,14 +114,55 @@ export const searchReplaceToolImpl: ClientToolImpl = async (
...
-    // Return success - applyToFile will handle the completion state
+    // Store success information for later LLM feedback
+    const successOutput: ContextItem[] = [
+      {
+        name: "Search and Replace Results",
+        description: `Successfully applied ${allBlocks.length} replacements to ${filepath}`,
+        content: [
+          `Successfully modified ${filepath}`,
+          `Applied ${allBlocks.length} search/replace operations:`,
+          ...appliedBlocks,
```
RomneyDa, line 125: I might be a bit more conservative on the amount of code returned to the model, maybe comment out the applied diffs output for now but leave the functionality there for later iteration?

gui/src/hooks/ParallelListeners.tsx
```diff
@@ -242,6 +243,9 @@ function ParallelListeners() {
     dispatch(updateIndexingStatus(data));
   });
 
+  // Track processed toolCallIds to prevent duplicates
+  const processedToolCalls = useRef(new Set<string>());
+
   useWebviewListener(
     "updateApplyState",
```
RomneyDa, line 239: I think it could be nice to move the bulk of the parallel listeners code into a utility thunk, since parallel listeners is getting a bit cluttered and `updateApplyState` handler uses quite a bit of redux state

RomneyDa:
note merge conflicts

chezsmithy:
@RomneyDa thanks! I'll clean up the PR and review your feedback.

gui/src/util/clientTools/searchReplaceImpl.ts
chezsmithy: Sounds good. I'll do some testing to see what works more effectively. Going to experiment with removing request to read before changes and see how this plays with that.

PR #6431, opened 2025-07-03
build: Remove test dependencies from the release

exigow:
I noticed that `video-recorder-junit5` dependency was probably mistakenly added as a non-test dependency. As a result, the `buildPlugin` artifact included a lot of unnecessary dependencies including an old, vulnerable Apache Commons IO ([CVE-2021-29425](https://www.mend.io/vulnerability-database/CVE-2021-29425)).

List of dependencies before fix:

```
annotations-13.0.jar
apiguardian-api-1.0.0.jar
awaitility-3.1.6.jar
commons-io-1.4.jar
commons-lang3-3.8.1.jar
hamcrest-core-1.3.jar
hamcrest-library-1.3.jar
instrumented-continue-intellij-extension-1.0.27.jar
junit-jupiter-api-5.4.1.jar
junit-platform-commons-1.4.1.jar
kotlin-stdlib-1.8.10.jar
kotlin-stdlib-common-1.9.10.jar
kotlin-stdlib-jdk7-1.9.24.jar
kotlin-stdlib-jdk8-1.9.24.jar
kotlinx-serialization-core-jvm-1.5.0.jar
kotlinx-serialization-json-jvm-1.5.0.jar
log4j-api-2.20.0.jar
log4j-core-2.20.0.jar
log4j-over-slf4j-1.7.36.jar
objenesis-2.6.jar
okhttp-4.12.0.jar
okio-jvm-3.6.0.jar
opentest4j-1.1.1.jar
owner-1.0.10.jar
owner-java8-1.0.10.jar
posthog-1.2.0.jar
searchableOptions-1.0.27.jar
sentry-8.14.0.jar
sentry-log4j2-8.14.0.jar
sentry-okhttp-8.14.0.jar
slf4j-api-1.7.36.jar
video-recorder-core-2.0.jar
video-recorder-junit5-2.0.jar
zt-exec-1.10.jar
```

After:

```
annotations-13.0.jar
instrumented-continue-intellij-extension-1.0.27.jar
kotlin-stdlib-1.8.10.jar
kotlin-stdlib-common-1.9.10.jar
kotlin-stdlib-jdk7-1.9.24.jar
kotlin-stdlib-jdk8-1.9.24.jar
kotlinx-serialization-core-jvm-1.5.0.jar
kotlinx-serialization-json-jvm-1.5.0.jar
log4j-api-2.20.0.jar
log4j-core-2.20.0.jar
okhttp-4.12.0.jar
okio-jvm-3.6.0.jar
[...]

PR #6824, opened 2025-07-28
fix: Add file tests + migrate to IntelliJ Platform Gradle Plugin v2

exigow:
This is a batch of several fixes that I described below (in short: unit tests for file operations + migration with some cleanup).

I could prepare these changes separately, but it would be artificial because everything is a sort-of consequence of adding these unit tests, which in a way forced this cleanup (we had a problem running JUnit 3 tests -> I switched to platform v2 -> bumped Gradle -> saw that we don’t need to include the standard library so I removed it -> added verification to check that I’m not breaking anything).

Changes:

#### Add file test

The tests use fixtures and operate on real files in a temp directory. This change will allow us to confidently migrate from `java.io.File` to `VFS`.

Note: These tests are headless and starts the Continue binary. This gives us many possibilities because we can write fast and valuable tests without `runIdeForUiTests`.

#### Clean-up multiple JUnit runners

The IntelliJ Platform plugin already includes classes like `BasePlatformTestCase` written in JUnit 3, which we should be using for unit testing. We also have JUnit 5 for e2e testing. As a result, we had two different runners, which caused problems because we couldn’t use classes like UsefulTest since they required a different runner. In order to fix that, I had to add `org.junit.vintage:junit-vintage-engine` to enable running both JUnit 3 and JUnit 5 tests.

Note: ideally, we should have a sourceSet named `testIntegration` with its own dependencies for E2E (robot, JUnit 5, video recorder). That would give us a clean separation between E2E tests and unit tests. That also would prevent accidental use of JUnit 5 classes in unit tests. We should do that in the future.

#### Migrate to IntelliJ Platform Gradle Plugin v2

The new platform simplifies plugin config and allows us to remove some of the custom code. I left the original Qodana and Kover configs, but looking at our workflow, i think it's currently dead code. If we don’t plan to use them, I think we can remove these dependencies in the future.

Note: along with this migration, I had to bump a wrapper to latest version.

#### Remove Kotlin stdlib

For some reason, we included the Kotlin standard library with the plugin, even though we had a property specifying we didn’t want to. I saw no objections to removing this dependency, so I decided to do that.

From now on, we rely on older stdlib that's bundled with 2022.3+.

More context: https://github.com/continuedev/continue/pull/6431#issuecomment-3031838189

#### Add plugin verification

This proves my point regarding the removal of the standard library dependency.

I added plugin verification against versions `2022.3.3` (oldest) and `2025.1.4` (newest). Everything is working.

Plugin can be verified with `./gradlew verifyPlugin`. We can consider adding this to CI later.

#### Simplify build scripts

I made a few small improvements that simplify the build scripts:
* Some properties were removed because they duplicated content from `plugin.xml` (e.g., supported version numbers, plugin ID, and similar - all were unnecessarily overridden inside `intellij {}`/`intellijPlatform {}`).
* I replaced custom `prepareSandbox` with extension of `PrepareSandboxTask`, so there is no need to declare each case individually.
* From now on, properties are read via delegation instead of calling our custom `properties(...)` function. This is type-safe (if a property is not declared, the build will fail).
* I removed the `kotlin("test")` dependency, which we were only using for `assertTrue`. :sweat_smile: 

#### Note about e2e

The e2e Autocomplete tests are very flaky, and unfortunately, I haven’t been able to find a solution yet. I will try to address this in upcoming fixes. I believe that ultimately we should have max one e2e test describing the happy path, and the rest of the features should be tested with unit tests. Our e2e tests are very inconvenient, which results in no one running them (because we hijack the input), so they should be kept to a minimum. The rest of the e2e tests were already disabled (I assume I know why :D), so I didn't touch them.

#### Reduce number of dependencies

Fun fact: with the removal of the standard library, we managed to remove all unnecessary dependencies except for Sentry and PostHog (which are included for obvious reasons).

    

---

## Summary by cubic
Added unit tests for file operations, migrated to IntelliJ Platform Gradle Plugin v2, removed the Kotlin stdlib dependency, and cleaned up build scripts and test runners.

- **New Features**
  - Added headless file operation tests using real files in temp directories.
  - Added plugin verification against multiple IntelliJ versions.

- **Refactors**
  - Migrated to IntelliJ Platform Gradle Plugin v2 and updated Gradle wrapper.
  - Removed unnecessary Kotlin stdlib dependency.
  - Simplified build scripts and unified test runners for better reliability.

### DIFF
--- a/extensions/intellij/CONTRIBUTING.md
+++ b/extensions/intellij/CONTRIBUTING.md
@@ -116,15 +116,15 @@ If in doubt, check out the
 official [IntelliJ Platform Plugin Template](https://github.com/JetBrains/intellij-platform-plugin-template).
 These templates are the most up-to-date examples of how to correctly customize the plugin build scripts.
 
-Also, check out 
-the [useful recipes](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-recipes.html) 
+Also, check out
+the [useful recipes](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-recipes.html)
 for common problems.
 
 ### Adding new extensions in `plugin.xml`
 
-There's a tool called [JetBrains Platform Explorer](https://plugins.jetbrains.com/intellij-platform-explorer) that 
-aggregates plugin metadata and allows you to filter by specific 
-extension points. If you're having trouble implementing a feature that's not officially documented, 
+There's a tool called [JetBrains Platform Explorer](https://plugins.jetbrains.com/intellij-platform-explorer) that
+aggregates plugin metadata and allows you to filter by specific
+extension points. If you're having trouble implementing a feature that's not officially documented,
 you can learn from other open source plugins.
 
 ### Reloading changes


exigow:
Branch rebased / conflicts solved.

EDIT: All green.

PR #6866, opened 2025-07-29
fix: migrate FileUtils to VFS

exigow:
I replicated the original behavior, for example, writeFile should automatically create dirs + overwrite the file (if exists), readText normalizes line endings + limits the output to 10k chars, etc. All those details should be covered by unit test.

Also, I removed some usages of UriUtils because if we treat "file URIs" as URLs, we can use refreshAndFindFileByUrl. It works, but we should keep that in mind.

### Extra: Manual testing

(I did it myself on top of the unit tests, but I’m posting the instructions here in case someone wants to reproduce it)

* writeFile - run Continue in agentic mode and ask smth like: "create scss/test.scss"
* listDir (+ getFileStats) - ask: "list files in project root"
* openFile/showFile - edit a rule using the toolbar (pen icon)
* fileExists - is invoked on startup multiple times
* readFile - is invoked on startup multiple times (put some files into .continue/*)

    

---

## Summary by cubic
Migrated all file operations in FileUtils to use the IntelliJ Virtual File System (VFS) for better integration and reliability.

- **Refactors**
  - Replaced direct file access and UriUtils with VFS methods for reading, writing, listing, and checking files.
  - Updated related code in IntelliJIDE to use the new FileUtils methods.
  - Expanded and updated unit tests to cover VFS-based behavior.

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/file/FileUtils.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/file/FileUtils.kt
@@ -2,14 +2,17 @@ package com.github.continuedev.continueintellijextension.`continue`.file
 
 import com.github.continuedev.continueintellijextension.FileStats
 import com.github.continuedev.continueintellijextension.FileType
+import com.intellij.openapi.application.runReadAction
 import com.intellij.openapi.application.runWriteAction
 import com.intellij.openapi.fileEditor.FileDocumentManager
 import com.intellij.openapi.fileEditor.FileEditorManager
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.TextRange
 import com.intellij.openapi.vfs.VfsUtil
 import com.intellij.openapi.vfs.VfsUtilCore
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
+import kotlin.math.min
 
 
 class FileUtils(
@@ -48,7 +51,10 @@ class FileUtils(
     fun readFile(fileUri: String, maxLength: Int = 100_000): String {
         val found = findFile(fileUri)
             ?: return ""
-        val text = VfsUtil.loadText(found, maxLength)
+        val text = runReadAction {
+            // note: document (if exists) is more up-to-date than VFS
+            readDocument(found, maxLength) ?: VfsUtil.loadText(found, maxLength)
+        }
         return normalizeLineEndings(text)
     }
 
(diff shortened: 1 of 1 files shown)


exigow:
Note: After these changes, most file operations now use VFS. However, there are still a few remaining usages of `java.io.File`. I decided to commit the current changes as a working checkpoint and will continue later.

exigow:
Note: I'll test it on :window: Windows tomorrow, but I'm pretty sure I didn't break anything since I kept the original OS-specific logic (like this URI authorities).

Also, I noticed that we have code for removing parameters from file URIs, but I didn't find any case where we'd actually receive something like that in a message. I decided to leave it as-is.

Patrick-Erichsen (approved):
Thanks for the detailed writeup + tests!

Patrick-Erichsen:
@exigow apologies, I think we already discussed this morning, but is this good to merge?

exigow:
> @exigow apologies, I think we already discussed this morning, but is this good to merge?

@Patrick-Erichsen, no worries! That's correct, we can merge, but there are some issues with the e2e tests. I'll need to take care of fixing those tests this week. They're either flaky, or I (?) introduced a regression at some point. However, I don't think it's anything serious, since I'm manually testing the plugin for the functionality being covered. I suspect something broke directly in the test itself.

Patrick-Erichsen (approved):
Approved, but guessing the current failing tests will require some additional commits


```
node:internal/modules/cjs/loader:1215
  throw err;
  ^

Error: Cannot find module 'esbuild'
Require stack:
- /home/runner/work/continue/continue/binary/build.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1212:15)
    at Module._load (node:internal/modules/cjs/loader:1043:27)
    at Module.require (node:internal/modules/cjs/loader:1298:19)
    at require (node:internal/modules/helpers:182:18)
    at Object.<anonymous> (/home/runner/work/continue/continue/binary/build.js:1:17)
    at Module._compile (node:internal/modules/cjs/loader:1529:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1613:10)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
    at Module._load (node:internal/modules/cjs/loader:1096:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:164:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/home/runner/work/continue/continue/binary/build.js' ]
}
```

exigow:
@Patrick-Erichsen, FYI it's :green_circle: fixed / now I know how to run the e2e tests locally. There was indeed an issue with `readFile` (https://github.com/continuedev/continue/pull/6866/commits/29ec8c84ac9919abca77c71e5994d81966c1d29c).

PR #6874, opened 2025-07-29
feat: Plug and play system message tool frameworks

RomneyDa:
## Description

More than half of the system message tools code can be plug and play with other ways to write/parse system message tools, such as the XML approach in the original PR.

Some models might do best with the XML or other approaches, or we might want to experiment with them. This PR pulls the tool codeblock format approach (```tool) into a plug and play `SystemMessageToolsFramework`, to allow easy other approaches later.
    

---

## Summary by cubic
Refactored the system message tools code to use a plug-and-play framework, making it easy to swap or add new tool codeblock formats like XML in the future.

- **Refactors**
  - Moved the tool codeblock logic into a `SystemMessageToolsFramework` interface and a default implementation.
  - Updated all tool definitions and system message generation to use the new framework.
  - Removed hardcoded codeblock logic and centralized tool parsing, formatting, and detection.
  - Updated tests and Redux logic to support the new framework structure.

### DIFF
--- a/core/tools/definitions/searchAndReplaceInFile.ts
+++ b/core/tools/definitions/searchAndReplaceInFile.ts
@@ -10,9 +10,6 @@ export interface SearchAndReplaceInFileArgs {
 export const NO_PARALLEL_TOOL_CALLING_INSRUCTION =
   "Note this tool CANNOT be called in parallel.";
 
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen:
Would it be better for @sestinj to review this? Think he might have more context from the original PR

sestinj:
Tbh I don't think this solves a problem for us right now. It's more forward looking. Seems like we could close for now and if it became needed ask Continue to re-apply the changes.

sestinj:
This seems fine at first glance though isn't completely necessary. Should be careful, probably test it well if we are going to merge it at all, and definitely do not expose this as an option quite yet.

Patrick-Erichsen (approved):

PR #6890, opened 2025-07-30
fix: :bug: Fix CodebaseIndexer Bugs

chezsmithy:
## Description

* Introducing codebase indexer refreshes based on configuration changes introduced a series of race condition bugs that were causing the abort signal become undefined during initalization resulting in a "signal undefined" error showing up during indexing. This was most noticible in IntelliJ but could also happen in VsCode. In IntelliJ it often had the side effect of breaking the startup causing the Chat window to never generate any LLM output.

* The pause codebase indexer on start VsCode setting no longer made much sense becaue it was tied to having a configuration and a pause request. As such I redefined it slightly and made sure that not only is it paused on start, it's also paused on configuration switches.

* There were some interesting data uniqueness issues during tag/cacheKey creation and indexing especially if you happended to swap between the intelliJ and vscode extension in the same repo. I applied a fix to the addTag function so that if the tag/cacheKey combination already exists, to ignore it as a soft error as it's likely one of the extensions already added it. This change did not affect the index content or status, and cleaned up a common indexing error I observed.

## Checklist

## Screenshots

## Tests

Performed the following manual tests:

* Set the VsCode pause on start setting and observe that the codebase indexing is always in a paused state on start, and on configuration changes.
* Started vscode indexing and manually paused the indexing. Stopped the extension and restarted. Noted it was still paused. Restarted an it finished sucessfully.
* Noted I had a local IntellliJ manged database in the state where a tag/cacheKey was duplicating during addTag. This was ignored and the index was updated with the soft bypass. No changes to the funcitonallity.
* Verified that during startup if the global abort signal in codeBaseIndexer gets set to undefined due to race condtions or duplicate funciton calls that the in process function continues as it's using a locally initalized version of the AbortController now in the inner loop.

    

---

## Summary by cubic
Fixed race conditions and tag duplication bugs in the codebase indexer to prevent startup errors and improve stability in both VSCode and IntelliJ.

- **Bug Fixes**
  - Ensured the abort signal is always defined during indexing to avoid "signal undefined" errors.
  - Updated pause-on-start behavior to also pause indexing on configuration changes.
  - Handled duplicate tag/cacheKey entries gracefully to prevent errors when switching between extensions.

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -204,24 +204,6 @@ export class Core {
     dataLogger.ideInfoPromise = ideInfoPromise;
     dataLogger.ideSettingsPromise = ideSettingsPromise;
 
-    void ideSettingsPromise.then((ideSettings) => {
-      // Index on initialization
-      void this.ide.getWorkspaceDirs().then(async (dirs) => {
-        // Respect pauseCodebaseIndexOnStart user settings
-        if (ideSettings.pauseCodebaseIndexOnStart) {
-          this.codeBaseIndexer.paused = true;
-          void this.messenger.request("indexProgress", {
-            progress: 0,
-            desc: "Initial Indexing Skipped",
-            status: "paused",
-          });
-          return;
-        }
-
-        void this.codeBaseIndexer.refreshCodebaseIndex(dirs);
-      });
-    });
-
     const getLlm = async () => {
       const { config } = await this.configHandler.loadConfig();
       if (!config) {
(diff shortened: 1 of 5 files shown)


Patrick-Erichsen:
Nice, thanks for the fixes here! The failing tests appear valid: https://github.com/continuedev/continue/actions/runs/16617533730/job/47013631965?pr=6890#step:6:423

I think the assertions just need to be updated to match your new logic, i.e. the tests are broken not your logic.

chezsmithy:
@Patrick-Erichsen yep, looks like I unlocked a few tests to run in new way. Grouped the tests and added cleanup. Ready for review.

Patrick-Erichsen (approved):

PR #6894, opened 2025-07-30
fix: apply Sentry and PostHog telemetry settings correctly

exigow:
Now we are very strict about telemetry config. If anything fails (plugin service is null, no/invalid config response, etc.), the flag falls back to `false`.

More details:
* Because we need to poll settings ourselves (via `getSerializedProfileInfo`), I created a new service that holds that setting.
* This service implements an interface to allow separation that enables us to write unit tests. Unit tests only cover PostHog telemetry behavior. I tried to write tests for Sentry, but it’s problematic because the initialization is static and I couldn’t mock it (`mockkStatic` doesn’t work). I decided not to spend more time on it, especially since the telemetry checks are identical to PostHog (it's just `if` statement). I think that should be enough.
* Status polling is lazy, so as long as no one asks for `allowAnonymousTelemetry` (so if no error occurs), polling is inactive. This is a side effect of having it as a service, but I think it’s a nice feature in this context.
* If telemetry is disabled but the user submits an error via the `Report to Continue` dialog, report will be sent anyway (because this is an explicit decision). This action is one-time and does not change the global telemetry settings.

To summarize, we now have three services:

* `ContinueTelemetryStatusService` - holds state in a non-blocking way and handles polling itself
* `ContinuePostHogService` - previously `TelemetryService`
* `ContinueSentryService` - previously `ContinueErrorService`

    

---

## Summary by cubic
Telemetry settings for Sentry and PostHog are now strictly enforced, so if config is missing or invalid, telemetry is always disabled by default. Added a new service to manage telemetry status and refactored telemetry code to use this service.

- **Refactors**
  - Replaced the old TelemetryService with ContinueTelemetryStatusService, ContinuePostHogService, and ContinueSentryService.
  - All telemetry checks now use a shared interface for easier testing and stricter control.
  - Added unit tests for PostHog telemetry behavior.


exigow:
I'm setting it back to Draft because I want to double-check this (trigger exceptions artificially, change settings multiple times, etc)

exigow:
I toggled telemetry on and off within a single session and it works (I was throwing exceptions via custom action).

```
(telemetry disabled)

2025-07-30 19:11:38,380 [ 161853]   WARN - #com.github.continuedev.continueintellijextension.error.ContinuePostHogService - PostHog capture was ignored because telemetry is disabled
2025-07-30 19:11:38,419 [ 161892]   WARN - #com.github.continuedev.continueintellijextension.error.ContinueSentryService - Sentry message report was ignored because telemetry is disabled

(telemetry enabled)

2025-07-30 19:14:50,658 [  25678]   WARN - #com.github.continuedev.continueintellijextension.error.ContinuePostHogService - Telemetry sent to PostHog: test
2025-07-30 19:14:50,707 [  25727]   WARN - #com.github.continuedev.continueintellijextension.error.ContinueSentryService - Message sent to Sentry: test
```

PR #6897, opened 2025-07-30
fix: build `packages` in CI step

Patrick-Erichsen:


Patrick-Erichsen:
Actually this is unnecessary because we already build the packages in a previous step

PR #6932, opened 2025-08-02
feat: :sparkles: MCP Prompts now display when inserted

chezsmithy:
## Description

Using /slash commands to insert MCP prompts now displays the prompt content consistent with other prompts like /init

MCP prompt content is loaded at config change at the same time we are loading the MCP prompt definition and are cached.

## Checklist

image

## Tests

Manual verification performed.

### DIFF
--- a/core/config/profile/doLoadConfig.ts
+++ b/core/config/profile/doLoadConfig.ts
@@ -18,6 +18,7 @@ import {
   SlashCommandDescWithSource,
   Tool,
 } from "../../";
+import { stringifyMcpPrompt } from "../../commands/slash/mcpSlashCommand";
 import { MCPManagerSingleton } from "../../context/mcp/MCPManagerSingleton";
 import CurrentFileContextProvider from "../../context/providers/CurrentFileContextProvider";
 import MCPContextProvider from "../../context/providers/MCPContextProvider";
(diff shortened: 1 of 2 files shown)


core/config/profile/doLoadConfig.ts
```diff
@@ -216,15 +217,39 @@ export default async function doLoadConfig(options: {
...
       const serverSlashCommands: SlashCommandDescWithSource[] =
-        server.prompts.map((prompt) => ({
-          name: prompt.name,
-          description: prompt.description ?? "MCP Prompt",
-          source: "mcp-prompt",
-          isLegacy: false,
-          mcpServerName: server.name, // Used in client to retrieve prompt
-          mcpArgs: prompt.arguments,
-        }));
+        await Promise.all(
```
Patrick-Erichsen, line 222: Only concern is this might take longer than we anticipate. Mind throwing a timer + console log for observability purposes?

Patrick-Erichsen (commented):
This is great! I haven't used MCP prompts much but very nice to have an actual UI here.

Patrick-Erichsen:
Merging main should hopefully resolve the errors here!

RomneyDa:
On second thought since it's such an edge case that mcp prompts will change between config loads I think I prefer your original approach so that there's no UI lag

PR #6933, opened 2025-08-02
fix: :green_heart: Remove clean on build

chezsmithy:
## Description

Before

    // Phase 3: Build packages that depend on other local packages
    await buildPackagesInParallel(["openai-adapters", "continue-sdk"], true);

After

    // Phase 3: Build packages that depend on other local packages
    await buildPackagesInParallel(["openai-adapters", "continue-sdk"]);

I'm not clear why we would need to clear node_modules before building these two packages.

In the local debug build the other stages don't wait for this Continue Build to complete, and as such when the cleanup runs it can sometimes break downstream build steps when things are running in parallel.

## Checklist

## Tests

### DIFF
--- a/scripts/build-packages.js
+++ b/scripts/build-packages.js
@@ -86,7 +86,7 @@ async function main() {
     await buildPackagesInParallel(["fetch", "config-yaml", "llm-info"]);
 
     // Phase 3: Build packages that depend on other local packages
-    await buildPackagesInParallel(["openai-adapters", "continue-sdk"], true);
+    await buildPackagesInParallel(["openai-adapters", "continue-sdk"]);
 
     console.log("🎉 All packages built successfully!");
   } catch (error) {


chezsmithy:
@Patrick-Erichsen I did some work to fix up the builds. I think I got to the bottom of why the ripgrep download would fail periodically as well based on API rate limiting. Moved a few things around but I feel really good about this. Noting you on this as you were the last one that touched the build system.

Patrick-Erichsen:
Thank you for taking the time to clean up my mess! 😅  I had even noticed the `ripgrep` errors but didn't think to check if I had incorrectly updated the secret passing logic with the new actions we introduced.

We ended up making a few changes to the build system in parallel and it didn't all come together cleanly. 

Also, I'm hopeful that the improvements around `vscode dist copy times` will fix some flake I've noticed with the non-required `PR Build` steps. The flake seems to almost always be at that stage of the build.

Patrick-Erichsen (approved):

PR #6723, opened 2025-07-20
feat: System Message tools

RomneyDa:
A remake of https://github.com/continuedev/continue/pull/5809 and https://github.com/continuedev/continue/pull/6395/

Adds support for tools in the form of system message instructions and response content parsing + tests
Tools are parsed as they come in, converted to ToolCalls, and on the way out converted back to text. Switching between native and non-native tools is interchangeable with this approach.
For now, this applies only to models that are NOT known to support native tools
Adds a warning to the Agent mode dropdown that shows for any models not known to be great with tools (including models that support native tools)
All models can now be used in Agent mode

PR #6934, opened 2025-08-02
fix: :ambulance: Tool descriptions are not being added for nativeTool enabled models

chezsmithy:
## Description

🚨  PR: #6723 is no longer added the full tool description when the model can use native tool calling. The tool description was refactored to have a description and a systemMessageDescription. For non native models, the tool descriptions are appended to the system message. For native tool calling models the description should include all the needed information. 

This change updates the tool description for native tool calling models by concating the tool description and the systemMessgeDescription. I've noted in testing this fixes the tool calling issues that have been introduced after #6723 was merged to main. 

I'd recommend merging this fix before releasing another pre-release. Also note this fix must be bundled with #6723 in any release to not cause a broken tool calling experience.

This should be viewed as a critical hotfix to the pre-release channel.

Note, I've also updated the parallel tool calling instruction to make it more forceful. I'm willing to put that into a separate PR if it might be risky. I've noted with this change I see parallel tool calls with claude sonnet 4.

## Checklist

## Tests

I've not added any new tests. I've performed inspection of the options.tools values by debugging and the system messages when nativeTool use is enabled.

    

---

## Summary by cubic
Fixed an issue where tool descriptions were missing for models with nativeTool enabled, ensuring all relevant tool information is included.

- **Bug Fixes**
  - Combined function and system message descriptions for native tools so both are shown to the model.

### DIFF
--- a/core/llm/defaultSystemMessages.ts
+++ b/core/llm/defaultSystemMessages.ts
@@ -60,8 +60,7 @@ ${EDIT_CODE_INSTRUCTIONS}
 export const DEFAULT_AGENT_SYSTEM_MESSAGE = `\
 <important_rules>
   You are in agent mode.
-
-  If you need to use multiple tools, you can call call multiple read only tools simultaneously.
+  Make parallel tool call unless the tool description specifically states it is not supported.
 
 ${CODEBLOCK_FORMATTING_INSTRUCTIONS}
 </important_rules>`;
@@ -72,6 +71,7 @@ export const DEFAULT_PLAN_SYSTEM_MESSAGE = `\
 <important_rules>
   You are in plan mode, in which you help the user understand and construct a plan.
   Only use read-only tools. Do not use any tools that would write to non-temporary files.
+  Make parallel tool call unless the tool description specifically states it is not supported.
   If the user wants to make changes, offer that they can switch to Agent mode to give you access to write tools to make the suggested updates.
 
 ${CODEBLOCK_FORMATTING_INSTRUCTIONS}
--- a/gui/src/redux/thunks/streamNormalInput.ts
+++ b/gui/src/redux/thunks/streamNormalInput.ts
@@ -3,7 +3,6 @@ import { LLMFullCompletionOptions, ModelDescription, Tool } from "core";
 import { getRuleId } from "core/llm/rules/getSystemMessageWithRules";
 import { ToCoreProtocol } from "core/protocol";
 import { BuiltInToolNames } from "core/tools/builtIn";
-import posthog from "posthog-js";
 import { selectActiveTools } from "../selectors/selectActiveTools";
 import { selectUseSystemMessageTools } from "../selectors/selectUseSystemMessageTools";
 import { selectSelectedChatModel } from "../slices/configSlice";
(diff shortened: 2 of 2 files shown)


chezsmithy:
@RomneyDa flagging you as a reviewer on this change as you developed #6723

chezsmithy:
Note, CI/CD is likely to fail on this PR until #6933 gets merged.

RomneyDa (changes requested):
@chezsmithy systemMessageDescription is included in the system message, see
https://github.com/continuedev/continue/blob/main/core/tools/systemMessageTools/buildToolsSystemMessage.ts#L112

System Message Description is an _alternative_ to the tools JSON, _only_ to be used for system message tools.

Could you reiterate the fix here?

RomneyDa:
@chezsmithy could merge parallel fix on separate pr!

chezsmithy:
@RomneyDa in my debugging for say AWS Bedrock with Sonnet. I noticed that the System Message doesn't include any of the tool definitions as it's native tool use, but then in the tools array passed to the model, the tool description is just the description field and not the systemMessageDescription. Thus the model gets a list of tools, with a very basic agent system message and is left to figure out how to use the tools from the shape of the payloads and a VERY basic tool description. This results in the model using the tools very poorly. Does that make sense?

This fix, re-introduces what would have been in the decription back to the tools array description field fixing tools use for native tools using models.

RomneyDa:
> @RomneyDa in my debugging for say AWS Bedrock with Sonnet. I noticed that the System Message doesn't include any of the tool definitions as it's native tool use, but then in the tools array passed to the model, the tool description is just the description field and not the systemMessageDescription. Thus the model gets a list of tools, with a very basic agent system message and is left to figure out how to use the tools from the shape of the payloads and a VERY basic tool description. This results in the model using the tools very poorly. Does that make sense?

The system message description is brief blurb + an example of calling the tool using system message tools, e.g. 
```tool
TOOL_NAME: read_file
BEGIN_ARG: filepath
test.js
END_ARG
```

It doesn't apply for native tools, it's mostly duplicate. In many cases, it's more concise. If there is information missing from the normal tool description which is in the system message description, we should add it to the normal tool description.

RomneyDa:
Another way of putting this is that when I added systemMessageDescriptions (specifically used for system message tools) I did not remove anything from the other descriptions. If I did, it was my mistake!

chezsmithy:
> Another way of putting this is that when I added systemMessageDescriptions (specifically used for system message tools) I did not remove anything from the other descriptions. If I did, it was my mistake!

Ah! Gotcha. So yeah we lost some significant content from the descriptions.

RomneyDa:
@chezsmithy which tools lost content?

RomneyDa:
I took a look through and it looks like specifically search and replace was effected negatively. Does this seem right to you?

RomneyDa:
I took a look through and it looks like in all cases the JSON descriptions are at least as thorough as the alternative system message ones, although the system message tools search and replace description seems to have missed an update during a merge. Any tools seem off to you?

RomneyDa:
Closing this for now but keeping a close eye on any agent issues so please reopen something if needed or if I missed something!

Also the parallel note would be a nice change, have also noticed claude tends to still do series calls

PR #6937, opened 2025-08-02
fix: add maximum 4 cache control blocks for anthropic

uinstinct:
## Description

Previously the following error was received when cache control was enabled for anthropic providers

`Error in Continue server: {"type":"error","error":{"type":"invalid_request_error","message":"A maximum of 4 blocks with cache_control may be provided. Found 5."}}`

This PR caps the total cache control blocks (including system messages, tools, and user messages) to be 4.

resolves CON-3176

references https://discord.com/channels/1108621136150929458/1400509002898477117

## Checklist

## Tests

---

## Summary by cubic
Limited the number of cache_control blocks sent to Anthropic to a maximum of 4 to prevent invalid request errors.

- **Bug Fixes**
  - Ensured that no more than 4 cache_control blocks are added across system, tools, and user messages.

### DIFF
--- a/packages/openai-adapters/src/apis/AnthropicCachingStrategies.ts
+++ b/packages/openai-adapters/src/apis/AnthropicCachingStrategies.ts
@@ -1,3 +1,5 @@
+const MAX_CACHING_MESSAGES = 4;
+
 // Caching strategy type - transforms a clean Anthropic body by adding cache_control
 export type CachingStrategy = (anthropicBody: any) => any;
 
@@ -9,7 +11,7 @@ const noCachingStrategy: CachingStrategy = (body) => body;
 
 // Strategy 2: System Messages Only (Current Implementation)
 const systemOnlyStrategy: CachingStrategy = (body) => {
-  let availableCacheMessages = 4;
+  let availableCacheMessages = MAX_CACHING_MESSAGES;
   if (body.system && Array.isArray(body.system)) {
     return {
       ...body,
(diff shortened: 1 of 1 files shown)


packages/openai-adapters/src/apis/AnthropicCachingStrategies.ts
```diff
@@ -9,13 +9,20 @@ const noCachingStrategy: CachingStrategy = (body) => body;
 
 // Strategy 2: System Messages Only (Current Implementation)
 const systemOnlyStrategy: CachingStrategy = (body) => {
+  let availableCacheMessages = 4;
```
Patrick-Erichsen: Could we set this as a global const in the file isntead?
```suggestion
  const MAX_CACHE_MESSAGES = 4;
  
  const systemOnlyStrategy: CachingStrategy = (body) => {
  let availableCacheMessages = MAX_CACHE_MESSAGES;
```

Just to make sure all strategies are using the same base.

Patrick-Erichsen (changes requested):
Thanks for adding the tests! Just had one nitpick on removing a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)).

One other question - is this the only place we need to make the update, or do we need to update the non-OAI-adapters version as well? i.e. `core/llm/llms/Anthropic.ts`

packages/openai-adapters/src/apis/AnthropicCachingStrategies.ts
uinstinct: implemented!

PR #6939, opened 2025-08-03
feat: relocate the codebase indexing tab

shssoichiro:
## Description

This moves the codebase indexing status from its current location underneath the Settings icon, to a new tab next to the Docs tab, inline with the others.

This improves the accessibility of the codebase index tab, making it easier to find and requiring only one click to reach. This also brings it in line with the docs indexing tab so that they are providing a more similar experience.

Additionally, this fixes the text on the disabled state to show "Enable Indexing", matching the text on the toggle in the Settings tab.

## Checklist

Settings view looks correct with Indexing tab removed:

Screenshot_20250803_032116

Codebase tab displays, and has a tooltip when hovered:

Screenshot_20250803_032133

When opened, the Codebase tab displays the same contents as the current Indexing tab:

Screenshot_20250803_032145

The disabled state also displays correctly and the text matches the settings toggle:

Screenshot_20250803_032558

## Tests

All existing tests pass, and manual verification was done (see screenshots).

    

---

## Summary by cubic
Moved the codebase indexing status to a new tab next to the Docs tab for easier access and updated the disabled state text to say "Enable Indexing".

### DIFF
--- a/gui/src/components/mainInput/Lump/LumpToolbar/BlockSettingsTopToolbar.tsx
+++ b/gui/src/components/mainInput/Lump/LumpToolbar/BlockSettingsTopToolbar.tsx
@@ -2,6 +2,7 @@ import {
   BookOpenIcon,
   ChatBubbleLeftIcon,
   ChevronLeftIcon,
+  CircleStackIcon,
   CubeIcon,
   EllipsisHorizontalIcon,
   ExclamationTriangleIcon,
@@ -48,6 +49,12 @@ const sections: Section[] = [
   { id: "models", title: "Models", tooltip: "Models", icon: CubeIcon },
   { id: "rules", title: "Rules", tooltip: "Rules", icon: PencilIcon },
   { id: "docs", title: "Docs", tooltip: "Docs", icon: BookOpenIcon },
+  {
+    id: "codebase",
+    title: "Codebase",
+    tooltip: "Codebase",
+    icon: CircleStackIcon,
+  },
   {
     id: "prompts",
     title: "Prompts",
--- a/gui/src/components/mainInput/Lump/sections/CodebaseSection.tsx
+++ b/gui/src/components/mainInput/Lump/sections/CodebaseSection.tsx
@@ -1,0 +1,26 @@
+import IndexingProgress from "../../../../pages/config/IndexingProgress";
+import { useAppSelector } from "../../../../redux/hooks";
+
+export function CodebaseSection() {
+  const config = useAppSelector((state) => state.config.config);
+  return (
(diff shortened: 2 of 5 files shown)


shssoichiro:
The CI error appears to be spurious: `Downloading ripgrep failed: Error: Request failed: 403`

RomneyDa (changes requested):
@shssoichiro the current direction is to slowly deprecate codebase RAG indexing, which is why this is more hidden than docs. The notch tabs are precious UX space and we've decided to not add codebase to this for now. Also, codebase indexing is designed to be in the background so the user doesn't have to think about it. If they do, there's probably an issue with the system. In which situations do you frequently find yourself visiting the indexing tab?

The copy fix (disable -> enable) would be a great fix, would love to merge that.

shssoichiro:
The issues I've found that led to wanting this tab more visible are indeed related to codebase indexing errors. The most frequent being that the codebase indexing will fail when VSCode initially loads (the common error seems to be `Cannot read properties of undefined (reading 'signal')`), so I have to visit the Indexing tab, see that often there is an error, and click Re-index, and then it works. Essentially, by hiding the tab, errors are also hidden from the user.

May I ask why the codebase indexing overall is being deprecated? I know that is outside the scope of this ticket, but I'm curious as it seems like a useful feature.

Pyroboomka:
@RomneyDa Can you tell us a bit more why are you moving into the direction of RAG deprecation? It seems like a very useful feature, at least for me. Is the plan is to make it more implicit (use under the hood usage as a tool) or you think that grep/search tools are more useful than codebase index?

RomneyDa:
@Pyroboomka @shssoichiro exactly, grep and glob in practice seem much more effective. We don't have strong evals on this but feedback has been pretty strong towards agent grep/glob vs codebase. Would definitely be interested in any case studies/evals that show otherwise!

RomneyDa:
@shssoichiro is there an issue for the signal undefined error? Maybe we could fix the error to prevent needing to reindex frequently

RomneyDa:
@shssoichiro talked with team, sounds like we won't move forward with this move but would be interested to fix the bug you mentioned.

I also opened another PR with the `Enable` copy fix and added you as coauthor
https://github.com/continuedev/continue/pull/7056

RomneyDa:
@shssoichiro we are also considering moving the docs tab into settings with the codebase indexing statuses. 
I will

RomneyDa:
@shssoichiro, @uinstinct is working on a fix for the indexing error
https://github.com/continuedev/continue/pull/7062

PR #7007, opened 2025-08-06
feat: improvements to docs indexing status visibility

shssoichiro:
## Description

- Add a line to the Docs tab for each doc displaying the current indexing status. Since indexing for some sites can take a long time, users may wonder why it is taking a very long time.
- If the indexing for a doc is completed, show the number of pages indexed.
- When clicking the number of pages indexed, show a tooltip with the list of pages. This improves visibility to see which pages in a site were indexed.

## Checklist

Screenshot_20250805_210609
<br/>
image
<br/>
image

## Tests

Completed manual testing. Added unit tests for GUI components. DocsService tests are still having mocking issues, so left those untouched+skipped.

    

---

## Summary by cubic
Improved the Docs tab by showing the current indexing status for each doc, including the number of pages indexed and a tooltip with the list of indexed pages.

- **UI Improvements**
  - Added a status line for each doc showing indexing progress or completion.
  - When indexing is complete, displays the number of pages indexed.
  - Clicking the page count opens a tooltip listing all indexed pages.

### DIFF
--- a/gui/src/components/mainInput/Lump/sections/docs/DocsDetailsDialog.tsx
+++ b/gui/src/components/mainInput/Lump/sections/docs/DocsDetailsDialog.tsx
@@ -1,21 +1,19 @@
 import { DocsIndexingDetails } from "core";
-import { usePostHog } from "posthog-js/react";
 import { useCallback, useContext, useEffect, useState } from "react";
 import { useDispatch } from "react-redux";
 import { Tooltip } from "react-tooltip";
 import { SecondaryButton } from "../../../..";
 import { IdeMessengerContext } from "../../../../../context/IdeMessenger";
-import { useAppSelector } from "../../../../../redux/hooks";
 import {
   setDialogMessage,
   setShowDialog,
 } from "../../../../../redux/slices/uiSlice";
+
 interface DocsDetailsDialogProps {
   startUrl: string;
 }
+
 function DocsDetailsDialog({ startUrl }: DocsDetailsDialogProps) {
-  const config = useAppSelector((store) => store.config.config);
-  const posthog = usePostHog();
   const dispatch = useDispatch();
 
   const ideMessenger = useContext(IdeMessengerContext);
@@ -49,7 +47,7 @@ function DocsDetailsDialog({ startUrl }: DocsDetailsDialogProps) {
   }, []);
 
   useEffect(() => {
-    fetchData();
+    void fetchData();
   }, []);
 
   let comp = <div>Loading...</div>;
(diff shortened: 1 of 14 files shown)


gui/src/components/mainInput/Lump/sections/docs/DocsDetailsDialog.tsx
shssoichiro, line 1: This file was not used anywhere and seemed to be safe to delete. It referenced chunks rather than pages, so it is likely to be less useful to users.

RomneyDa (changes requested):
@shssoichiro the deleted DocsDetailsDialog (which used the existing `getDetails` function provided more or less the same data (but chunks instead of page urls). This PR could just show that dialog, or if you prefer just the URLs, it should cleanly remove the getDetails functionality. Note that `getDetails` was a bit more comprehensive, you may want to look into the differences.

RomneyDa:
@shssoichiro I think a nice implementation would be to open a virtual file in the IDE listing all the URLs, rather than having a massive tooltip. I originally made a tooltip and then moved it to the DocsDetailsDialog, which was later hidden because it was a bit noisy in the UI. Thoughts on virtual file approach?

shssoichiro:
I took a look at the `DocsDetailsDialog`, the big difference seems to be that it is listing the indexed content alongside each chunk. How useful do you think that would end up being other than for debugging? Currently the pages in the list are clickable, so the user can open them in the browser.

I think if we are following the tooltip approach, just showing the pages list feels a lot cleaner, but if we move to a virtual file approach then that would be better for showing the indexed content alongside as in the `DocsDetailsDialog`. Thoughts?

RomneyDa:
Ah yes it was originally for debugging so you could see what was being indexed, since e.g. the CheerioCrawler has odd indexing output for some sites with h3 headers, etc.

Now I understand it's more of a feature to be able to click the links. Could we either re-add a little eye icon to a docs status that opens the docs details dialog for debugging, or fully clean it up by removing `getDetails` message/function/types/etc before merging this?

shssoichiro:
Since we have the code in place already, it does seem useful to be able to use the existing dialog for debugging. I'll add an icon to allow opening it. Thanks.

shssoichiro:
I've added in the icon to display the dialog. Since we are restricted to the sidepane area, it unfortunately leaves the dialog a bit cramped, but the sidepan can be expanded if needed. The alternative would be to rewrite the docs dialog as text/markdown to display in a virtual file tab, but that wasn't much prettier, so I took the path of using the existing component.

image
image

shssoichiro:
Thanks, added.

PR #7058, opened 2025-08-08
fix: function definition chunking to respect max chunk size

uinstinct:
## Description

When constructing function definition chunk, utilize the max chunk size.

- When a function is inside a class, use class header + collapsed body
- Else, fall back to the function signature + collapsed body
- Else, fall back to first line of signature + collapsed body
- Else, return just the collapsed body

resolves CON-2250

## Checklist

## Tests

### DIFF
--- a/core/indexing/chunk/code.ts
+++ b/core/indexing/chunk/code.ts
@@ -128,25 +128,46 @@ async function constructFunctionDefinitionChunk(
   maxChunkSize: number,
 ): Promise<string> {
   const bodyNode = node.children[node.children.length - 1];
-  const funcText =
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
@uinstinct can you add a unit test that verifies this case?

RomneyDa (approved):
Fixes underlying cause of chunking failures, nice

RomneyDa:
@uinstinct I think you may need to clean up the async encoder after the tests, or they will run forever in CI. Tests ran 37 minutes here before I cancelled ^

PR #7071, opened 2025-08-09
Fix: Watch the block YAML files in the global directory

Jazzcort:
## Description

File watching now includes block YAML files in the global directory, removing the need for manual config reloads.

## Checklist

## Tests

Check if the config gets reloaded automatically after modifying the block YAML files in the global directory.

    

---

## Summary by cubic
Block YAML files in the global directory are now watched for changes, so config reloads happen automatically without manual steps.

- **Bug Fixes**
 - Added file watcher for block YAML files to trigger config reloads when modified.

### DIFF
--- a/extensions/vscode/src/extension/VsCodeExtension.ts
+++ b/extensions/vscode/src/extension/VsCodeExtension.ts
@@ -6,11 +6,7 @@ import { EXTENSION_NAME, getControlPlaneEnv } from "core/control-plane/env";
 import { Core } from "core/core";
 import { FromCoreProtocol, ToCoreProtocol } from "core/protocol";
 import { InProcessMessenger } from "core/protocol/messenger";
-import {
-  getConfigJsonPath,
-  getConfigTsPath,
-  getConfigYamlPath,
-} from "core/util/paths";
+import { getContinueGlobalPath } from "core/util/paths";
 import { v4 as uuidv4 } from "uuid";
 import * as vscode from "vscode";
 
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
@Jazzcort thinking we fix this inside the `files/changed` message in core instead, as it's IDE-agnostic and already has similar logic, and if yaml detection isn't working there then it likely isn't working for jetbrains as well

Jazzcort:
@RomneyDa Doesn't the `files/changed` only monitor files at workspace level? The global directory `~/.continue` is monitored by the file watcher here. It was only watching `config.yaml`, `config.json`, and `config.ts` before this patch so I extended it to also watch those block directories. Let me know if I'm not correct. 😄

RomneyDa:
I think `files/changed` is just called whenever a file is changed in the IDE - it's based off of an IDE event. Might be wrong, have you verified otherwise?

EDIT, just verified the message does come through for files outside the workspace

It seems like two issues could be causing this to not reload config
1. [Object object] in profile URI

image

2. The `URI.equal` expression in `handleFilesChanged` seems to return true when they're not the same

RomneyDa:
@Jazzcort fixed here which seems to fix config reloading for editing files outside the workspace when using hub assistants at least!
https://github.com/continuedev/continue/pull/7156

Jazzcort:
@RomneyDa Thanks for the update! I'm on a vacation right now and will be back next Wednesday. I'll test it against local block files at the global directory and if it can also catch the changes with auto reload I'll do another PR to delete the file watching logic here since it's handled by files/changed entirely.

RomneyDa:
@Jazzcort sounds good! I think we might want to leave the file watching logic for `config.yaml` just because we know of other clients that edit it so there are cases where it is changed outside the IDE, but the other ones seem more edge case-y. Will close this for now, and let me know if you find functionality still broken/feel free to reopen!

PR #7073, opened 2025-08-10
Thinking_for_deepseek_and_openrouter

ferenci84:
## Summary of Changes

This PR introduces proper handling of "thinking blocks" for OpenAI and Deepseek providers, which return reasoning content in separate fields within their streaming responses rather than embedding it in the main content.

## Key Changes

1. **Thinking Role Handling**:
   - Added support for "thinking" role messages that appear in streaming responses
   - Modified `openaiTypeConverters.ts` to:
     - Return `null` for "thinking" role messages when converting to OpenAI format (as they're not standard)
     - Filter out null messages when constructing chat bodies
     - Detect and convert `reasoning_content` or `reasoning` fields from API responses to "thinking" role messages

2. **Stream Processing Improvements**:
   - Added `forceStreamChat` flag to `BaseLLM` (default: false)
   - Set `forceStreamChat = true` for Deepseek and OpenRouter providers
   - Modified the `streamChat` method to always use `_streamChat` directly when `forceStreamChat` is enabled, rather than falling back to `_streamComplete`

## Why These Changes?

Previously, some models would unnecessarily route through `_streamComplete`, which only yields strings. This prevented proper extraction of thinking blocks as separate entities since all content was treated as a single string stream.

By ensuring these providers use the direct `_streamChat` pathway:
1. We can properly identify and separate thinking content from the main response
2. The UI can potentially display thinking processes differently (e.g., in a dedicated "thinking" section)
3. We maintain compatibility with providers that implement extended reasoning features like Anthropic's "extended thinking"

This change is particularly important for models that provide transparent reasoning processes, allowing Continue to properly render and utilize these intermediate thinking steps rather than treating them as part of the final response.

## Checklist


ferenci84:

jfouret:
Nice,
After checkout/build/install, It works well on my side also with OpenRouter and both GPT-5 and claude sonnet 4
image

jfouret:
```
diff --git a/core/llm/index.ts b/core/llm/index.ts
index edbfecbdb..b4d2c6cb8 100644
--- a/core/llm/index.ts
+++ b/core/llm/index.ts
@@ -9,7 +9,6 @@ import {
 import Handlebars from "handlebars";
 
 import { DevDataSqliteDb } from "../data/devdataSqlite.js";
-import { Logger } from "../util/Logger.js";
 import { DataLogger } from "../data/log.js";
 import {
   CacheBehavior,
@@ -31,6 +30,7 @@ import {
   TemplateType,
   Usage,
 } from "../index.js";
+import { Logger } from "../util/Logger.js";
 import mergeJson from "../util/merge.js";
 import { renderChatMessage } from "../util/messageContent.js";
 import { isOllamaInstalled } from "../util/ollamaHelper.js";
@@ -1037,9 +1037,11 @@ export abstract class BaseLLM implements ILLM {
               { ...body, stream: false },
               signal,
             );
-            const msg = fromChatResponse(response);
-            yield msg;
-            completion = this._formatChatMessage(msg);
+            const messages = fromChatResponse(response);
+            for (const message of messages) {
+              completion += this._formatChatMessage(message);
+              yield message;
+            }
           } else {
             // Stream true
             const stream = this.openaiAdapter.chatCompletionStream(
diff --git a/core/llm/llms/OpenAI.ts b/core/llm/llms/OpenAI.ts
index 27006aea1..6ab06db2f 100644
--- a/core/llm/llms/OpenAI.ts
+++ b/core/llm/llms/OpenAI.ts
@@ -16,6 +16,7 @@ import {
   fromChatCompletionChunk,
   LlmApiRequestType,
   toChatBody,
+  fromChatResponse,
 } from "../openaiTypeConverters.js";
 
 const NON_CHAT_MODELS = [
@@ -356,7 +357,10 @@ class OpenAI extends BaseLLM {
         return; // Aborted by user
       }
       const data = await response.json();
-      yield data.choices[0].message;
+      const messages = fromChatResponse(data);
+      for (const message of messages) {
+        yield message;
+      }
       return;
     }
 
@@ -454,3 +458,4 @@ class OpenAI extends BaseLLM {
 }
 
 export default OpenAI;
+
diff --git a/core/llm/openaiTypeConverters.ts b/core/llm/openaiTypeConverters.ts
index 75ea7ee91..c11675cbd 100644
--- a/core/llm/openaiTypeConverters.ts
+++ b/core/llm/openaiTypeConverters.ts
@@ -153,11 +153,22 @@ export function toFimBody(
   } as any;
 }
 
-export function fromChatResponse(response: ChatCompletion): ChatMessage {
-  const message = response.choices[0].message;
+export function fromChatResponse(response: ChatCompletion): ChatMessage[] {
+  const choice = response.choices[0];
+  const message = choice.message;
+  const messages: ChatMessage[] = [];
+
+  const reasoning =
+    (choice as any)?.reasoning_content || (choice as any)?.reasoning;
+  if (reasoning) {
+    messages.push({
+      role: "thinking",
+      content: reasoning,
+    });
+  }
   const toolCall = message.tool_calls?.[0];
   if (toolCall) {
-    return {
+    messages.push({
       role: "assistant",
       content: "",
       toolCalls: message.tool_calls
@@ -170,13 +181,15 @@ export function fromChatResponse(response: ChatCompletion): ChatMessage {
             arguments: (tc as any).function?.arguments,
           },
         })),
-    };
+    });
+  } else {
+    messages.push({
+      role: "assistant",
+      content: message.content ?? "",
+    });
   }
 
-  return {
-    role: "assistant",
-    content: message.content ?? "",
-  };
+  return messages;
 }
 
 export function fromChatCompletionChunk(
@@ -208,6 +221,11 @@ export function fromChatCompletionChunk(
         toolCalls,
       };
     }
+  } else if ((delta as any)?.reasoning_content || (delta as any)?.reasoning) {
+    return {
+      role: "thinking",
+      content: (delta as any)?.reasoning_content || (delta as any)?.reasoning,
+    };
   }
 
   return undefined;
```

What would you think of this ? fromChatResponse returning array in place of single object so that we can properly manage thinking and non thinking message in non-stream.

ferenci84:
@RomneyDa

> harmony format uses reasoning, not reasoning_content, I believe, e.g. gpt-oss

Some models use the naming `reasoning_content`. I cannot recall the exact models, but I the cases I tested contained both naming. See this, for example: https://docs.vllm.ai/en/v0.9.1/features/reasoning_outputs.html

> I think we should not use "force chat" but instead force set useLegacyCompletions to false

I quote an AI-generated summary of my investigation:

*** START OF AI-GENERATED SUMMARY ***

`useLegacyCompletionsEndpoint` cannot replace `forceStreamChat` because they operate at different levels, and `useLegacyCompletionsEndpoint` is already set to `false` in both Deepseek and OpenRouter classes.

**The Problem:** When `templateMessages` exists, `BaseLLM.streamChat()` bypasses `_streamChat()` entirely (pre-PR code):

```typescript
// BaseLLM.streamChat() - original code before PR
if (this.templateMessages) {
  // Uses _streamComplete → doesn't process thinking blocks
  for await (const chunk of this._streamComplete(prompt, signal, options)) {
    // ...
  }
} else {
  // Uses _streamChat → processes thinking blocks via openaiTypeConverters
}
```

**Where `useLegacyCompletionsEndpoint` is used:** Only in `OpenAI._streamChat()`, which never gets reached when `templateMessages` exists:

```typescript
// OpenAI._streamChat() - never reached when templateMessages exists
if (NON_CHAT_MODELS.includes(model) || this.useLegacyCompletionsEndpoint || options.raw) {
  // Legacy completions endpoint
} else {
  // Chat completions endpoint  
}
```

**Evidence:** Both Deepseek and OpenRouter already have `useLegacyCompletionsEndpoint: false` in their default options, yet the thinking blocks issue still occurred.

**Why `forceStreamChat` is needed:** It ensures `_streamChat()` is used instead of `_streamComplete()` so that thinking blocks get processed through `openaiTypeConverters.ts`. Setting `useLegacyCompletionsEndpoint = false` has no effect because that code path is bypassed when `templateMessages` exists.

*** END OF AI-GENERATED SUMMARY ***

Do you think a different naming would be better instead of forceStreamChat, or maybe a permanent condition that applies for all providers that avoids going through the completions endpoint in specific cases?

> remove any type casting

Updated to a different typecast:
```
  const delta = chunk.choices?.[0]?.delta as
    | (ChatCompletionChunk.Choice.Delta & {
        reasoning?: string;
        reasoning_content?: string;
      })
    | undefined;
```

ferenci84:
@RomneyDa  - Added two updates after noticing thinking blocks were missing in Agent and Plan modes:

1. **Fixed missing thinking block support** - Discovered that models like R1 (via OpenRouter) use the OpenAI adapter in Agent/Plan modes, but thinking blocks weren't being captured. Extracted chunk processing into a shared function to avoid ESLint max-depth violations while ensuring both streaming and non-streaming responses properly handle thinking/reasoning content.

2. **Performance optimization** - Replaced string concatenation with array operations to avoid O(n²) complexity on long streaming responses.

ferenci84:
@jfouret I made a similar solution to the one you suggested. Now all these scenarios work (tested with R1 via OpenRouter):
- Chat mode with streaming
- Chat mode non-streaming
- Plan/Agent mode (using OpenAI adapter) with streaming
- Plan/Agent mode (using OpenAI adapter) non-streaming

jfouret:
Thank you @ferenci84 very nice work.

ferenci84:
@RomneyDa 
Due to this: [preserveReasoning](https://openrouter.ai/docs/use-cases/reasoning-tokens#preserving-reasoning-blocks), I added preserveReasoning option.

The previous behaviour was to strip all the reasoning, but it seems to be a bad idea for many cases when tools are used.

In most cases models ignore the thinking blocks, but some may use them, for example GPT-5 returns an encrypted thinking block at the end, that I suspect will be used in some cases. In some special cases (e.g. switching models in the same conversation), they may cause problems. Now I made the default behavior to preserve the thinking blocks in the openai compatible endpoints, but this behavior can be disabled by setting preserveReasoning to false. Please let me know if it's OK, or if you have any other suggestions.

ferenci84:
With the latest changes, one can safely switch between most models using the openrouter provider (the same switching didn't work with the anthropic provider if prevous thinking blocks was generated by open source models).

What I also noticed that the GPT-5 do not show reasoning when called via the openai provider.

core/llm/openaiTypeConverters.ts
```diff
@@ -208,11 +301,68 @@ export function fromChatCompletionChunk(
...
+
+    if (existingIndex === -1) {
+      // No existing item with this type, add new item
+      result.push({ ...deltaItem });
+    } else {
+      // Merge with existing item of the same type
+      const existingItem = result[existingIndex];
+
+      for (const [key, value] of Object.entries(deltaItem)) {
+        if (value === null || value === undefined) continue;
```
RomneyDa, line 385: nitpick, seems like prettier formatting not applied here

core/llm/openaiTypeConverters.ts
```diff
@@ -48,7 +68,39 @@ export function toChatMessage(
...
+              {
+                signature: prevMessage.signature,
+              },
+            ]
+          : undefined);
+
+      msg.reasoning_details = reasoningDetails || [];
+
+      if (
+        options.model.includes("claude") &&
```
RomneyDa: a bit too model-specific here (I know we're not exactly perfect with this principal but trying to reduce)

could also remove `delete` usage if possible

RomneyDa (changes requested):
@ferenci84 appreciate the in-depth considerations.

I really like the change of using array.join instead of string concatenation, probably massively reduces compute for long streams.

@`forceStreamChat`
The AI-generated summary pointed out the underlying issue, which is that templateMessages is causing it to bypass `useLegacyCompletionsEndpoint`. It is pre-PR code but I do think the settings are directly contradictory and we should fix the existing issue rather than introducing a new one.

@`preserveReasoning` I think we could just always preserve reasoning for now and leave this setting out but add a `TODO` comment and/or issue. I think the most likely reason thinking was stripped before was just lack of support for it (i.e. typescript said handle this, someone just stripped it for now).

ferenci84:
Regarding forceStreamChat, do you have a specific recommendation?
The problem here is that streamChat uses the complete endpoint sometimes. Can we remove this bevaviour by default from the default streamChat implementation? In that case if it's important for a provider to go through the complete endpoint, the provider class need to override the streamChat method. Do you think it's a good solution?

--
Ferenci Zoltán László

fbricon:
FYI for Ollama, we can parse both thinking and content messages while streaming or not, without forcing anything, see https://github.com/continuedev/continue/commit/ca5aadda7388e30431dba3b52ee0eb6539513c61

RomneyDa:
@ferenci84 thanks for the updates, will try to get this in soon

RomneyDa (approved):

ibrahim-ava:
@ferenci84 thank you for doing this, hope it can be merged soon 🙏

RomneyDa:
Note for future reference the root issue of the forceStreamChat/templateMessages issue is that deepseek and openrouter are not included in the `PROVIDER_HANDLES_TEMPLATING` array

RomneyDa:
This PR has been merged into main via https://github.com/continuedev/continue/pull/7891

PR #7096, opened 2025-08-12
chore: move system message tools out of experimental

uinstinct:
## Description

Move system message tools or virtual tool calls from experimental to the core feature section.

resolves CON-3262

## Checklist

## Tests

### DIFF
--- a/core/config/migrateSharedConfig.ts
+++ b/core/config/migrateSharedConfig.ts
@@ -130,6 +130,15 @@ export function migrateJsonSharedConfig(filepath: string, ide: IDE): void {
         effected = true;
       }
 
+      const { onlyUseSystemMessageTools, ...withoutOnlyUseSystemMessageTools } =
+        migratedExperimental as any;
+      if (onlyUseSystemMessageTools !== undefined) {
+        shareConfigUpdates.onlyUseSystemMessageTools =
+          onlyUseSystemMessageTools;
+        migratedExperimental = withoutOnlyUseSystemMessageTools;
+        effected = true;
+      }
+
       if (Object.keys(migratedExperimental).length > 0) {
         config = {
           ...withoutExperimental,
--- a/core/config/sharedConfig.ts
+++ b/core/config/sharedConfig.ts
@@ -13,6 +13,7 @@ export const sharedConfigSchema = z
     allowAnonymousTelemetry: z.boolean(),
     disableIndexing: z.boolean(),
     disableSessionTitles: z.boolean(),
+    onlyUseSystemMessageTools: z.boolean(),
 
     // `experimental` in `ContinueConfig`
     useChromiumForDocsCrawling: z.boolean(),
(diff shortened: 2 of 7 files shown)


core/config/migrateSharedConfig.ts
```diff
@@ -130,6 +130,15 @@ export function migrateJsonSharedConfig(filepath: string, ide: IDE): void {
         effected = true;
       }
 
+      const { onlyUseSystemMessageTools, ...withoutOnlyUseSystemMessageTools } =
```
RomneyDa, line 133: @uinstinct this is unnecessary as `onlyUseSystemMessageTools` is not used in JSON config and is not a security concern

core/config/sharedConfig.ts
```diff
@@ -153,6 +153,10 @@ export function modifyAnyConfigWithSharedConfig<
   if (sharedConfig.disableSessionTitles !== undefined) {
     configCopy.disableSessionTitles = sharedConfig.disableSessionTitles;
   }
+  if (sharedConfig.onlyUseSystemMessageTools !== undefined) {
+    (configCopy as any).onlyUseSystemMessageTools =
```
RomneyDa, line 157: remove any type cast

core/index.d.ts
```diff
@@ -1607,6 +1606,7 @@ export interface JSONModelDescription {
 export interface SerializedContinueConfig {
   env?: string[];
   allowAnonymousTelemetry?: boolean;
+  onlyUseSystemMessageTools?: boolean;
```
RomneyDa, line 1609: Don't add to JSON config, is a user setting

core/index.d.ts
```diff
@@ -1638,6 +1638,8 @@ export type ContinueRcJson = Partial<SerializedContinueConfig> & {
 export interface Config {
   /** If set to true, Continue will collect anonymous usage data to improve the product. If set to false, we will collect nothing. Read here to learn more: https://docs.continue.dev/telemetry */
   allowAnonymousTelemetry?: boolean;
+  /** If set to true, use system message tools instead of native tool calling */
+  onlyUseSystemMessageTools?: boolean;
```
RomneyDa, line 1642: Same here, should not be on the `Config` intermediate type passed to config.ts

RomneyDa (changes requested):
@uinstinct I understand why you added to `SerializedContinueConfig` and `Config` because they were on `ExperimentalConfig` already, but I think the goal of this is to make it on by default for models that don't support Agent, not just to move the setting around under the hood. I think moving it off `experimental` under the hood is probably unnecessary

chezsmithy:
@RomneyDa so the thinking is if no native tools then try system message tools. Is there thinking that a list of models that work with system message tools might be needed too?

RomneyDa:
@chezsmithy more or less all chat models work with system message tools! Even e.g. llama 2b or granite 2b can use the tools decenly. The exception being the search and replace tool, which we already only include for specific models. Any models you have in mind?

uinstinct:
closing in favor of https://github.com/continuedev/continue/pull/7236

bdougie:
There is a reference in the docs that will need to be changed 

https://docs.continue.dev/features/agent/model-setup#how-system-message-tools-work

PR #7099, opened 2025-08-12
Continuous ai review bot

tomasz-stefaniak:
## Description

The bot is triggered when the PR opens or when a draft PR is converted to a regular PR.

## Checklist

## Tests

---

## Summary by cubic
Added a GitHub Actions workflow to automatically run the Continue CLI for code review on pull requests and pushes to main.

- **New Features**
 - Installs and runs the Continue CLI with a hardcoded prompt to review code quality.
 - Uses the CONTINUE_API_KEY secret for authentication.

### DIFF
--- a/.github/workflows/continue-review.yaml
+++ b/.github/workflows/continue-review.yaml
@@ -54,7 +54,7 @@ jobs:
             # For issue_comment event on a PR
             PR_NUMBER="${{ github.event.issue.number }}"
           fi
-          
+
           # Get PR diff
           gh pr diff $PR_NUMBER > pr_diff.txt
 
@@ -86,13 +86,27 @@ jobs:
           ## Code Review Summary
 
           ### ✅ Strengths
-          - [List positive aspects]
+          - **[Aspect]**: [Description of what was done well]
 
           ### ⚠️ Issues Found
-          - [List any problems with severity levels]
+
+          Only include the severity subheaders below if you actually found issues at that level:
+
+          #### Critical
+          - **[Issue Title]**: [Description of critical issues that must be fixed before merging]
+
+          #### High
+          - **[Issue Title]**: [Description of high-priority issues that should be fixed]
+
+          #### Medium
+          - **[Issue Title]**: [Description of issues that should be addressed]
+
+          #### Low
+          - **[Issue Title]**: [Description of minor issues or nice-to-have improvements]
+
 
           ### 💡 Suggestions
-          - [List improvement recommendations]
+          - **[Suggestion Title]**: [Description of improvement recommendations]
 
           ### 🚀 Overall Assessment
           [Provide overall recommendation: APPROVE, REQUEST_CHANGES, or COMMENT]


tomasz-stefaniak:
@continue-review please review

.github/workflows/continue-review.yaml
```diff
@@ -0,0 +1,164 @@
+name: Continue CLI Code Review
+
+on:
+  pull_request:
+    types: [opened, ready_for_review]
```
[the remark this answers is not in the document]
bdougie: In my experience. Only "ready to review" is great to limit credit spend, but I don't think contributors will start with a draft. So makes sense to have this as opened, but probably makes sense"ready to review" unnecessary

.github/workflows/continue-review.yaml
```diff
@@ -0,0 +1,186 @@
...
+        with:
+          fetch-depth: 0 # Fetch full history for better context
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install Continue CLI
+        run: npm install -g @continuedev/cli@1.4.25
```
Patrick-Erichsen, line 46: Do we want to pin to this CLI version?

.github/workflows/continue-review.yaml
```diff
@@ -0,0 +1,186 @@
...
+          1. **Code Quality**: Are there any bugs, performance issues, or code smells?
+          2. **Best Practices**: Does the code follow established patterns and conventions?
+          3. **Security**: Are there any potential security vulnerabilities?
+          4. **Testing**: Are appropriate tests included or updated?
+          5. **Documentation**: Is documentation adequate for the changes?
+          6. **Architecture**: Do the changes fit well with the existing codebase structure?
+
+          You can use the available tools to explore the codebase and understand context better.
+
+          Format your response as a markdown code review with the following structure:
```
Patrick-Erichsen, line 84: My hunch is that this will get annoying with it following the exact structure even when unnecessary. Might be helpful to add a comment like "You do not need to follow this exact structure, use it for inspiration while still keeping your review concise and focused."

.github/workflows/continue-review.yaml
```diff
@@ -0,0 +1,186 @@
...
+          cat >> review_prompt.txt << 'EOF'
+
+          Please analyze these changes and provide a comprehensive code review. Consider:
+
+          1. **Code Quality**: Are there any bugs, performance issues, or code smells?
+          2. **Best Practices**: Does the code follow established patterns and conventions?
+          3. **Security**: Are there any potential security vulnerabilities?
+          4. **Testing**: Are appropriate tests included or updated?
+          5. **Documentation**: Is documentation adequate for the changes?
+          6. **Architecture**: Do the changes fit well with the existing codebase structure?
```
Patrick-Erichsen, line 80: Since we have the repo cloned, will the agent be following our rules in the `.continue` folder? If not, could we just `cat` them into the context window or something?

tomasz-stefaniak: I thought about this but we have a lot of different rules with different triggers, we'd have to either put all the rules in the context (noisy), or handle rule resolution for each file individually (complicated). I'm not against it but it feels out of the scope of this PR.

.github/workflows/continue-review.yaml
tomasz-stefaniak: I made it more rigid on purpose - the odds are most people won't read this thoroughly but will skim it. The more consistent the structure, the easier it will be to find the exact data you need.

.github/workflows/continue-review.yaml
tomasz-stefaniak: Yeah, that's by design. I think we should pretty much always do this when using a dependency with ad-hoc npm install. Otherwise, we may end up using a version with a breaking change, or even a version with a security issue.

tomasz-stefaniak:
@continue-review hello

PR #7112, opened 2025-08-13
feat: tool call status icon for search and replace UI

RomneyDa:
## Description
- Shows red X if search and replace fails, and green checkmark if it is completed
- Clicking opens virtual file with output, just like other tool call UIs
Also tweaks spacing a bit


Patrick-Erichsen (approved):
Nice improvements here!

PR #7116, opened 2025-08-13
fix: standardize dev data schemas and add missing fields

Patrick-Erichsen:
## Summary
• Standardized `modelTitle` → `modelName` across all dev data schemas for consistency
• Added missing `modelProvider` field to `chatFeedback` event schema  
• Added missing `modelProvider` and `modelName` fields to `nextEditWithHistory` event schema

Fixes CON-3401, CON-3400, CON-3402, CON-3424, CON-3080

---

## Summary by cubic
Standardized all dev data schemas to use modelName instead of modelTitle, and added missing modelProvider fields for better consistency and completeness. This addresses requirements from CON-3401, CON-3400, and CON-3402.

### DIFF
--- a/core/data/log.vitest.ts
+++ b/core/data/log.vitest.ts
@@ -35,6 +35,7 @@ const TEST_AGENT_INTERACTION_EVENT: DevDataLogEvent = {
     completion: "Hello, world!",
     modelProvider: "openai",
     modelName: "gpt-4",
+    modelTitle: "gpt-4",
     sessionId: "1234",
     tools: ["test-tool1"],
   },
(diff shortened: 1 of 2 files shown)


RomneyDa:
Do you think we could keep modelTitle and modelName rather than replacing? I want to be cautious about removing fields that have been there for a while

Patrick-Erichsen:
@RomneyDa good call, updated to have both `modelName` and `modelTitle`

RomneyDa (approved):
Updates so all the events have at least the same fields
Noting duplicate modelName/modelTitle field

PR #7161, opened 2025-08-14
feat: :sparkles: Add experimental range tool for file operations

chezsmithy:
## Description

Introduce **experimental** range tool for file operations. This provides streamlined ways to interact with files where there is a need to pull just targetted content into the LLM context without reading the entire file at onece. This will be particularlly useful when analyzing or comparing multiple files. The new tools can run in parallel vs multiple terminal tool operations speeding up processing in multiple use cases.

Note, I looked at add a tail tool or allowing negative range processing (bottom up), but this didn't appear feasible as there wasn't a good way to do this without understanding the file line count. Researching ways to get line counts without reading the entire file into memory which could be extremely costly due to size seemed non trival with current architecture.

## Checklist

## Tests

Extensive unit and integration tests have been added for the new tool.

    

---

## Summary by cubic
Added new range tool to allow reading specific lines or the first N lines from files, making it easier and faster to analyze targeted file content without loading entire files.

- **New Features**
 - Added a tool for reading a custom range of lines from a file using positive line numbers.
 - Tool supports parallel file operations for quicker processing.
 - Includes validation and error handling for non-existent files and input parameters.
 - Added as experimental tool.


Patrick-Erichsen:
Neat idea! I think we should add them as experimental tools to begin with so we can play around with them a bit before shipping to all users.

How does the model know how many lines to read at the beginning of a file, or the range? 

In general, curious what your inspiration here was - any particular behavioral patterns you've noticed that led you to think that these tools would be helpful?

chezsmithy:
> Neat idea! I think we should add them as experimental tools to begin with so we can play around with them a bit before shipping to all users.
> 
> How does the model know how many lines to read at the beginning of a file, or the range?
> 
> In general, curious what your inspiration here was - any particular behavioral patterns you've noticed that led you to think that these tools would be helpful?

@Patrick-Erichsen inspired by a use case to process large files including logs. These files might have headers in the first row, and row sampling can help the LLM discover format quickly and then create strategies to process log data. These strategies might involve hte development of python, or terminal usage to process the logs in bulk. With the addition of parallel tool calls processing multiple log files can be done quickly. For example, you can ask for the LLM to take mutliple log files, sample the first 50 lines and look for anomolies in server startup times, or errors.

I'd be happy to introduce them as experimental. Could you perhaps point me to a resource I can look at to add the specific flags? I'd also be happy to add this to the contributing guide as a strategy to add new tools if that helps! Thanks for the awesome question.

chezsmithy:
Referencing this use case in discussions: https://github.com/continuedev/continue/discussions/7113

I might want to tune the error message from a large file read from the readFile tool to redirect it to use the range tool to help. Curious what the best way to solve for that might be.

RomneyDa:
@chezsmithy thoughts on adding these as optional boolean args on existing read file tool? Might be overkill what do you think?

chezsmithy:
> @chezsmithy thoughts on adding these as optional boolean args on existing read file tool? Might be overkill what do you think?

I experimented with that but noted the model started asking for the range to review at times when it could just read the whole file. Optional parameters seem to "sometimes" confuse the model into trying to optimize by putting the human in the middle.

Patrick-Erichsen:
@chezsmithy neat! I've been wondering what the right strategy for parsing large log files is. I run into context window limitations pretty frequently when working with CI logs for example. 

I wonder if there's a way we could/should make that use case more first-class. Maybe we put these as args on readFile like Dallin mentioned, but explicitly told the model to use them only when reading large files such as logs?

RomneyDa:
@chezsmithy realized I never pointed to experimental tools
You can add it to `getConfigDependentToolDefinitions` 

What are the main reasons to split these tools up? Do you think the readFileRange tool could achieve both functions with a little extra prompting to start from line 1 to read the head?

chezsmithy:
> @chezsmithy thoughts on adding these as optional boolean args on existing read file tool? Might be overkill what do you think?

I like the cleanness of the readtool without overloading with more parameters. I think introducing as experimental and doing some tests in the real world will be interesting.

chezsmithy:
> @chezsmithy realized I never pointed to experimental tools You can add it to `getConfigDependentToolDefinitions`
> 
> What are the main reasons to split these tools up? Do you think the readFileRange tool could achieve both functions with a little extra prompting to start from line 1 to read the head?

@RomneyDa I think having a single tool makes sense. More experiments seems to have claude using this tool and asserting line 1 as the start when mentioning "pull top 5 lines of file X", "get the header row of file X".

chezsmithy:
> @chezsmithy neat! I've been wondering what the right strategy for parsing large log files is. I run into context window limitations pretty frequently when working with CI logs for example.
> 
> I wonder if there's a way we could/should make that use case more first-class. Maybe we put these as args on readFile like Dallin mentioned, but explicitly told the model to use them only when reading large files such as logs?

@Patrick-Erichsen I think with prompts you could for example note you are working with large files and avoid using the readFile Tool and instead use the range tool to get data. Once it's enabled experimentally I'd be interested in feedback on how we might tune things. I found other interesting use cases like data sampling where the range tool is often used. Say, like sample 10 rows from "test.csv" and show me just the average temperature.

chezsmithy:
@Patrick-Erichsen ready for review.

PR #7163, opened 2025-08-15
fix: :lock: Gate search_web tool to hub subscribers

chezsmithy:
## Description

When using continue in a local only mode without a hub account or free trial the search_web tool will reach out to a continue owned proxy and attempt a web search. The use of this proxy is denied as it requires a paid or trial hub subscription to use this feature. However, this raised a potential security concern that content of the web search might be captured by the proxy.

To eliminate any possiblity of this happening and to allow for the inclusion of hub subscription gated tools, I'm proposing this fix which creates a Hub subscription required tool group which is only enabled when a hub subscription is active.

I will require some testing and validation support from the continue team as I don't have a subscription to help validate the use case when the tools should be enabled. I'd also love some help deciding if e2e tests should be added and how that might be accoplished for this use case.

## AI Code Review

## Checklist

## Tests

Unit tests have been added.
Visual inspection / manual testing has been perfomed for the use case when no trial or subscription is active. I would need help with the counter case.
Note the question about potential e2e tests and help with that.


chezsmithy:
@sestinj including you on this PR for visibility as it is security related. Note I also alerted your security email of this potential issue.

RomneyDa:
> The use of this proxy is denied as it requires a paid or trial hub subscription to use

@chezsmithy note that the `search_web` tool does not require any subscription, it is free to use. If you see an error something like
`Error in Continue free trial server: Request failed with status 402. You have exceeded your credits limit. Please top up to keep using Exa at [dashboard.exa.ai](http://dashboard.exa.ai/)`

That just means we need to bump our limit a bit on our end

However, the security concern is valid
@sestinj @TyDunn thoughts?

chezsmithy:
> @chezsmithy given security concern I think this is a valid change
> 
> we have access to the user's signed in state in both GUI and config handler, we could make this exclusion a few lines in `getConfigDependentToolDefinitions`

Agree, considered that option as well. It's a much easier change. Wanted to give transparency to why a tool might not be available. But if we want a quick security fix, agree the route suggested is faster and less risky.

chezsmithy:
> > The use of this proxy is denied as it requires a paid or trial hub subscription to use
> 
> @chezsmithy note that the `search_web` tool does not require any subscription, it is free to use. If you see an error something like `Error in Continue free trial server: Request failed with status 402. You have exceeded your credits limit. Please top up to keep using Exa at [dashboard.exa.ai](http://dashboard.exa.ai/)`
> 
> That just means we need to bump our limit a bit on our end
> 
> However, the security concern is valid @sestinj @TyDunn thoughts?

Yes, our users are seeing the credit error from the tool. However, we don't want them using this tool if information can escape from our highly regulated platforms.

chezsmithy:
@RomneyDa @TyDunn : https://github.com/continuedev/continue/pull/7165

Inspired by @RomneyDa 's comment.

chezsmithy:
@TyDunn should we keep this PR and adapt it in a way that we could have a set of "opt-in" hub tools that might have less secure profiles. I could imagine there could be BYOK organizations that might want control over this tool as well. I'm also happy to close it, and open an issue to track the need for a future PR as well.

TyDunn:
@chezsmithy Sorry for the delay. This one slipped through the cracks. Since we shipped the simpler fix, I say we close this and open an issue for it

PR #7164, opened 2025-08-15
fix: rule colocation uri logic

Patrick-Erichsen:
This fixes a category of errors with rule colocation due to file URI things.

Previously, for colocated rules without a glob pattern, just used the full file URI, but this fails on the minimatch comparisons, eg. `minimatch("home/**/*", "file:///path/to/home/**/*)` returns `false`.

To fix this, we grab the URI path relative to the workspace dir.

Additional work here:
- Convert remaining `.yaml` rules to `.md`
- Delete unnecessary rules
- Make some rules manually requested
- Rename `filePath` to `fileUri `
  - Note that this was in `config-yaml` package, but I verified that in hub repo we weren't using the `filePath` property anywhere

### Testing

- Include `nested-folder/helloNested.py` in a message
- Confirm the `rules.md` rule is included
    

---

## Summary by cubic
Fixed rule colocation logic by using URI paths relative to the workspace, so file-based rules match patterns correctly. Also converted all rule files from YAML to Markdown, deleted unused rules, updated some rules to be manually requested, and renamed filePath to uriPath throughout the config logic.


Patrick-Erichsen:
Cubic really nailed the PR summary here

Patrick-Erichsen:
@RomneyDa left a todo comment for the `path.dirname` update for now.

```
  // TODO: This is likely causing bugs for Windows users.
  // We can't access `core/util/uri.ts` from this package though.
```

.continue/rules/intellij-test-standards.md
```diff
@@ -0,0 +1,20 @@
+---
+name: IntelliJ Test Standards
+description: Standards for writing tests in the IntelliJ plugin
+alwaysApply: false
+globs: "**/test/**/*.kt"
+---
+
```
RomneyDa, line 7: 1. Totally non blocking but I think rules in this format confuse the model quite a bit when they are appended together with other rules

E.g. Single header with bulleted list is better, or have a more clear differentiator like
<jetbrains_testing_rules>

2. saying `jetbrains` is probably better than `intellij`, rectangles and squares

RomneyDa (approved):
See comment, lgtm

PR #7165, opened 2025-08-15
fix: :lock: Simpler fix to exclude search_web tool to signed in users

chezsmithy:
## Description

Excludes the search_web tool so it's only avaiable to signed in Hub Users. Excludes it availability to users who might be using local only features. This is due to a potential security issue with the search web tool and how it posts content to an external proxy. For highly regulated users of the extension we would not want any kind of information posted to a third party proxy.

## AI Code Review

## Checklist

## Tests

Tests added to validate the tool is excluded properly based on the configuration and signed in state.

### DIFF
--- a/core/config/profile/doLoadConfig.ts
+++ b/core/config/profile/doLoadConfig.ts
@@ -93,6 +93,7 @@ export default async function doLoadConfig(options: {
   const uniqueId = await ide.getUniqueId();
   const ideSettings = await ideSettingsPromise;
   const workOsAccessToken = await controlPlaneClient.getAccessToken();
+  const isSignedIn = await controlPlaneClient.isSignedIn();
 
   // Migrations for old config files
   // Removes
@@ -285,6 +286,7 @@ export default async function doLoadConfig(options: {
       rules: newConfig.rules,
       enableExperimentalTools:
         newConfig.experimental?.enableExperimentalTools ?? false,
+      isSignedIn,
     }),
   );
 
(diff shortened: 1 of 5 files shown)


chezsmithy:
A simplier fix proposed by @RomneyDa then what was more broadly proposed here: https://github.com/continuedev/continue/pull/7163

Patrick-Erichsen:
Pulling in @TyDunn for an opinion here.

TyDunn:
@chezsmithy @Patrick-Erichsen Thanks for the PR and pulling me in! This seems reasonable. I like this simple approach better. 

From my understanding of the implementation, this is just checking if the user is signed into the Hub (which is the right behavior--it should not matter if you are on free trial, paid, BYOK, etc). 

However, there is a comment that says `// Web search is only available for signed-in users (free trial and paid subscribers)`

chezsmithy:
> @chezsmithy @Patrick-Erichsen Thanks for the PR and pulling me in! This seems reasonable. I like this simple approach better.
> 
> From my understanding of the implementation, this is just checking if the user is signed into the Hub (which is the right behavior--it should not matter if you are on free trial, paid, BYOK, etc).
> 
> However, there is a comment that says `// Web search is only available for signed-in users (free trial and paid subscribers)`

@TyDunn I've clarified the comment to note // Web search is only available for signed-in users

Patrick-Erichsen:
Might want to update the docs around this eventually but I think it's okay for now.

PR #7213, opened 2025-08-18
test: describe temporary workaround for failing runIdeForUiTests

exigow:
Removing the configuration from the home directory doesn’t help either. I haven’t found another way, and there must be some arbitrary code in the binary that somehow identifies me and fetches this configs (even after `~/.continue` removal).

Those e2e tests are rarely used (it's single smoke test mostly for CI), so the workaround is sufficient, but ideally we should understand why this happens (we need to debug the TS `core` and find out what takes precedence over `CONTINUE_GLOBAL_DIR`).

Solves CON-3298
    

---

## Summary by cubic
Documented a temporary workaround in runIdeForUiTests for flaky e2e runs when CONTINUE_GLOBAL_DIR is ignored if a local Continue config is present. The comment advises running tests offline as a short-term fix while we investigate the root cause (CON-3298).

### DIFF
--- a/.github/workflows/continue-detailed-review.yaml
+++ b/.github/workflows/continue-detailed-review.yaml
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Run Detailed PR Review
-        uses: continuedev/continue/actions/detailed-review@b3f4b82600f603ca28c65a319b80ec36c4644ed7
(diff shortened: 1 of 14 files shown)


exigow:
@Patrick-Erichsen, this is an old/unimportant PR (it's just a comment in the build scripts), but I think we should put this detail in case someone wants to run the e2e tests locally.

RomneyDa (approved):

PR #7224, opened 2025-08-18
feat: update default-included context providers

RomneyDa:
## Description
Includes more context providers by default, as part of soft-deprecating the `context` block
- Unifies the parts of JSON and YAML config loading that are the same
- Increases the number of providers included by default

CON-3397
    

---

## Summary by cubic
Loads more context providers by default and centralizes provider loading for JSON/YAML via a new utility. This soft-deprecates the context block since common providers are auto-included.

- **New Features**
  - Auto-includes default providers: terminal, diff, file, current-file, problems, rules.
  - Auto-adds docs provider when docs are configured (if not already present).
  - Sets workOsAccessToken on continue-proxy when present.

- **Refactors**
  - Moved provider loading to loadConfigContextProviders and reused across YAML/JSON loaders.
  - Removed ad hoc fallbacks and tests tied to specific providers; docs indexing no longer depends on a docs provider being present.
  - Unknown provider names are now collected as non-fatal config errors.


core/config/loadContextProviders.ts
```diff
@@ -0,0 +1,75 @@
...
+        new cls({
+          name: config.name,
+          ...config.params,
+        }),
+      );
+    }
+  }
+
+  // Add from defaults if not found in config
+  const DEFAULT_PROVIDERS: IContextProvider[] = [
```
Patrick-Erichsen: nitpick but global consts are easier to spot/update when in the global scope rather than inline in the fn

core/config/loadContextProviders.ts
```diff
@@ -0,0 +1,75 @@
...
+import TerminalContextProvider from "../context/providers/TerminalContextProvider";
+
+/*
+    Loads context providers based on configuration
+    - default providers will always be loaded, using config params if present
+    - other providers will be loaded if configured
+
+    Note:
+    - MCPContextProvider is added in doLoadConfig if any resources are present
+    - DocsContextProvider is added in doLoadConfig if docs configs are present and it hasn't been added yet
```
Patrick-Erichsen: To avoid this, could we just wait to call `loadConfigContextProviders` until we have the data to pass that will allows us to conditionally include `MCPContextProvider` and `DocsContextProvider`? Or maybe we just always include them by default regardless of config? 

Would be ideal if we didn't have the logic spread out.

RomneyDa: We do have the information for docs, I'll add a boolean!

I think it makes sense to leave MCP resources since MCP is loaded after JSON/YAML loading but it's awkward to streamline the other ones between JSON and YAML since the configs look different, so tradeoff

core/config/loadContextProviders.ts
RomneyDa: since it's instantiating classes, will make it non-global looking

Patrick-Erichsen (approved):

PR #7234, opened 2025-08-19
feat: org policies for indexing and mcp

RomneyDa:
## Description
Adds support for org policies `allowMcpServers` and `allowCodebaseIndexing`
If `allowMcpServers` is set to false, MCP servers are shutdown on config loading and prevented from loading, and user settings toggle is disabled
If `allowCodebaseIndexing` is set to false, `config.disableIndexing` is set to true and user settings toggle is disabled

Requires support on continue proxy to work

image


core/config/ConfigHandler.ts
```diff
@@ -202,6 +202,8 @@ export class ConfigHandler {
           message: `Error loading Continue Hub assistants${e instanceof Error ? ":\n" + e.message : ""}`,
         });
       }
+    } else {
+      PolicySingleton.getInstance().policy = null;
```
RomneyDa, line 206: wasn't clearing org policy if not signed in

core/context/mcp/MCPManagerSingleton.ts
```diff
@@ -48,6 +48,21 @@ export class MCPManagerSingleton {
...
 
+  async shutdown() {
+    if (this.connections.size > 0) {
+      await Promise.allSettled(
+        Array.from(this.connections.entries()).map(([id, connection]) => {
+          try {
+            connection.abortController.abort();
+            void connection.client.close();
+          } finally {
+            this.connections.delete(id);
```
tingwai, line 59: this seems like a race condition

core/context/mcp/MCPManagerSingleton.ts
```diff
@@ -48,6 +48,21 @@ export class MCPManagerSingleton {
     this.connections.delete(id);
   }
 
+  async shutdown() {
+    if (this.connections.size > 0) {
+      await Promise.allSettled(
+        Array.from(this.connections.entries()).map(([id, connection]) => {
+          try {
+            connection.abortController.abort();
+            void connection.client.close();
```
tingwai, line 57: this needs a `catch` block to actually catch any errors

tingwai (changes requested):
Looks good otherwise!

core/context/mcp/MCPManagerSingleton.ts
RomneyDa: should be okay to close after deleting from map

core/context/mcp/MCPManagerSingleton.ts
RomneyDa: Promise.allSettled will catch these

core/context/mcp/MCPManagerSingleton.ts
tingwai: oh I see, I think I misread earlier

core/context/mcp/MCPManagerSingleton.ts
tingwai: ah you're right

tingwai (approved):

PR #7236, opened 2025-08-19
Virtual tools move 2

uinstinct:
## Description

Make system message tools on by default and use whenever native tools are absent.

closes https://github.com/continuedev/continue/pull/7096

resolves CON-3262

## AI Code Review

## Checklist

## Tests

### DIFF
--- a/gui/src/redux/selectors/selectUseSystemMessageTools.ts
+++ b/gui/src/redux/selectors/selectUseSystemMessageTools.ts
@@ -1,4 +1,5 @@
 import { shouldAutoEnableSystemMessageTools } from "core/config/shouldAutoEnableSystemMessageTools";
+import { modelSupportsNativeTools } from "core/llm/toolSupport";
 import { selectSelectedChatModel } from "../slices/configSlice";
 import { RootState } from "../store";
 
@@ -30,6 +31,11 @@ export function selectUseSystemMessageTools(state: RootState): boolean {
   // If auto-detection has a preference, use it (takes priority)
   if (autoSetting !== undefined) {
     return autoSetting;
+  }
+
+  // For native supported models, use native tools
+  if (modelSupportsNativeTools(selectedModel)) {
+    return false;
   }
 
   // If no auto-preference, use manual setting or default to false
--- a/gui/src/pages/config/UserSettingsForm.tsx
+++ b/gui/src/pages/config/UserSettingsForm.tsx
@@ -93,7 +93,7 @@ export function UserSettingsForm() {
   const enableExperimentalTools =
     config.experimental?.enableExperimentalTools ?? false;
   const onlyUseSystemMessageTools =
-    config.experimental?.onlyUseSystemMessageTools ?? false;
+    config.experimental?.onlyUseSystemMessageTools ?? true;
   const codebaseToolCallingOnly =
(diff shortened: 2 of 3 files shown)


RomneyDa (changes requested):
@uinstinct 
If someone sets onlyUseSystemMessageTools = true, it should not use native tools, whether the model supports it or not. The setting should be default to false, but system message tools should _always_ serve as a fallback for native tools.

i.e.
onlyUseSMTools = true + supports native tools  => use SM tools
onlyUseSMTools = false + supports native tools  => use Native tools
onlyUseSMTools = true + does not support native tools  => use SM tools
onlyUseSMTools = false + does not support native tools  => use SM tools

the current exception is openrouter, but we should probably rearrange to just say openrouter for the given models doesn't support native tools, the logic is redundant

gui/src/redux/selectors/selectUseSystemMessageTools.ts
```diff
@@ -32,6 +34,11 @@ export function selectUseSystemMessageTools(state: RootState): boolean {
     return autoSetting;
   }
 
+  // when model does not support native tools, use system message tools
```
RomneyDa: @uinstinct if manualSetting is false, should return false whether native tools is supported or not

uinstinct: > onlyUseSMTools = false + does not support native tools => use SM tools

If we do that, we would be eventually using the older implementation

RomneyDa (changes requested):
@uinstinct this looks great! I think just need to add the supportsTools false to openrouter in tool support (besides claude/kimi)

gui/src/redux/thunks/streamNormalInput.ts
```diff
@@ -151,8 +150,12 @@ export const streamNormalInput = createAsyncThunk<
     const supportsNativeTools = modelSupportsNativeTools(selectedChatModel);
 
     // Use the centralized selector to determine if system message tools should be used
-    const useSystemTools = selectUseSystemMessageTools(state);
+    let useSystemTools =
```
RomneyDa:
```suggestion
    const useNativeTools = state.config.config.experimental?.onlyUseSystemMessageTools ? false : modelSupportsNativeTools(selectedChatModel)
```

RomneyDa: can clean up logic significantly, remove `useSystemtools` etc.

core/llm/toolSupport.ts
```diff
@@ -248,9 +248,14 @@ export const PROVIDER_TOOL_SUPPORT: Record<string, (model: string) => boolean> =
...
-        ["vision", "math", "guard", "mistrallite", "mistral-openorca"].some(
-          (part) => model.toLowerCase().includes(part),
-        )
+        [
+          "vision",
+          "math",
+          "guard",
+          "mistrallite",
+          "mistral-openorca",
+          "claude",
```
RomneyDa: should return true for claude I think?

uinstinct: apologies. corrected now!

RomneyDa (dismissed):

RomneyDa (dismissed):

RomneyDa (approved):

PR #7128, opened 2025-08-13
fix: pass no proxy hosts in IntelliJ Plugin

vldF:
## Description

After https://github.com/continuedev/continue/pull/6752, some users met issues with connection (probably, https://github.com/continuedev/continue/issues/7092 and https://github.com/continuedev/continue/issues/7081). This fix passes proxy exceptions from IntelliJ Idea settings to the binaries.

## AI Code Review

- **Team members only**: AI review runs automatically when PR is opened or marked ready for review
- Team members can also trigger a review by commenting `@continue-review`

## Checklist

- [X] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [X] The relevant docs, if any, have been updated or created
- [X] The relevant tests, if any, have been updated or created

## Test

You need to disable TCP transport if you would like to run the plugin from sources. See https://github.com/continuedev/continue/pull/6752

PR #7238, opened 2025-08-19
fix: sanitize proxy settings + add NO_PROXY support + add unit tests

exigow:
Previously, proxy settings parsing was very _"garbage-in, garbage-out"_ and since the HttpConfigurable class was written in Java, some settings could be null. We didn’t guard against that and simply concatenated proxy + port, which sometimes produced invalid values like `null:8080`.

The implementation has been updated to sanitize + skip entries that cannot be used as env vars. In the worst case, the result is simply an empty map and the process env vars are unaffected.

Co-author: @vldF
Related to: https://github.com/continuedev/continue/pull/7128
    

---

## Summary by cubic
Sanitize IntelliJ proxy settings and add NO_PROXY support so the Continue binary gets valid env vars and respects proxy exceptions. Adds unit tests for enabled/disabled/null/blank cases.

- **Bug Fixes**
  - Validate PROXY_HOST and skip null/blank to avoid values like "null:8080".
  - When proxy is disabled, return an empty env map so process env stays unchanged.

- **New Features**
  - Forward IntelliJ proxy exceptions as NO_PROXY.
  - Use toContinueEnvVars() to set HTTP_PROXY and NO_PROXY together.

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/process/ContinueBinaryProcess.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/continue/process/ContinueBinaryProcess.kt
@@ -1,8 +1,8 @@
 package com.github.continuedev.continueintellijextension.`continue`.process
 
-import com.github.continuedev.continueintellijextension.proxy.ProxySettings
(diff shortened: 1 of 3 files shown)


vldF:
Thank you!

Patrick-Erichsen (approved):

PR #7254, opened 2025-08-20
chore: remove ids from tooltips

uinstinct:
## Description

Initially TooTips required to have an id associated with 2 elements to display it. This required 2 independent elements one with `id` and another with `data-tooltip-id`.
This PR introduces a single Tooltip which removes the usage of `id` and the displaying element can be passed as children to the tooltip

- refactors ToolTips to match new props
- keeps all of the UI and other content the same

resolves CON-3586

## AI Code Review

## Checklist

## Tests

---

## Summary by cubic
Replaced id-based tooltip wiring with a simpler, id-less API. Tooltips now wrap the trigger element and accept content directly, keeping UI and behavior the same.

- **Refactors**
  - New API: <ToolTip content="...">{trigger}</ToolTip> with auto-generated ids.
  - Removed data-tooltip-id and id usage across components.
  - Preserves existing options (place, className, style, clickable, delays).
  - Simplified helpers (e.g., InfoHover now accepts msg as a string).

- **Migration**
  - Wrap the target element with <ToolTip content="..."> instead of pairing an element with data-tooltip-id and a separate <ToolTip id>.
  - Move any previous tooltip children into the content prop.
  - Update InfoHover to pass msg as a string.

### DIFF
--- a/gui/src/components/AcceptRejectDiffButtons.tsx
+++ b/gui/src/components/AcceptRejectDiffButtons.tsx
@@ -58,33 +58,31 @@ export default function AcceptRejectAllButtons({
       className="flex flex-row items-center justify-evenly gap-3 px-3"
       data-testid="accept-reject-all-buttons"
     >
-      <button
-        className="text-foreground flex cursor-pointer flex-row flex-wrap justify-center gap-1 border-none bg-transparent p-0 text-xs opacity-80 hover:opacity-100 hover:brightness-125"
-        onClick={() => handleAcceptOrReject("rejectDiff")}
-        data-testid="edit-reject-button"
-        data-tooltip-id="reject-shortcut"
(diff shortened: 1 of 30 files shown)


tomasz-stefaniak (commented):
Just a sanity check - have you verified that this doesn't change the layout? It could, result in a slightly different HTML and alignment issues, I want to make sure it's been visually inspected before merging.

uinstinct:
> Just a sanity check - have you verified that this doesn't change the layout? It could, result in a slightly different HTML and alignment issues, I want to make sure it's been visually inspected before merging.

Yes! I checked most of them including simple ones, nested divs inside tooltips, hidden properties, clickables, etc. 
All seem to be working fine!

tingwai (approved):
Reviewed together with @uinstinct!

PR #7280, opened 2025-08-21
fix: :bug: Fix lsTool Exclusions

chezsmithy:
## Description

Noting that there are excessive fallbacks to the terminal tool to try and discover certain files and folders in the repo. In looking at the code the lsTool is using the walkDir function which is shared with the codebase indexer. There are extensive exlusions in place for the codebase indexer which makes sense, however for the lsTool these exclusions are confusing for the LLM when files that should exist are just not found. 

This change allows the lsTool to find all files but still considers the default ignore mechanisms.

## AI Code Review

## Checklist

## Tests

No tests were added. Manual testing was performed in the testing sandbox. For example before the change the dist directory is not returned. After the change the dist directory is included.

    

---

## Summary by cubic
Fixes lsTool to stop inheriting codebase indexer exclusions, so it returns full directory listings and avoids terminal fallbacks. It now includes directories like dist/ and build/ while still honoring default ignore rules (e.g., .gitignore).

### DIFF
--- a/core/tools/implementations/lsTool.ts
+++ b/core/tools/implementations/lsTool.ts
@@ -1,3 +1,5 @@
+import ignore from "ignore";
+
 import { ToolImpl } from ".";
 import { walkDir } from "../../indexing/walkDir";
 import { resolveRelativePathInDir } from "../../util/ideUtils";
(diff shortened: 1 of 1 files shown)


RomneyDa:
@chezsmithy I agree with this change, has been a major thorn. TLDR I'll open a quick PR to address security implications before we merge this


Especially given recent AI security breaches, we should update the behavior to be the following:
- Split the default ignores into security concerns vs additional indexing concerns.
- Codebase indexing respects default ignores = security + additional
- lsTool can _see_ everything
- Prevent the readFile and readCurrentlyOpenFile tools from reading files that match security concerns, with a message to the model (can't read this, it's a security concern)

This would not fully prevent terminal commands from being able to bypass this but that's already the case

RomneyDa:
@chezsmithy https://github.com/continuedev/continue/pull/7302
This PR will stay the same, let's just merge that one first!

RomneyDa (approved):

Issue #5504, opened 2025-05-04
idea log full of JS_QUERY_POOL_SIZE warns


mosinnik:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:
- Continue version:
- IDE version:
- Model:
- config:
  

  
  OR link to assistant in Continue hub:
```

### Log output

```Shell

```

exigow:
Thanks @mosinnik,

I discovered a hint that might help reproduce this problem: 
* if we close the IDE with the toolbar open and then start the IDE again, there are no errors (probably because toolbar is already open - this is standard behavior because IDE remembers your previous layout)
* if we close the IDE with the toolbar hidden, then start the IDE again and open the toolbar manually, these exceptions appear

It has something to do with the order of initialization. I will look on this.

QianKuang8:
Hi @exigow, is there any progress in this issue?
I met few times the issue https://github.com/continuedev/continue/issues/5774 which seems like these two are similar.

exigow:
Hi @QianKuang8, it's fixed in https://github.com/continuedev/continue/pull/7291 👍

It's a draft, but practically finished (I'm testing the features now).

QianKuang8:
That's so great. Hope it can solve most issues.

PR #7291, opened 2025-08-21
fix: fix JCEF browser issues

exigow:
Solves CON-2469
Solves CON-2979
Solves #5504

### Off-screen rendering is the new default

Off-screen rendering is now enabled by default. This is motivated by the fact that most users will benefit from it,
but the minority with issues can disable it in the Continue settings. I believe this is the correct strategy,
because disabling it causes a flickering window for the majority.

Note: I tried to mask it by preloading the HTML with a background color and by changing the background color of the
browser component. That worked, but only reduced the flashing time and didn't completely eliminate the problem.
Enabling Off-screen rendering seems to be the only robust solution.

### Fix JS_QUERY_POOL_SIZE errors

After decoupling the browser from the toolbar and moving it to a dedicated service, there is now full control over
when and what is initialized. The setting is now applied correctly.

### Rework actions

I've noticed that most of the actions require a working, active Continue toolbar.
I've created the `ContinueToolbarAction` abstraction to remove duplicated logic. I've also removed `actions.utils`.

### Minor changes

* Fix bug in `CoreMessenger.handleMessage` (some pass-through messages had message type used as message ID)
* Add `browser` package and moved browser-related classes there, including `CustomHandlerFactory` (package `factories` is now removed). `CustomHandlerFactory` implementation is untouched.
* Add `project.getBrowser()` extension method (this is mostly for readability).
* Some loggers now use a shorter identifier (simple class name) without the full package, because our package name is 
  unusually long, and it was hard to read the logs.
* Remove commented-out `TerminalActivityTrackingService`.
* Create `JcefErrorPanel` and put `jcef_error.html` there.

### DIFF
--- a/extensions/intellij/build.gradle.kts
+++ b/extensions/intellij/build.gradle.kts
@@ -96,8 +96,7 @@ intellijPlatformTesting {
                         "-Dapple.laf.useScreenMenuBar=false",
                         "-Didea.trust.all.projects=true",
                         "-Dide.show.tips.on.startup.default.value=false",
-                        "-Dide.browser.jcef.jsQueryPoolSize=10000",
-                        "-Dide.browser.jcef.contextMenu.devTools.enabled=true"
+                        "-Dide.browser.jcef.sandbox.enable=false"
                     )
                 }
             }
@@ -145,5 +144,6 @@ tasks {
 
     test {
         useJUnitPlatform()
+        jvmArgumentProviders += CommandLineArgumentProvider { listOf("-Dide.browser.jcef.sandbox.enable=false") }
     }
 }


exigow:
I tested all the features (actions, focusing, chat functionalities) and it seems to be OK, except for one action that doesn’t recognize the command `focusContinueInputWithoutClear`, however, this issue existed before. I squashed all the "wip" commits into a single one with all the details.

exigow:
I'm switching it back to draft because we need to discuss OSR (let's wait with merge).

exigow:
✅ ~TODO: Remove OSR from settings (it should be enabled by default with no option to change).~

I've enabled OSR by default and removed the option. I also made sure there are no outdated comments in the readme/contribute docs.

Patrick-Erichsen (dismissed):

Patrick-Erichsen:
@exigow I think the test failure here is no longer flake. Might be a test that we can just remove though, lmk your thoughts.

exigow:
@Patrick-Erichsen, you're right, it was failing test (compilation error caused by changes in our browser code). :sweat_smile: 

I think it's better to remove it. The test was fine, but almost everything was mocked there and the test was heavily dependent on the handler's code (+ everything around). It made some sense, but ideally there should be as few mocks as possible and it should be tested in a more black-box manner.

If you think we should keep it, let me know - I'll try to fix it.

exigow:
Interesting error on CI (it's from e2e test I suppose):

`[0826/175633.869512:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /home/runner/.gradle/caches/8.14.3/transforms/83a0ed050b837f55f39183dd9bb4af55/transformed/ideaIC-2024.1/jbr/lib/chrome-sandbox is owned by root and has mode 4755.`

Test is passing locally. :face_exhaling: I'm investigating this right now.

exigow:
:green_circle: 

(fix: https://github.com/continuedev/continue/pull/7291/commits/2b085c15c09f6070e2b1519d81517d21540b783c)

Patrick-Erichsen (approved):

PR #7294, opened 2025-08-21
feat: scroll to first diff after apply completes

uinstinct:
## Description

Scroll to the first diff after apply completes irrespective of where the cursor is.

resolves CON-2190

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/b546124a-84d4-48db-b194-5be7a4062e92

## Tests

---

## Summary by cubic
Scroll to the first diff after applying changes, centering the editor on that line regardless of the current cursor. Addresses Linear CON-2190.

- New Features
  - Capture the first changed line before clearing decorations and reveal it InCenter after apply.
  - Works even if the cursor was moved elsewhere during the apply.


extensions/vscode/src/diff/vertical/handler.ts
```diff
@@ -97,6 +97,17 @@ export class VerticalDiffHandler implements vscode.Disposable {
       false,
     );
 
+    // Store the first changed line position before clearing decorations
+    let firstChangedLine: number | null = null;
+    if (accept) {
+      const diffBlocks =
```
uinstinct: @RomneyDa had a doubt here: this works fine when we accept all the diffs at once using the green tick button. when accepting them one by one using the accept phrase which appears on the editor, it `this.editorToVerticalDiffCodeLens.get(this.fileUri)` is null. Do you have any pointers what I could be missing?

uinstinct: no worries... found it today!

RomneyDa (changes requested):
@uinstinct the goal is to stream to the first diff after apply finishes streaming, not after it is closed. The view should not move when diffs are accepted/rejected

PR #7302, opened 2025-08-21
feat: prevent unsafe file reads

RomneyDa:
## Description

Prevents tools, autocomplete, and some other services from reading files which are likely security concerns
Achieves this by splitting defaultignores into security vs indexing. Indexing now ignores security + additional ignores, and the security ignores are used in various places to prevent unsafe reads

CON-3622

    

---

## Summary by cubic
Introduce security-focused ignore rules and enforce them across autocomplete, NextEdit, context providers, and tools to prevent unsafe file reads/edits. Addresses CON-3622 and reduces the risk of leaking secrets.

- **New Features**
  - Split ignores into security vs indexing; added default patterns for env files, keys/certs, secret dirs, and credential stores.
  - Autocomplete/NextEdit: skip suggestions/edits for security-matched files; getAutocompleteContext throws early; VS Code RecentlyVisitedRanges ignores them.
  - Context providers: OpenFiles and Problems redact content when a file is a security concern.
  - Tools: read/edit/search-and-replace (CLI and GUI) now block on sensitive paths via throwIfFileIsSecurityConcern with clear error messaging.
  - Ignore utilities: added defaultFileAndFolderSecurityIgnores and string constants; combined defaults still used for indexing.


core/context/providers/FileContextProvider.ts
```diff
@@ -36,6 +37,21 @@ class FileContextProvider extends BaseContextProvider {
       await extras.ide.getWorkspaceDirs(),
     );
 
+    if (isSecurityConcern(relativePathOrBasename)) {
```
sestinj, line 40: Given that the user needs to explicitly choose the file themselves, I don't think we should check here

core/context/providers/ProblemsContextProvider.ts
```diff
@@ -28,6 +29,14 @@ class ProblemsContextProvider extends BaseContextProvider {
           problem.filepath,
           workspaceDirs,
         );
+        if (isSecurityConcern(relativePathOrBasename)) {
```
sestinj, line 32: this also doesn't feel that important. the file needs to be explicitly opened by the user, AND they need to select the context provider themselves

tomasz-stefaniak (approved):

extensions/cli/src/tools/searchAndReplace/index.ts
```diff
@@ -95,7 +97,7 @@ Each string in the diffs array can contain multiple SEARCH/REPLACE blocks, and a
   preprocess: async (args) => {
     // Get and validate args
     const { filepath, diffs } = parseSearchAndReplaceArgs(args);
-
+    throwIfFileIsSecurityConcern(filepath);
```
sestinj, line 100: I don't think we ever show the model the output of this tool, which means we don't need to throw. I absolutely expect this to be a problem in codebase scaffolding scenarios where the model might want to create a .env for you that has examples (and doesn't think to use .example.env

extensions/cli/src/ui/UserInput.tsx
```diff
@@ -254,6 +255,7 @@ const UserInput: React.FC<UserInputProps> = ({
       // Read the file content and notify parent component
       if (onFileAttached) {
         try {
+          throwIfFileIsSecurityConcern(filePath);
```
sestinj, line 258: remove

RomneyDa:
@sestinj resolved your comments here
https://github.com/continuedev/continue/pull/7310

PR #7310, opened 2025-08-22
fix: cleanup file security pr

RomneyDa:
## Description
Removes cases of preventing file reads that were unneccessary
Reinstates `./continue` as a security concern for API keys


chezsmithy:
I think I noted venv get added which is a common python directory for a virtual environment for actual installing and running a python project. Should it be redacted?

chezsmithy:
If .continue is added would that inhibit using continue to help you setup MCP servers, or prompts/rules?

RomneyDa:
> If .continue is added would that inhibit using continue to help you setup MCP servers, or prompts/rules?

Correct, I moved it back to indexing ignore only. I was thinking for API keys but people shouldn't put inline API keys in workspace .continue

RomneyDa:
> I think I noted venv get added which is a common python directory for a virtual environment for actual installing and running a python project. Should it be redacted?

Seems right, moved!

sestinj (approved):

PR #7315, opened 2025-08-22
feat: continue tool calls after rejection

uinstinct:
## Description

Continue streaming when a tool call is rejected. This PR introduces an experimental setting which will resume tool calls. It will skip the tool call if not necessary otherwise will try another approach.

resolves CON-2844

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/4fce1d36-03de-4c20-906f-683097683e99

## Tests

---

## Summary by cubic
Enable continuing the conversation after a tool call is rejected. Adds an opt-in setting that treats canceled tool calls as complete and resumes streaming (CON-2844).

- **New Features**
  - New UI setting: “Stream after tool rejection” (default off). Stored as ui.continueAfterToolRejection and added to shared config/types.
  - On cancel, appends a tool message (“User skipped the tool call”) and a “Tool Call Rejected” context item, then auto-continues the LLM response.
  - Tool-call completion logic considers “canceled” as done when the setting is enabled, so chats don’t get stuck.


gui/src/redux/thunks/cancelToolCall.ts
```diff
@@ -21,6 +34,36 @@ export const cancelToolCallThunk = createAsyncThunk<
...
+          },
+        ],
+      }),
+    );
+  }
+
   // Dispatch the actual cancel action
   dispatch(cancelToolCallAction({ toolCallId }));
+
+  if (continueAfterToolRejection) {
```
RomneyDa: Use streamResponseAfterToolCall here or will break parallel tool calls

gui/src/redux/thunks/cancelToolCall.ts
```diff
@@ -21,6 +34,36 @@ export const cancelToolCallThunk = createAsyncThunk<
...
+    dispatch(
+      updateToolCallOutput({
+        toolCallId,
+        contextItems: [
+          {
+            icon: "problems",
+            name: "Tool Call Rejected",
+            description: "User skipped the tool call",
+            content: DEFAULT_USER_REJECTION_MESSAGE,
+            hidden: false,
```
RomneyDa: I might hide this one `hidden: true`

gui/src/redux/thunks/streamResponseAfterToolCall.ts
```diff
@@ -71,7 +75,12 @@ export const streamResponseAfterToolCall = createAsyncThunk<
             item.toolCallStates?.some((tc) => tc.toolCallId === toolCallId),
         );
 
-        if (areAllToolsDoneStreaming(assistantMessage)) {
+        if (
+          areAllToolsDoneStreaming(
+            assistantMessage,
+            state.config.config.ui?.continueAfterToolRejection ?? false,
```
RomneyDa: since it's false by default don't need to include ?? false here, just make param boolean | undefined

gui/src/redux/thunks/cancelToolCall.ts
```diff
@@ -1,15 +1,28 @@
...
+If the tool call is essential, try an alternative approach.
+If no alternatives exist, offer to pause here.`;
 
 export const cancelToolCallThunk = createAsyncThunk<
   void,
   { toolCallId: string },
   ThunkApiType
 >("chat/cancelToolCall", async ({ toolCallId }, { dispatch, getState }) => {
   const state = getState();
+  const continueAfterToolRejection =
```
RomneyDa, line 22: same here no need for ?? false since undefined is falsy

gui/src/redux/thunks/cancelToolCall.ts
```diff
@@ -21,6 +34,36 @@ export const cancelToolCallThunk = createAsyncThunk<
...
   // Dispatch the actual cancel action
   dispatch(cancelToolCallAction({ toolCallId }));
+
+  if (continueAfterToolRejection) {
+    const rejectionMessage: ChatMessage = {
+      role: "tool",
+      content: "User skipped the tool call",
+      toolCallId,
+    };
+    dispatch(streamUpdate([rejectionMessage]));
```
RomneyDa: streamResponseAfterToolCall will also handle this logic

RomneyDa (changes requested):
See comments!

RomneyDa (approved):

PR #7322, opened 2025-08-22
feat: Adding the isExecutionResultSearchable property to the Tool indicates whether the tool's execution results are checkable

xiaohuanxiong3:
## Description

Adding the isExecutionResultSearchable property to the Tool indicates whether the tool's execution results are checkable
Remove the duplicate onClick function of ToggleWithIcon in SimpleToolCallUI

## AI Code Review

## Checklist

## Tests


tingwai:
Thank you for the PR, would you mind providing screenshots showing the current vs the new look?

xiaohuanxiong3:
@tingwai My local test result is that the current look and the new look are consistent, given this, do I still need to provide screenshots of the current look and the new look?

RomneyDa (changes requested):
@xiaohuanxiong3 I think for now we've used `hidden: true` on tool output context items. If all `isExecutionResultCheckable` are set to true, what is the added functionality?

xiaohuanxiong3:
@RomneyDa 
    `isExecutionResultSearchable` is used to indicate that for certain tools, their execution results (typically query results) are not displayed by default, but users can check their execution results by clicking on the tool invocation tag.
    The `hidden` property of `ContextItem` is used to indicate whether the partial result of a tool call is displayable. In `SimpleToolCallUI`, only ContextItems with hidden set to false are selected for display.
    Some tools like `runTerminalCommandTool` display their execution results by default, and the display method differs from tools like `readFile`. Therefore, I think using the hidden attribute cannot achieve the desired goal effectively.
    As for `isExecutionResultSearchable`, my thought is that if `isExecutionResultSearchable` is true but `toolCallIcon` is null, we should provide a default icon. It would be cleaner to remove the `isExecutionResultSearchable` property

RomneyDa:
@xiaohuanxiong3 just to clarify is the only change here code cleanup at this point? Besides the duplicate onClick
I realize it's nice to have icon names with the tool definition but I'm also not sure I like the string Icon names in core because it feels like mixing frontend and backend a bit

xiaohuanxiong3:
@RomneyDa So to speak. But it also makes tool development more convenient because you don't need to add icon and tool name mappings to specific places in the gui module when developing tools like ReadFile

RomneyDa:
note to reviewers, please squash and merge

RomneyDa (approved):
Talked with @Patrick-Erichsen, we think it's a net positive. @xiaohuanxiong3 thanks for the contribution!

PR #7323, opened 2025-08-22
fix: grep escaped chars and empty results

RomneyDa:
## Description
- Adds handling for improperly escaped chars in grep patterns post JSON-parsing, along with explanations for the model in case this fails
- Makes truncation warning context items for various tools less ugly by removing unnecessary description and shortening name
- Bumps grep token limit from ~1000 to ~1500
- Adds meaningful no-result context items for glob and grep queries that return zero results

CON-3953
CON-3497


chezsmithy:
@RomneyDa funny. We had the same idea. https://github.com/continuedev/continue/pull/7370

I went way down the rabbit hole to solve for args and more focused searches as well.

PR #7329, opened 2025-08-22
fix: add Run Extension (standalone) config

exigow:
Currently, we only provide `Run Extension` with `USE_TCP` and `GUI_URL` env vars to connect with localhost server. The compound `Run Continue` config relies on this. If someone (like me :sweat_smile:) wants to run the IDE without starting the server, they have to create a custom config manually (it's basically `runIde` without any customs) or remember not to commit changes in the default config (after removing these params).

I noticed that I use my config about 95% of the time, because often I don't need to debug the server (usually, if the changes only concern the plugin code). The plugin starts faster, has a smaller footprint, and I don’t have to think about the rest of the compound `Run Continue` config.

To sum-up, after this changes we would have:

* `Run Extension (use TCP)`, default config, AKA Continue-over-TCP (previously named `Run Extension`)
* `Run Extension (standalone)`, extra config, AKA Continue-over-STDIO

    

---

## Summary by cubic
Split the IntelliJ extension run config into two: standalone (no server) and use TCP (with USE_TCP/GUI_URL). Updated the compound Run Continue configs to target the TCP variant, making standalone development simpler and faster.

- **New Features**
  - Added "Run Extension (standalone)" using runIde with no env vars.
  - Added "Run Extension (use TCP)" with USE_TCP=true and GUI_URL.
  - Updated "Run Continue" and "Run Continue (CE)" to use the TCP config.
  - Removed the legacy "Run Extension" config.


exigow:
(I tested the configs :heavy_check_mark:)

exigow:
The summary by cubic is completely off the mark :smile:

Patrick-Erichsen:
Screenshot 2025-08-26 at 8 59 32 AM

This one really doesn't want to get merged 😅

Issue #7337, opened 2025-08-22
Session storage can grow too large.


etherandrius:
### Before submitting your bug report

- [ ] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: linux vs code backend / mac VS code frontend running in Chromium.
- Continue version: v1.1.65
- IDE version: VS Code web running in a browser
```

### Description

Continue stores the current session in localStorage. For longer conversations this context can grow too large and hit QuotaExceededError 

https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#web_storage
```
Browsers can store up to 5 MiB of local storage, and 5 MiB of session storage per origin.
```

<img width="806" height="56" alt="Image" src="https://github.com/user-attachments/assets/a96ca360-9992-446e-aa81-3c351dc6b77a" />

My use-case is pretty long conversations and large amount of tool usage. Every so often I get `QuotaExceededError` and have to clear localStorage.

In the most recent case I had 76 messages before the error was hit.
<img width="538" height="33" alt="Image" src="https://github.com/user-attachments/assets/ceb3fcd7-df65-4a19-ba19-512cb6ec3588" />

The biggest culprit seems to be the `promptLogs` key which stores all the tools and their properties. I have about 30 large tools with complex input schemas and long descriptions. `appliedRules` key is not helping either.

<img width="684" height="149" alt="Image" src="https://github.com/user-attachments/assets/59a4aca4-8084-4d2e-8fd5-49412822d305" />

If I delete the two keys I get a 30x smaller size

<img width="835" height="68" alt="Image" src="https://github.com/user-attachments/assets/c2f232e3-14f2-4919-846d-901062e33fa8" />

etherandrius:
Attempted fix
 - https://github.com/continuedev/continue/pull/7338 - removing completionOptions
 - https://github.com/continuedev/continue/pull/7343 - removing appliedRule text from history

[...]

PR #7343, opened 2025-08-22
Fix: Remove applied rules from chat history

etherandrius:
## Description

https://github.com/continuedev/continue/issues/7337

### DIFF
--- a/gui/src/redux/slices/sessionSlice.ts
+++ b/gui/src/redux/slices/sessionSlice.ts
@@ -7,7 +7,6 @@ import {
 } from "@reduxjs/toolkit";
 import { JSONContent } from "@tiptap/react";
 import {
-    AppliedRule,
   ApplyState,
   AssistantChatMessage,
   ChatHistoryItem,


gui/src/components/mainInput/belowMainInput/RulesPeek.tsx
```diff
@@ -45,29 +47,37 @@ const getSourceLabel = (rule: RuleWithSource): string => {
...
+      rule.source === "default-agent"
+    ) {
+      ideMessenger.post("openUrl", DEFAULT_SYSTEM_MESSAGES_URL);
+    } else {
+      ideMessenger.post("config/openProfile", {
+        profileId: undefined,
+        element: { sourceFile: (rule as any).sourceFile },
+      });
     }
   };
```
etherandrius: This is not ideal, but it's consistent with `RuleCard: React.FC<RuleCardProps>` on click will just open the rules files in the editor.

etherandrius:
@RomneyDa linked back to the issue

core/index.d.ts
```diff
@@ -497,10 +497,21 @@ export interface ChatHistoryItem {
   toolCallStates?: ToolCallState[];
   isGatheringContext?: boolean;
   reasoning?: Reasoning;
-  appliedRules?: RuleWithSource[];
+  appliedRules?: AppliedRule[];
   conversationSummary?: string;
 }
 
+export interface AppliedRule {
```
RomneyDa: Can use Omit<RuleWithSource, "rule">

RomneyDa (changes requested):
@etherandrius see nitpick
Overall agree with this, will be nice to reduce UI a bit here too

gui/src/components/mainInput/Lump/sections/RulesSection.tsx
```diff
@@ -29,10 +29,33 @@ interface RuleCardProps {
...
+      profileId: undefined,
+      element: { sourceFile: (rule as any).sourceFile },
+    });
+  }
+};
+
 const RuleCard: React.FC<RuleCardProps> = ({ rule }) => {
   const dispatch = useAppDispatch();
-  const ideMessenger = useContext(IdeMessengerContext);
-  const mode = useAppSelector((store) => store.session.mode);
```
etherandrius, line 35: unused.

gui/src/components/mainInput/Lump/sections/RulesSection.tsx
```diff
@@ -29,10 +29,33 @@ interface RuleCardProps {
   rule: RuleWithSource;
 }
 
+export const openRules = async (rule: Omit<RuleWithSource, "rule">) => {
+  const ideMessenger = useContext(IdeMessengerContext);
```
etherandrius, line 33: `ideMessenger` this was define outside the function before. I assume the `useContext(IdeMessengerContext)` call is basically free and we can call it inside the function.

gui/src/redux/thunks/streamNormalInput.ts
```diff
@@ -282,7 +282,7 @@ export const streamNormalInput = createAsyncThunk<
             ...(appliedRules.length > 0 && {
               rules: appliedRules.map((rule) => ({
                 id: getRuleId(rule),
-                rule: rule.rule,
+                rule: "", // TODO: remove rule key from type. The contents should be present in the prompt
```
etherandrius, line 285: a bit funky.

core/llm/rules/getSystemMessageWithRules.ts
```diff
@@ -361,6 +361,9 @@ export const getSystemMessageWithRules = ({
 
   return {
     systemMessage,
-    appliedRules,
+    appliedRules: rules.map(fullRule => {
+      const { rule, ...rest } = fullRule // destruct rule key
+      return rest;
+    }),
```
etherandrius, line 367: A bit funky as well. 

Ideally this method would not take `availableRules: Omit<RuleWithSource, "rule">[]` and read rules from the file. 
This would also fix the issue of rules sometimes being outdated <- maybe a FLUP.

gui/src/components/mainInput/Lump/sections/RulesSection.tsx
RomneyDa: Will need to pass ideMessenger for this one since can't use hooks outside of top level of component

core/index.d.ts
```diff
@@ -497,7 +497,7 @@ export interface ChatHistoryItem {
   toolCallStates?: ToolCallState[];
   isGatheringContext?: boolean;
   reasoning?: Reasoning;
-  appliedRules?: RuleWithSource[];
+  appliedRules?: Omit<RuleWithSource, "rule">[];
```
RomneyDa, line 500: Can keep the AppliedRule type and just say `type AppliedRule = Omit<RuleWithSource, "rule">` to prevent having `Omit`s all over

core/llm/rules/getSystemMessageWithRules.ts
RomneyDa: This function could still just return RuleWithSource, shouldn't effect storage if omitted at the point of redux

note that config should reload if a rule file is updated, so should be up to date in most cases

gui/src/components/mainInput/Lump/sections/RulesSection.tsx
```diff
@@ -29,10 +29,33 @@ interface RuleCardProps {
   rule: RuleWithSource;
 }
 
+export const openRules = async (rule: Omit<RuleWithSource, "rule">) => {
```
RomneyDa, line 32: could take RuleWithSource | AppliedRule

gui/src/redux/thunks/streamNormalInput.ts
RomneyDa: Any specific reason to remove rules from dev data? Have you had issues with verbose dev data?

RomneyDa (changes requested):
1. Main blocking feedback is to remove use of useContext in utility function
2. create new type with Omit and pass the new type around to prevent duplicate `Omit`s 
3. Leave rule content on dev data unless specific issues with dev data storage - let me know your thoughts on this?

RomneyDa:
@etherandrius bump, note failing prettier/tests!

RomneyDa:
@etherandrius bump!

RomneyDa:
@etherandrius note that the root localStorage issue will also be solved by https://github.com/continuedev/continue/pull/7944

RomneyDa:
I think this is a good change, since this is a bit stale re-worked here!
https://github.com/continuedev/continue/pull/8594

PR #7353, opened 2025-08-22
fix: keep assistant menu open when clicking 'Reload agents'

sestinj:
Prevents the assistants dropdown menu from closing when the user clicks the 'Reload agents' option by using stopPropagation() to prevent event bubbling to the parent component that handles menu closure.

Previously, clicking 'Reload agents' would immediately close the menu, making it impossible to see the loading state or perform subsequent actions without reopening the menu. Now the menu stays open as expected for this action.
    

---

## Summary by cubic
Keeps the Assistants dropdown open when "Reload assistants" is clicked, so the loading state stays visible and you can take follow-up actions without reopening the menu. Prevents click event bubbling to the parent close handler.


tingwai (changes requested):
Test PR, ignore

PR #7370, opened 2025-08-24
Codex/refactor-grep-implementation-in-extensions

chezsmithy:
## Description

Performed extensive analysis of common grep scenarios being attempted by the LLM and researched common features enabled in grep tools in other code assistants. 

Implementing recommended changes to allow the LLM to request specific arguments like -A and -B or -l for example to research code just before or just after.

Implementating scoping by add an optional path parameter so the LLM doesn't need to always search the whole repo which might result in truncation. Also allow scoping searches to specific files to avoid the need to always read the whole file.

## AI Code Review

## Checklist

## Tests

Added exstensive tests.
Perfomed extensive manual testing by asking the LLM to generate common lsTool type queries against the test repo and observed the outcomes expecting full passes.

### DIFF
--- a/core/tools/definitions/grepSearch.ts
+++ b/core/tools/definitions/grepSearch.ts
@@ -21,7 +21,7 @@ export const grepSearchTool: Tool = {
         query: {
           type: "string",
           description:
-            "The search query to use. Must be a valid ripgrep regex expression, escaped where needed",
+            "The search query to use. Must be a valid ripgrep regex expression. For word boundaries use \\b, for literal parentheses use \\(, for alternations use (pattern1|pattern2). Examples: 'function.*add', '\\bclass\\b', 'add.*\\(', '\\b(def|function)\\b'",
         },
         args: {
           type: "array",
@@ -29,12 +29,20 @@ export const grepSearchTool: Tool = {
           description:
             "Optional additional ripgrep arguments, for example ['-n', '-w']",
         },
+        path: {
+          type: "string",
+          description:
+            "Optional path to search in - can be a specific file (e.g., 'src/main.ts') or directory (e.g., 'src/'). If not provided, searches entire project.",
+        },
       },
     },
   },
   defaultToolPolicy: "allowedWithoutPermission",
   systemMessageDescription: {
     prefix: `To perform a grep search within the project, call the ${BuiltInToolNames.GrepSearch} tool with the query pattern to match. For example:`,
-    exampleArgs: [["query", ".*main_services.*"]],
+    exampleArgs: [
+      ["query", ".*main_services.*"],
+      ["query", "throw new Error"],
+    ],
   },
 };
--- a/core/tools/definitions/grepSearch.ts
+++ b/core/tools/definitions/grepSearch.ts
@@ -23,6 +23,12 @@ export const grepSearchTool: Tool = {
           description:
             "The search query to use. Must be a valid ripgrep regex expression, escaped where needed",
         },
+        args: {
+          type: "array",
+          items: { type: "string" },
+          description:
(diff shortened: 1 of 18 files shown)


core/tools/definitions/grepSearch.ts
```diff
@@ -21,14 +21,28 @@ export const grepSearchTool: Tool = {
...
       },
     },
   },
   defaultToolPolicy: "allowedWithoutPermission",
   systemMessageDescription: {
     prefix: `To perform a grep search within the project, call the ${BuiltInToolNames.GrepSearch} tool with the query pattern to match. For example:`,
-    exampleArgs: [["query", ".*main_services.*"]],
+    exampleArgs: [
+      ["query", ".*main_services.*"],
+      ["query", "throw new Error"],
```
RomneyDa, line 45: path?
EDIT: to clarify, I mean the system message tools example for this will not be valid. missing args "path" and "args"

chezsmithy:
@RomneyDa thoughts on these additional enhancements? I wanted to solve for a number of use cases I saw attempted but failed. The big ones were -A -B, -l and path scoped searches. Note often I see failed grep attempts by Claude Sonnet with fallback to full file reads or terminal use.

core/tools/implementations/grepSearch.ts
```diff
@@ -40,13 +43,25 @@ function splitGrepResultsByFile(content: string): ContextItem[] {
...
-
-  const { query, warning } = prepareQueryForRipgrep(rawQuery);
+  const extraArgs = getOptionalStringArrayArg(args, "args");
+  const path = args.path as string | undefined;
 
   let results: string;
+  let queryWarning: string | undefined;
+
   try {
+    // const { query, warning } = prepareQueryForRipgrep(rawQuery);
```
chezsmithy, line 53: @RomneyDa in my testing, I noted adding this check in before processing reduces the number of successful queries slightly. Does it make sense to include it? Without it I see a higher degree of success, now this is with Claude Sonnet 4 which might be more reliable at building regex.

RomneyDa: In which cases does it reduce successful queries? Might be a specific bug with the prepareQueryForRipgrep function? We could look at fixing/adding a test case. The preparation was introduced because even claude 4 regularly outputs queries in a way where the JSON parsed pattern is invalid regex. Not sure if it's a prompting issue or model issue or parsing issue but `prepareQueryForRipgrep` resolves most cases

Open to other approaches especially if seems to be a prompting issue or JSON parsing issue

RomneyDa (changes requested):
note comment from first review about the systemMessageDescription
Partly requesting changes because we're still not sure we want another tool that uses array args for now, because of issues with array args when using system message tools, and many models just aren't great with array args. Seems like a definite improvement for e.g. claude 4. I think we'd want a simple pattern grep search tool available for some models. Might make sense to more or less duplicate and use config-dependent to only add the fleshed out grep search tool for great models. Thoughts?

chezsmithy:
@RomneyDa got it! Why don't I look at if it's possible to scrape the args and path out of the query string too. I like the idea of maybe two tools and applying by model. I'll take your lead on what you are doing for the search and replace work if that makes sense?

RomneyDa:
@chezsmithy a single query string of args could be interesting, the `shell-quote` package used in CLI could be helpful for parsing this

Then maybe we could just update the argument description based on if it's a proficient model or not, e.g. poor models see "ripgrep regex expression" where better models might see "string of ripgrep args" or similar. That might be too clever. A fully separate tool that takes an array of `rg` args as implemented here seems good if it's gated to a few models now - yes similar to multi edit tool.

Issue #7072, opened 2025-08-09
@Codebase doesn't include relevant files


Darkspeed294:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS 15,1 and Arch linux
- Continue version: latest + pre-release
- IDE version: VScode 1,103
- Model: Qwen3-32b-A3B GGUF
- config:
  
name: Local Assistant
version: 1.0.0
schema: v1
models:
  - name: Qwen3 32b
    provider: llama.cpp
    model: qwen3-coder-32b
    apiBase: http://192.168.2.61:5000
    roles:
      - chat
      - edit
      - apply
      - embed
    codebase:
      rootDir : ${workspaceFolder}
      ignoreGlobs:
      - "**/node_modules/**"
      - "**/.git/**"
      - "**/dist/**"
      - "**/__pycache__/**"
context:
  - provider: open
  - provider: code
  - provider: docs
  - provider: diff
  - provider: file
  - provider: terminal
  - provider: problems
  - provider: folder
  - provider: codebase
  - provider: currentFile
```

### Description

I am having a problem with codebase. Both on macOS and my linux computer running the LLM. Every time I add @Codebase to the prompt, it only fetches partial files. Even when I specifically state which file, it pulls the same files. In my scenario, I have 3 files: index.html, styles.css and scripts.JS. When the @Codebase command is utilized within the prompt, it only contexts the html and script file, not the css file.

I've checked the index.sqlite file with sqlitedatabase browser, and the tables clearly show that the files are in the codebase index. I've tried re-indexing on every step, different versions of continue, deleted the index file and rebuild it from scratch, tried to make a rule to automatically include all files, reinstalled extension/vscode/OS but none of them seem to work. every prompt I have to manually insert each file that is relevant to my problem, but sometimes I don't know what is causing the problem so I don't know what to insert.

[...]

PR #7375, opened 2025-08-24
fix: CSS files not being indexed (#7072)

su0as:
## Description
CSS files were not being indexed because codeChunker expects code structures (classes/functions) that don't exist in CSS. The chunker would return zero chunks, preventing CSS files from being indexed.

This fix routes CSS, HTML, JSON and similar non-code files to basicChunker instead of codeChunker, ensuring they are properly indexed while maintaining intelligent chunking for actual code files.

Root Cause's Found
- CSS files were being processed by codeChunker which expects code structures (classes/functions)
- CSS doesn't have these structures, so codeChunker returns 0 chunks
- No chunks = no indexing = no retrieval

Screenshot 2025-08-24 at 11 53 20 PM

## Tests
•  Manually verified CSS files are now indexed and retrievable via @Codebase 
•  Confirmed all existing chunk tests pass (10/10)
•  TypeScript compilation succeeds with no errors
•  Prettier formatting applied

    

---

## Summary by cubic
Fixes CSS files not being skipped during indexing by sending non-code files to the basic chunker so they produce chunks and can be retrieved. Code files still use the code chunker.

- **Bug Fixes**
  - Route css, html/htm, json, toml, yaml/yml to basicChunker.
  - Verified CSS indexing and retrieval; existing chunk tests pass.
  - No changes to behavior for code files.

### DIFF
--- a/core/indexing/chunk/chunk.ts
+++ b/core/indexing/chunk/chunk.ts
@@ -22,7 +22,22 @@ export async function* chunkDocumentWithoutId(
     return;
   }
   const extension = getUriFileExtension(fileUri);
-  if (extension in supportedLanguages) {
+  // Files that should use basicChunker despite having tree-sitter support
+  // These files don't have code structure (classes/functions) that codeChunker expects
+  const NON_CODE_EXTENSIONS = [
+    "css",
+    "html",
+    "htm",
+    "json",
+    "toml",
+    "yaml",
+    "yml",
+  ];
+
+  if (
+    extension in supportedLanguages &&
+    !NON_CODE_EXTENSIONS.includes(extension)
+  ) {
     try {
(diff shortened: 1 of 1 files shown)


su0as:
@RomneyDa Initially, this appeared to be a retrieval issue - CSS files were in the database but not being returned by @Codebase queries. However, CSS files were never being chunked/indexed at all.

Route CSS, HTML, JSON and similar non-code files to basicChunker instead of codeChunker. These files have structure but not code constructs, so they need simple line-based chunking rather than AST-based chunking.
This minimal change (adding one array check) ensures all file types are properly indexed while maintaining intelligent chunking for actual code files.

Can you review it and Provide if any changes is needed.

RomneyDa (approved):
This seems right, can't nontrivially remove from supportedLanguages since it is used for autocomplete, etc.

PR #7385, opened 2025-08-25
fix: show expansion toggle when adding a snippet via Cmd+L

fbricon:
When selecting code in a file and pressing Cmd+L, the selected snippet in the chat view doesn't show the toggle icon to expand the snippet. 

no-expand-toggle

The problem seems to be :

1. Component mounts: ExpandableToolbarPreview renders
2. useEffect runs: The useEffect with the ResizeObserver executes
3. contentRef.current is null: At this point, contentRef.current is still null because React hasn't yet attached the ref to the DOM element
4. ResizeObserver setup fails: The condition if (contentRef.current) is false, so the observer is never attached
5. No measurements: Since the observer is never attached, contentDims never gets updated from { width: 0, height: 0 }
Chevron never shows: contentDims.height > MAX_PREVIEW_HEIGHT is always false

The timing issue is that the useEffect runs before React has had a chance to:
1. Render the ScrollableContent div
2. Attach the contentRef to that div
3. Make contentRef.current point to the actual DOM element

The fix consists in using a callback ref instead of useRef. This ensures the ref callback is called exactly when the DOM element is ready.

After the fix:

fixed-expand-toggle


RomneyDa (approved):
Nice react state fix!

PR #7389, opened 2025-08-25
fix: render Ollama thinking messages

fbricon:
When calling gpt-oss:20b running on ollama, the thinking part is not displayed. We can see blank "assistant" messages in the Continue console. Turns out ollama puts the thinking part into a separate "thinking" field, but with the "assistant" role:

Screenshot 2025-08-25 at 16 07 17 

Small code change to read that property allows us to see the model reasoning:

https://github.com/user-attachments/assets/d3581d5e-873a-4f77-aa85-a13654e3df78

    

---

## Summary by cubic
Show Ollama model reasoning by handling the thinking field and emitting a thinking message, fixing blank assistant messages in the console.

- **Bug Fixes**
  - Parse res.message.thinking (when role is assistant) and return a ThinkingChatMessage.
  - Prevents blank assistant outputs for gpt-oss:20b and other Ollama models.

### DIFF
--- a/core/llm/llms/Ollama.ts
+++ b/core/llm/llms/Ollama.ts
@@ -9,6 +9,7 @@ import {
   CompletionOptions,
   LLMOptions,
   ModelInstaller,
+  ThinkingChatMessage,
 } from "../../index.js";
 import { renderChatMessage } from "../../util/messageContent.js";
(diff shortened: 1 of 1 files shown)


RomneyDa (approved):
Nice addition of thinking support for ollama

PR #7391, opened 2025-08-25
feat: create .envrc

Patrick-Erichsen:
## Summary by cubic
Add .envrc to prepend the repo’s scripts/ folder to PATH. This lets you run project scripts directly (e.g., my-script) when entering the directory with direnv.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+export PATH="$PWD/scripts:$PATH"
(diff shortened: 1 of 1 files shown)


metcalfc:
PATH_add scripts 

Is a safer way to do this. Direnv will handle adding and removing the directory as you enter or leave the project.

PR #7409, opened 2025-08-26
Fix: Updated ideSettings didn't get passed properly

Jazzcort:
## Description

The updating process for ConfigHandler works as expected, but some dependent objects under it don’t always reflect the latest updates. For example, both the globalLocalProfileManager and the controlPlaneClient continue using the original ideSettingsPromise created when the ConfigHandler was initialized.

When updateIdeSettings is called, the ideSettingsPromise inside ConfigHandler is updated, but these downstream entities don’t pick up the changes and still reference the outdated version.

To address this, instead of passing a direct reference to the Promise, we should pass a function that always resolves to the latest Promise. This ensures that both the globalLocalProfileManager and controlPlaneClient always retrieve the current ideSettings.

It’s worth noting that this isn’t an issue for PlatformProfileLoader, since it gets recreated during cascadeInit after every ideSettings update. However, the controlPlaneClient is only recreated when the sessionInfo changes—not when ideSettings changes—so this patch is still necessary to keep it up to date.

## Checklist

I make the doLoadConfig (core/config/profile/doLoadConfig.ts) function send the ideSettings it received whenever it gets called to the toast

### Before this patch

https://github.com/user-attachments/assets/6f945d84-2928-489d-8e95-58b057cb39cc

### After the patch

https://github.com/user-attachments/assets/4f0b91f1-031a-485c-9f94-c2a934d4ad98


Jazzcort:
Another solution would be just requesting the `ideSettings` from the `ide`, but I don't know if this would raise any performance concerns. However, we've made a few requests to `ide` inside the `doLoadConfig` function already. With this solution, we don't need to pass the `ideSettingsPromise` as a parameter anymore. Also, we don't need to pass the `ideSettings` to the `updateIdeSettings` function, it becomes more like a function to notify the `ConfigHandler`.

I'm fine with either directions. 😄

Patrick-Erichsen:
Thanks for the fix here @Jazzcort ! 

> However, we've made a few requests to ide inside the doLoadConfig function already. With this solution, we don't need to pass the `ideSettingsPromise` as a parameter anymore.

I think this would be cleaner, I'm not sure why we have the `ideSettingsPromise` if we're able to just request the latest settings directly from the `ide`. I doubt it's a performance concern.

cc @RomneyDa , any thoughts?

Jazzcort:
Thanks for the review @Patrick-Erichsen !

If we decide to get rid of the `ideSettingsPromise`, we might also want to do the same to the`sessionInfoPromise`. 😁

Let's wait for some feedback from @RomneyDa and we can decide which solution to go for. I also prefer requesting directly from `ide` because it's cleaner.

Patrick-Erichsen:
Would be great to remove `sessionInfoPromise` too! But yeah I have a hunch there's a reason we had to go that route, just not sure why.

Patrick-Erichsen:
Chatted with the team, seems like it should be safe to remove the promises here 👍

Jazzcort:
Sounds great! Will ping you when I'm done with updating this patch! 😁

Jazzcort:
@Patrick-Erichsen For the LocalProfileLoader, I'm thinking if we should pass the getter function for getting the most current ControlPlaneClient (current version) or we should just recreate the `globalLocalProfileManager` whenever we do `cascadeInit` for "Control plane session info update". The LocalProfileLoader owned by the `globalLocalProfileManager` is the only LocalProfileLoader that could have a stale ControlPlaneClient when the singn-in status changes. All the other LocalProfileLoader will get recreated with the new ControlPlaneClient during `cascadeInit.` Any preference? 😁

Patrick-Erichsen:
> The LocalProfileLoader owned by the globalLocalProfileManager is the only LocalProfileLoader that could have a stale ControlPlaneClient when the singn-in status changes. All the other LocalProfileLoader will get recreated with the new ControlPlaneClient during cascadeInit

If I'm understanding the tradeoffs correctly, if we recreate the `globalLocalProfileManager` everything would be a lot cleaner but we might have issues with a stale `LocalProfileLoader`?

Jazzcort:
>If I'm understanding the tradeoffs correctly, if we recreate the globalLocalProfileManager everything would be a lot cleaner but we might have issues with a stale LocalProfileLoader?
  
If we recreate the globalLocalProfileManager during every cascadeInit, it resolves the issue of holding onto a stale LocalProfileLoader. This approach also treats both PlatformProfileLoader and LocalProfileLoader consistently, instead of passing a getter function to one and a direct ControlPlaneClient reference to the other. I’m leaning toward this solution, where we make ControlPlaneClient an immutable property of both loaders. That way, whenever the sign-in status changes, we simply recreate all of them with the updated status.

@Patrick-Erichsen  If you also think this would be a better approach, I just need a small tweak on this patch and everything is good to go! 😁

Patrick-Erichsen:
> This approach also treats both PlatformProfileLoader and LocalProfileLoader consistently, instead of passing a getter function to one and a direct ControlPlaneClient reference to the other. I’m leaning toward this solution

@Jazzcort  this makes sense to me, let's do it. Thanks for the detailed writeup!

Jazzcort:
@Patrick-Erichsen Should be good to go!

RomneyDa (approved):
Talked to @Patrick-Erichsen, looks good. Also pulled and tested, works

PR #7431, opened 2025-08-27
Fix: runTerminalCommand respect command aliases

etherandrius:
## Description

Coming from Claude code the terminal tool feels a bit flake. The environment is too different from what I would expect in a lot of cases.

I think the biggest offender is not respecting aliases which a user may rely on. See minimal example of the change:
image

## Checklist

## Tests

Tested manually, the change is minimal. Everyone should have bash installed on their machine.

### DIFF
--- a/extensions/cli/src/tools/runTerminalCommand.ts
+++ b/extensions/cli/src/tools/runTerminalCommand.ts
@@ -40,7 +40,7 @@ export const runTerminalCommandTool: Tool = {
   },
   run: async ({ command }: { command: string }): Promise<string> => {
     return new Promise((resolve, reject) => {
-      const child = spawn("sh", ["-c", command]);
+      const child = spawn("bash", ["-i", "-c", command]);
       let stdout = "";
       let stderr = "";
       let timeoutId: NodeJS.Timeout;


Patrick-Erichsen (approved):
Thanks for the patch here! 

cc @chezsmithy for any thoughts

extensions/cli/src/tools/runTerminalCommand.ts
```diff
@@ -40,7 +40,7 @@ export const runTerminalCommandTool: Tool = {
   },
   run: async ({ command }: { command: string }): Promise<string> => {
     return new Promise((resolve, reject) => {
-      const child = spawn("sh", ["-c", command]);
+      const child = spawn("bash", ["-i", "-c", command]);
```
chezsmithy, line 43: I think we want -l here vs -i. Why don't I put up a PR that matches to what's in the current non-cli terminal and we can have a look.

Great question! The -i and -l flags for bash control different aspects of how the shell initializes:

-l (Login Shell)
Purpose: Treats the shell as a "login shell"
Configuration files read (in order):
/etc/profile (system-wide)
~/.bash_profile, ~/.bash_login, or ~/.profile (first one found)
~/.bash_logout (when exiting)
When used: Simulates logging into the system
Environment: Sets up full user environment (PATH, environment variables, etc.)
-i (Interactive Shell)
Purpose: Treats the shell as "interactive"
Configuration files read:
~/.bashrc (typically)
Features enabled:
Command line editing
Job control
PS1 prompt processing
History expansion
When used: For shells that read commands from a user
Key Differences
Aspect	(Login)	(Interactive)
Config files	,	
Purpose	Full environment setup	Interactive features
Typical use	SSH login, initial terminal	Subshells, new terminal tabs
PATH setup	✅ Full PATH from profile	❌ May not have full PATH

chezsmithy:
@etherandrius and @Patrick-Erichsen here is an even more comprehensive fix including updated tests for the CLI. This should resolve the issue @etherandrius kindly identfied by using -l (which includes $PATH) and it has fixes for platform specific tests for windows and mac that are failing on other branches. 

https://github.com/continuedev/continue/pull/7437

etherandrius:
Thanks @chezsmithy.

closing this out

PR #7446, opened 2025-08-28
fix: split thinking and chat messages when Ollama is not streaming

fbricon:
Following up on #7389, I noticed that Ollama, when not streaming, returns both thinking and chat content in the same message, which my previous code didn't handle, basically only showing thinking and no content  with the following settings:

```yaml
name: Local Assistant
version: 1.0.0
schema: v1
models:
  - name: gpt-oss:20b
    provider: ollama
    model: gpt-oss:20b
    defaultCompletionOptions: 
      stream: false
```
    

---

## Summary by cubic
Fixes Ollama non‑streaming responses by splitting thinking and assistant content into separate messages so chat content is no longer dropped. Also preserves tool calls and yields messages correctly in both streaming and non‑streaming modes.

- **Bug Fixes**
  - Split non‑streaming assistant responses into thinking and chat messages.
  - Preserve and emit tool_calls with generated IDs.
  - Update converters to return arrays and yield each message safely.
  - Guard against unexpected role=tool messages.

### DIFF
--- a/core/llm/llms/Ollama.ts
+++ b/core/llm/llms/Ollama.ts
@@ -422,31 +422,35 @@ class Ollama extends BaseLLM implements ModelInstaller {
       signal,
     });
 
-    function convertChatMessage(res: OllamaChatResponse): ChatMessage {
+    function convertChatMessage(res: OllamaChatResponse): ChatMessage[] {
       if ("error" in res) {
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen (approved):
Thanks for the follow up fixes!

PR #7454, opened 2025-08-28
Delete config_schema.json

Patrick-Erichsen:
`config.json` is deprecated so this can be removed.
    

---

## Summary by cubic
Removed extensions/vscode/config_schema.json in the VS Code extension. config.json is deprecated and no longer used, so the schema file is unnecessary.


RomneyDa (commented):
if the goal is to reduce developer time could we e.g. hide it from vs code search results instead?

Note will need to update build scripts etc.

Issue #7121, opened 2025-08-13
java.io.IOException: Stream closed after upgrading Jetbrains IDE.


BvdMerwe:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: Jetbrains 1.0.33
- IDE version: PhpStorm Build #PS-252.23892.419
- Model: 
- config:
  
name: Continue Config
version: 0.0.1
models:
  - name: Anthropic
    provider: anthropic
    model: claude-4-sonnet-20250514
    apiKey: XXX
    roles:
      - chat
      - edit
      - apply
  - name: Finn mini
    provider: openai
    model: XXX
    apiKey: XXX
    apiBase: XXX
    roles:
      - chat
      - edit
      - apply
  - name: Finn 4o
    provider: openai
    model: XXX
    apiKey: XXX
    apiBase: XXX
    roles:
      - chat
      - edit
  - name: Finn/Claude
    provider: anthropic
    model: anthropic.claude-3-5-sonnet-20240620-v1:0
    apiKey: XXX
    apiBase: XXX
    roles:
      - chat
      - edit
  - name: Finn/Gemini
    provider: openai
    model: XXX
    apiKey: XXX
    apiBase: XXX
    roles:
      - chat
      - edit
context:
  - uses: continuedev/folder-context
  - uses: continuedev/docs-context
```

### Description

When starting up the IDE, continue immediately throws an internal IDE error:

```
Unhandled exception in Dispatchers.IO

java.io.IOException: Stream closed
	at java.base/java.lang.ProcessBuilder$NullOutputStream.write(ProcessBuilder.java:447)
	at java.base/java.io.OutputStream.write(OutputStream.java:167)
	at java.base/java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:125)
	at java.base/java.io.BufferedOutputStream.implFlush(BufferedOutputStream.java:252)
[...]

Issue #7018, opened 2025-08-06
Continue plugin for PyCharm fails to load correctly


djpetti:
### Before submitting your bug report

- [ ] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Ubuntu 24
- Continue version: 1.0.30
- IDE version: PY-252.23892.439
- Model: Codestral-202501
- config:
  
name: Local Assistant
version: 1.0.0
schema: v1
models:
  - name: Gemma 12b
    provider: ollama
    model: gemma3:12b
    apiBase: http://bsail-jetson.ad.ufl.edu:11434
    roles:
      - chat
  - name: Codestral
    provider: openai
    apiKey: <redacted>
    model: mistralai/codestral-2501
    apiBase: https://openrouter.ai/api/v1
    roles:
      - edit
      - apply
      - autocomplete
    defaultCompletionOptions:
      contextLength: 262144
  - name: All-MiniLM
    provider: ollama
    model: all-minilm:latest
    apiBase: http://bsail-jetson.ad.ufl.edu:11434
    roles:
      - embed

context:
  - provider: code
  - provider: docs
  - provider: diff
  - provider: terminal
  - provider: problems
  - provider: folder
  - provider: codebase
  
  OR link to assistant in Continue hub:
```

### Description

Upon opening the IDE after installing the plugin, I get this error:

```
Unhandled exception in Dispatchers.IO

java.io.IOException: Stream closed
	at java.base/java.lang.ProcessBuilder$NullOutputStream.write(ProcessBuilder.java:447)
	at java.base/java.io.OutputStream.write(OutputStream.java:167)
	at java.base/java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:125)
	at java.base/java.io.BufferedOutputStream.implFlush(BufferedOutputStream.java:252)
	at java.base/java.io.BufferedOutputStream.flush(BufferedOutputStream.java:246)
[...]

Issue #5505, opened 2025-05-04
IDEA java.io.IOException: Stream closed


mosinnik:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: win 10
- Continue version: 1.0.15
- IDE version: 2025.3.5
- Model:
- config:
  

  
  OR link to assistant in Continue hub:
```

### Description

absolutely uninformative logs from plugin. Add full exception printing everywhere it possible to understand cause of error

### Log output

```Shell
2025-05-04 16:22:49,086 [5729339]   INFO - STDOUT - Error writing to Continue core: java.io.IOException: Stream closed
2025-05-04 16:22:49,086 [5729339]   INFO - STDOUT - Error writing to Continue core: java.io.IOException: Stream closed
2025-05-04 16:22:49,086 [5729339]   INFO - STDOUT - Error writing to Continue core: java.io.IOException: Stream closed
2025-05-04 16:22:49,086 [5729339]   INFO - STDOUT - Error writing to Continue core: java.io.IOException: Stream closed
2025-05-04 16:22:49,086 [5729339]   INFO - STDOUT - Error writing to Continue core: java.io.IOException: Stream closed
2025-05-04 16:22:49,086 [5729339]   INFO - STDOUT - Error writing to Continue core: java.io.IOException: Stream closed
2025-05-04 16:22:49,086 [5729339]   INFO - STDOUT - Error writing to Continue core: java.io.IOException: Stream closed
```

tomasz-stefaniak:
@mosinnik when does this issue occur? Is it still reproducible on the latest pre-release version of the JetBrains extension?

mosinnik:
@tomasz-stefaniak I don't know why it happen because of no additional info in logs. So it just happen sometimes.
Add printing stacktraces everywhere

eilensm:
I get this error when I don't use my laptop for some hours and reopen it.

Issue #3981, opened 2025-02-05
Continue for Pycharm floods logs with writing to Continue core error


Jonnokc:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [ ] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: 15.1.1 (24B91)
- Continue version: 0.0.83
- IDE version:2024.3.2
- Model: N/A
- config.json:
  
{
  "models": [
    {
      "model": "claude-3-5-sonnet-20240620",
      "provider": "anthropic",
      "systemMessage": "You are an expert software developer. You give helpful, accurate, concise, and high quality responses.",
      "apiKey": "xxxx",
      "title": "Claude 3.5 Sonnet"
    },
    {
      "model": "gpt-4o",
      "title": "GPT-4o",
      "systemMessage": "You are an expert software developer. You give helpful, accurate, concise, and high quality responses.",
      "apiKey": "xxxx",
      "provider": "openai"
    },
    {
      "model": "gpt-4o-mini",
      "title": "GPT-4o Mini",
      "systemMessage": "You are an expert software developer. You give helpful, accurate, concise, and high quality responses.",
      "apiKey": "xxxx",
      "provider": "openai"
    }
  ],
  "contextProviders": [
    {
      "name": "code",
      "params": {}
    },
    {
      "name": "docs",
      "params": {}
    },
    {
      "name": "diff",
      "params": {}
    },
    {
      "name": "terminal",
      "params": {}
    },
    {
      "name": "problems",
      "params": {}
    },
    {
      "name": "folder",
      "params": {}
    },
    {
      "name": "codebase",
      "params": {}
    }
  ],
  "slashCommands": [
    {
      "name": "edit",
      "description": "Edit selected code"
    },
    {
      "name": "comment",
      "description": "Write comments for the selected code"
    },
    {
      "name": "share",
      "description": "Export the current chat session to markdown"
    },
    {
      "name": "cmd",
      "description": "Generate a shell command"
    },
    {
      "name": "commit",
      "description": "Generate a git commit message"
    }
  ]
}
```

### Description

diagnostic logs are flooded with records like the following:

```
[...]

PR #7463, opened 2025-08-28
fix: recover Continue process after unexpected exceptions

exigow:
Solves CON-3587
Solves #7018
Solves #7121
Solves #5505
Solves #3981, and possibly other `Stream closed`-like errors

## Main change

Now, if some exception bubbles up at any stage (whether it's read, write, anything else within `processScope`), we catch it via `CoroutineExceptionHandler` and we restart the process + back-off. Type of exception is not important / now it's more like black-box approach.

This simplifies things quite a bit, since we now have only single piece of code for process recovery.

## Detailed changes

* The process now recovers from exceptions globally, without specifying the type of exception (more of a black-box approach).
* Remove `CoreMessengerManager`, since process recovery is now implemented in `ContinueProcessHandler`.
* Add a back-off utility for calculating delays between process recovery retries (just a cleanup).
* Restore `ContinuePluginService.dispose()` (because `Disposable` was accidentally removed during the last changes in the Next Edit - that's not critical, but it was causing exceptions when closing the application/during plugin disposal).
* Remove the artificial delay between reading messages (from now on, blocking happens directly on `readLine`).

## Manual testing

* I didn't notice any regressions
* The `Restart Process` action works correctly, and the binary restarts immediately
* Killing the process externally causes the binary to restart
* Closing the IDE does not orphan the process


exigow:
@Patrick-Erichsen, :green_circle: + WIP commits squashed

Patrick-Erichsen (approved):
Great cleanup here! The backoff calculator is a nice touch.

PR #7467, opened 2025-08-28
feat: added UI for global rule creation 

panyamkeerthana:
## Description

Add a rule scope dropdown in the RulesSection that lets users choose between creating workspace rules or global rules. This sets up the base for managing rules that work across all projects, with updates in the backend and protocol to support global rule creation.

- added a rule scope dropdown in RulesSection with isGlobalMode state.
- added backend function to create global rule files and reload config.
- extended protocol and core handler (config/addGlobalRule) to support global rule creation.
- added info hover component
- generated unique rule names for global rules

https://github.com/user-attachments/assets/7da71a36-78d1-48c5-95d4-e545ba8967b9

    

---

## Summary by cubic
Adds a toggle in RulesSection to choose between creating workspace rules or global rules, enabling global rule files to be created directly from the UI. Connects the UI to a new core handler to write the rule in the global directory and reload config.

- **New Features**
  - UI toggle: “Create global rules” in RulesSection; ExploreBlocksButton routes to global rule creation when enabled.
  - Protocol/Core: Added config/addGlobalRule message; createNewGlobalRuleFile writes to the global path, opens the file, and triggers config reload.
  - Pass-through: Webview passthrough updated to allow the new message.


Patrick-Erichsen:
@prakashsharma91 mind sharing a screenshot of what this looks like with the new rules section in settings?

panyamkeerthana:
https://github.com/user-attachments/assets/8bdf9cf9-ce91-4b88-8222-18d9dbe4f97d

Patrick-Erichsen:
Nice ! I could see this being useful for all of the "new block" buttons on the settings page. 

My only nitpick is that I think the select should be incorporated into the "Plus" button itself. We have a button similar to what I'm describing on the hub:

Screenshot 2025-09-19 at 12 53 04 PM

panyamkeerthana:
@Patrick-Erichsen I added a new `DropdownButton.tsx` component that can be reusable for all of the new block buttons on the settings page. Here's what it looks like in the Rules section:

Screenshot 2025-09-22 at 15 28 47

Patrick-Erichsen (changes requested):
Thanks for implementing the new button! A few nitpicks/polish requests

- The border color/thickness is not the same for the other "add" buttons in settings, it's thinner/lighter
- Nitpick but I prefer just "Add rule" for the tooltip to keep things simple/consistent
- This advanced button should only be visible when the user has "Local agent" select. For hub agents the button doesn't modify anything on your local filesystem, it just opens a link to the hub
- Instead of "Globally" we should say "Global"
- There is a bug with the filepath logic, I clicked "Current workspace" and it correctly added it to the `.continue` in my workspace, but then I clicked "Globally" it still added to current workspace. Eventually it sometimes does add globally but it's flaky.
- I think the hover effect should be the same as the other buttons, eg a ghost button, rather than the orange fill I'm getting here
Screenshot 2025-09-22 at 2 19 31 PM

Patrick-Erichsen:
Looks great, thank you again for the attention to detail here! Excited to get this one merged 🚀

Patrick-Erichsen (approved):

PR #7483, opened 2025-08-29
Feat: Untangle NES and autocomplete logic

Jazzcort:
## Description

This commit refactors the ContinueCompletionProvider to separate the logic for Next Edit Suggestions (NES) from the standard inline autocomplete functionality.

Previously, the two features were intertwined, leading to complex conditional logic. This was particularly problematic when interacting with VS Code's native completion widget (selectedCompletionInfo), where changing the completion range is restricted.

Key changes include:
  - The main provideInlineCompletionItems method is now split into two distinct paths for NES and standard autocomplete.

  - A mayChangeRange parameter was added to processSingleLineCompletion to prevent replacing the rest of a line when the VS Code completion widget is visible.

  - The willDisplay function was split into willDisplayForNES and willDisplayForAutocomplete to handle the distinct validation requirements of each path.

## Checklist

---

## Summary by cubic
Separated Next Edit Suggestions (NES) from standard inline autocomplete in the VS Code provider to reduce complexity and fix conflicts with the native completion widget. This makes suggestions more reliable and predictable.

- **Refactors**
  - Split provideInlineCompletionItems into distinct NES and autocomplete paths.
  - Added mayChangeRange to processSingleLineCompletion to respect widget constraints.
  - Replaced willDisplay with willDisplayForNES and willDisplayForAutocomplete.

- **Bug Fixes**
  - Prevents unintended range replacement when the VS Code completion widget is visible.
  - Ensures inline completions extend selectedCompletionInfo and use its range when required.

### DIFF
--- a/core/autocomplete/util/processSingleLineCompletion.ts
+++ b/core/autocomplete/util/processSingleLineCompletion.ts
@@ -39,29 +39,26 @@ function diffPatternMatches(
 type DiffPartType = "+" | "-" | "=";
 
 export function processSingleLineCompletion(
-  lastLineOfCompletionText: string,
+  completionText: string,
   currentText: string,
   cursorPosition: number,
+  mayChangeRange: boolean,
 ): SingleLineCompletionResult | undefined {
-  const diffs: DiffType[] = Diff.diffWords(
-    currentText,
-    lastLineOfCompletionText,
-  );
+  const diffs: DiffType[] = Diff.diffWords(currentText, completionText);
 
   if (diffPatternMatches(diffs, ["+"])) {
     // Just insert, we're already at the end of the line
     return {
-      completionText: lastLineOfCompletionText,
+      completionText,
     };
   }
 
-  if (
-    diffPatternMatches(diffs, ["+", "="]) ||
-    diffPatternMatches(diffs, ["+", "=", "+"])
-  ) {
-    // The model repeated the text after the cursor to the end of the line
+  if (mayChangeRange && diffPatternMatches(diffs, ["+", "=", "+"])) {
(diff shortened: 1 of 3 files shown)


Jazzcort:
@jpoly1219 Our main idea is to separate NES and autocomplete logic, so the code looks cleaner and we can customize different handling process for each logic. I'm not sure if NES still needs the completion text to be preprocessed by `processSingleLineCompletion` and the modified range. From my observation, NES does not use the modified range at all. Let me know what do you think about this. 😁

jpoly1219:
@Jazzcort thanks for the PR! I just realized that you sent me an email as well. Thank you for your patience :')

If I'm not mistaken, we need `processSingleLineCompletions` for next edit as well. There are cases where the model returns a single line, inline completion that has to be rendered properly with regards to the surrounding code. Your changes also contains changes that could potentially impact stability regarding race conditions between events and properly handling selection changes/cursor movements (deleting GhostTectAcceptanceTracker could induce lots of interesting runtime bugs).

I want to hold off on merging changes to next edit until I finish documentation and get the rest of the team be confident on how next edit works. This actually happens to be my last week at Continue, so I'm currently more focused on getting the team up to speed with existing next edit logic. I'm currently in the process of adding a ton of comments in the codebase so that other people can understand how the system works. We also need proper e2e tests before we start accepting contributor changes with confidence.

Your point is sound -- we feel the same pain and are working to refactor this internally :') I would greatly appreciate it if you could hold your thoughts until we are ready to accept contributor PRs. If you are interested, we would greatly appreciate your help in writing e2e tests for next edit! We'd be very happy to merge your PR if you could contribute an e2e test to confirm that next edit works as expected.

PR #7508, opened 2025-09-01
feat: enable thinking/reasoning toggle for ollama models

fbricon:
This PR enables users to toggle thinking for Ollama models that support it. In this recording, you can see qwen3 reasoning is not enabled by default, but setting reasoning : true in the defaultCompletionOptions allows it:
```yaml
  - name: qwen3:8b
    provider: ollama
    model: qwen3:8b
    defaultCompletionOptions:
      reasoning: true
  ```
 

https://github.com/user-attachments/assets/c0dba04b-2b88-4ccf-8da4-c4c15b783fa2

- shows reasoning toggle if model supports reasoning or config explicitly forces it
- reasoning session state properly resets when changing models
- reasoning status passed as think:true to ollama
- updated docs to mention support of reasoning configuration for ollama models

Hardcoding reasoning capabilities on ollama is tricky: e.g. qwen3:8b has reasoning capabilities but think=true causes an Ollama error if the version pulled into ollama is too old. Pulling a newer version will fix it.

Also worth noting some models, like gpt-oss:20b will just ignore the think:false flag passed to ollama. It supports different levels of thinking, but needs to be set in the system prompt. See https://huggingface.co/openai/gpt-oss-120b#reasoning-levels

### CODE REVIEW

**Comment by fbricon** [comment]
I need to fix those gui tests

File: core/llm/autodetect.ts [hunk 1]
```diff
@@ -126,6 +131,26 @@ function modelSupportsImages(
...
+
+function modelSupportsThinking(
+  model: ModelDescription | null | undefined,
+): boolean {
+  if (!model) {
+    return false;
+  }
+  if ("anthropic" === model.underlyingProviderName) {
+    return true;
+  }
```
Patrick-Erichsen, line 143: I believe older models such as 3.5 Sonnet don't support reasoning? 

Although I'm a little confused, this file is just for Ollama autodetect, no?

core/llm/llms/Ollama.ts
```diff
@@ -87,6 +87,7 @@ interface OllamaChatOptions extends OllamaBaseOptions {
   tools?: OllamaTool[]; // the tools of the chat, this can be used to keep a tool memory
   // Not supported yet - tools: tools for the model to use if supported. Requires stream to be set to false
   // And correspondingly, tool calls in OllamaChatMessage
+  think?: boolean; // if true the model will be prompted to think about the response before generating it
```
Patrick-Erichsen, line 90: Should this be `reasoning` to match other configs?

gui/src/redux/thunks/streamNormalInput.ts
```diff
@@ -121,12 +121,17 @@ export const streamNormalInput = createAsyncThunk<
...
-        reasoningBudgetTokens:
-          selectedChatModel.completionOptions?.reasoningBudgetTokens ?? 2048,
+        reasoning: !!state.session.hasReasoningEnabled,
+        ...(state.session.hasReasoningEnabled &&
+          selectedChatModel.underlyingProviderName !== "ollama" && {
+            // Ollama doesn't support limiting reasoning tokens at this point
+            reasoningBudgetTokens:
+              selectedChatModel.completionOptions?.reasoningBudgetTokens ??
+              2048,
+          }),
```
Patrick-Erichsen: This logic is a bit hard to follow, could we pull it out into a helper fn?

Patrick-Erichsen (changes requested):
Just a few questions/nitpicks, thanks for the contribution @fbricon ! Appreciate the detailed description and screen recording.

core/llm/autodetect.ts
fbricon: I just moved that existing check from [gui/src/components/mainInput/InputToolbar.tsx](https://github.com/continuedev/continue/blob/f0bc32a6d663a1c38b3f6c184f767b7acd2ba8d2/gui/src/components/mainInput/InputToolbar.tsx#L142)

And no, this file seems to handle different models from different providers

core/llm/llms/Ollama.ts
fbricon: No,  It's the expected parameter by the Ollama API, i.e. what will be put in the http request, so we must keep `think`

fbricon: I've replaced "thinking" with "reasoning" in other places, to be more consistent

gui/src/redux/thunks/streamNormalInput.ts
fbricon: done

Patrick-Erichsen (approved):
Thanks for making those updates! Lmk if this is good to merge and I'll do so.

fbricon:
I can't test the Anthropic models so if you can do that and I didn't break anything, then go ahead and merge it

RomneyDa:
@fbricon have verified this works with e.g. claude 4.1 opus local. Appreciate the contribution!

PR #7518, opened 2025-09-02
fix: stop reading on EOF

exigow:
Before this change, if the process returned EOF (if the line was `null` - that's how `BufferedReader` works), the loop kept trying to read forever.


exigow:
(@Patrick-Erichsen, FYI this bugfix quite a serious one, but luckily, I'm actively dogfooding the main branch, so I was able to spot this quickly enough)

exigow:
Change of plans

Issue #7519, opened 2025-09-02
remote config sync broken in Jetbrains


pczekaj:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: MacOS
- Continue version: 1.0.33
- IDE version: IntelliJ 2025.2.1
```

### Description

Continue on Jetbrains started to send `POST` requests to `sync` endpoint while Visual Studio Code still sends `GET`
probably due to change introduced in:
https://github.com/continuedev/continue/commit/60aa531650bed4f8393c6a4244e6446f4ca6e6d2

Also even if request hits some backend then continue is has problems with deserialization:

```
Serializer for class 'ContinueRemoteConfigSyncResponse' is not found.
Please ensure that class is marked as '@Serializable' and that the serialization compiler plugin is applied.
```

### To reproduce

point remote config in continue dev config to for example ngrok, on Jetbrains you will see incoming POST request while on VS Code it will be GET

### Log output

```Shell
2025-09-02 15:54:16,587 [ 393043] SEVERE - #c.i.u.c.SchedulingWrapper - Serializer for class 'ContinueRemoteConfigSyncResponse' is not found.
Please ensure that class is marked as '@Serializable' and that the serialization compiler plugin is applied.

kotlinx.serialization.SerializationException: Serializer for class 'ContinueRemoteConfigSyncResponse' is not found.
Please ensure that class is marked as '@Serializable' and that the serialization compiler plugin is applied.

	at kotlinx.serialization.internal.Platform_commonKt.serializerNotRegistered(Platform.common.kt:92)
	at kotlinx.serialization.SerializersKt__SerializersKt.noCompiledSerializer(Serializers.kt:294)
	at kotlinx.serialization.SerializersKt.noCompiledSerializer(Unknown Source)
	at com.github.continuedev.continueintellijextension.services.ContinueExtensionSettings.syncRemoteConfig(ContinueExtensionSettingsService.kt:253)
[...]

PR #7525, opened 2025-09-02
fix: fix settings sync

exigow:
Solves #7519

Changes:
* Use GET.
* Do not rely on `kotlinx.serialization` because we don't have a plugin for it (it was single usage, so there's no point in adding it). Use `gson` like everywhere else. 

**Note, why this bug appeared:** some time ago we removed `kotlinx.serialization` dependency, but this deserialization problem only appears at runtime (+ we didn't have automated tests for it), so it was overlooked.


exigow:
Thanks @pczekaj for spotting this!

Issue #6461, opened 2025-07-04
Continue should limit number of docs being indexed at one time


shssoichiro:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Currently Continue will attempt to index all docs at once. If the user has a large number of docs (e.g. 15+), then attempting to index them all at once will choke the model provider, leading to individual embed responses taking 4+ minutes and the indexing never completing.

### Solution

Continue should only allow X number of docs to index at one time (somewhere between 3-5 seems sane), and queue the rest to start once others have finished.

Jazzcort:
@shssoichiro @RomneyDa I've been working on making the chunks embedding process concurrent to boost the docs embedding speed which is sort of related to this enhancement. I'm very interested in working on implementing a queuing system for this as well! 😄

shssoichiro:
Can be closed due to indexing being deprecated

PR #7557, opened 2025-09-03
Feat: Limit concurrent indexing to prevent resource exhaustion

Jazzcort:
## Description

This commit addresses a performance bug where indexing multiple documentation sources simultaneously would overload the backend inference service. The previous implementation fired all indexing tasks at once, leading to resource contention that choked the embedding provider and drastically slowed down individual tasks.

To resolve this, an EmbeddingManager has been introduced to manage and throttle the indexing workload:
  - Task Queuing: The manager places all new indexing requests into a queue instead of executing them immediately.

  - Concurrency Control: It processes a limited number of jobs from the queue in parallel (defaulting to 3), ensuring the backend is not overwhelmed.

  - User Configuration: As part of this fix, a new setting has been added to the UI, allowing users to configure the maximum number of concurrent processes.

This change leads to a more stable and efficient indexing process, improving overall performance and reliability.

## Checklist

https://github.com/user-attachments/assets/32378b57-964c-41ce-8136-a5326e2c018b

## Tests

You can test this feature with the UI panel.

    

---

## Summary by cubic
Limits concurrent documentation indexing to prevent backend overload by queuing tasks and capping parallel jobs (default 3). Adds a UI control to configure the max concurrency for more stable and predictable performance.

- **New Features**
  - Queue-based indexing with a concurrency cap (default 3) and cancel-aware progress updates.
  - UI control to adjust the limit; value persisted via GlobalContext.
  - Core/webview/IntelliJ protocol to get/update the max concurrency.

- **Refactors**
  - Moved crawl/chunk/embed workflow from DocsService into EmbeddingManager; DocsService now delegates and triggers processing.
  - Prevents duplicate indexing per startUrl and handles aborts cleanly.

### DIFF
--- a/core/indexing/docs/DocsService.ts
+++ b/core/indexing/docs/DocsService.ts
@@ -26,17 +26,13 @@ import {
   getLanceDbPath,
 } from "../../util/paths";
 import { Telemetry } from "../../util/posthog";
+import DocsCrawler from "./crawlers/DocsCrawler";
 
-import {
-  ArticleWithChunks,
-  htmlPageToArticleWithChunks,
-  markdownPageToArticleWithChunks,
-} from "./article";
-import DocsCrawler, { DocsCrawlerType, PageData } from "./crawlers/DocsCrawler";
 import { runLanceMigrations, runSqliteMigrations } from "./migrations";
 
 import type * as LanceType from "vectordb";
 import { LLMError } from "../../llm";
+import { EmbeddingManager } from "./EmbeddingManager";
 
 // Purposefully lowercase because lancedb converts
 export interface LanceDbDocsRow {
@@ -177,6 +173,7 @@ export default class DocsService {
 
   private docsIndexingQueue = new Set<string>();
   private lanceTableNamesSet = new Set<string>();
+  private embeddingManager = new EmbeddingManager();
 
   private config!: ContinueConfig;
   private sqliteDb?: Database;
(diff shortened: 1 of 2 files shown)


Jazzcort:
This feature addresses https://github.com/continuedev/continue/issues/6461

gui/src/pages/config/IndexingSettingsSection.tsx
```diff
@@ -1,9 +1,53 @@
...
 import { useAppSelector } from "../../redux/hooks";
 import IndexingProgress from "./IndexingProgress";
 
 export function IndexingSettingsSection() {
   const config = useAppSelector((state) => state.config.config);
+  const ideMessenger = useContext(IdeMessengerContext);
+  const [numConcurrent, setNumConcurrent] = useState(3);
+
+  const handleUpdate = async (num: number) => {
+    await ideMessenger.request("docs/updateMaxNumberOfConcurrentIndexing", num);
```
RomneyDa: Add optimistic update to prevent UI lag
Preferably change to use existing sharedConfig approach

RomneyDa (changes requested):
@Jazzcort main feedback is that the approach can be simplified by using the existing sharedConfig infrastructure for this setting - no need for additional message types - and then it's already on config which docs manager already references and which is readily available throughout the GUI with a robust updating system

>const toWait = 100 * this.docsIndexingQueue.size + 50;

This line in the existing code should already limits concurrency somewhat based on the number of docs in the queue. I think the mixing of this existing delay with the new setting could cause confusing results. Adjusting this logic could be a simpler albeit nonconfigurable way to solve the problem, or maybe provide a quick analysis of why this should be configurable on a user setting level as opposed to hardcoded, on a model level, etc.?

Note that the @docs context provider is deprecated and we're limiting contributions to it. This PR fixes a known issue with massive concurrency for embeddings so would be a good one to get in.

Jazzcort:
> Adjusting this logic could be a simpler albeit nonconfigurable way to solve the problem, or maybe provide a quick analysis of why this should be configurable on a user setting level as opposed to hardcoded, on a model level, etc.?

After some thinking, hardcoding the concurrent number might be better because this feature just adds a prioritizing mechanism to the docs indexing, the overall time to index all docs should still stay the same. Increasing the concurrent number is actually reduce the effect that this feature adds on. @RomneyDa If you are also leaning to this way, I can just remove the configurable feature. 

> const toWait = 100 * this.docsIndexingQueue.size + 50;

I think it's not necessary to make waiting pause proportional to the queue size, but it does need a pause so it won't freeze the UI. (I tried it with no waiting time, and the UI had some unpredictable delay while clicking)

Jazzcort:
@RomneyDa I removed the configurable feature and hardcoded the concurrent number to 3. Let me know if there are any changes needed. Feel free to drop this pr if you guys think there is no need to add in patch for deprecated feature. 😁

RomneyDa:
@Jazzcort thanks for approaching this with depth. Since the feature is deprecated I think we will close this one. I'd definitely be open to merging something that addresses the same concurrency issue with a few lines of code somehow, might be tricky to whittle it down.

PR #7577, opened 2025-09-04
fix: dispose JCEF JSQuery handler correctly

exigow:
The JCEF browser wasn’t registered as a Disposable when held by the service. Service was disposed, but async JSQuery handler tried to run some code on a disposed browser (browsers are disposed automatically by default according to docs), throwing exceptions when opening another project.

This fix registers both the browser and the handler as disposable. I tested this change manually, and opening project after project (with the same IDE) is no longer causing exceptions.

Side note: I think it would be useful to have an e2e test at some point that _"opens a project, closes it, and then opens another one"_ to trigger that mechanism. Such a test should catch this issue early.


exigow:
Rebased with main to fix a failing CI (compilation error).

PR #7589, opened 2025-09-05
chore: disable fail-fast for cli workflow

tingwai:
There's some flakey tests in cli test suite, let's not fail fast so one failed matrix won't immediately stop all other running matrixes.[

[`fail-fast` documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategyfail-fast)

[Example of flakey test](https://github.com/continuedev/continue/actions/runs/17478278546/job/49643268150?pr=7587) which causes all other running tests to stop. Causes a painful cycle of re-running tests

### DIFF
--- a/.github/workflows/cli-pr-checks.yml
+++ b/.github/workflows/cli-pr-checks.yml
@@ -59,6 +59,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         node-version: [18, 20, 22, 24]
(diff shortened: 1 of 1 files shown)


tingwai:
[Flakey test strikes again](https://github.com/continuedev/continue/actions/runs/17479860511/job/49648032021?pr=7589) on this PR build, but `fail-fast` flag is working now

PR #7617, opened 2025-09-06
Add per-tab model persistence and selection functionality

ferenci84:
## Description

This PR adds per-tab model persistence and selection functionality. The changes enable each tab to remember and restore its selected AI model when switching between tabs. This improves the user experience by allowing users to maintain different model selections for different conversations.

Key changes:
- Added modelTitle property to Tab interface to store the selected model per tab
- Implemented setTabModel action in tabsSlice to update a tab's model title
- Enhanced tab creation to include the currently selected model
- Added functionality to restore the model when switching to a tab

## AI Code Review

## Checklist

## Tests

---

## Summary by cubic
Remember and restore the selected model per tab so each conversation can use its own model. Switching tabs now updates the global selection to match the tab’s saved model.

- **New Features**
  - Added modelTitle to Tab and a setTabModel reducer to store the model per tab.
  - New tabs inherit the current model; switching tabs restores that tab’s model via updateSelectedModelByRole.
  - Persist model changes into the active tab with a TabBar effect.


RomneyDa (changes requested):
@ferenci84 I think this would be a great feature for any sessions, not just through tabs. What if we implemented it per session rather than per tab, so that users not using tabs could have the same benefits?

sestinj:
I think I agree with dallin here. keeping sessions as the primary source of state simplifies a lot

RomneyDa:
@ferenci84 I started a discussion to gain some clarity on how to implement this
https://github.com/continuedev/continue/discussions/7986

I think any per-session implementation that isn't tabs-only will differe enough from this PR that we may as well start fresh. I've referenced this PR in the discussion and will close this for now.

PR #7644, opened 2025-09-09
instance id to fix metrics

sestinj:
## Description

instance id to fix metrics. before this the metrics from different cn processes could overwrite each other


extensions/cli/src/telemetry/telemetryService.ts
```diff
@@ -567,6 +580,16 @@ class TelemetryService {
     return this.config.sessionId;
   }
 
+  public recordSlashCommand(commandName: string) {
+    if (!this.isEnabled()) return;
+
+    console.log("ADD: ", commandName);
```
RomneyDa: @sestinj remove this?

PR #7675, opened 2025-09-10
fix: HOTFIX diff line color overflow

RomneyDa:
## Description
Allows diff colors in find replace display to fill width

Before:

image

After:

image

    

---

## Summary by cubic
Fixes diff highlight width in Find & Replace so added/removed colors span the full line and don’t overflow in the scroll container. Applies w-fit with min-w-full to the pre element to align backgrounds with the container.


uinstinct:
works awesome! cleaner solution also.

Issue #6107, opened 2025-06-13
Can't reference local block inside ~/.continue folder using relative filepath


TelpeNight:
### Before submitting your bug report

- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: macOS
- Continue version: 1.0.21
- IDE version: JetBrains
```

### Description

This bug should be easy to fix.

I've tried to reference local block as:

```yaml
docs:
  - uses: ./blocks/docs/new-doc.yaml
```

But it fails with message `Assistant includes blocks that don't exist`.

After playing around it, reading logs and source code I've discovered the following:

```
Failed to unroll block ./blocks/docs/new-doc.yaml: ENOENT: no such file or directory, open 'Users/name/.continue/blocks/docs/new-doc.yaml' 
```

This file exists and full path to it will work. But notice that abs file path misses leading slash. 

I think that the issue is in the following code:

https://github.com/continuedev/continue/blob/00911c6055f8c23f08bcb12065ed2c5b04bc7209/core/config/yaml/loadYaml.ts#L113

Note, that `getCleanUriPath` with strip leading slash from `packageIdentifier.filePath`. Hope this will help. Can't wait to use relative paths for better organized workflow)

### Log output

```Shell
Failed to unroll block ./blocks/docs/new-doc.yaml: ENOENT: no such file or directory, open 'Users/name/.continue/blocks/docs/new-doc.yaml'
```

jtabox:
Have you had any luck with this issue?

I've been trying to setup local model files and other blocks but it's virtually an unusable functionality as nothing seems to work. I'm getting the same annoying message, that `Assistant includes blocks that don't exist`.

I'm just trying to get the `config.yaml`

```yaml
# ~/.continue/models/config.yaml :
models:
  - uses: gemini-2-5-pro
    with:
      GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
```

to read a local block:

```yaml
# ~/.continue/models/gemini-2-5-pro.yaml :
name: gemini-2-5-pro
version: 1.0.4
schema: v1
models:
  - name: gemini-2-5-pro
[...]

PR #7686, opened 2025-09-10
feat: support `uses` blocks locally

### LINKED ISSUE #5943: config.yaml: `uses:` with relative path doesn't work

- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Linux (and others?)
- Continue version:git main and 1.1.41
```

### Description

Relative urls to blocks to include with `uses:` don't work, because it's trying to resolve against an absolute path to config.yaml, but stripping off the leading /.

As I read it, the call to getCleanUriPath() in:

https://github.com/continuedev/continue/blob/ceb2e5256032516919ae8f49b26bee2b8abdb194/core/config/yaml/loadYaml.ts#L113

is unnecessary and harmful.

### To reproduce

A 'uses:' statement with a relative path

``` yaml
[...]
models:
  - uses: ./include/myModel.yaml
```

Gives "Assistant includes blocks that don't exist" - debugging shows that it is trying to load `var/home/otaylor/.continue/include/myModel.yaml` with the missing initial `/`. An explicit:

``` yaml
[...]
models:
  - uses: file:///var/home/otaylor/.continue/include/myModel.yaml
```

works fine.

### Log output

```Shell

```

uinstinct:
## Description

Add support for having local `uses` block in config.yaml defined with `file://path/to/file`

resolves CON-3869
closes https://github.com/continuedev/continue/issues/5943
closes https://github.com/continuedev/continue/issues/6107

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/5ca5cb48-0a57-486f-802d-040e40516f49

## Tests

---

## Summary by cubic
Support local file includes in config.yaml via file:// uses blocks, so assistants can load models, MCP servers, prompts, and rules from nearby files. Addresses CON-3869; closes #5943 and #6107.

- New Features
  - Unrolls file:// uses in models, mcpServers, prompts, and rules during config load, including overrideConfigYaml.
  - Resolves local URIs with localPathOrUriToPath and derives registry root from the config file directory.
  - Adds a test that loads a rule from a local block to verify end-to-end resolution.

### DIFF
--- a/core/config/yaml/loadYaml.ts
+++ b/core/config/yaml/loadYaml.ts
@@ -104,7 +104,8 @@ async function loadConfigYaml(options: {
       prompts?: any[];
       rules?: any[];
     };
-    const maybeHasUsesBlocks = (
+    /*** Checks if any of the assistant sections contain a "uses" block which starts with `file://`*/
+    const maybeHasLocalUsesBlocks = (
       assistant: MinimalAssistantSections | undefined,
     ): boolean => {
       if (!assistant) return false;
@@ -112,13 +113,19 @@ async function loadConfigYaml(options: {
       return sections.some(
         (section) =>
           Array.isArray(assistant[section]) &&
-          (assistant[section] as any[]).some(
-            (item: any) => item && typeof item === "object" && "uses" in item,
-          ),
+          (assistant[section] as any[]).some((item: any) => {
+            if (!item || typeof item !== "object" || !("uses" in item))
+              return false;
+            const uses = (item as any).uses;
+            if (typeof uses === "string") {
+              return uses.startsWith("file://");
+            }
+            return false;
+          }),
       );
     };
 
-    if (maybeHasUsesBlocks(config)) {
+    if (maybeHasLocalUsesBlocks(config)) {
       const registryClient = await getRegistryClient();
       const unrolled = await unrollBlocks(
         config as any,
--- a/core/config/yaml/loadYaml.vitest.ts
+++ b/core/config/yaml/loadYaml.vitest.ts
@@ -1,8 +1,20 @@
 import {
   AssistantUnrolledNonNullable,
+  PackageIdentifier,
   validateConfigYaml,
 } from "@continuedev/config-yaml";
-import { describe, expect, it } from "vitest";
+import path from "path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { LLMLogger } from "../../llm/logger";
+import { testControlPlaneClient, testIde } from "../../test/fixtures";
+import {
+  addToTestDir,
+  setUpTestDir,
+  tearDownTestDir,
+  TEST_DIR_PATH,
+} from "../../test/testDir";
+import { localPathToUri } from "../../util/pathToUri";
+import { loadContinueConfigFromYaml } from "./loadYaml";
 
 describe("MCP Server cwd configuration", () => {
   describe("YAML schema validation", () => {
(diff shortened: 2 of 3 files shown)


uinstinct:
@RomneyDa currently this implementation makes all uses blocks in the config relative when a `uses: file://path/to/file` is present. Do we want to support a combination of `uses: package-name/block-name` and `uses: file://path/to/file`?
In that case, we might need to change how [slugs](https://github.com/continuedev/continue/blob/aec308612871911cce670412f49c7470c606c654/packages/config-yaml/src/interfaces/slugs.ts) currently work

RomneyDa:
@uinstinct yes, it should support both

core/config/yaml/loadYaml.ts
```diff
@@ -95,6 +96,50 @@ async function loadConfigYaml(options: {
...
+    };
+    /*** Checks if any of the assistant sections contain a "uses" block which starts with `file://`*/
+    const maybeHasLocalUsesBlocks = (
+      assistant: MinimalAssistantSections | undefined,
+    ): boolean => {
+      if (!assistant) return false;
+      const sections = ["models", "mcpServers", "prompts", "rules"] as const;
+      return sections.some(
+        (section) =>
+          Array.isArray(assistant[section]) &&
```
RomneyDa, line 115: I believe all this type checking should be unnecessary since we use zod to load the YAML. Might requires some nullable checks but otherwise should be typesafe

core/config/yaml/loadYaml.ts
```diff
@@ -95,6 +96,50 @@ async function loadConfigYaml(options: {
...
+            }
+            return false;
+          }),
+      );
+    };
+
+    if (maybeHasLocalUsesBlocks(config)) {
+      const registryClient = await getRegistryClient();
+      const unrolled = await unrollBlocks(
+        config as any,
```
RomneyDa, line 131: remove any type casting

core/config/yaml/loadYaml.ts
```diff
@@ -95,6 +96,50 @@ async function loadConfigYaml(options: {
 
   if (overrideConfigYaml) {
     config = overrideConfigYaml;
+
+    // If overrideConfigYaml still contains any local uses blocks, unroll them here
+    type MinimalAssistantSections = {
+      models?: any[];
```
RomneyDa, line 102: no any types needed

ShaunaGordon:
What I'd personally *really* love is slugified local references (like `uses: models/model.yml`), that automatically pulls from the relevant folder within `~/.continue` (that's honestly what I'd thought it'd do when I was first learning about how to import stuff in Continue), but that's at best a ways off, with the current global import nature of those folders, unless we can add a folder like "imports" or something that doesn't get auto-imported and use that as the import root, but that feels a little clunky.

Alternatively, relative paths with `~/.continue` as the root would be super nice, too, which would allow users to create our own "imports" folder or something (assuming the global import mechanism limits its scope to the specific folders) and sort of simulate the slugification, just with an extra segment, and if a structure is defined, could be easy enough to separate from Hub slugs, by virtue of having three segments instead of two (e.g. `uses: local/models/model.yml`), or, absent a defined structure, a quick look in the local `~/.continue`/`workspace/.continue` for the presence of the top level segment and fall back to Hub in its absence.

(It's possible the above is out of scope, and if so, the current behavior in this PR can be worked into something similar, albeit a bit clunkier and more fragile, but it's definitely a solid start. 👍🏼 )

RomneyDa (changes requested):
@uinstinct let's find an approach that allows using hub slugs in the same assistant. I think you can handle it at the registry level

RomneyDa:
@ShaunaGordon I think we could easily add the relative path from ~/.continue relative folder if ends in .yml/.yaml!

RomneyDa:
Closing in favor of https://github.com/continuedev/continue/pull/7723

PR #7690, opened 2025-09-10
test: use intellij-ide-starter for e2e tests

exigow:
Solves CON-3723

Extra / details:

* E2E tests are now in a separate source set called `testIntegration`.
* Update the CI and docs accordingly.
* Remove robot dependencies and remove the custom `runIdeForUiTests` task. The IDE startup is now automatic/per test (+ instance can be reused if we want).
* A new DSL for E2E tests means we no longer need to add `wait` everywhere, because each closure opening ensures that the window is ready to use. As a result, the tests are slightly faster. The only exception is waiting for inline completion.
* The testing task now depends on `prepareSandbox`, because an E2E test requires specifying where the plugin can be copied from (hence this dependency, since we need to make sure these files exist, see: `CONTINUE_PLUGIN_DIR`).
* The test framework is specified in a different version than the platform version, because older versions don't have the required drivers (`241` vs `243`).
* Move `test-config/config.json` to directory with integration tests.

Known problems:

* I had to upgrade the Kotlin to `2.1.0` because the starter relies on the new SDK. I didn’t find a quick solution for this, but if `./gradlew verifyPlugin` works, we shouldn’t worry about it.
* The IDE doesn’t recognize the `testIntegration` directory as a "test" folder, but as a regular "src" folder. This causes a problem because running the test a second time via the IDE (right click -> run test) triggers the cache, and the test isn’t executed since the sources haven’t changed. This problem doesn’t occur when running the test via `./gradlew testIntegration`, but it’s a bit inconvenient.

### DIFF
--- a/extensions/intellij/.run/Run e2e tests.run.xml
+++ b/extensions/intellij/.run/Run e2e tests.run.xml
@@ -1,7 +1,0 @@
-<component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="Run e2e tests" type="CompoundRunConfigurationType">
-    <toRun name="Run Test IDE" type="GradleRunConfiguration" />
-    <toRun name="Run Tests" type="GradleRunConfiguration" />
-    <method v="2" />
-  </configuration>
-</component>


exigow:
Peek 2025-09-16 22-52

exigow:
Note: `verifyPlugin` is :green_circle: green.

Patrick-Erichsen (approved):
Screenshot 2025-09-19 at 12 54 12 PM

👌

PR #7692, opened 2025-09-10
refresh workspace dirs when folders are added

mq200:
## Description

When folders are moved/copied into a workpace in VSCode, they are not added to Continue's workspaceDirs (except for the 1st time folders are added, as an extension restart is triggered the 1st time).

The PR adds a listener to  [WorkspaceFoldersChangeEvent](https://code.visualstudio.com/api/references/vscode-api#WorkspaceFoldersChangeEvent)

From the VSCode API doc:
> Note: this event will not fire if the first workspace folder is added, removed or changed, because in that case the currently executing extensions (including the one that listens to this event) will be terminated and restarted so that the (deprecated) rootPath property is updated to point to the first workspace folder.
> 

## AI Code Review

## Checklist

---

## Summary by cubic
Fixes missed reindexing when folders are added or removed in a VS Code workspace. The extension now refreshes workspace directories and updates the codebase index without a restart.

- **Bug Fixes**
  - Listen to workspace folder changes and emit folders/changed with affected URIs.
  - Core invalidates walkDirCache, refreshes workspace dirs, and reindexes changed URIs when indexing is enabled.
  - Added refreshWorkspaceDirs to the IDE protocol and implemented it in VS Code and FileSystem IDEs.
  - VsCodeIdeUtils now refreshes its cached workspace directory list.


extensions/vscode/src/extension/VsCodeExtension.ts
```diff
@@ -503,6 +503,14 @@ export class VsCodeExtension {
       });
     });
 
+    vscode.workspace.onDidChangeWorkspaceFolders(async (event) => {
+      this.core.invoke("folders/changed", {
```
RomneyDa: can set ide utils.workspace dirs directly here. Might need to play with some private/public but I think worth not adding an extra message

RomneyDa (commented):
@mq200 this would be great.

1. Can you implement this for jetbrains as well? or at least add a TODO comment to add to jetbrains?
2. I think we could simplify the messaging a bit and remove the `refreshWorkspaceDirs` ide method. You can set workspace dirs within the VS Code IDE directly on folders changed.
3. nitpick but I think the message could be named more clearly to mean workspace dirs rather than folders with a workspace dir. E.g. "workspaceDirsChanged"

his implementation

mq200:
@RomneyDa okay, I'll add the changes for jetbrains

RomneyDa (approved):

PR #7541, opened 2025-09-03
fix: :recycle: Refactor Compaction Summary

chezsmithy:
## Description

Integrate Conversation Summary into the Chat History eliminating the current issues where the Conversation History overlaps the Mentions menu.

Create a more fluid integration of the conversation summary making it look like an assistant response.

When generating the summary, block the enter button to avoid having two requests active.

Move the generating summary message to the Lump Toolbar to unify it with Generating and Applying messages.

## AI Code Review

- **Team members only**: AI review runs automatically when PR is opened or marked ready for review
- Team members can also trigger a review by commenting `@continue-general-review` or `@continue-detailed-review`

## Checklist

- [X] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [X] The relevant docs, if any, have been updated or created
- [X] The relevant tests, if any, have been updated or created

## Screen recording or screenshot

[ When applicable, please include a short screen recording or screenshot - this makes it much easier for us as contributors to review and understand your changes. See [this PR](https://github.com/continuedev/continue/pull/6455) as a good example. ]

## Tests

Added new UI tests.
Performed manual testing.

    

---

## Summary by cubic
Integrates conversation summaries directly into the chat timeline as assistant-style messages, fixes UI overlap with the Mentions menu, and shows a unified “Generating Summary” state in the Lump toolbar. Input is disabled during compaction to prevent double submissions.

- **Bug Fixes**
  - Prevented Mentions menu overlap by rendering summaries inline in chat.
  - Blocked Enter/submit while a summary is generating to avoid parallel requests.

- **Refactors**
  - Removed ConversationSummary component; Chat now renders summaries with an assistant icon.
  - Moved the “Previous Conversation Compacted” indicator to Chat and show it only for the latest summary.
  - Added a Lump toolbar state for “Generating Summary” and disabled input/editor while compaction is active.
  - Introduced setConversationSummary in session slice; used compactionLoading to drive UI state.
  - Added UI tests for compaction flows and summary display.

PR #7695, opened 2025-09-11
feat: improve compaction UI, show compaction message as status message

tingwai:
## Description

Previously compaction messages were not very clear and would disappear when its done compacting so I've
- refactored the existing loading animation we have into a `ActionStatus` component so that when it is compacting there's an animation so users won't think its stuck
- `/compact` slash command now also uses this new component
- ESC key will interrupt compaction both when its automatic compaction or when `/compact` is called
- uses new message queue to queue messages while compacting

For video demo purposes I've set my compaction threshold extremely low so every new message causes a compaction.

https://github.com/user-attachments/assets/26bc40d6-5876-411d-8db8-146b8c950643

### DIFF
--- a/extensions/cli/src/ui/hooks/useChat.ts
+++ b/extensions/cli/src/ui/hooks/useChat.ts
@@ -1,4 +1,5 @@
 /* eslint-disable max-lines */
+/* eslint-disable max-statements   */
 import type { ChatHistoryItem, Session } from "core/index.js";
 import { useApp } from "ink";
 import { useEffect, useRef, useState } from "react";
@@ -526,7 +527,7 @@ export function useChat({
       } catch (error) {
         // If compaction fails, remove the queued message from display
         setQueuedMessages((prev) =>
-          prev.filter((msg) => msg !== queuedMessageForDisplay)
+          prev.filter((msg) => msg !== queuedMessageForDisplay),
         );
         throw error; // Re-throw to maintain error handling
       } finally {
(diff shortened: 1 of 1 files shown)


extensions/cli/src/stream/streamChatResponse.autoCompaction.test.ts
```diff
@@ -147,7 +147,7 @@ describe("handleAutoCompaction", () => {
       "Auto-compacting...",
     );
     expect(mockCallbacks.onSystemMessage).toHaveBeenCalledWith(
-      "✓ Chat history auto-compacted successfully.",
+      "Chat history auto-compacted successfully.",
```
tingwai, line 150: Removed checkmarks to be consistent with other system messages, we don't use check marks anywhere else

extensions/cli/src/ui/hooks/useChat.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable max-lines */
```
tingwai, line 1: I tried refactoring `useChat.ts` to reduce line numbers but it got way too messy, would be better to tackle that another time / PR

chezsmithy:
Did something similar here: https://github.com/continuedev/continue/pull/7541

RomneyDa:
> Did something similar here: #7541

nice, taking a look. just for clarity @tingwai did it for the CLI

PR #7696, opened 2025-09-11
fix: HOTFIX edit apply state tool output 

RomneyDa:
## Description
Assigns tool output at the right phase of edit tool cycle and fixes currently broken tests
    

---

## Summary by cubic
Fixes the edit tool so output is generated during the apply-state phase, not on invocation, ensuring accurate success/failure messages and unblocking failing tests.

- **Bug Fixes**
  - Move tool output to handleApplyStateUpdate with hidden "Edit Success"/"Edit Failed" context items tied to toolCallId.
  - editImpl now defers response (respondImmediately: false) and relies on apply-state updates.
  - Only query open files if relative path resolution fails, improving file path matching.


RomneyDa:
squash and merge pls

PR #7705, opened 2025-09-11
feat: close listbox on click outside

uinstinct:
## Description

Close the listbox only when a click outside the listbox options is detected for AssistantAndOrg Select and Mode Select

resolves  CON-3879

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/415c47e4-4d64-45e9-89c7-aaab43d1e6e7

## Tests

---

## Summary by cubic
Keep Assistant/Org and Mode listboxes open until the user clicks outside; adds a shared useClickOutside hook to control closing behavior. Resolves CON-3879.

- **Bug Fixes**
  - Added useClickOutside to close listboxes only on outside clicks.
  - Made ListboxOptions controlled via listboxOpen + static and refs; removed programmatic Escape close.
  - Removed onClose props and stop closing on option selection; buttons now open the listbox, and it closes only on outside click.
  - Updated navigation/actions (e.g., New Assistant, Rules) to explicitly close the listbox after handling.


uinstinct:
> @uinstinct I think the current goal based on linear ticket is to keep the listbox open when you select an assistant

right. actually the listbox closes only when the user clicks outside of it

RomneyDa:
@uinstinct I think we should only do this for the agents listbox since the mode one is simple (e.g. doesn't have agents that change based on selected org)
Also, I think this should be achievable with some much simple listbox properties or CSS, could you double check that there isn't a super easy prop somewhere?

uinstinct:
actually [headless-ui's listbox](https://github.com/tailwindlabs/headlessui/discussions/975) does not provide open state as well - this is somewhat a makeshift to open/close the listbox with states

uinstinct:
about the css change, there is another way to hide the listbox by hiding the listboxbutton. 

similar to the change in this PR - https://github.com/continuedev/continue/pull/7943/files#diff-609a928b942fcaee5c271604f63b51b97859641947cf6faab4bf9d1a7617dfb1R53-R61

https://github.com/tailwindlabs/headlessui/blob/6b5709ad803255cfbdc8360b5c930bf3d73ae16d/packages/%40headlessui-react/src/components/listbox/listbox.tsx#L564C14-L565

RomneyDa (dismissed):

uinstinct:
closing as this is currently possible on latest main

PR #7709, opened 2025-09-11
fix: use common process.env values to launch MCP

fbricon:
My nodejs versions are managed by asdf. Continue fails to resolve npx to launch MCPs because it doesn't pass the proper system environment variables (basically, just process.env.PATH is passed): 

Screenshot 2025-09-11 at 14 44 03

The proposed fix merges options.transport.env with all process.env variables before launching the MCP. I can't see a reason why other env vars would be ignored by default. But, if that is an issue, we could get away with adding env.HOME and maybe a couple other variables.
    

---

## Summary by cubic
Fixes MCP launch failures caused by missing env vars. MCP processes now inherit the full process.env, so tools like npx resolve correctly with asdf and similar managers.

- **Bug Fixes**
  - Pass all process.env values to stdio transports and merge with options.transport.env.
  - Keep using PATH from the user shell on non-Windows when available.
  - Prevent failures from incomplete environments (e.g., asdf-managed Node).


core/context/mcp/MCPConnection.ts
```diff
@@ -347,24 +347,27 @@ class MCPConnection {
...
+          Object.entries(process.env).filter(([_, v]) => v !== undefined) as [
+            string,
+            string,
+          ][],
+        );
+
+        const env = {
+          ...processEnv,
+          ...(options.transport.env ?? {}),
+        };
```
panyamkeerthana: do we want to pass every environment variable to the subprocess or just a few important ones?

RomneyDa:
@fbricon @panyamkeerthana merging this has been delayed because of security concerns (i.e. @panyamkeerthana's comment). Just want to reduce the chance a rogue MCP goes haywire on users. 

I'd +1 to getting 80% of the way there with HOME and other common vars that are usually safe

@fbricon would any vars specifically fix your case? Would that be transferable to most users?

RomneyDa:
@fbricon wondering if just need to add `asdf` to the `getEnvPathFromUserShell` utility?

>  const command = `${process.env.SHELL} -l -c 'for f in ~/.zprofile ~/.zshrc ~/.bash_profile ~/.bashrc; do [ -f "$f" ] && source "$f" 2>/dev/null; done; echo $PATH'`;

fbricon:
I switched to using common ENV vars. That should be enough for most MCPs hopefully. 
I don't think asdf requires special treatment.

RomneyDa:
@fbricon I have a feeling that eventually we'll end up just doing your original full pass through implementation, but while we wrap our heads around MCP security this should work. Thanks for the contribution

RomneyDa (approved):

PR #7714, opened 2025-09-11
fix: show config errors for selected assistant, remove duplicate fatal

RomneyDa:
…l error
Before:
duplicate fatal error notices
image

no fatal error explanation
image

After:
removes duplicate
image

shows fatal errors in config agents section
image

## Description

## AI Code Review

## Checklist

## Tests

---

## Summary by cubic
Fixes duplicate fatal error banners and shows configuration errors for the selected assistant in the Agents section, so users see a single, clear error with details. Addresses CON-3899.

- **Bug Fixes**
  - Display config errors for the active assistant by using store-backed errors when the profile is selected.
  - Remove duplicate fatal error notice by only rendering the global indicator outside the chat route (chat keeps its own indicator).

### DIFF
--- a/gui/src/components/Layout.tsx
+++ b/gui/src/components/Layout.tsx
@@ -1,5 +1,5 @@
 import { OnboardingModes } from "core/protocol/core";
-import { useContext, useEffect, useMemo, useState } from "react";
+import { useContext, useEffect, useState } from "react";
 import { Outlet, useLocation, useNavigate } from "react-router-dom";
 import styled from "styled-components";
 import { CustomScrollbarDiv } from ".";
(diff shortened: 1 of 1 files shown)


gui/src/components/Layout.tsx
```diff
@@ -54,6 +54,8 @@ const Layout = () => {
 
   const showDialog = useAppSelector((state) => state.ui.showDialog);
   const isInEdit = useAppSelector((store) => store.session.isInEdit);
+  const isHome =
+    location.pathname === "/" || location.pathname === "/index.html";
```
tomasz-stefaniak: Could you use the `ROUTES` object here? (There is no `index.html` in `ROUTES` but can easily be added)

RomneyDa: good idea, will do!

gui/src/components/Layout.tsx
```diff
@@ -54,6 +54,12 @@ const Layout = () => {
 
   const showDialog = useAppSelector((state) => state.ui.showDialog);
   const isInEdit = useAppSelector((store) => store.session.isInEdit);
+  const isHome = useMemo(() => {
+    return (
+      location.pathname === ROUTES.HOME ||
+      location.pathname === ROUTES.HOME_INDEX
+    );
+  }, []);
```
tomasz-stefaniak: Boolean checks are cheap enough that a memo here is probably overkill. I'm also not sure if this is going to behave is expected - the dependency array is empty so this value is never going to be updated, is that by design?

RomneyDa: fixed

PR #7727, opened 2025-09-12
feat: fix flickering by splitting and stylizing text so each row is rendered separately

tingwai:
## Description

Addresses flickering issue in cn. The root cause was when a component was larger than the height of the terminal is displayed, any redraws will cause flickering. This was an issue when llm responds with a large text or diff and the loading animation would cause everything to flash. 

In order to fix this this PR splits all the content by width of terminal and renders each line separately so no piece is larger than height of the terminal. This requires a big refactor of inverting our previous tsx styling where a large piece of text can be put straight inside a <Text>. 

Here we first split chunks of text by column width (intelligently split at spaces so words won't get split between lines). Then segment each line if necessary for stylizing, eg syntax highlighted code can have multiple colors in one line. Finally pass this to be rendered in separate rows. 

## AI Code Review

## Checklist

## Tests

---

## Summary by cubic
Eliminates CLI chat flickering by pre-processing markdown and tool outputs into styled, terminal-width rows so each row renders independently. Messages and tool results now render instantly without reflow during typing or resize.

- **Bug Fixes**
  - Removed flicker in assistant messages and tool results by splitting content into per-row segments.
  - Prevented re-render storms on resize by avoiding in-render markdown processing.

- **Refactors**
  - Added MarkdownProcessor to convert content into StyledSegments and split them into rows; introduced StyledSegmentRenderer.
  - Added ToolResultProcessor to expand tool outputs (diffs, bash, checklists) into pre-styled rows; titles generated via getToolCallTitleSegments.
  - Updated useChat flow: processHistoryForTerminalDisplay, splitMessageContent for user/assistant, streaming callbacks split content by terminal width.
  - Simplified MemoizedMessage to render pre-styled segments and tool-result rows; removed nested, heavy components.
  - Deprecated components moved to .deprecated: MarkdownRenderer, ColoredDiff, ToolResultSummary, ChecklistDisplay.
  - StaticChatContent and TUIChat now pass terminal width; SessionSelector input handling improved (supports Ctrl+C/D/Esc).
  - Tests updated for new split-message behavior; dependency bump ink to 6.3.0.


sestinj (commented):
This is very well done, but from what I'm seeing here we are going to have to write a custom splitter for every new component we ever want to build. Am I missing something? It feels near impossible to maintain if the splitting can't be automated, mostly in the sense that LLMs would struggle to learn this pattern compared to how good they are at normal React

tingwai:
Let me go back to the drawing board and see how we can approach this flickering in a better developer experience way. I have an idea, maybe there's a way to keep our existing React components and have a post processing step to decompose them and be rendered line by line to avoid the flickering

PR #7733, opened 2025-09-14
feat: adapt duplicate toolCallId

xiaohuanxiong3:
## Description

When a large language model makes tool calls in a conversation, if the same toolCallId appears, the existing code will update the first toolCallState, resulting in an abnormal tool call status displayed on the chat page, or other unknown exceptions
Duplicate toolCallId occurs when provider is openrouter and for some models or providers

## AI Code Review

## Checklist

## Tests

### DIFF
--- a/gui/src/redux/selectors/selectToolCalls.ts
+++ b/gui/src/redux/selectors/selectToolCalls.ts
@@ -50,7 +50,9 @@ export const selectApplyStateByToolCallId = createSelector(
     (_store: RootState, toolCallId: string) => toolCallId,
   ],
   (applyStates, toolCallId) => {
-    return applyStates.states.find((state) => state.toolCallId === toolCallId);
+    return applyStates.states.findLast(
+      (state) => state.toolCallId === toolCallId,
+    );
   },
 );
 
--- a/gui/src/redux/thunks/streamResponseAfterToolCall.ts
+++ b/gui/src/redux/thunks/streamResponseAfterToolCall.ts
@@ -69,7 +69,7 @@ export const streamResponseAfterToolCall = createAsyncThunk<
 
         // Check if we should continue streaming based on tool call completion
         const history = getState().session.history;
-        const assistantMessage = history.find(
+        const assistantMessage = history.findLast(
           (item) =>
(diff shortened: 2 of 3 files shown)


RomneyDa (changes requested):
@xiaohuanxiong3 I think there should only ever be one apply state and tool call with the id, could you clarify what the difference is here between find and findLast in these cases? We might have an underlying bug causing duplicate tool calls or failing to merge them.

xiaohuanxiong3:
@RomneyDa This PR handles the situation where two different tool calls have the same toolCallId. In this case, two different applyStates or toolCallStates may have the same toolCallId. FindLast or find from back to front will find the latest, correct applyState or toolCallState.

RomneyDa:
@xiaohuanxiong3 tool calls and apply states should be unique - there should only be one apply state for a given tool call id and no two tool calls should have the same ID. What is a situation e.g. config in which this could occur?

xiaohuanxiong3:
@RomneyDa Currently, both applyState and toolCallState use the toolCallId returned by the large language model. When the large language model uses the same toolCallId when making tool calls, two applyState or toolCallState will use the same toolCallId. OpenRouter may have different tool calls using the same toolCallId. These tool calls are usually of the same type, such as multiple runTerminalCommand tool calls.

xiaohuanxiong3:
@RomneyDa 
image
image
image
image
image
image

RomneyDa:
@xiaohuanxiong3 appreciate the in-depth feedback here. It sounds like this is happening specifically with openrouter or maybe a specific model that has this `functions.tool_name: index` tool call id format? Are you able to share which model/API is doing this? To me it seems like we could go ahead and push your fix but it's more of an upstream issue since these conversations are cumulative and tool call ids should not be duplicated

xiaohuanxiong3:
@RomneyDa moonshotai/kimi-k2-0905 provider: chutes/fp8

Patrick-Erichsen (approved):

RomneyDa:
@xiaohuanxiong3 thanks again, will keep an eye on this model

PR #7748, opened 2025-09-15
perf: memoize input box to avoid re-render on each keystroke

sestinj:
## Description

This PR stops the input toolbar and TipTapEditor from re-rendering on each keystroke. This might have caused performance problems for a small number of users.

It also introduces a component / React hook that can be used for debugging re-rendering issues in the future.
    

---

## Summary by cubic
Prevents input toolbar and editor from re-rendering on every keystroke, improving typing responsiveness. Adds a small debug hook to visualize re-renders during development.

- **Performance**
  - Memoized ContinueInputBox, InputToolbar, and TipTapEditor with React.memo and custom comparators.
  - Stabilized toolbarOptions with useMemo to avoid identity-driven renders.
  - Set activeKey only for Meta/Control/Alt to reduce state updates while typing.
  - Safely update TipTap placeholder only if the extension is present.

- **New Features**
  - Added useRenderDebug hook and RenderCountBadge to inspect re-renders.
  - Added CSS helpers: rerender-flash animation and render-count badge styles.


gui/src/components/mainInput/ContinueInputBox.tsx
```diff
@@ -69,16 +69,20 @@ function ContinueInputBox(props: ContinueInputBoxProps) {
...
         hideImageUpload: false,
         hideUseCodebase: true,
         hideSelectModel: false,
         enterText:
           editModeState.applyState.status === "done" ? "Retry" : "Edit",
-      }
-    : {};
+      } as ToolbarOptions;
+    }
+    // Stable empty object to avoid re-renders from identity changes
```
RomneyDa, line 83: will register as change every time I believe? - empty objects aren't stable

RomneyDa (approved):
Cubic feedback seems worthy of a look

PR #7772, opened 2025-09-16
fix: do not show indexing options unless config loads

uinstinct:
## Description

When config is loading, in the ui we were showing that indexing is being initialized (which was perceived as indexing was starting by users). This PR fixes that by showing it only when the config has finished loading.

resolves CON-3946

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/64a0ffb1-097c-44d5-a545-be97bdaa9f01

https://github.com/user-attachments/assets/fc02ac92-c74c-448a-8d07-1a492bbeaa7c

## Tests

### DIFF
--- a/gui/src/pages/config/sections/IndexingSettingsSection.tsx
+++ b/gui/src/pages/config/sections/IndexingSettingsSection.tsx
@@ -7,7 +7,7 @@ import Alert from "../../../components/gui/Alert";
 import { Card, Divider } from "../../../components/ui";
 import { IdeMessengerContext } from "../../../context/IdeMessenger";
 import { useAppDispatch, useAppSelector } from "../../../redux/hooks";
-import { updateConfig } from "../../../redux/slices/configSlice";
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
- enabling indexing is not actually dependent on config loading since it lives in global context. It makes sense to hide the indexing UI but probably not the toggle. Thoughts on this?
- I think it would be nice to show some kind of loading text or indicator when loading rather than nothing. Maybe "Config has not loaded successfully" or similar
- also makes sense to hide/disable when config has fatal errors
- It seems like more things throughout the app could use this logic. What if instead of the JSON.stringify EMPTY_CONFIG comparison we add a simple value to redux `config.hasInitialized` which we trigger in ParallelListeners once initialized? or similar

uinstinct:
closing as it was resolved [in this thread](https://xenaditya1gma-xkk5787.slack.com/archives/C090WHD3SRH/p1758826408279489?thread_ts=1758633415.142589&cid=C090WHD3SRH)

PR #7773, opened 2025-09-16
fix: remove empty lines from start/end in find and replace

uinstinct:
## Description

Remove empty lines in FindAndReplace lines content which would otherwise be perceived as padding.

resolves CON-3916

## AI Code Review

## Checklist

**before**
image

**after**
image

## Tests

---

## Summary by cubic
Removes leading and trailing empty lines from Find & Replace previews to avoid visual padding and misleading context. Addresses CON-3916.

- **Bug Fixes**
  - Added removeEmptyLines util (fromStart/fromEnd) to strip blank lines.
  - Applied trimming to startLines and endLines in FindAndReplaceDisplay without changing diff logic.

### DIFF
--- a/gui/src/pages/gui/ToolCallDiv/FindAndReplace.test.tsx
+++ b/gui/src/pages/gui/ToolCallDiv/FindAndReplace.test.tsx
@@ -19,9 +19,16 @@ vi.mock("../../../components/ui", () => ({
   useFontSize: () => 14,
 }));
 
-vi.mock("../../../util/clientTools/findAndReplaceUtils", () => ({
-  performFindAndReplace: vi.fn(),
-}));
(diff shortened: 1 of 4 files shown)


RomneyDa (changes requested):
@uinstinct looks like some failing tests. Note the line removal logic could be simplified a bit somthing like
```
export function trimEmptyLines(lines: string[], fromEnd = false): string[] {
  const arr = fromEnd ? [...lines].reverse() : [...lines];
  let i = 0;
  while (i < arr.length && arr[i].trim() === ""){
   i++;
  }
  const result = arr.slice(i);
  return fromEnd ? result.reverse() : result;
}
```

but approach not critical

Let's add some unit tests

RomneyDa (approved):
Nice!

PR #7783, opened 2025-09-16
bugfix: make tool calls clickable

Patrick-Erichsen:
Resolves https://github.com/continuedev/continue/discussions/7768
    

---

## Summary by cubic
Make tool call status text clickable to open the tool’s output in a virtual file. This lets users quickly view full tool output without expanding the UI.

- **Bug Fixes**
  - Clicking the tool call message opens a "Tool Output" virtual file with rendered output (when available).
  - Applies to both SimpleToolCallUI and ToolCallDisplay; shows pointer/hover states when clickable.
  - Only clickable when the call is done, canceled, or errored; in-progress calls are not clickable.


RomneyDa (approved):

RomneyDa (changes requested):
I think SimpleToolCall already has this functionality, this approach duplicates for SimpleToolCall

PR #7811, opened 2025-09-17
fix(cli): append /v1 to ollama endpoint

uinstinct:
## Description

When using a custom ollama base in config.yaml for cn, `/v1` needed to be appended at the end of the url, whereas the extension did not need it. This created requirement of 2 different configs for the same ollama model. This PR fixes this.

resolves CON-3909

## AI Code Review

## Checklist

## Tests

### DIFF
--- a/packages/openai-adapters/src/index.ts
+++ b/packages/openai-adapters/src/index.ts
@@ -141,6 +141,15 @@ export function constructLlmApi(config: LLMConfig): BaseLlmApi | undefined {
     case "lmstudio":
       return openAICompatible("http://localhost:1234/", config);
     case "ollama":
+      // for openai compaitability, we need to add /v1 to the end of the url
+      // this is required for cli (for core, endpoints are overriden by core/llm/llms/Ollama.ts)
+      if (config.apiBase && !config.apiBase.endsWith("v1")) {
+        if (config.apiBase.endsWith("/")) {
+          config.apiBase += "v1/";
+        } else {
+          config.apiBase += "/v1/";
+        }
+      }
       return openAICompatible("http://localhost:11434/v1/", config);
     case "mock":
       return new MockApi();
--- a/packages/openai-adapters/src/test/main.test.ts
+++ b/packages/openai-adapters/src/test/main.test.ts
@@ -309,4 +309,25 @@ describe("Configuration", () => {
     );
     expect((azure as OpenAIApi).openai.apiKey).toBe("sk-xxx");
   });
+
+  it('should have correct default API base for "ollama"', () => {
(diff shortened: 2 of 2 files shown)


Patrick-Erichsen (dismissed):

RomneyDa (changes requested):
I believe this logic will fail if the config ends with `v1/` -> will end up as `v1/v1`

uinstinct:
> I think this logic will fail if the config ends with `v1/` -> will end up as `v1/v1`

nice catch. fixed!

RomneyDa (changes requested):
@uinstinct will still fail for some cases like
myhostv1/ (will not append v1 to it) or anything with query params (may be be used for ollama api wrappers) or with upper case url ending in V1 or V1/ (regex is case sensitive currently)

Let's do a robust implementation that that matches /v1/ or /v1 in just the url path, isn't case sensitive, and adds v1/ or /v1/ respectively to the URL path. I might pull this into a separate util since there is a use case for other providers

packages/openai-adapters/src/util/appendPathToUrl.ts
RomneyDa, line 1: @uinstinct this will still fail for e.g. http://my.endpointv1, will end up as http://my.endpointv1/

packages/openai-adapters/src/util/appendPathToUrl.ts
```diff
@@ -0,0 +1,17 @@
+export function appendPathToUrlIfNotPresent(
+  urlString: string,
+  pathWithoutSlash: string,
+) {
+  const url = new URL(urlString);
+  if (!url.pathname.endsWith("/")) {
+    url.pathname += "/";
+  }
+  if (!url.pathname.endsWith(pathWithoutSlash + "/")) {
```
RomneyDa, line 9:
```suggestion
  if (!url.pathname.endsWith("/" + pathWithoutSlash + "/")) {
```

RomneyDa (changes requested):

packages/openai-adapters/src/util/appendPathToUrl.ts
uinstinct: actually it would only append '/' to the pathname

image


have written [some test cases](https://github.com/continuedev/continue/blob/d3d75e43b6d95e6021e31d45aadb7042d9c2b9c3/packages/openai-adapters/src/test/main.test.ts/#L335-L351) for this case

packages/openai-adapters/src/util/appendPathToUrl.ts
RomneyDa: You're right!

PR #7842, opened 2025-09-18
fix: make config-yaml compatible with browser

sestinj:
## Description

avoid importing node:os to make config-yaml compatible with browser for remote-config-server
    

---

## Summary by cubic
Make config-yaml work in the browser by removing the node:os dependency and safely handling "~" path expansion. This unblocks remote-config-server in browser contexts.

- **Bug Fixes**
  - Replaced os.homedir() with environment-based helpers that read HOME/USERPROFILE only if available; no Node imports.
  - Expanded only a leading "~"; if no home dir is found, the path is left unchanged.
  - Added tests for both cases (HOME present and absent).

### DIFF
--- a/packages/config-yaml/src/interfaces/slugs.test.ts
+++ b/packages/config-yaml/src/interfaces/slugs.test.ts
@@ -158,4 +158,62 @@ describe("PackageIdentifier", () => {
       "Unknown URI type: invalid",
     );
   });
+
+  it("should expand leading tilde when HOME is available", () => {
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen (approved):
I think we can get rid of this from the package.json since I believe the only place we were using node imports was in the file you updated.

```json
    ".": {
      "browser": "./dist/browser.js",
      "node": "./dist/index.js",
      "default": "./dist/index.js"
    }
  },
```

PR #7859, opened 2025-09-19
Enable the terminal command for all remote names

etherandrius:
## Description

Enabling the terminal command for all remotes; I'm pretty sure this is no longer needed?

The fallback says it's not enabled because it does not work for SSH but at the same time the feature is explicitly enabled for `ssh-remote` 

See definition of `vscode.env.remoteName`
```
        /**
         * The name of a remote. Defined by extensions, popular samples are `wsl` for the Windows
         * Subsystem for Linux or `ssh-remote` for remotes using a secure shell.
         *
         * *Note* that the value is `undefined` when there is no remote extension host but that the
         * value is defined in all extension hosts (local and remote) in case a remote extension host
         * exists. Use {@link Extension.extensionKind} to know if
         * a specific extension runs remote or not.
         */
        export const remoteName: string | undefined;
```

### DIFF
--- a/core/tools/implementations/runTerminalCommand.ts
+++ b/core/tools/implementations/runTerminalCommand.ts
@@ -1,7 +1,6 @@
 import iconv from "iconv-lite";
 import childProcess from "node:child_process";
 import os from "node:os";
-import util from "node:util";
 // Automatically decode the buffer according to the platform to avoid garbled Chinese
 function getDecodedOutput(data: Buffer): string {
   if (process.platform === "win32") {
@@ -43,8 +42,6 @@ import {
 } from "../../util/processTerminalStates";
 import { getBooleanArg, getStringArg } from "../parseArgs";
 
-const asyncExec = util.promisify(childProcess.exec);
-
 // Add color-supporting environment variables
 const getColorEnv = () => ({
   ...process.env,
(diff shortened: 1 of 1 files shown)


core/tools/implementations/runTerminalCommand.ts
```diff
@@ -1,7 +1,6 @@
 import iconv from "iconv-lite";
 import childProcess from "node:child_process";
 import os from "node:os";
-import util from "node:util";
```
etherandrius, line 4: not used

core/tools/implementations/runTerminalCommand.ts
```diff
@@ -43,8 +42,6 @@ import {
 } from "../../util/processTerminalStates";
 import { getBooleanArg, getStringArg } from "../parseArgs";
 
-const asyncExec = util.promisify(childProcess.exec);
```
etherandrius, line 46: not used

core/tools/implementations/runTerminalCommand.ts
```diff
@@ -55,391 +52,363 @@ const getColorEnv = () => ({
...
 export const runTerminalCommandImpl: ToolImpl = async (args, extras) => {
   const command = getStringArg(args, "command");
   // Default to waiting for completion if not specified
   const waitForCompletion =
     getBooleanArg(args, "waitForCompletion", false) ?? true;
 
-  const ideInfo = await extras.ide.getIdeInfo();
   const toolCallId = extras.toolCallId || "";
 
-  if (ENABLED_FOR_REMOTES.includes(ideInfo.remoteName)) {
```
etherandrius, line 79: Actual change is removing this if. 

All lines below are just indentation changes

core/tools/implementations/runTerminalCommand.ts
```diff
@@ -55,391 +52,363 @@ const getColorEnv = () => ({
...
-  await extras.ide.runCommand(command);
-  return [
-    {
-      name: "Terminal",
-      description: "Terminal command output",
-      content:
-        "Terminal output not available. This is only available in local development environments and not in SSH environments for example.",
-      status: "Command failed",
-    },
-  ];
```
etherandrius, line 444: Removing fall-back

RomneyDa:
@etherandrius I think the original reason for this logic was to enable running commands in the remote environment, i.e. less as a fallback for environments that don't support it and more of a way to enable running commands in remotes at all (through IDE interface). It seems like this has slowly been watered down so not all commands are run locally, partly because the support for local commands is so much better now (partial output, backgrounding, etc.)

@chezsmithy you've done a lot of work here, noticed your reaction, do you have any thoughts on running commands remote vs locally when in a remote filesystem?

chezsmithy:
@RomneyDa there are explicit ways to use local vscode with remote terminals. I've had a few ideas I've tinkered with but it gets odd when vscode makes file creates/edits locally but runs terminal commands remotely. See a dev container use case for example. 

This is also quite different from what jetbrains expects using their remote ssh support.

This PR might work fine in both tools but I'm not sure without testing.

My take so far has been to install the entire extension in the remote host for vscode. IM not a huge jetbrains user (tho we have many in our organization).

RomneyDa:
@chezsmithy thanks!

@etherandrius what was the original intent for this PR, does it fix a specific problem with commands being run remotely when they shouldn't for a given `remote`?

RomneyDa:
Closing this; stale/conflicts/failing tests
@etherandrius feel free to reopen and address comments, I've also added a ticket for myself to take another look into this!

PR #7267, opened 2025-08-20
feat: multi edit and single search and replace tools

RomneyDa:

PR #7875, opened 2025-09-20
feat: restore trimmed/whitespace search and replace fallbacks 

RomneyDa:
This PR reintroduces whitespace and trimmed fallbacks and cleans up/unifies search and replace logic

More granular error codes for edit tools
Uses ContinueError type with ContinueErrorReason enum
Adds posthog telemetry for CLI tool errors
Also removed an unused error step container component
changes posthog event names from gui_tool_call_outcome and gui_tool_call_decision to tool_call_outcome and tool_call_decision, respectively
unifies search and replace logic between extensions and CLI
Removes ability to create a file with multi edit and empty old_string because it's confusing and pretty much fully separate functionality

Note
Jaro Winkler was removed here https://github.com/continuedev/continue/pull/7240
Other fallbacks were removed here https://github.com/continuedev/continue/pull/7267
This PR replaces trimmed/whitespace fallbacks

---

## Summary by cubic
Restores trimmed and whitespace-tolerant find/replace fallbacks and unifies search-and-replace logic across Core, CLI, and GUI. Adds structured error codes and telemetry so failures are easier to diagnose, and simplifies multi-edit behavior.

- **New Features**
  - Restored whitespace-ignored and trimmed matching fallbacks for find/replace (consistent in CLI and GUI).
  - Centralized find/replace and multi-edit in core (shared match, validate, and replace utilities); removed legacy CLI search-and-replace code.
  - Added ContinueError with reason codes; propagated through CLI/GUI and included in telemetry (tool_result, tool_call_decision/outcome).
  - Stricter validation for edit args and multi-edit; clearer errors for missing fields, no match, or multiple matches.
  - Extensive tests for matching strategies and multi-edit.

- **Migration**
  - Replace any use of the removed CLI search-and-replace tool with edit or multiEdit.
  - Do not create files via multiEdit using an empty old_string; use writeFile instead.
  - If you read error objects, switch to ContinueError and check error.reason (the old ContinueError interface was removed).


core/util/errors.ts
```diff
@@ -10,3 +10,44 @@ export function getRootCause(err: any): any {
...
+  // General File
+  FileAlreadyExists = "file_already_exists",
+  FileNotFound = "file_not_found",
+  FileWriteError = "file_write_error",
+  FileIsSecurityConcern = "file_is_security_concern",
+  ParentDirectoryNotFound = "parent_directory_not_found",
+
+  // Other
+  Unspecified = "unspecified",
+  Unknown = "unknown",
```
tingwai, line 52: we should add a comment explaining the difference so the next person knows when to use which. I see a mix of both 

Seems like this code snippet from this PR summarizes how unspecified vs unknown convention should be used?
```
          : e instanceof Error
            ? new ContinueError(ContinueErrorReason.Unspecified, e.message)
            : new ContinueError(ContinueErrorReason.Unknown, String(e)),
```

gui/src/redux/thunks/streamResponse_toolCalls.test.ts
```diff
@@ -567,7 +567,7 @@ describe("streamResponseThunk - tool calls", () => {
     // Verify telemetry events for auto-approved tool execution
     // Use partial matching to allow additional fields (e.g. model) in payload
     expect(mockPosthog.capture).toHaveBeenCalledWith(
-      "gui_tool_call_decision",
+      "tool_call_decision",
```
tingwai, line 570: For my own education, why are we changing the name? Are we able to stitch the data back together without losing previous historical data?

tingwai (approved):

tingwai:
Nice work on refactoring the errors!

core/util/errors.ts
RomneyDa: Agree! Will add this on a followup PR today

gui/src/redux/thunks/streamResponse_toolCalls.test.ts
RomneyDa: Yes, we have the IDE type on the event

PR #7878, opened 2025-09-20
Fix/another bug2

Akashbellary:
This is 3rd fix for test cases


sestinj:
@Akashbellary it looks like there are a lot of changes in this PR. Without further explanation I'm going to have to close it. Please let me know if you had something in mind and we're happy to potentially discuss it before you go through all the work of putting up a PR!

PR #7888, opened 2025-09-22
feat(cli): enhance /init command to create custom slash command

sestinj:
## Summary
- Updated the `/init` slash command prompt to instruct the model to create a custom slash command file
- The slash command file is created at `.continue/rules/review.md` with proper frontmatter structure
- The review prompt provides instructions for common code review tasks

## Changes
- Modified `extensions/cli/src/commands/init.ts` to include instructions for creating a slash command file
- The slash command follows the required structure with `invokable: true` in the frontmatter
- The review prompt focuses on code quality, performance, security, error handling, testing, and documentation

## Test Plan
- [ ] Run `/init` command in the CLI
- [ ] Verify that both `AGENTS.md` and `.continue/rules/review.md` files are created
- [ ] Confirm the slash command file has correct frontmatter structure
- [ ] Test that the `/review` command becomes available after file creation

🤖 Generated with [Claude Code](https://claude.ai/code)
    

---

## Summary by cubic
Enhanced /init to also scaffold a /review slash command by generating .continue/rules/review.md with invokable frontmatter, alongside AGENTS.md. This makes a ready-to-use code review command available right after initialization.

- New Features
  - Updated CLI init prompt to create both AGENTS.md and a review slash command file.
  - Generates .continue/rules/review.md (invokable: true) with repo-aware review guidance.
  - Enables the /review command by default after running /init.


Patrick-Erichsen (changes requested):
Instead of a hardcoded string/filepath instructions we should use the same logic that we have in `core/tools/implementations/createRuleBlock.ts`

Otherwise this will almost certainly break at some point in the future

PR #7889, opened 2025-09-22
Fix/npmci

Akashbellary:
sub-branch to fix test case

    

---

## Summary by cubic
Fixes the HTTP Context Provider to safely handle missing or null uri fields from API responses, preventing crashes and restoring HTTP requests when invoked from suggestions/@ commands. Adds a focused test suite to ensure robust handling of different response shapes.

- **Bug Fixes**
  - Add null/shape checks for item.uri and omit it when absent; supports array and single-object responses.
  - Introduce HttpContextProvider tests covering uri present/missing/null and array responses.

- **Dependencies**
  - Update @types/jest to 29.5.14 (core, VS Code) and 30.0.0 (root).
  - Update ts-jest to 29.4.4.
  - Add is-localhost-ip to root dependencies.


sestinj:
Hey @Akashbellary thanks for the PR! Can you share more about the goals here? Given that this feature is used quite infrequently and that there were substantial changes/mocking required to test it, I am not confident that this is something we would want to merge.

sestinj:
Given the state of the other PRs (which this depends on), I'm going to close this as well. Would love to have the conversation about what we want to accomplish in the GitHub issue

PR #7904, opened 2025-09-22
fix: use json schema for tool definitions

RomneyDa:
## Description
Uses normal JSON schema for tool definitions - main change being moving required from a boolean at param level to an array of key names at the level above
This fixes incomplete json schema tool definition for tools with nested args (currenly multi edit) for CLI


extensions/cli/src/tools/multiEdit.ts
```diff
@@ -65,18 +65,39 @@ WARNINGS:
...
+      },
+      edits: {
+        type: "array",
+        description:
+          "Array of edit operations to perform sequentially on the file",
+        items: {
+          type: "object",
+          required: ["old_string", "new_string"],
+          properties: {
+            old_string: {
```
RomneyDa, line 84: This is the main reason for this PR
Model should see nested object schema with old_string, new_string, replace_all

PR #7908, opened 2025-09-22
Revert "Fix image drop from main editor"

RomneyDa:
Reverts continuedev/continue#7205

PR #7910, opened 2025-09-22

RomneyDa:
## Description

https://github.com/continuedev/continue/pull/7908
Reopening this with requested changes:
- remove extraneous build script changes
- add jetbrains support or handling for jetbrains which doesn't have read file from data uri implemented
- remove console.log's


sestinj:
@aadarshkt it doesn't appear that I have access to push to this PR, could you either change this or run the following and add a commit?

```bash
git checkout main -- .vscode/launch.json
git checkout main -- .vscode/tasks.json
```

PR #7944, opened 2025-09-24
chore: remove chat history from redux persistent storage

uinstinct:
## Description

Remove session reducer's history item from redux's persistent storage which was stored in vscode's session storage (having cases exceeding 5mb limit).

closes CON-3773

## AI Code Review

## Checklist

## Tests

---

## Summary by cubic
Removed persisting chat history in Redux to VS Code session storage to avoid hitting the 5 MB limit and prevent errors on large sessions. Closes CON-3773.

- **Bug Fixes**
  - Stop persisting session.history and remove its migration so history is no longer saved or rehydrated.

- **Refactors**
  - Remove unused redux-logger import and setup from the store.

### DIFF
--- a/gui/src/hooks/ParallelListeners.tsx
+++ b/gui/src/hooks/ParallelListeners.tsx
@@ -107,6 +107,7 @@ function ParallelListeners() {
         await handleConfigUpdate(true, result.content);
       }
       dispatch(setConfigLoading(false));
+      await dispatch(loadLastSession());
     }
     void initialLoadConfig();
     const interval = setInterval(() => {
@@ -241,10 +242,6 @@ function ParallelListeners() {
 
   useEffect(() => {
     migrateLocalStorage(dispatch);
-  }, []);
-
-  useEffect(() => {
-    void dispatch(loadLastSession());
   }, []);
 
   return <></>;


RomneyDa (changes requested):
@uinstinct when I tried this earlier it caused failure to load last session

uinstinct:
> @uinstinct when I tried this earlier it caused failure to load last session, do we need to add loading of last session on startup from JSON?

I see. There is a slight delay on the first load.
Added a loading indicator!


https://github.com/user-attachments/assets/6699d53d-380c-4c1d-82b8-af00f6d88e8d

RomneyDa:
@uinstinct I mean on startup it won't load the last session like it currently does. The last session button would operate as before

RomneyDa (changes requested):
See above comments

uinstinct:
https://github.com/user-attachments/assets/b599fdd4-3821-4c1a-bd15-1f6acf3173b4

implemented!

gui/src/redux/store.ts
```diff
@@ -108,13 +104,6 @@ const persistedReducer = persistReducer<ReturnType<typeof rootReducer>>(
 export function setupStore(options: { ideMessenger?: IIdeMessenger }) {
   const ideMessenger = options.ideMessenger ?? new IdeMessenger();
 
-  const logger = createLogger({
```
RomneyDa, line 111: This can be used if you uncomment the `.concat(logger)` below. For debugging

gui/src/redux/store.ts
```diff
@@ -35,9 +34,7 @@ const rootReducer = combineReducers({
 
 const saveSubsetFilters = [
   createFilter("session", [
-    "history",
     "id",
-    "lastSessionId",
```
RomneyDa, line 40: We still need to persist lastSessionId

gui/src/redux/thunks/session.ts
```diff
@@ -146,20 +142,17 @@ export const loadRemoteSession = createAsyncThunk<
...
+export const loadLastSession = createAsyncThunk<void, void, ThunkApiType>(
   "session/loadLast",
-  async ({ saveCurrentSession }, { extra, dispatch, getState }) => {
-    const state = getState();
-
-    if (state.session.id && saveCurrentSession) {
+  async (_, { extra, dispatch, getState }) => {
+    let lastSessionId = getState().session.lastSessionId;
+
+    const lastSessionResult = await extra.ideMessenger.request("history/list", {
```
RomneyDa: I would just used persisted lastSessionId still rather than listing history. If there's noLast sessionId we start a new session

RomneyDa: That being said might just comment it out since we will want some version of this soon for per-workspace sessions!

gui/src/hooks/ParallelListeners.tsx
```diff
@@ -240,6 +243,10 @@ function ParallelListeners() {
     migrateLocalStorage(dispatch);
   }, []);
 
+  useEffect(() => {
+    void dispatch(loadLastSession());
```
RomneyDa: Since this is pretty critical loading could we add some retry logic and error handling? Maybe retry with exponential backoff or a few times or similar

RomneyDa (changes requested):

gui/src/hooks/ParallelListeners.tsx
uinstinct: do we need to retry for redux dispatch? because loadLastSession has error catching

RomneyDa: I mean, retry the loading of the session, since the dispatch won't fail, if loadLastSession catches an error for this initial load (maybe pass a boolean or unwrap dispatch) then retry once or twice

RomneyDa: @uinstinct alternatively we could loadLastSession after initial config load, as a way to detect that core is up and running (will always be with vs code but not necessarily with jetbrains). I.e. remove the use effect and add load last session into the various things triggered on initial config load

uinstinct: Implemented!

PR #7979, opened 2025-09-25
feat: husky for prettier

sestinj:
## Description

add a pre-commit hook to run husky
    

---

## Summary by cubic
Set up a Husky pre-commit hook with lint-staged so Prettier formats staged files automatically. Also expanded .prettierignore to skip generated VSCode e2e outputs and other build folders.

- **Dependencies**
  - Added husky (^9.1.7) and lint-staged (^15.5.2).

- **Migration**
  - Run npm install to install hooks (prepare script is included).
  - If hooks don’t run, execute npm run prepare.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-npx lint-staged
+npx lint-staged --no-install


RomneyDa:
We did this at some point and removed because the husky job takes like 30 seconds on any commit and logs a bunch, as opposed to the targeted format by `format on save`. Just double checking that this will be quick and targeted?

sestinj:
yeah the command here only formats staged files, I tried it and it took 1-2 seconds

PR #7993, opened 2025-09-26
feat: auto-accept agent edits setting auto accept edit tools

uinstinct:
## Description

When auto-accept agent edit setting is turned on, automatically accept the edit tools including findandreplace, multiedit and editexistingfile.

resolves CON-3918

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/40f4fc0f-f8d8-4b39-b22a-52bcb1b8e100

https://github.com/user-attachments/assets/c78398f1-277b-461d-8f7e-f53eab65b6f9

## Tests

### DIFF
--- a/gui/src/redux/thunks/streamNormalInput.ts
+++ b/gui/src/redux/thunks/streamNormalInput.ts
@@ -29,6 +29,7 @@ import { AppThunkDispatch, RootState, ThunkApiType } from "../store";
 import { constructMessages } from "../util/constructMessages";
 
 import { modelSupportsNativeTools } from "core/llm/toolSupport";
+import { BuiltInToolNames } from "core/tools/builtIn";
 import { addSystemMessageToolsToSystemMessage } from "core/tools/systemMessageTools/buildToolsSystemMessage";
 import { interceptSystemToolCalls } from "core/tools/systemMessageTools/interceptSystemToolCalls";
 import { SystemMessageToolCodeblocksFramework } from "core/tools/systemMessageTools/toolCodeblocks";
@@ -105,6 +106,8 @@ async function handleToolCallExecution(
 ): Promise<boolean> {
   // Return whether all tools were auto-approved
   const newState = getState();
+  const autoAcceptEditToolDiffs =
+    newState.config.config.ui?.autoAcceptEditToolDiffs;
   const toolSettings = newState.ui.toolSettings;
   const allToolCallStates = selectCurrentToolCalls(newState);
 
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
@uinstinct I agree YOLO mode should be expanded to include policies. With this implementation the policy behavior would conflict with the tool policy in the settings page. I think we could do this approach but only if we also show UI (maybe a tooltip/icon) to indicate that the tool policy is forced to automatic by the auto accept edits setting (and show it as automatic and disabled in the policies)

uinstinct:
> @uinstinct I agree YOLO mode should be expanded to include policies. With this implementation the policy behavior would conflict with the tool policy in the settings page. I think we could do this approach but only if we also show UI (maybe a tooltip/icon) to indicate that the tool policy is forced to automatic by the auto accept edits setting (and show it as automatic and disabled in the policies)

added!

image

gui/src/pages/config/components/ToolPolicyItem.tsx
```diff
@@ -28,12 +29,24 @@ interface ToolPolicyItemProps {
...
+    (state) => state.config.config.ui?.autoAcceptEditToolDiffs,
+  );
+  const isAutoAcceptedToolCall =
+    isAutoEditToolCall(props.tool.function.name) && autoAcceptEditToolDiffs;
+
+  const policy = isAutoAcceptedToolCall
+    ? "allowedWithoutPermission"
+    : toolPolicy;
+
+  console.log("debug1 policy->", policy, "and", props.tool.function.name);
```
RomneyDa: remove console.log

gui/src/redux/thunks/streamNormalInput.ts
```diff
@@ -268,6 +268,7 @@ export const streamNormalInput = createAsyncThunk<
       activeTools,
       generatedCalls3,
       toolPolicies,
+      state3.config.config.ui?.autoAcceptEditToolDiffs ?? false,
```
RomneyDa: no need to default to false here can just pass undefined

gui/src/pages/config/components/ToolPolicyItem.tsx
```diff
@@ -100,6 +114,20 @@ export function ToolPolicyItem(props: ToolPolicyItemProps) {
                     <InformationCircleIcon className="h-3 w-3 flex-shrink-0 cursor-help text-yellow-500" />
                   </ToolTip>
                 ) : null}
+                {isAutoAcceptedToolCall ? (
+                  <ToolTip
+                    place="bottom"
+                    className="flex flex-wrap items-center"
+                    content={
+                      <p className="m-0 p-0">
+                        Tool policy is now automatic as you have turned on
```
RomneyDa: Could simplify to e.g. `Auto-Accept Agent Edits setting is on`

RomneyDa (changes requested):
@uinstinct just the console log otherwise lgtm!

RomneyDa (approved):

PR #7994, opened 2025-09-26
fix: config loading set to false before being fetched

uinstinct:
## Description

Do not set config loading to false before config from hub is fetched.

In case of local config, config is already loaded in core before gui initializes (due [to the constructor call](https://github.com/continuedev/continue/blob/e1455c69f66c9b291211fe889addd0d540bf8012/core/config/ConfigHandler.ts/#L86)).
In case of profile config, config takes time to load (due to data fetching from remote server) and profileId is not yet set. 

resolves CON-3946

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/53a743a1-b997-4c02-918a-89e64fad2c56

https://github.com/user-attachments/assets/3d259f82-c632-4278-88b2-d163fb80f768

## Tests

---

## Summary by cubic
Fixes premature clearing of the config loading state when fetching config from the hub. The UI now stays in “Loading” until remote profile config is ready, preventing the empty-state flash and addressing CON-3946.

- **Bug Fixes**
  - Keep configLoading true when profileId is null and result.content.result.configLoadInterrupted; wait for configUpdate before clearing loading.
  - Update SelectedAssistantButton to prioritize the loading spinner over the “Create your first agent” message.


gui/src/components/AssistantAndOrgListbox/SelectedAssistantButton.tsx
```diff
@@ -23,6 +23,8 @@ export function SelectedAssistantButton({
   const buttonPadding = isSidebar ? "px-2 py-1.5" : "px-0 py-0.5";
   const buttonStyle = isSidebar ? {} : { fontSize: fontSize(-3) };
 
+  console.log("debug1 config loading", configLoading);
```
tingwai: PR LGTM, can you remove this first?

tingwai: Oh just saw you'll be out, I can remove it

uinstinct:
> @uinstinct I'm wondering if this will lead to infinite loading for local folks or folks with e.g. an error in their yaml schema. Why is the method for detecting config not yet loaded to check `configLoadInterrupted`?

Rechecked it. It does not lead to infinite loading in incorrect yaml schema

https://github.com/user-attachments/assets/002edd21-cdcb-4609-ae2f-5ddddbdbddf9


The reason why configLoadInterrupted is true is [because profileId is null](https://github.com/continuedev/continue/blob/5292eeb764a09742794652bcb83ed901ceeaa1a5/core/config/ConfigHandler.ts/#L573). We do the initial fetch and load of the config in the [ConfigHandler constructor call](https://github.com/continuedev/continue/blob/5292eeb764a09742794652bcb83ed901ceeaa1a5/core/config/ConfigHandler.ts/#L86).

In case of local only config, this is immediately available (since core is initialized before gui and that ConfigLoader constructor is already called). In case of remote config, it takes a while to initialize due to the http calls (and [config/getSerializedProfileInfo is already called in gui](https://github.com/continuedev/continue/blob/5292eeb764a09742794652bcb83ed901ceeaa1a5/gui/src/hooks/ParallelListeners.tsx/#L100)).

RomneyDa:
@uinstinct in your video, it doesn't lead to infinite loading because a profile Id was selected. I understand that profileId === null and configLoadingInterrupted = true can be used to detect loading conditions with the current profileId setup, but the condition feels fragile since configLoadingInterrupted = true in other cases as well, e.g.
image, and there is no first principles reason that profileId would not be null in other cases as well. So you would expect situations in which profileId is null and config load is interrupted and we would _not_ set configLoading = true.

Can we come up with a more robust way to detect this?

PR #7895, opened 2025-09-22
feat: remove config.ts support and esbuild runtime packaging across repo

joffeoja:
## Summary

This PR removes support for `config.ts` and all associated `esbuild` runtime packaging logic across the repository.

Per discussion in [#4173](https://github.com/continuedev/continue/pull/4173#issuecomment-3313126955), the only remaining reason for including `esbuild` in `core` was to compile `config.ts`. That feature was deprecated more than 6 months ago, so it is now fully removed.

---

### Core
- Removed logic for compiling and importing `~/.continue/config.ts` from `load.ts`.
- Added fail-fast detection: if a legacy `config.ts` is present, Continue will stop and instruct the user to migrate to `config.yaml` or `config.json`.
- Dropped `getTsConfigPath` and other helpers in `util/paths.ts`.
- Adjusted comments, error messages, and tests (`ConfigHandler.vitest.ts`).
- Cleaned up `core/package.json`:
  - Removed `esbuild` runtime dependency.
  - Retained build-time plugins where still required (e.g. `@sentry/esbuild-plugin`).

### Documentation
- Updated `configuration.mdx`:
  - Removed all references to `config.ts`.
  - Clarified that YAML is now the recommended format.
  - Noted that `config.ts` is fully removed as of 2025.

---

## Motivation

- Reduce maintenance overhead by removing `esbuild` from the runtime.
- Align codebase with current supported configuration formats (`config.yaml` and `config.json`).
- Make future dependency management and security scanning simpler.

---

## Security

This PR removes any **embedded esbuild runtime** from our VSIX and core. That:

- Eliminates scanner findings tied to older esbuild executables and their **transitive Go stdlib CVEs**, which motivated upstream discussion in [esbuild#3802](https://github.com/evanw/esbuild/issues/3802).  
- Avoids exposure to the **dev server CORS issue** ([GHSA-67mh-4wv8-2f99](https://github.com/advisories/GHSA-67mh-4wv8-2f99)) since we do **not** ship or run esbuild’s dev server in end-user artifacts.  
- Ensures esbuild is used **only as a build-time tool**, never packaged for end users.

**Supersedes [#7790](https://github.com/continuedev/continue/pull/7790)**, which attempted to mitigate by upgrading the embedded binary; this PR removes the runtime binary entirely.

---

## Related Issue

Refs #4173

PR #7790, opened 2025-09-16
Esbuild upgrade

joffeoja:
## Description

Re-upgrade esbuild to **0.24.2** after the revert in [PR #4173](https://github.com/continuedev/continue/pull/4173).

- Add Node-only resolution to `extensions/vscode/scripts/esbuild.js` so `transformers.js` consistently picks `onnxruntime-node` instead of `onnxruntime-web`.
- This avoids the missing WASM artifacts that caused the previous revert.
- Update `core/package.json`, `extensions/vscode/package.json`, and packaging scripts to use esbuild 0.24.2 and fetch matching binaries.

### Security context

The VSIX package gets flagged by vulnerability scanners due to the **old esbuild executable** bundled inside it. This is tracked upstream in [esbuild#3802](https://github.com/evanw/esbuild/issues/3802).

By upgrading from 0.17.19 → 0.24.2 (built with Go 1.23.1), we reduce or eliminate findings related to outdated Go stdlib dependencies. Specifically:

- **CVE-2023-45288** — HTTP/2 continuation DoS (`x/net/http2`)
- **CVE-2023-45289** — cookie/header forwarding issue (`net/http`)
- **CVE-2023-45290** — multipart form memory exhaustion (`net/http`)
- **CVE-2024-24783** — x509 panic on unknown key algorithm
- **CVE-2024-24784** — net/mail display-name parsing bug
- **CVE-2024-24785** — html/template JSON marshal error handling
- **CVE-2024-24789**, **CVE-2024-24790** — additional Go stdlib fixes
- **CVE-2024-45336**, **CVE-2024-45341** — flagged for `<0.24.2-r1`

⚠️ Note: **GHSA-67mh-4wv8-2f99** (dev server CORS permissive origin) still affects `esbuild ≤ 0.24.2`. This PR is a safe intermediate step; full mitigation comes with `esbuild ≥ 0.25.0`. Since Continue does not ship or expose esbuild’s dev server in the VSIX, exposure is limited.

## AI Code Review

- Will be run automatically by the Continue team.

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)  
- [ ] The relevant docs, if any, have been updated or created  
- [ ] The relevant tests, if any, have been updated or created  

## Screen recording or screenshot

N/A — change affects build scripts and dependency versions.

## Tests

- Verified `npm run vscode-extension:package` produces a working `.vsix`.
[...]

PR #8016, opened 2025-09-29
Remove bundled esbuild from VS Code extension, keep build-time only

joffeoja:
## Summary

Removes the bundled `esbuild` binary from the VS Code extension. The extension no longer ships `esbuild.exe` or `@esbuild/*` platform packages in the `.vsix`. Build-time `esbuild` usage is still retained.

## Details

- **Runtime esbuild removal**
  - Deleted all staging/copy logic for `esbuild` and `@esbuild/*` in prepackage scripts.
  - Dropped `esbuild` from runtime dependencies.
  - Added `.vscodeignore` entries to block `esbuild` binaries or `*.exe` from slipping into the packaged VSIX.

- **Build-time esbuild retained**
  - `scripts/esbuild.js` still uses `esbuild` for bundling during development/build.
  - `esbuild` remains a devDependency.

- **Config.ts behavior**
  - `config.ts` is still supported but no longer auto-installs `esbuild`.
  - If a user loads `config.ts` without having `esbuild`, the extension shows a clear error with instructions:
    ```
    config.ts has been deprecated and esbuild is no longer automatically installed by Continue.
    To use config.ts, run: npm i esbuild@<supported-version> --prefix ~/.continue
    ```

- **Other changes**
  - Renamed `download-copy-sqlite-esbuild.js` → `download-copy-sqlite.js`.
  - Removed unused `copyTreeSitterTagQryFiles` helper.
  - Updated lockfiles to reflect dependency changes.

## Verification

- Built and packaged extension locally:
  - ✅ `npm run package` produces a `.vsix`.
  - ✅ No `esbuild` or `.exe` files in the VSIX (`unzip -l continue-*.vsix | rg esbuild` → no hits).


joffeoja:
Hi @Patrick-Erichsen and @RomneyDa, here is the reworked PR for removing the esbuild executable from the vscode extension along with the downloading logic, while keeping config.ts support (assuming end user installs esbuild themselves).

The first upload had all tests pass but when I added the load.ts changes there were some fails. I don't see how they are related though :thinking: Edit: back to green.

Let me know your thoughts on this third attempt at fixing the "vulnerable" esbuild situation.

RomneyDa (dismissed):
@joffeoja appreciate the rework!! Looks good to me

extensions/vscode/.vscodeignore
```diff
@@ -24,4 +24,9 @@ e2e
...
 server/exe/**
 !out/node_modules/**
-!node_modules/@vscode/ripgrep/bin/rg
\ No newline at end of file
+!node_modules/@vscode/ripgrep/bin/rg
+
+# Do not ship esbuild in VSIX
+out/node_modules/esbuild/**
+out/node_modules/@esbuild/**
+**/*.exe
```
sestinj: ignoring all .exe's seems like overkill

core/config/load.ts
```diff
@@ -696,102 +695,43 @@ function escapeSpacesInPath(p: string): string {
...
+        "node_modules",
+        "esbuild",
+      );
+      const candidate = require.resolve("esbuild", { paths: [userEsbuild] });
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      require(candidate);
+      return true; // available via ~/.continue
+    } catch {
+      // Not available → show friendly instructions and opt out of building
+      await ide.showToast(
```
sestinj, line 724: It looks like this is going to get shown to every user with a `config.ts` file, which is basically every user. I don't think we should bother everyone with this.

Instead we should find a subtle way to notify users that are _actually_ using `config.ts`. For example, we could place a Codelens in the config.ts file, or we could check whether `config.ts` has the default file contents

sestinj (changes requested):

extensions/vscode/.vscodeignore
joffeoja: Done

core/config/load.ts
joffeoja: Ok, I have wrapped the toast logic with a check for a customized config.ts

joffeoja:
Hi @sestinj @RomneyDa & @Patrick-Erichsen, any update on this one? Would you like me to file an issue in order to track this work?

core/config/load.ts
```diff
@@ -866,8 +806,15 @@ async function buildConfigTsandReadConfigJs(ide: IDE, ideType: IdeType) {
...
-  await tryBuildConfigTs();
+  // Only bother with esbuild if config.ts is actually customized
+  if (currentContent.trim() !== DEFAULT_CONFIG_TS_CONTENTS.trim()) {
+    const ok = await handleEsbuildInstallation(ide, ideType);
+    if (!ok) {
+      // esbuild not available → we already showed a friendly message; skip building
+      return;
+    }
+    await tryBuildConfigTs();
+  }
```
Patrick-Erichsen, line 817: I think the goal from @sestinj 's feedback was just to provide the deprecation message to user's that have modified their config.ts , not to attempt to use their locally installed esbuild.

The general goal here is rip out all the config.ts logic but give a heads up to those affected.

Could we modify this to instead just show the toast with that in mind?

joffeoja: @RomneyDa @Patrick-Erichsen @sestinj : The current code is based on feedback from Dallin in my now closed PR:
https://github.com/continuedev/continue/pull/7895#pullrequestreview-3269749999

> Spoke with the team further about this and we're thinking **let's remove esbuild module installation while still supporting config.ts by assuming users already have esbuild installed** and showing them an error with instructions if they don't e.g. config.ts has been deprecated and esbuild is no longer automatically installed by Continue. To use config.ts, you can run npm i [esbuild@x.x.x](mailto:esbuild@x.x.x) in ~/.continue`. Sorry for the partial bait and switch on this

As you requested I kept all the config.ts related logic, still letting the end user use config.ts if they have installed esbuild themselves. This PR leaves a deprecation message and the next logical step would be to remove all code pertaining to config.ts.

This PR supersedes my two earlier attempts:
~~https://github.com/continuedev/continue/pull/7790~~: Upgrade esbuild to less vulnerable version
~~https://github.com/continuedev/continue/pull/7895~~: Remove esbuild + config.ts support

RomneyDa: @joffeoja chatted with @Patrick-Erichsen, sounds like we'll move forward with this

joffeoja: @RomneyDa @Patrick-Erichsen Sounds good. Do I have to do anything more on my part?

PR #8018, opened 2025-09-29
feat: rules and MCP fixes and UI updates

RomneyDa:
## Description
- New Tools section UI moves tool policies for an MCP server into the MCP server Tools section
- Hides settings Tools section resources and prompts if none are present, reduce confusion since Tools is the most common usage
- Changes `ruleFile` and `promptFile` to `sourceFile` for uniformity through the codebase
- Simplifies MCP disconnect logic to use a `disabled` status
- Fixes edit block opening wrong thing for mcp servers, rules, and prompts in various situations and unifies logic
- Fixes inaccurate parts of the init prompt
- Fixes navigation to settings pages from fatal error
- Fixes toggle switch click propagation
- Fixes various bad tailwind classname merging
- Removes unused Prompts section (was merged into rules)
    

---

## Summary by cubic
Hide MCP Prompts and Resources in the Config > Tools view when none exist to reduce clutter and keep Tools as the focus. Aligns with CON-4022.

- **New Features**
  - Render Prompts only when server.prompts.length > 0.
  - Render Resources only when server.resources.length > 0 or server.resourceTemplates.length > 0.


RomneyDa:
Doing this on a separate pr

PR #8077, opened 2025-10-03
Add invokable rule support to cli

tomasz-stefaniak:
## Description

Add invokable rule support to cli

## AI Code Review

## Checklist

## Tests

### DIFF
--- a/extensions/cli/src/commands/commands.ts
+++ b/extensions/cli/src/commands/commands.ts
@@ -146,7 +146,10 @@ export function getAllSlashCommands(
       category: "assistant" as const,
     })) || [];
 
-  return [...systemCommands, ...assistantCommands];
+  // Get invokable rule commands
+  const invokableRuleCommands = getInvokableRuleSlashCommands(assistant);
+
+  return [...systemCommands, ...assistantCommands, ...invokableRuleCommands];
 }
 
 /**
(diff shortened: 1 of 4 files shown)


extensions/cli/src/commands/commands.ts
```diff
@@ -163,3 +166,33 @@ export function getAssistantSlashCommands(
...
+      // Only include rules with invokable: true
+      return rule.invokable === true;
+    })
+    .map((rule) => {
+      // TypeScript now knows rule is an object with invokable: true
+      const ruleObj = rule as any;
+      return {
+        name: ruleObj.name || "",
+        description: ruleObj.description || "",
+        category: "assistant" as const,
```
babblebey, line 195: ### Some Concerns x Clarifying Question 🤔 

I can understand that this (rules) mostly firsthand would be coming from the continue-hub, which means the name x command just might seem unique but incases where we are reading invokable from local...

- How do we handle conflicts in same slash-command???
- Well yes, since you cannot have 2 files with same name in local, there might not be conflict... but what happens when a rule from the continue-hub meets with an invokable one in local???
- Think what also happens when a systemCommand, assistantCommand conflicts an invokableRuleCommand?

I'm looking forward to having invokablePromptsCommand support though (see https://github.com/continuedev/continue/discussions/8080), this where I figure the different `init` slash command...

- One's a system prompt in the IDE extension that "Initialize Codebase"
- Thee other's the slash command that "creates the AGENTS.md file" in the continue-cli

tomasz-stefaniak: @babblebey this is not a problem (yet) because we currently do not support local commands in the CLI. That said, there is an implicit order of resolution that goes roughly like this:

  1. System commands (highest priority)
  2. Assistant prompts
  3. Invokable rules (lowest priority)

I believe it is a part of our roadmap to also add local slash commands to the CLI, and when that moment comes we may need to come up with a better resolution system.

babblebey: It's not a problem until its a problem 😉 

Good to know... and I definitely look forward to the local prompt support though... It give that feel of portability of continue with projects.

extensions/cli/src/slashCommands.ts
```diff
@@ -233,6 +233,21 @@ export async function handleSlashCommands(
     return { newInput };
   }
 
+  // Check for invokable rules
+  const invokableRule = assistant.rules?.find((rule) => {
+    // Handle both string rules and rule objects
+    if (!rule || typeof rule === "string") {
+      return false;
+    }
+    const ruleObj = rule as any;
```
RomneyDa, line 242: Can avoid any type with RuleObject type or using zod schema from config-yaml

PR #8110, opened 2025-10-06
fix: await mcp loading in headless

RomneyDa:
## Description

So that mcps load in cicd

remake of https://github.com/continuedev/continue/pull/8028


RomneyDa:
This was done with a better approach on a different PR at some point

PR #8115, opened 2025-10-07

RomneyDa:
https://github.com/continuedev/continue/pull/7910
    

---

## Summary by cubic
Fixes image drag-and-drop in the main TipTap editor so users can drop local files and images from VS Code and have them insert correctly.

- **Bug Fixes**
  - Handle drop/dragover inside a ProseMirror plugin and hide the overlay on drop.
  - Support VS Code “vscode-resource” image drops: parse HTML, resolve the file path, request the file as a data URL, and insert it as an image node.
  - Add readFileAsDataUrl protocol and VS Code handler to read files and return data URLs (png/jpeg).
  - Simplify DragOverlay and make it non-interactive to avoid blocking drops.
  - Gate behavior behind modelSupportsImages to avoid unsupported models.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,7 +18,7 @@
       ],
       "pauseForSourceMap": false,
       "outFiles": ["${workspaceFolder}/extensions/vscode/out/extension.js"],
-      "preLaunchTask": "vscode-extension:build-without-watch",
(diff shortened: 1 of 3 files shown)


RomneyDa:
@aadarshkt closing this for now, feel free to take another swing at it!
One of the reasons I haven't picked this up is because the code is vs-code specific

PR #8138, opened 2025-10-08
perf(cli): long start due to synchronous config init in normal flow run

uinstinct:
## Description

When doing `runNormalFlow`, we need the config only for [rules injection](https://github.com/continuedev/continue/blob/f0d02e704273b5f926cd4e7d5aca2f5a4d1d343b/extensions/cli/src/onboarding.ts/#L298-L300).

We already fetch the config again when [initializing services for config](https://github.com/continuedev/continue/blob/f0d02e704273b5f926cd4e7d5aca2f5a4d1d343b/extensions/cli/src/services/index.ts/#L166-L167).

resolves CON-4174

## AI Code Review

## Checklist

reduction of init time by approx 2 seconds

**before**

image

**after**
image

## Tests

### DIFF
--- a/extensions/cli/src/onboarding.ts
+++ b/extensions/cli/src/onboarding.ts
@@ -280,24 +280,35 @@ export async function initializeWithOnboarding(
   authConfig: AuthConfig,
   configPath: string | undefined,
   rules?: string[],
-): Promise<OnboardingResult> {
+) {
   const firstTime = await isFirstTime();
 
-  let result: OnboardingResult;
-
   if (firstTime) {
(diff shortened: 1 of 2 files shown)


uinstinct:
> @uinstinct cubic comment seems right
> 
> Do we actually need the rules injected in config for onboarding? Could we just not load config/only load model or similar?

Right. 
I dove a little deeper and rules injection is unnecessary because we are injecting the rules in the SystemMessageService. Also running the normalFlow is unnecessary because all parts of it are done in ConfigService, MCPService and AuthService in service initialization.

PR #8221, opened 2025-10-13
fix: treat premature close during tool calls as a tool error

RomneyDa:
## Description
Catches premature close error during streaming if there are tool calls, and marks them as errored with instructions to break up tool call into small chunks. This is because the common premature close case is during create/edit operations and only happens in large outputs/files

image

image

### DIFF
--- a/core/llm/defaultSystemMessages.ts
+++ b/core/llm/defaultSystemMessages.ts
@@ -71,8 +71,6 @@ ${BRIEF_LAZY_INSTRUCTIONS}
 
 However, only output codeblocks for suggestion and demonstration purposes, for example, when enumerating multiple hypothetical options. For implementing changes, use the edit tools.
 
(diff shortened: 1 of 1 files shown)


core/llm/defaultSystemMessages.ts
```diff
@@ -71,6 +71,8 @@ ${BRIEF_LAZY_INSTRUCTIONS}
 
 However, only output codeblocks for suggestion and demonstration purposes, for example, when enumerating multiple hypothetical options. For implementing changes, use the edit tools.
 
+If a file edit or creation operation will be quite large, split it into multiple smaller tool calls to prevent errors with long output streams.
```
sestinj: I think this is going to have consequences we don't want. Mostly worried the models will behave in ways that are different from their optimal.

PR #8234, opened 2025-10-14
fix: check key type for onboarding modal

RomneyDa:
## Description

    

---

## Summary by cubic
Fixes the onboarding modal trigger by correctly detecting free trial API keys vs models add-on, so the “out of starter credits” modal only shows for true free trial users.

- **Bug Fixes**
  - Added usesFreeTrialApiKey to identify free trial keys via secret type.
  - Updated isOutOfStarterCredits to use the free trial key check.
  - Switched streamChat and useCredits to the new check to prevent false triggers for models add-on users.

### DIFF
--- a/core/config/usesFreeTrialApiKey.ts
+++ b/core/config/usesFreeTrialApiKey.ts
@@ -29,6 +29,31 @@ export function usesCreditsBasedApiKey(
   return false;
 }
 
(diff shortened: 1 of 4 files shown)


RomneyDa:
This was fixed by @uinstinct

PR #8277, opened 2025-10-16
fix: jetbrains browser freezing fix attempt 1

RomneyDa:
## Description

    

---

## Summary by cubic
Reduce JCEF WebView freezing in IntelliJ by disabling off-screen rendering and adding an idle maintenance routine to keep the browser responsive.

- **Bug Fixes**
  - Switched to on-screen rendering (setOffScreenRendering(false)).
  - Added a maintenance scheduler (runs every 30 min); after 1 hour idle, triggers browser GC and adjusts JS query pool size.
  - Tracked last interaction on sendToWebview to measure idle time.
  - Ensured clean shutdown of the scheduler during dispose.


RomneyDa:
closing as was a shot in the dark attempt and probably not the right approach

Issue #2651, opened 2024-10-24
Gemini output swallows square bracket


robertpiosik:
### Before submitting your bug report

- [X] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [X] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [X] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:
- Continue version:
- IDE version:
- Model:
- config.json:
```

### Description

One square bracket is swallowed with gemini models.

mubiru-simeon:
Same issue here. IT keeps ignoring one ]. I thought it was a continue problem but i saw it happen once with Cline too. And i was using Gemini pro on that one.

wilksu:
When i use gemini web,i have same issue，i lost [] in code block.

spottenn:
I am consistently having this issue with about 20% of ending square brackets. I'm working in TypeScript and react native. I have the problem with Gemini 2.0 Flash, Flash thinking, Pro, and Gemini 2.5 Pro. The problem has persisted across multiple versions of Continue (at least Jetbrains 1.0.5 to 1.0.10). It occurs on both VScode and Webstorm. 

> Same issue here. IT keeps ignoring one ]. I thought it was a continue problem but i saw it happen once with Cline too. And i was using Gemini pro on that one.

I wonder if there is a way to mitigate it since you had seen it only once with Cline. Do you use Cline as frequently as Continue? Are you still having the issue in both?


> When i use gemini web,i have same issue，i lost [] in code block.

I wonder if this is the same issue or not. My use is only losing ending square brackets, both with empty and non-empty arrays. Your issue seems to be described here. https://support.google.com/gemini/thread/329152739/gemini-outputting-empty-square-brackets-is-broken?hl=en 

My guess is that the issues are related, but not identical. 

@mubiru-simeon @wilksu @robertpiosik Are you still having the issue with the square brackets? Have any of you found a workaround? Also see my questions above.

spottenn:
Relevant environment info where the issue has occurred
- OS: Windows 10 & 11
- Continue version: At least Jetbrains 1.0.5 to 1.0.10, maybe earlier
- IDE version: Webstorm 2024.3.5
[...]

PR #8280, opened 2025-10-16
fix: bracket stripping in gemini responses

uinstinct:
## Description

Use streamSse instead of processGeminiResponse when receiving responses from Gemini streamChat .

resolves CON-4266
closes https://github.com/continuedev/continue/issues/2651

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/5bdc06f0-455f-41c1-97ef-dccdae2eabb8

https://github.com/user-attachments/assets/20589cfd-e968-4567-9170-ecb45f76acfc

## Tests

---

## Summary by cubic
Switched Gemini streamChat to use streamSse for proper SSE handling and direct chunk streaming, fixing broken/laggy responses and simplifying the streaming path.

- **Bug Fixes**
  - Iterate streamSse(response) and yield chunks directly.
  - Remove processGeminiResponse in the Gemini streaming flow.

### DIFF
--- a/core/llm/llms/Gemini.ts
+++ b/core/llm/llms/Gemini.ts
@@ -312,75 +312,57 @@ class Gemini extends BaseLLM {
   }
 
   public async *processGeminiResponse(
-    stream: AsyncIterable<string>,
+    response: Response,
   ): AsyncGenerator<ChatMessage> {
-    let buffer = "";
-    for await (const chunk of stream) {
-      buffer += chunk;
-
-      const parts = buffer.split("\n,");
-
-      let foundIncomplete = false;
-      for (let i = 0; i < parts.length; i++) {
-        const part = parts[i];
-        let data: GeminiChatResponse;
-        try {
-          data = JSON.parse(part) as GeminiChatResponse;
-        } catch (e) {
-          foundIncomplete = true;
-          continue; // yo!
-        }
-
-        if ("error" in data) {
-          throw new Error(data.error.message);
-        }
-
-        // In case of max tokens reached, gemini will sometimes return content with no parts, even though that doesn't match the API spec
-        const contentParts = data?.candidates?.[0]?.content?.parts;
-        if (contentParts) {
-          const textParts: MessagePart[] = [];
-          const toolCalls: ToolCallDelta[] = [];
-
-          for (const part of contentParts) {
-            if ("text" in part) {
-              textParts.push({ type: "text", text: part.text });
-            } else if ("functionCall" in part) {
-              toolCalls.push({
-                type: "function",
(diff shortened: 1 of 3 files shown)


uinstinct:
> Wondering how this is working when it removes the translation of gemini chunks into chat chunks?

Ah, missed it. It works for normal chat messages but not for tool calls. Fixed that!

> Let's keep processGeminiResponse/implement the fix for processGeminiResponse since it is used by VertexAI
> Let's also implement for openai adapters!

implemented!

RomneyDa (changes requested):
Something still feels off here, I think streamSse already does the buffer splitting which makes e.g. this line redundant:
`const parts = buffer.split("\n,");`

Thinking you should be able to remove streamResponse entirely from openai adapter and just use streamSse?

I think Vertex LLM class usage of processGeminiResponse would also require using streamSse. Alternatively, you could change processGeminiResponse to take response directly and use streamSse on it

uinstinct:
> Something still feels off here, I think streamSse already does the buffer splitting which makes e.g. this line redundant:
const parts = buffer.split("\n,");

it is needed for [parsing the gemini chat response](https://github.com/continuedev/continue/blob/31c7383e92913b2fee4b2603cde04394a9691921/core/llm/llms/Gemini.ts/#L323-L332)

> Thinking you should be able to remove streamResponse entirely from openai adapter and just use streamSse?

> I think Vertex LLM class usage of processGeminiResponse would also require using streamSse. Alternatively, you could change processGeminiResponse to take response directly and use streamSse on it

implemented!

RomneyDa:
> > Something still feels off here, I think streamSse already does the buffer splitting which makes e.g. this line redundant:
> > const parts = buffer.split("\n,");
> 
> it is needed for [parsing the gemini chat response](https://github.com/continuedev/continue/blob/31c7383e92913b2fee4b2603cde04394a9691921/core/llm/llms/Gemini.ts/#L323-L332)
> 

What I meant is I think streamSse already splits chunks by new lines so the split at "\n," is extraneous
https://github.com/continuedev/continue/blob/579bcbfcdda4dfe46d10d92f4b7eb329ecfeaebf/packages/fetch/src/stream.ts#L124-L150

core/llm/llms/Gemini.ts
```diff
@@ -317,15 +317,6 @@ class Gemini extends BaseLLM {
     let buffer = "";
     for await (const chunk of stream) {
```
RomneyDa: why don't we just streamSse(stream) directly in processGeminiReponse?

uinstinct: implemented!

PR #8298, opened 2025-10-16
feat: allow file reads outside IDE workspace (w/ permission)

Patrick-Erichsen:
closes CON-4324 CON-3616 CON-4465
    

## Summary by cubic
Allow reading and listing files outside the IDE workspace with explicit permission. Adds robust path resolution for absolute, tilde (~/), and file:// paths, and enforces stricter access policies for non-workspace paths.

- **New Features**
  - Added pathResolver to normalize user-provided paths and detect workspace boundaries; supports relative, absolute, tilde, and file:// URIs (with tests).
  - Introduced evaluateFileAccessPolicy: files outside the workspace always require permission; integrated via preprocessArgs + evaluateToolCallPolicy for ls, readFile, readFileRange, and viewSubdirectory.
  - Core now supports tool arg preprocessing and passes processed args to policy evaluation; terminal command display uses processed args.
  - Updated tool implementations to use resolveInputPath with clearer errors and consistent display paths; added a CLI path resolver for local runs.

### DIFF
--- a/core/indexing/docs/crawlers/DocsCrawler.test.ts
+++ b/core/indexing/docs/crawlers/DocsCrawler.test.ts
@@ -19,7 +19,7 @@ const TIMEOUT_MS = 1_000_000_000;
 // so that we don't delete the Chromium install between tests
 ChromiumInstaller.PCR_CONFIG = { downloadPath: os.tmpdir() };
 
-describe("DocsCrawler", () => {
+describe.skip("DocsCrawler", () => {
   let config: ContinueConfig;
   let mockIde: FileSystemIde;
   let chromiumInstaller: ChromiumInstaller;


core/core.ts
```diff
@@ -1102,14 +1102,26 @@ export class Core {
...
+        let processedArgs = args;
+        if (tool.preprocessArgs) {
+          try {
+            processedArgs = await tool.preprocessArgs(args, { ide: this.ide });
+          } catch (e) {
+            // If preprocessing fails, use original args
+            console.warn(`Failed to preprocess args for ${toolName}:`, e);
+            processedArgs = args;
+          }
+        }
```
Patrick-Erichsen: @RomneyDa - we now preprocess args before they reach the `evaluateToolCallPolicy`

This seems like the most sus part of the PR, but we need access to the IDE to make the runtime check for file permissions.

Patrick-Erichsen: Update: this was indeed sus, I think the correct fix was done here: [bf9a19a](https://github.com/continuedev/continue/pull/8298/commits/bf9a19ada539d4a2fc2665c7a87969fe7ae8035f)

--
Problem & Solution Summary

  Problem: preprocessArgs was running twice - once in GUI, then again in core.ts during policy evaluation because GUI was passing parsedArgs instead of processedArgs to the
  policy handler.

  Solution:
  - Fixed evaluateToolPolicies.ts to pass processedArgs (line 42)
  - Removed duplicate preprocessing from core.ts (lines 1105-1114)

  Now preprocessing happens only once, and the preprocessed args (with _resolvedPath etc.) are correctly used for policy evaluation.

--

I tried this and things were still functioning correctly, eg `evaluateToolCallPolicy` on `readFileTool` was being passed the extra property we we add in the `preprocessArgs` fn on the tool def.

gui/src/redux/thunks/evaluateToolPolicies.ts
```diff
@@ -38,14 +38,14 @@ async function evaluateToolPolicy(
     )?.defaultToolPolicy ??
     DEFAULT_TOOL_SETTING;
 
-  // Use already parsed arguments
-  const parsedArgs = toolCallState.parsedArgs || {};
+  // Use preprocessed arguments if available, otherwise fall back to parsed arguments
+  const args = toolCallState.processedArgs || toolCallState.parsedArgs || {};
```
RomneyDa: processedArgs do not necessarily include parsedArgs as a subset, could lead to confusion

core/tools/definitions/readFileRange.ts
```diff
@@ -49,4 +52,23 @@ export const readFileRangeTool: Tool = {
...
+    return {
+      ...args,
+      _resolvedPath: resolvedPath,
+    };
+  },
+  evaluateToolCallPolicy: (
+    basePolicy: ToolPolicy,
+    parsedArgs: Record<string, unknown>,
+  ): ToolPolicy => {
+    const resolvedPath = parsedArgs._resolvedPath as any;
```
RomneyDa: can we remove any type casting somehow?

core/tools/definitions/readFile.ts
```diff
@@ -32,4 +35,23 @@ export const readFileTool: Tool = {
...
+    return {
+      ...args,
+      _resolvedPath: resolvedPath,
+    };
+  },
+  evaluateToolCallPolicy: (
+    basePolicy: ToolPolicy,
+    parsedArgs: Record<string, unknown>,
+  ): ToolPolicy => {
+    const resolvedPath = parsedArgs._resolvedPath as any;
```
RomneyDa: can we remove any type casting somehow?

core/util/pathResolver.ts
```diff
@@ -0,0 +1,134 @@
...
+async function isPathWithinWorkspace(
+  ide: IDE,
+  absolutePath: string
+): Promise<boolean> {
+  const workspaceDirs = await ide.getWorkspaceDirs();
+  const normalizedPath = path.normalize(absolutePath).toLowerCase();
+
+  for (const dir of workspaceDirs) {
+    const normalizedDir = path.normalize(dir).toLowerCase();
+    if (normalizedPath.startsWith(normalizedDir)) {
```
RomneyDa: findUriInDirs logic already handles tons of edge cases with this. I would convert it to URI and use findUriInDirs rather than finding absolute path in workspace dirs (which are URIs). This is nice because findUriInDirs already determinis if it's inside any of the dirs or not

gui/src/redux/thunks/evaluateToolPolicies.ts
```diff
@@ -38,14 +38,18 @@ async function evaluateToolPolicy(
     )?.defaultToolPolicy ??
     DEFAULT_TOOL_SETTING;
 
-  // Use already parsed arguments
-  const parsedArgs = toolCallState.parsedArgs || {};
+  // Merge parsed and preprocessed arguments to ensure we have both original args and any added metadata
+  // processedArgs may add metadata like _resolvedPath but should include all parsedArgs fields
+  const args = {
```
RomneyDa: Can we pass them through separately instead of merging? Not sure they are expected to be overlapping/mergable necessarily

core/util/pathResolver.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  return null;
+}
+
+/**
+ * Normalizes a path for display purposes.
+ * Contracts home directory to ~ on Unix-like systems.
+ */
+export function normalizeDisplayPath(fullPath: string): string {
```
RomneyDa: This utility seems unused, should it be used or removed?

core/util/pathResolver.ts
```diff
@@ -0,0 +1,124 @@
...
+      uri,
+      displayPath,
+      isAbsolute: true,
+      isWithinWorkspace,
+    };
+  }
+
+  // Expand tilde paths
+  let expandedPath = trimmedPath;
+  if (trimmedPath.startsWith("~/")) {
```
RomneyDa: noting no tilde support for windows, probably fine since tilde is usually unix, but e.g. ~\this\that would be missed here

core/util/pathResolver.ts
```diff
@@ -0,0 +1,124 @@
...
+  const isAbsolute =
+    path.isAbsolute(expandedPath) ||
+    // Windows network paths
+    expandedPath.startsWith("\\\\") ||
+    // Windows drive letters (C:, D:, etc.)
+    /^[a-zA-Z]:/.test(expandedPath);
+
+  if (isAbsolute) {
+    // For Windows network paths, handle specially
+    if (expandedPath.startsWith("\\\\")) {
```
RomneyDa: this check is duplicate, can move windows network path handling outside/above other absolute paths

core/util/pathResolver.ts
```diff
@@ -0,0 +1,124 @@
...
+      const networkPath = expandedPath.replace(/\\/g, "/");
+      const uri = "file:" + networkPath; // file://server/share format
```
RomneyDa: note networkPath is not normalized to URI path style, probably fine but could add URI.normalize here

core/util/pathResolver.ts
```diff
@@ -0,0 +1,124 @@
...
+    return {
+      uri,
+      displayPath: expandedPath,
+      isAbsolute: true,
+      isWithinWorkspace,
+    };
+  }
+
+  // Fall back to relative path resolution within workspace
+  const workspaceUri = await resolveRelativePathInDir(expandedPath, ide);
```
RomneyDa, line 73: One issue here is that resolveRelativePathInDir checks if file exists while isUriWithinWorkspace does not. e.g. isUriWithinWorkspace is just doing string parsing, while resolveRelativePathInDir has a file exists check. This means you will not get file exists error if not in workspace (or there will be duplicate exists checks, not sure)

extensions/cli/src/util/pathResolver.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Resolves user-provided paths for CLI context.
+ * Handles absolute paths, tilde paths, and relative paths.
+ */
+export function resolveInputPath(inputPath: string): string | null {
+  // Trim whitespace
+  const trimmedPath = inputPath.trim();
+
+  // Expand tilde paths
+  let expandedPath = trimmedPath;
+  if (trimmedPath.startsWith("~/") || trimmedPath.startsWith("~\\")) {
```
RomneyDa, line 15: Nitpick but could merge this section of logic with a `expandTildePaths` util or similar

gui/src/redux/thunks/evaluateToolPolicies.ts
RomneyDa: nonblocking

core/util/pathResolver.ts
RomneyDa: This is the only blocking feedback

gui/src/redux/thunks/evaluateToolPolicies.ts
Patrick-Erichsen: just leaving as const parsedArgs = toolCallState.parsedArgs || {}; for now

core/util/pathResolver.ts
Patrick-Erichsen: i dont think that is a valid windows path

```
Windows doesn't interpret ~ as the home directory like Unix/Linux systems do. Valid Windows paths typically start with:
  - A drive letter: C:\this\that
  - A UNC path: \\server\share\this\that
  - A relative path: this\that or .\this\that

  To reference the user's home directory in Windows, you'd use environment variables like %USERPROFILE%\this\that or %HOMEPATH%\this\that.
```

Patrick-Erichsen:
After pulling in `untildify` and `normalize-path`
Screenshot 2025-10-17 at 2 31 58 PM

RomneyDa (dismissed):

Patrick-Erichsen:
Note that I added a `.skip` to `DocsCrawler.test.ts` - it flaked 3+ times on me and this feature is deprecated so I think we should just skip the test rather than try to resolve the flake.

PR #8303, opened 2025-10-17
fix: prevent anthropic 3.5 models from use thinking

uinstinct:
## Description

Sonnet 3.5 and Haiku 3.5 do not support thinking and hence should not auto enable the thinking tag. This PR fixes that.

resolves CON-4428

## AI Code Review

## Checklist

**before**

image

**after**

image

## Tests

### DIFF
--- a/core/llm/autodetect.ts
+++ b/core/llm/autodetect.ts
@@ -143,7 +143,11 @@ function modelSupportsReasoning(
     return false;
   }
   // do not turn reasoning on by default for claude 3 models
-  if (model.model.includes("claude") && !model.model.includes("-3-")) {
+  if (
+    model.model.includes("claude") &&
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
Let's add claude 3 to exclusion.
It might make more sense to change it to  `if model.includes("claude") && ...` instead of checking underlying provider name, since the same issue will happen for other providers like bedrock etc

uinstinct:
> Let's add claude 3 to exclusion.

makes sense. it can be turned on through the config

> It might make more sense to change it to if model.includes("claude") && ... instead of checking underlying provider name, since the same issue will happen for other providers like bedrock etc

implemented

RomneyDa:
Now it will miss 3.5 model strings (some apis use 3.5, some 3-5

PR #8339, opened 2025-10-19

aadarshkt:
## Description

Removed console log and changed launch configuration

## AI Code Review

## Checklist

## Tests

---

## Summary by cubic
Fixes image drag-and-drop in the main editor. Drops now work for files and for images dragged from VS Code (vscode-resource URLs), which are read and inserted as data URLs.

- **Bug Fixes**
  - Handle drops in a ProseMirror plugin and insert images when the selected model supports images.
  - Resolve images dragged from VS Code (file+.vscode-resource URLs) by reading the file and converting to a data URL.
  - Make the drag overlay non-blocking and hide it reliably on drop/leave.

- **Refactors**
  - Added readFileAsDataUrl protocol and VS Code handler (PNG/JPEG) to return data URLs.
  - Added helpers to parse vscode-resource URLs and extract them from dropped HTML.
  - Simplified DragOverlay component and set pointer-events: none.
  - Added a gui:build VS Code task and increased Node memory to avoid OOM during builds.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,7 +18,7 @@
       ],
       "pauseForSourceMap": false,
       "outFiles": ["${workspaceFolder}/extensions/vscode/out/extension.js"],
-      "preLaunchTask": "vscode-extension:build-without-watch",
+      "preLaunchTask": "vscode-extension:build-with-packages",
       "env": {
         // "CONTROL_PLANE_ENV": "local",
         "CONTINUE_GLOBAL_DIR": "${workspaceFolder}/extensions/.continue-debug"
--- a/gui/src/components/mainInput/TipTapEditor/utils/imageUtils.ts
+++ b/gui/src/components/mainInput/TipTapEditor/utils/imageUtils.ts
@@ -41,11 +41,7 @@ export async function handleVSCodeResourceUrl(
     return undefined;
   }
 
-  console.log("Extracted filepath:", filepath);
-
   try {
-    console.log("Requesting readFileAsDataUrl for filepath:", filepath);
-    console.log("About to call ideMessenger.request...");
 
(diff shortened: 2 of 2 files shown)


RomneyDa:
@aadarshkt note failing prettier checks!

.vscode/tasks.json
```diff
@@ -174,6 +174,25 @@
         }
       ]
     },
+    {
```
RomneyDa, line 177: I think we can leave this build script out.

extensions/vscode/src/extension/VsCodeMessenger.ts
```diff
@@ -686,6 +686,18 @@ export class VsCodeMessenger {
         });
     });
 
+    this.onWebviewOrCore("readFileAsDataUrl", async (msg) => {
+      const { filepath } = msg.data;
+      const fileUri = vscode.Uri.file(filepath);
+      const fileContents = await vscode.workspace.fs.readFile(fileUri);
+      const fileType =
+        filepath.split(".").pop() === "png" ? "image/png" : "image/jpeg";
```
RomneyDa, line 694: JPEG fallback will cause various types to fail, could we add a few more common fileTypes here?

RomneyDa (changes requested):
@aadarshkt appreciate the rework, added some comments and could you check on the cubic feedback?

.vscode/tasks.json
aadarshkt: Sure will make this changes this weekend.

RomneyDa:
@aadarshkt this is a bit stale and pending updates, I will close for now and free to re-open when ready! I think it would be a great addition to the VS Code extension.

PR #8376, opened 2025-10-21
fix: mcp agent policies and deduplicate blocks when merging unrolled assistants

RomneyDa:
## Description
Currently agents fail because default account models have a secrets issue
    

---

## Summary by cubic
Deduplicates blocks when merging unrolled assistants to prevent duplicate models, rules, prompts, and more. Incoming blocks take priority, and env/requestOptions are merged predictably.

- **Bug Fixes**
  - Merge arrays across all block types (models, context, rules, prompts, docs, data, mcpServers) with deduplication via BlockDuplicationDetector; incoming items are kept, duplicates are filtered.
  - Filter nulls and set block arrays to undefined when empty after merge.
  - Merge env with incoming overriding base; merge requestOptions with base preferred and headers combined.
  - Add comprehensive Jest tests covering dedup, ordering, env/requestOptions behavior, and edge cases.

### DIFF
--- a/extensions/cli/src/services/index.test.ts
+++ b/extensions/cli/src/services/index.test.ts
@@ -1,7 +1,8 @@
 import { vi } from "vitest";
 
-import { initializeServices, services } from "./index.js";
 import { MCPService } from "./MCPService.js";
+
+import { initializeServices, services } from "./index.js";
 
 describe("initializeServices", () => {
   describe("mode conversion", () => {


extensions/cli/src/services/ToolPermissionService.ts
```diff
@@ -61,55 +64,80 @@ export class ToolPermissionService
...
+          tool: t.name,
+          permission: "exclude",
+        }));
+        policies.push(...disallowed);
       }
     }
 
-    if (policies.length > 0) {
-      logger.debug(`Generated ${policies.length} agent file tool policies`);
+    const hasMcp = !!parsedTools.mcpServers?.length;
```
sestinj, line 130: i feel like this is pretty dense code. if it has to be that dense, could it get an extra comment or two?

sestinj (changes requested):

sestinj (approved):

PR #8416, opened 2025-10-23
fix(cli): exclude ask first tools in headless

RomneyDa:
The model should never see tools which it will not be able to call. Replaces the failure/exit with a simple denial in the case of tool policy evaluation, and excludes tools if "ask" and headless
    

---

## Summary by cubic
Prevents the model from seeing tools it can’t call. In headless mode, tools with “ask” permission are excluded, and policy evaluation returns a denial instead of exiting.

- **Bug Fixes**
  - Exclude “ask” tools from the request set in headless so the model never requests uncallable tools.
  - Replace headless permission exit with a clean denial from policy evaluation.
  - Recompute tools on each streaming iteration to reflect mode changes immediately.

- **Refactors**
  - Introduced getRequestTools and getAllAvailableTools; removed getAllTools and filterExcludedTools.
  - checkToolPermission now takes explicit permissions and uses ALL_BUILT_IN_TOOLS for dynamic evaluation.
  - Centralized ChatCompletionTool/MCP tool conversion; updated tests to use the new APIs.

### DIFF
--- a/docs/troubleshooting.mdx
+++ b/docs/troubleshooting.mdx
@@ -7,7 +7,8 @@ description: "Comprehensive guide to resolving common issues with Continue, incl
 2. [Try the latest pre-release](#download-the-latest-pre-release)
 3. [Download an older version](#download-an-older-version)
 4. [Resolve keyboard shortcut issues](#keyboard-shortcuts-not-resolving)
-5. [Check FAQs for common issues](/faqs)
+5. [MCP Server connection issues](#mcp-server-connection-issues)
+6. [Check FAQs for common issues](/faqs)
 
 ## Check the logs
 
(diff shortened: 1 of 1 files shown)


extensions/cli/src/ui/__tests__/TUIChat.planModeBug.test.tsx
Patrick-Erichsen, line 1: Does this one need to be kept?

Patrick-Erichsen (approved):

extensions/cli/src/ui/__tests__/TUIChat.planModeBug.test.tsx
RomneyDa: The test was there to produce a bug and was skipped (I guess because the bug was fixed)

PR #7891, opened 2025-09-22
OpenAI Responses API integration to enable GPT-5 features

ferenci84:
## Description

# OpenAI Responses API integration to enable GPT‑5 features

This PR adds first-class support for the OpenAI Responses API (streaming and non‑streaming) and routes supported models (o‑series, GPT‑5) through it. The Responses API is required to fully use GPT‑5 features like native reasoning, structured output items, and robust tool calling.

## What changed
- Responses API request/response handling for both streaming and non‑streaming flows
- Support for reasoning items and structured outputs; tool calls with proper argument streaming
- Persisted provider metadata to correctly pair reasoning and output items across turns
- Type‑safe request bodies and simplified event handling
- Kept Chat Completions for legacy/non‑supported models (backward compatible)

## Impact
- Unlocks GPT‑5 capabilities end‑to‑end in Continue
- More reliable tool calling and reasoning visibility
- Minimal UI/state updates to surface new metadata and avoid duplicate output

No breaking changes for existing non‑o‑series models.

Please note that the changes was made on top of other PR changes https://github.com/continuedev/continue/pull/7073 as it contained changes required by this feature.

## AI Code Review

- **Team members only**: AI review runs automatically when PR is opened or marked ready for review
- Team members can also trigger a review by commenting `@continue-general-review` or `@continue-detailed-review`

## Checklist

- [X] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [X] The relevant docs, if any, have been updated or created
- [X] The relevant tests, if any, have been updated or created

## Screen recording or screenshot

[ When applicable, please include a short screen recording or screenshot - this makes it much easier for us as contributors to review and understand your changes. See [this PR](https://github.com/continuedev/continue/pull/6455) as a good example. ]

## Tests

[ What tests were added or updated to ensure the changes work as expected? ]

PR #8417, opened 2025-10-24
feat: impl Responses API in oai-adapters

Patrick-Erichsen:
## Overview

This PR patches over the logic from https://github.com/continuedev/continue/pull/7891 into openai-adapters. It converts Responses objects back into chat completion objects for streaming, implements ternary logic for streaming vs non-streaming defined in `packages/openai-adapters/src/apis/base.ts`, is non-stateful, and generally tries to adhere to the logic from #7891 

## Testing

Create a `gpt5-codex-openai.yaml` somewhere, e.g. `continue/` root
```yaml
name: GPT5 Codex Manual Test
version: 1.0.0
schema: v1

models:
  - name: openai-gpt5-codex
    provider: openai
    model: gpt-5-codex
    apiKey: <KEY>
    roles:
      - chat
```

then run `npm run start -- --config /path/to/continue//gpt5-codex-openai.yaml` and perform general smoke testing.


extensions/cli/src/stream/streamChatResponse.test.ts
```diff
@@ -4,6 +4,20 @@ import type { ChatHistoryItem } from "core/index.js";
 import type { ChatCompletionChunk } from "openai/resources/chat/completions.mjs";
 import { vi } from "vitest";
 
+vi.mock("fdir", () => ({
```
RomneyDa: what is this mock doing, what is fdir?

packages/openai-adapters/src/apis/openaiResponses.ts
```diff
@@ -0,0 +1,812 @@
...
+  ResponseUsage,
+} from "openai/resources/responses/responses.js";
+
+const RESPONSES_MODEL_REGEX = /^(?:gpt-5|gpt-5-codex|o)/i;
+
+export function isResponsesModel(model: string): boolean {
+  return !!model && RESPONSES_MODEL_REGEX.test(model);
+}
+
+function ensureText(value: string | null | undefined): string {
```
RomneyDa: this function could just be `value || " "` inline

packages/openai-adapters/src/apis/openaiResponses.ts
```diff
@@ -0,0 +1,812 @@
...
+}
+
+function convertAudioPart(
+  part: ChatCompletionMessageParam["content"],
+): ResponseInputAudio | undefined {
+  if (
+    Array.isArray(part) ||
+    !part ||
+    typeof part !== "object" ||
+    (part as any).type !== "input_audio"
```
RomneyDa: It feels like the types on these could be more precise and clean up the code a bit.
e.g. convertAudioPart could take a ChatCompletionAudioPart or something, the check on `type` is duplicate. Applies to other converters as well

packages/openai-adapters/src/apis/openaiResponses.ts
```diff
@@ -0,0 +1,812 @@
...
+            const callId = toolCall.id ?? `tool_call_${index}`;
+            const functionCall: any = {
+              type: "function_call",
+              call_id: callId,
+              name: func.name ?? "",
+              arguments: func.arguments ?? "{}",
+            };
+            if (
+              typeof toolCall.id === "string" &&
+              toolCall.id.startsWith("fc_")
```
RomneyDa, line 365: why this startsWith fc_ check? wondering if it might cause issue if someone starts a convo with one model and then switches to codex. 
Is fc_ required?

RomneyDa (dismissed):
Looks good, had several nitpicks but none are blocking. It would be nice to eliminate the >20 `as any` type casts

packages/openai-adapters/src/apis/openaiResponses.ts
Patrick-Erichsen: Hm good question. I forget the context but I was having issues with tool calls related to an `fc_` prefix on the tool call id, I'm assuming "function call" for tools.

packages/openai-adapters/src/apis/openaiResponses.ts
Patrick-Erichsen: this one was pretty bad 😅

extensions/cli/src/stream/streamChatResponse.test.ts
Patrick-Erichsen:
```
fdir (fast directory crawler used by FileIndexService to index files)
```

But, I removed it and tests still passed so not needed it appears.

packages/openai-adapters/src/apis/openaiResponses.ts
Patrick-Erichsen: removed

PR #8473, opened 2025-10-28
feat: instant edit for find/replace tools

RomneyDa:
## Description
We've been using diff streaming approaches built for when model streams an apply response in. Added utils for instant apply. See the before an after here:

https://github.com/user-attachments/assets/0150a04f-d929-4300-8dac-59fdd4fa43fc

And after for jetbrains
https://github.com/user-attachments/assets/21ea1ac1-9190-4356-bf91-16dcd0658db6

---
## Summary by cubic
Instant apply for find/replace edits in VS Code and JetBrains. Diffs are computed once and applied immediately, replacing streaming for these tools.

- **New Features**
  - Core: exposed myersDiff via a new getDiffLines protocol for synchronous diff generation.
  - JetBrains: added instantApplyDiffLines in DiffStreamHandler and switched ApplyToFileHandler to use it.
  - VS Code: added VerticalDiffManager.instantApplyDiff and used it in ApplyManager for search/replace; auto-scrolls to the first change and updates apply state.

- **Refactors**
  - Moved getFirstChangedLine to a shared util and removed duplicate handler code.
  - Added GetDiffLinesPayload and registered "getDiffLines" message type; removed streaming paths in instant-apply scenarios.

### DIFF
--- a/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/DiffStreamHandler.kt
+++ b/extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/DiffStreamHandler.kt
@@ -215,7 +215,7 @@ class DiffStreamHandler(
                 }
                 DiffLineType.NEW -> {
                     // Insert line
-                    val offset = editor.document.getLineStartOffset(currentLine)
+                    val offset = if (currentLine >= editor.document.lineCount) editor.document.textLength else editor.document.getLineStartOffset(currentLine)
                     editor.document.insertString(offset, text + "\n")
                     currentLine++
                 }


extensions/intellij/src/main/kotlin/com/github/continuedev/continueintellijextension/editor/DiffStreamHandler.kt
```diff
@@ -131,6 +132,138 @@ class DiffStreamHandler(
...
+            val diffLines = parseDiffLinesResponse(response) ?: run {
+                println("Error: Invalid response format for getDiffLines")
+                ApplicationManager.getApplication().invokeLater {
+                    setClosed()
+                }
+                return@request
+            }
+            
+            ApplicationManager.getApplication().invokeLater {
+                WriteCommandAction.runWriteCommandAction(project) {
```
Patrick-Erichsen, line 161: How confident are you that this is the right approach here?

RomneyDa: 3/10
seems to work

PR #8489, opened 2025-10-28
fix: instant reject all for VS Code

RomneyDa:
## Description
Instead of file operations for each individual diff, unified reject all with one file operation

First reject is before, second is after:

https://github.com/user-attachments/assets/9f57f180-24a3-4025-b5b3-758adf138c92

---
## Summary by cubic
Make “Reject All” in VS Code instant by applying all diffs in one edit instead of per-diff operations. This speeds up large files and keeps undo history clean.

- **Bug Fixes**
  - Added unifiedRejectAll() to rebuild final content and replace the document in a single atomic edit (undo stops disabled).
  - Processes removed/added ranges in reverse line order to avoid index shifts and ensure stable results.

### DIFF
--- a/extensions/vscode/src/diff/vertical/handler.ts
+++ b/extensions/vscode/src/diff/vertical/handler.ts
@@ -136,12 +136,7 @@ export class VerticalDiffHandler implements vscode.Disposable {
       // Accept all: delete all the red ranges and clear green decorations
       await this.deleteRangeLines(removedRanges.map((r) => r.range));
     } else {
-      // Reject all: Re-insert red lines, delete green ones
-      for (const r of removedRanges) {
-        await this.deleteRangeLines([r.range]);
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/diff/vertical/handler.ts
```diff
@@ -604,6 +599,63 @@ export class VerticalDiffHandler implements vscode.Disposable {
...
+    await this.ensureCurrentFileIsFocused();
+
+    const removedRanges = this.removedLineDecorations.ranges;
+    const addedRanges = this.addedLineDecorations.ranges;
+
+    interface LineOperation {
+      type: "removed" | "added";
+      line?: string; // Only for removed lines
+      range: vscode.Range;
+    }
```
Patrick-Erichsen, line 614: I didn't even know you could do define an interface inside a function lol

Kind of awkward but nbd

PR #8497, opened 2025-10-29
chore: deprecate continue.telemetryEnabled in vscode preferences

uinstinct:
## Description

The gui settings for telemetry enabling was in sync with the vscode ide settings only when the ide restarted. This PR fixes that.

Resolves CON-4319

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/61509bab-132b-452d-aa60-a15d2b4d2e73

https://github.com/user-attachments/assets/f34518d1-f3f8-4c2f-a99a-1aca7e325f4b

## Tests

---
## Summary by cubic
Deprecates the continue.telemetryEnabled setting in VS Code and points users to Telemetry > Allow Anonymous Telemetry so the in-app toggle and IDE settings stay in sync (CON-4319). This ensures changes apply immediately without a restart.

<sup>Written for commit d0cd4193c0e6142e6df8372860149a2b14936a7b. Summary will update automatically on new commits.</sup>

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -89,7 +89,9 @@
         "continue.telemetryEnabled": {
           "type": "boolean",
           "default": true,
-          "markdownDescription": "Continue collects anonymous usage data, cleaned of PII, to help us improve the product for our users. Read more  at [continue.dev › Telemetry](https://docs.continue.dev/telemetry)."
+          "markdownDescription": "Continue collects anonymous usage data, cleaned of PII, to help us improve the product for our users. Read more  at [continue.dev › Telemetry](https://docs.continue.dev/telemetry).",
+          "markdownDeprecationMessage": "This setting is deprecated and will be removed in a later version. Please use the _Telemetry > Allow Anonymous Telemetry_ setting in the extension's User Settings page.",
+          "deprecationMessage": "Please use the _Telemetry > Allow Anonymous Telemetry_ setting in the extension's User Settings page."
         },
         "continue.showInlineTip": {
           "type": "boolean",


RomneyDa (changes requested):
@uinstinct when ide settings are changed it comes through as an ide settings update which causes a config reload, which accounts for the ide setting when setting config.allowAnonymousTelemetry. It should already sync properly, although it might result in the toggle being un-toggleable if set to false in IDE settings (probably fine)

Could you explain the background a bit more?

uinstinct:
> @uinstinct when ide settings are changed it comes through as an ide settings update which causes a config reload, which accounts for the ide setting when setting config.allowAnonymousTelemetry. It should already sync properly, although it might result in the toggle being un-toggleable if set to false in IDE settings (probably fine)
> Could you explain the background a bit more?

Right. The scenario mentioned by you happens for syncing vscode settings page --> gui settings page configuration. The current PR adds the ability to for syncing gui settings page --> vscode settings page which was absent.

RomneyDa:
I think we've more or less deprecated the allowAnonymousTelemetry IDE setting or at least it's not the preferred way to configure. The issue with this approach is that it prevents people from just editing the JSON because now it's out of sync. E.g. the _only_ way to change allowAnonymousTelemetry is now through the toggle, which would create a security concern affecting the preferred config approach (globalContext)

Potentially a better UI would be to disable the toggle if set by the UI?

uinstinct:
> I think we've more or less deprecated the allowAnonymousTelemetry IDE setting or at least it's not the preferred way to configure. The issue with this approach is that it prevents people from just editing the JSON because now it's out of sync. E.g. the _only_ way to change allowAnonymousTelemetry is now through the toggle, which would create a security concern affecting the preferred config approach (globalContext)
> 
> Potentially a better UI would be to disable the toggle if set by the UI?

True. We should have one setting and that should be the toggle in extension gui settings page. 
About config.json 's allowAnonymousTelemetry setting, we have deprecated it for a long time and guess we can now remove it. Thoughts? (IMO, all the deprecated config.json settings should be removed together later in a major release)

RomneyDa:
@uinstinct the current functionality is that if allowAnonymousTelemetry is detected in config.json it migrates the value to shared config and then _removes_ it from config.json. This can stay the same.

uinstinct:
@RomneyDa  Now I have removed the isTelemetryEnabled setting from vscode and jetbrains. Hence currently the only way to set allowAnonymousTelemetry is now through the settings page toggle, config.json and org policy

uinstinct:
Added a deprecation message

image

PR #8535, opened 2025-10-31
fix(cli): prevent auto compaction loop and prune messages until valid

uinstinct:
## Description

Prevent infinite looping when auto compaction happens. 
Also prune messages when auto compaction fails and context length is invalid/exceeding, until chat history is within context length.

resolves CON-4681
resolves CON-4455

## AI Code Review

## Checklist

(both having context length set as 300 in the config.yaml)

https://github.com/user-attachments/assets/118f3ab8-2de5-49ee-bbad-79d78e4fa6aa

https://github.com/user-attachments/assets/078fd68d-67ff-4bd7-96f9-e003bd95faa6

## Tests

### DIFF
--- a/extensions/cli/src/stream/streamChatResponse.ts
+++ b/extensions/cli/src/stream/streamChatResponse.ts
@@ -8,6 +8,7 @@ import type {
   ChatCompletionTool,
 } from "openai/resources.mjs";
 
+import { pruneLastMessage } from "../compaction.js";
 import { services } from "../services/index.js";
 import { posthogService } from "../telemetry/posthogService.js";
 import { telemetryService } from "../telemetry/telemetryService.js";
@@ -130,6 +131,7 @@ interface ProcessStreamingResponseOptions {
 }
 
 // Process a single streaming response and return whether we need to continue
+// eslint-disable-next-line max-statements
 export async function processStreamingResponse(
   options: ProcessStreamingResponseOptions,
 ): Promise<{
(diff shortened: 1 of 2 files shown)


uinstinct:
Was looking at [`countChatHistoryItemTokens`](https://github.com/continuedev/continue/blob/3dd4d2d5012501f873af07e91c2f3ad48d816930/extensions/cli/src/util/tokenizer.ts#L30). Why do we not count the `systemMessage` and the tool call results?

From [the ticket](https://uploads.linear.app/19d8d766-77d3-47f0-9ebc-0c37b5fe85a6/f9425eeb-65ff-465d-ac4c-94c8b1001bad/82225787-6cc5-4385-87b0-47c1af859ccd), it looks like there is some counting mistake that is happening because before getting the 500 error from the server, we should have received our internal _Context length validation failed_ error in the tui.

extensions/cli/src/stream/streamChatResponse.ts
```diff
@@ -138,18 +140,22 @@ export async function processStreamingResponse(
...
   // Validate context length before making the request
-  const validation = validateContextLength(chatHistory, model);
+  let validation = validateContextLength(chatHistory, model);
+
+  // Prune last messages until valid
+  while (chatHistory.length > 1 && !validation.isValid) {
+    const prunedChatHistory = pruneLastMessage(chatHistory);
+    if (prunedChatHistory.length === chatHistory.length) break;
+    chatHistory = prunedChatHistory;
+    validation = validateContextLength(chatHistory, model);
```
RomneyDa: This approach means it will count the tokens of every message N times where N is number of messages pruned. This could be very inefficient if a large message comes in the front and kicks a bunch out

PR #8656, opened 2025-11-10
fix: warnings and errors for unresolved MCP secrets

RomneyDa:
- For CLI, add warnings for unresolved secrets, or throw error if in headless
- For extensions, make warnings more explanatory/up-to-date and move to mcp connection class

---
## Summary by cubic
Detect unresolved MCP secrets and surface clear errors in extensions; in headless CLI, throw an error to prevent misconfigured runs. Secret checks are now done at connection time for more accurate feedback.

- **Bug Fixes**
  - Show an error when MCP server config has unresolved secrets and list their names.
  - In headless CLI, treat unresolved secrets as errors and mark MCP status as error.
  - Add guidance links for personal/org secrets and “Include in Env”.

- **Refactors**
  - Move secret validation from YAML load to MCPConnection and CLI service.
  - Use getTemplateVariables and decodeSecretLocation to extract secret names from server config.

<sup>Written for commit 0b011ea5d65bf2adb168e710da483cfac6952c97. Summary will update automatically on new commits.</sup>

### DIFF
--- a/extensions/cli/src/services/MCPService.ts
+++ b/extensions/cli/src/services/MCPService.ts
@@ -4,10 +4,7 @@ import {
 } from "@continuedev/config-yaml";
 import { type AssistantConfig } from "@continuedev/sdk";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
-import {
-  SSEClientTransport,
-  SseError,
-} from "@modelcontextprotocol/sdk/client/sse.js";
+import { SSEClientTransport } from "@modelcontextprotocol/sdk/client/sse.js";
 import { StdioClientTransport } from "@modelcontextprotocol/sdk/client/stdio.js";
 import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
 import {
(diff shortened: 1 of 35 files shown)


core/config/yaml/loadYaml.ts
```diff
@@ -237,22 +236,6 @@ export async function configYamlToContinueConfig(options: {
     sourceFile: doc.sourceFile,
   }));
 
-  config.mcpServers?.forEach((mcpServer) => {
```
RomneyDa, line 240: removed this since it only applied to args. Moved to mcp loading

extensions/cli/src/services/MCPService.ts
```diff
@@ -248,7 +252,23 @@ export class MCPService
...
     this.updateState();
 
+    const vars = getTemplateVariables(JSON.stringify(serverConfig));
+    const unrendered = vars.map((v) => {
+      return decodeSecretLocation(v.replace("secrets.", "")).secretName;
+    });
+
     try {
+      if (unrendered.length > 0) {
+        const message = `${serverConfig.name} MCP Server has unresolved secrets: ${unrendered.join(", ")}
```
RomneyDa, line 326: This seems duplicate but the message is different enough for CLI since process.env is supported that I decided to leave them separate.

Issue #5992, opened 2025-06-04
Support configurable thinking output formats


sestinj:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

Some LLM providers like vLLM let you change the output format of thinking. Right now Continue expects only the standard format from OpenAI, Anthropic, etc. but doesn't have a way of supporting arbitrary thinking tag formats.

For reference: https://docs.vllm.ai/en/latest/features/reasoning_outputs.html#streaming-chat-completions

Iamshankhadeep:
@sestinj I will be working on this issue next.

vickywane:
@Iamshankhadeep did you get to work on this? -- I'm interested in giving it a go

TheNha:
+1

josh900:
+1

paulincai:
@Iamshankhadeep "next" as in "next year"?!

AyRickk:
Here is a PR for this request : https://github.com/continuedev/continue/pull/8901

RomneyDa:
Hi all, #8901 is stale and I decided to close it for now. But @AyRickk made a working solution. This issue is still open for contributions.

Abdul-nazeer:
Hi @RomneyDa , I’d like to work on this issue.

I noticed that there was a previous PR (#8901) addressing this, but it seems to have been closed. I’ll review that approach and see how it can be improved or adapted to support configurable thinking output formats (e.g., for vLLM and similar providers).

Proposed approach:

* Analyze current handling of thinking/reasoning outputs in Continue
* Identify assumptions tied to specific provider formats (OpenAI/Anthropic)
* Introduce a configurable parsing mechanism for different thinking output formats
* Ensure backward compatibility with existing providers

Please let me know if this aligns or if there are any specific directions I should follow.

Yash-200608:
Hey everyone, is this still open for contributions? I see previous PRs like #8901 were closed because they were too backend-specific. I'd love to take a crack at adding a generalized configuration option (maybe via config.json) so users can define custom thinking tags. Let me know if anyone is already working on this!

mason-san:
Hey @sestinj , is this issue still open? Would like to contribute as my first issue!

PR #8901, opened 2025-11-26
feat: add configurable thinking output format support for vLLM

AyRickk:
Related to https://github.com/continuedev/continue/issues/5992

---
## Summary by cubic
Adds configurable thinking output format support for vLLM by parsing custom tags in streamed responses, while preserving vLLM’s native reasoning_content during streaming. Models using tags like <think>…</think> or <reasoning>…</reasoning> now emit proper “thinking” and “assistant” chunks.

- **New Features**
  - Add VllmOptions: thinkingOpenTag and thinkingCloseTag to configure custom thinking tags (both required).
  - Implement ThinkingTagExtractor to split streamed content into thinking vs. regular content, with partial-tag handling and flush on stream end.
  - Override vLLM streaming to emit “thinking” chunks first and pass through assistant chunks; use direct SSE to preserve reasoning_content. Behavior is unchanged when options are not set.
  - Add unit and integration tests; update vLLM provider and reference docs with configuration examples.

<sup>Written for commit d99b93c635fe649c7b92f0e8105c2a2bd3d73556. Summary will update automatically on new commits.</sup>

### DIFF
--- a/core/llm/thinkingTagIntegration.vitest.ts
+++ b/core/llm/thinkingTagIntegration.vitest.ts
@@ -5,8 +5,8 @@ import { ThinkingTagExtractor } from "./thinkingTagExtractor";
 
 /**
  * Mock vLLM for testing thinking tag extraction during streaming.
- * Since the thinking tag extraction is now vLLM-specific, we mock the Vllm class
- * instead of BaseLLM.
+ * We override the OpenAI parent's _streamChat (via super.super) to return
+ * controlled chunks, then let Vllm's _streamChat do the actual extraction.
  */
 class MockVllm extends Vllm {
   private mockChunks: ChatMessage[] = [];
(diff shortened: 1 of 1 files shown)


AyRickk:

AyRickk:
recheck

core/index.d.ts
```diff
@@ -687,6 +687,12 @@ export interface LLMOptions {
 
   sourceFile?: string;
   isFromAutoDetect?: boolean;
+
+  // Thinking output format options
+  // These allow configuring custom tags to extract thinking content from the response
+  // For example, vLLM can use <think>...</think> tags instead of the standard reasoning_content field
+  thinkingOpenTag?: string;
```
RomneyDa: what if we just did `thinkingTagName` e.g. `think` instead of always saying `<think>` and `</think>`. Would you expect cases where the end tag name would be different?

core/llm/index.ts
```diff
@@ -84,6 +84,134 @@ export function isModelInstaller(provider: any): provider is ModelInstaller {
 
 type InteractionStatus = "in_progress" | "success" | "error" | "cancelled";
 
+/**
+ * Helper class to extract thinking content from custom tags during streaming.
+ * This is used for providers like vLLM that support custom thinking output formats.
+ */
+export class ThinkingTagExtractor {
```
RomneyDa: Move class to separate file

docs/customize/model-providers/more/vllm.mdx
```diff
@@ -104,4 +104,45 @@ Continue automatically handles vLLM's response format (which uses `results` inst
...
+
+<Tabs>
+   <Tab title="YAML">
+   ```yaml title="config.yaml"
+   models:
+     - name: DeepSeek R1 Distill
+       provider: vllm
+       model: deepseek-ai/DeepSeek-R1-Distill-Qwen-7B
+       apiBase: http://localhost:8000/v1
+       thinkingOpenTag: "<think>"
```
RomneyDa, line 124: I think we don't want to add this as a top level YAML property but we could add under e.g.`chatOptions`

core/llm/index.ts
```diff
@@ -1163,6 +1332,12 @@ export abstract class BaseLLM implements ILLM {
     let usage: Usage | undefined = undefined;
     let citations: null | string[] = null;
 
+    // Create thinking tag extractor if custom tags are configured
+    const thinkingExtractor =
```
RomneyDa: If this is a VLLM-specific feature could we move this into the VLLM class somehow?

core/llm/index.ts
```diff
@@ -1029,7 +1197,8 @@ export abstract class BaseLLM implements ILLM {
       completion,
       thinking,
```
RomneyDa, line 1029: feels odd that output has thinking and thinkingChunk

core/index.d.ts
AyRickk: some models provide a closing tag but not a opening one, an example is new Qwen3 models that assumes that reasoning is always present so they begin with reasoning without explicit `<think>` tag but keep the closing one

AyRickk:
@RomneyDa is it ok ?

RomneyDa:
@AyRickk what are your thoughts on this?

RomneyDa (changes requested):
@AyRickk status from previous review
✅ functionality is siloed to the VLLM class
✅ Follow up on/clean up test/docs commits. This is a new feature and we are tweaking when it runs (i.e. we might not run for community PRs) but I think in this case it made sense
🟡  use a single thinkingTagName not duplicate tag configs (unless it would be normal for them to be different in VLLM, I'm not sure) - I saw your comment, but I think still the think tag name would always be the same, so it might make more sense to say `thinkTagName` and `alwaysIncludesReasoning` (boolean) or similar, which could clean up much of this logic
🟡 On first glance I wonder if the tag parsing logic could be significantly more concise, maybe worth another look/approach?
❌ Update config yaml schema as in comment, ❌ and in config-yaml package (or it will not work for yaml configs)

I'm concerned about the maintainability of this, I think we could merge if we kept it to a couple hundred lines of simple think tag parsing etc.

RomneyDa:
@AyRickk Thanks for the contribution here. This has gone a bit stale, closing for now. Definitely still open for contributions on this. See above feedback.

PR #8927, opened 2025-11-28
feat: auto approve parallel read only builtin tools

uinstinct:
## Description

Previously when calling tools in parallel, all tool required approval. This PR changes that behaviour by when calling tools in parallel, the readonly builtin tools which were auto approved get called without waiting.

closes CON-3922

## AI Code Review

## Checklist

**before**

image

**after**

image

## Tests

---
## Summary by cubic
Auto-approve and immediately run read-only built-in tools during parallel tool calls. Addresses CON-3922 and removes approval delays and UI flicker for safe built-ins.

- **New Features**
  - Auto-exec built-in read-only tools marked allowedWithoutPermission during parallel runs.
  - Filters by BUILT_IN_GROUP_NAME and tool.readonly; dispatches with isAutoApproved: true.
  - Updates stream/UI state to avoid flicker while other tools wait for approval.

<sup>Written for commit 202431772a1f997bf7ee5474070299e501077d97. Summary will update automatically on new commits.</sup>

### DIFF
--- a/gui/src/pages/gui/ToolCallDiv/index.tsx
+++ b/gui/src/pages/gui/ToolCallDiv/index.tsx
@@ -29,7 +29,7 @@ export function ToolCallDiv({
 
   const shouldShowGroupedUI = toolCallStates.length > 1 && isStreamingComplete;
   const activeCalls = toolCallStates.filter(
-    (call) => call.status !== "canceled",
+    (call) => call.status !== "canceled" && call.status !== "done",
   );
 
   const renderToolCall = (toolCallState: ToolCallState) => {


RomneyDa (changes requested):
@uinstinct I think this implementation is really clean, should be good on the thunk side. I noticed a sort-of bug in the screenshot which may or may not be related: it says "Pending 2" calls when it has already called the read only one. Could you check on that?

uinstinct:
> @uinstinct I think this implementation is really clean, should be good on the thunk side. I noticed a sort-of bug in the screenshot which may or may not be related: it says "Pending 2" calls when it has already called the read only one. Could you check on that?

nice eyes 👀! fixed

image

RomneyDa (approved):

Issue #8683, opened 2025-11-11
Slash commands require double Enter


indidginus:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I'm not able to find a related conversation on [GitHub discussions](https://github.com/continuedev/continue/discussions) that reports the same bug
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 1.5.11
- IDE version: 1.105.1
- Model: ollama glm-4.6:cloud
- config:
  

  
  OR link to agent in Continue hub:
```

### Description

When using slash commands in the Continue CLI, I need to press Enter twice for the command to execute. I am using powershell terminal in VS Code. 

### Log output

```Shell

```

uinstinct:
It should go something like this: enter `/` command, a list of options open, navigate to the desired slash command, press Enter, the slash command fills up in the input, press Enter again to execute.

Is this the path you are going though?

indidginus:
yes

indidginus:
/ -> command -> enter. then enter again on same command

RomneyDa:
I think this was set up because of the `/title` command.
Let's make it so that it auto-submits the message for all but the `/title` command which requires additional input

RomneyDa:
@indidginus thanks for the report!

indidginus:
No problem and thanks for looking into it :)

RomneyDa:
Fixed by @uinstinct here
https://github.com/continuedev/continue/issues/8683

RomneyDa:
*Here https://github.com/continuedev/continue/pull/8928

PR #8928, opened 2025-11-28
feat: submit slash command in one enter

uinstinct:
## Description

Previously it requires 2 enter keys to submit slash commands. This PR helps it to be done in one enter key.

Fixes https://github.com/continuedev/continue/issues/8683
resolves CON-4937

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/a2ff270a-d2f6-4bc3-9cbb-ec4665cdcde8

https://github.com/user-attachments/assets/4bf56a23-6edc-4960-847f-4953b383ce26

## Tests

---
## Summary by cubic
Submit slash commands in the CLI with one Enter press, removing the previous two-step flow. Implements Linear CON-4937.

- **New Features**
  - Enter submits the highlighted command; Tab selects without submitting. /title stays select-only and Enter selects it.
  - Adds the command to history; enqueues if a response is generating, otherwise submits; then clears the input and resets UI state.

<sup>Written for commit 4cf337448d29077fd40ef4eb2311b48a9a8400ab. Summary will update automatically on new commits.</sup>

### DIFF
--- a/extensions/cli/spec/shell-mode.md
+++ b/extensions/cli/spec/shell-mode.md
@@ -27,6 +27,8 @@ Interaction with other input helpers
   - "/" slash command UI is disabled.
 - When in slash command mode (input starts with /):
   - "@" file search suggestions are disabled.
+  - Enter submits the highlighted slash command directly (except /title, which requires Tab to select first).
+  - Tab selects the highlighted command without submitting.
 
 Submission behavior
 


RomneyDa (changes requested):
@uinstinct I think the cubic feedback is valid, is it possible to queue it up instead of submit immediately?

uinstinct:
> @uinstinct I think the cubic feedback is valid, is it possible to queue it up instead of submit immediately? I didn't fully test but there might be a simple way to use the queuing method rather than the onsubmit

I wanted your thoughts on this one - enqueuing is the more bulletproof option and upon testing, submitting command directly works fine (even if we change to suitable models) with most of the slash commands.

have add support for enqueuing in the last commit

RomneyDa (changes requested):
>- Enter submits the highlighted slash command directly (except /title, which requires Tab to select first).
> - Tab selects the highlighted command without submitting.
  
@uinstinct I think the descriptions here don't match behavior, since Enter will select title as well it just won't submit it (if that's not the behavior it should be)

uinstinct:
> > * Enter submits the highlighted slash command directly (except /title, which requires Tab to select first).
> > * Tab selects the highlighted command without submitting.
> 
> @uinstinct I think the descriptions here don't match behavior, since Enter will select title as well it just won't submit it (if that's not the behavior it should be)

It should prompt the user to the title of the chat if not provided. Something like this:


https://github.com/user-attachments/assets/8d6b8895-eabe-4ffd-b576-9a20cabaf516

RomneyDa:
@uinstinct I just mean the description in the shell mode spec
>Enter submits the highlighted slash command directly (except /title, which requires Tab to select first).
Isn't right because you can do /title with Enter not tab

That spec isn't critical, will go ahead and merge!

PR #8973, opened 2025-12-02
fix: handle JSON contents of create_new_file

### LINKED ISSUE #8972: create_new_file fails on pure JSON contents

### Relevant environment info

```Markdown
- OS: MacOS
- Continue version: 	1.3.24
- IDE version: 1.107
- Model: M2
```

### Description

When the model calls `create_new_file`, and the `contents` argument is, itself, valid JSON, the contents argument is an object, not a string, because this snippet in `core/context/retrieval/pipelines/BaseRetrievalPipeline.ts`:

```ts
      const parsed = JSON.parse(responseContent);
      toolCalls = parsed.tools || [];
```

Will have deeply parsed the response.   This causes `getStringArg()` to throw an error.  Instead, if the type of `args[argName]` is 'object', it should be coerced back to a string with `JSON.stringify()`

I'll submit a PR shortly for this

### To reproduce

1. Enable tool calls
2. Ask a model in agent mode: "Create a new file called sample.json"

### Log output

```Shell

```

mcowger:
Handle case where JSON content was parsed into an object by the tool call parser.  If the arguments to the tool call are valid JSON (e.g. the model attempts to create a .json file) the earlier call to JSON.parse() will have deeply parsed the returned contents. If that has happened, convert back to string.

Fixes: #8972

## Description

## AI Code Review

## Checklist

NA

## Tests

Added 6 tests to `core/tools/parseArgs.vitest.ts` for new check.

---
## Summary by cubic
Fixes failures when create_new_file receives JSON content parsed into an object by ensuring contents is always a string. Prevents crashes when creating .json files. Fixes #8972.

- **Bug Fixes**
  - In getStringArg, JSON.stringify object/array values to safely handle parsed JSON.
  - Throw a clear error for non-string values after conversion.
  - Added tests covering contents object/array handling and edge cases.

<sup>Written for commit e04c21f9c883ad62d6c15661d715635f1a69c50e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/core/tools/parseArgs.ts
+++ b/core/tools/parseArgs.ts
@@ -26,19 +26,16 @@ export function getStringArg(
 
   let value = args[argName];
 
-  // Handle case where JSON content was parsed into an object by the tool call parser.
+  // Handle case where JSON was parsed into an object by the tool call parser.
   // If the arguments to the tool call are valid JSON (e.g. the model attempts to create a .json file)
-  // the earlier call to JSON.parse() will have deeply parsed the returned contents.
+  // the earlier call to JSON.parse() will have deeply parsed the returned arguments.
   // If that has happened, convert back to string.
   if (typeof value === "object" && value !== null) {
-    // Special handling for contents parameter which should always be a string
-    if (argName === "contents") {
+    try {
       value = JSON.stringify(value);
       return value;
-    } else {
-      throw new Error(
-        `Argument \`${argName}\` must be a string, not an object`,
-      );
+    } catch (e) {
+      //Swallow this, because it might be fine later.
     }
   }
 
--- a/core/tools/parseArgs.vitest.ts
+++ b/core/tools/parseArgs.vitest.ts
@@ -176,13 +176,6 @@ describe("getStringArg", () => {
     expect(result).toBe('["item1","item2",{"key":"value"}]');
   });
 
-  it("should throw error when non-contents parameter is an object", () => {
-    const args = { filename: { invalid: "object" } };
-    expect(() => getStringArg(args, "filename")).toThrowError(
-      "Argument `filename` must be a string, not an object",
-    );
-  });
-
   it("should handle contents parameter that is already a string", () => {
     const args = { contents: "already a string" };
     const result = getStringArg(args, "contents");


RomneyDa (changes requested):
@mcowger could we add this special handling in createFileImpl instead of here? 

OR, I think you could just remove special handling for `contents` and always make sure it's a string within `getStringArg`!

mcowger:

mcowger:
> @mcowger could we add this special handling in createFileImpl instead of here?
> 
> OR, I think you could just remove special handling for `contents` and always make sure it's a string within `getStringArg`!

Good call.   Removed the special case handling.

RomneyDa (commented):
@mcowger this looks good to me. I haven't been able to replicate the failing case locally - it seems like claude and openai models mostly escape JSON strings correctly. Could you share which model you find this error often happens with?

EDIT I see `M2` in the issue but wasn't sure which model that was. Model + provider combo would be helpful

mcowger:
I can replicate on Synthetic.new.  It doesn't happen for me on OpenRouter either.

Happy to provide a key privately for you to test so you dont have to give them money.

RomneyDa:
@mcowger which model are you working with?
EDIT this was a silly question, I see M2 on openrouter now
https://openrouter.ai/minimax/minimax-m2

mcowger:
> @mcowger which model are you working with?

Minimax M2 (note, I cannot replicate this on OpenRouter).  I cannot replicate this via GLM4.6 on the same provider, so it appears to be an interaction between the inference and the model and continue.

mcowger:
I captured traces of the differences in the output:

https://gist.github.com/mcowger/3aea3a2314d9f5540e7bbf68d9cafcd7

GLM outputs double escaped strings, while M2 does not.

That should be enough to replicate with?

RomneyDa:
This seems like could be a valid issue for other models as well, and if the tool implementation is looking for a string arg specifically, like the create new file tool, this fix seems fine. 

I guess you would still see issues with MCP tools when they expect string JSON and receive JSON object. The solution might be to detect `"type": "string"` in the tool definition and use `getStringArg` to replace at the time of calling the MCP tool. Do you use MCP tools at all? I could open a separate issue for that

RomneyDa:
Got it, thanks for the traces. It's no clear why the model would escape differently between providers (as far as I understand this would only happen because of something on the provider side) but I could see this happening in other situations as well and this fix won't hurt.

mcowger:
Agree its a provider side issue, and I've filed a bug with them.

I dont use MCPs much, but I agree it could impact them as well.

Issue #8877, opened 2025-11-25
ls tool can list os file system root contents when policy is set to "Automatic" and dirPath is '.'


wavu-238:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I'm not able to find a related conversation on [GitHub discussions](https://github.com/continuedev/continue/discussions) that reports the same bug
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Arch Linux
- Continue version: 1.2.11
- IDE version: VS Code 1.104.3
- Model: mocked predefined response
```

### Description

When policy for ls tool is set to ```automatic``` and model tries to list contents of ```/``` directory Continue will ask if it should list the files. This is good and expected.

But if the model tries to read ```.``` directory the check will be omitted and Continue will list file system root contents without asking.

### To reproduce

I have created an Ollama api interceptor and sent predefined messages as responses to verify the problem. To reproduce the behavior of ls tool check it on the following mocked responses

For model output:
```
"    ```tool\n    TOOL_NAME: ls\n    BEGIN_ARG: dirPath\n    /\n    END_ARG\n    ```"
```
You will get proper safe behavior

For output:
```
"    ```tool\n    TOOL_NAME: ls\n    BEGIN_ARG: dirPath\n    .\n    END_ARG\n    ```"
```
You will escape the evaluation policy

### Log output

```Shell

```

robertm1771:
## Related Issue: WSL + `code .` Path Resolution

### Environment
- **OS:** Windows 25H2 + WSL Ubuntu
- **Continue version:** pre-release
- **IDE:** VS Code [your version] (Remote - WSL)
- **Model:** LM Studio openai/gpt-oss-20b for chat and qwen2.5-coder-7b-instruct code agent

### Description

I'm experiencing what appears to be the same relative path resolution bug in a WSL environment.

When opening VS Code from WSL using `code .`, Continue resolves the workspace path incorrectly—pointing to the wrong directory (possibly the Windows filesystem root or incorrect WSL path) instead of the actual workspace.

### To Reproduce

1. Open WSL terminal
[...]

Issue #8744, opened 2025-11-16
Fail to call Qwen3-Coder model with tool calling enabled


ontecAI:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I'm not able to find a related conversation on [GitHub discussions](https://github.com/continuedev/continue/discussions) that reports the same bug
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 1.2.10
- IDE version: VS Code
- Model: Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8
- config:
  
  - name: qwen3-coder-30b
    provider: vllm
    model: Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8
    apiBase: http://ai01-dx-mr300-b04.dev1.xxx.at:7777/v1/
    apiKey: xxxx
    roles:
      - apply
      - chat
      - edit
    capabilities:
      - tool_use

  - name: qwen3-coder-30b-litellm
    provider: openai
    model: qwen3-coder-30b-fp8
    apiBase: https://litellm.dev1.xxx.at/v1/
    apiKey: xxxx
    roles:
      - apply
      - chat
      - edit
    capabilities:
      - tool_use
  
  OR link to agent in Continue hub:
```

### Description

[@continuedev] error: Error streaming response: list index out of range {"context":"llm_stream_chat","model":"Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8","provider":"vllm","useOpenAIAdapter":true,"streamEnabled":true,"templateMessages":true}

[Extension Host] Error: Error streaming response: list index out of range
	at parseDataLine (c:\Users\mbuchner\.vscode\extensions\continue.continue-1.2.10-win32-x64\out\extension.js:131119:15)
	at parseSseLine (c:\Users\mbuchner\.vscode\extensions\continue.continue-1.2.10-win32-x64\out\extension.js:131136:33)
	at streamSse (c:\Users\mbuchner\.vscode\extensions\continue.continue-1.2.10-win32-x64\out\extension.js:131151:37)
[...]

Issue #6220, opened 2025-06-20
Groq Tool Call is not working


ghdoergeloh:
### Before submitting your bug report

- [ ] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Mac
- Continue version: 1.0.13
- IDE version: VS Code 1.101.0
- Model: qwen-qwq-32b  or llama-3.1-8b-instant
- config:
  
%YAML 1.1
---
name: Local Assistant
version: 1.0.0
schema: v1

groq: &groq
  provider: groq 
  # provider: openai <- this one works
  # apiBase: https://api.groq.com/openai/v1 <- this one works
  apiKey: ...

models:
  - name: Groq - Qwen QwQ 32B
    <<: *groq
    model: qwen-qwq-32b
    roles:
      - chat
      - rerank
      - summarize
    capabilities:
      - tool_use
    defaultCompletionOptions:
      contextLength: 131072
```

### Description

I tried qwen-qwq-32b from groq with a simple tool call. It should list all files in the current directory.

The output was empty. The logs in groq showed a successful request (200) and a number of output tokens (~400 Tokens).

Continue Console showed a successful request but no tool call output nor another output token.

### To reproduce

1. Use my yaml config (add your apiKey).
2. Choose the Configured Model.
3. Ask to list all files of the current directory.

### Log output

```Shell

```

ghdoergeloh:
Error still exists, but workaround with openai provider and different base url is still working.

PR #8985, opened 2025-12-03
fix: lsTool path management and improve workspace file resolution

andreapizzato:
## Description

Code changes:
- removed custom path rewriter for `lsTool` using the IDE/Platform-provided path resolution
- fixed an issue where a file could be resolved within the workspace but being reported as non-existent

## Targeted issues

Here's a few of the issues I've found which should be cleared with this change:
- #8877 
- #8744 
- #6220

## AI Code Review

## Checklist

The best demo of this fix is to compare the current main with the branch running the same task; you can [view the demo](https://youtu.be/iHbcTkK2vgc) I just recorded.

## Tests

I actually had to remove some tests because they were testing the removed function; that specific functionality is tested elsewhere both in the codebase and in the supporting vscode/node SDKs.

---
## Summary by cubic
Switch lsTool to the IDE path resolver and fix workspace URI checks to prevent false “directory/file not found” errors. Addresses #8877, #8744, #6220.

- **Bug Fixes**
  - Use IDE-provided path resolution in lsTool; remove custom path rewriter.
  - Verify file existence using the resolved workspace URI, not the raw input.
  - Return a clear error when dirPath is missing.

<sup>Written for commit 9a7d15b0a293eed8af5f88f7265d61e8d17531fb. Summary will update automatically on new commits.</sup>

### DIFF
--- a/core/util/pathResolver.ts
+++ b/core/util/pathResolver.ts
@@ -20,7 +20,7 @@ async function isUriWithinWorkspace(ide: IDE, uri: string): Promise<boolean> {
   const workspaceDirs = await ide.getWorkspaceDirs();
   const { foundInDir, uri: foundUri } = findUriInDirs(uri, workspaceDirs);
 
-  // Check both: within workspace path AND the resolved file exists
+  // Check both: within workspace path AND resolved file exists
   if (foundInDir !== null) {
     return await ide.fileExists(foundUri);
   }


andreapizzato:

core/util/pathResolver.ts
```diff
@@ -18,11 +18,11 @@ export interface ResolvedPath {
  */
 async function isUriWithinWorkspace(ide: IDE, uri: string): Promise<boolean> {
   const workspaceDirs = await ide.getWorkspaceDirs();
-  const { foundInDir } = findUriInDirs(uri, workspaceDirs);
+  const { foundInDir, uri: foundUri } = findUriInDirs(uri, workspaceDirs);
```
RomneyDa, line 21: This should be unnecessary, foundUri and original uri should be the same

core/tools/implementations/lsTool.vitest.ts
```diff
@@ -12,30 +12,6 @@ const mockExtras = {
...
 
-test("resolveLsToolDirPath handles undefined path", () => {
-  expect(resolveLsToolDirPath(undefined)).toBe("/");
-});
-
-test("resolveLsToolDirPath handles empty string", () => {
-  expect(resolveLsToolDirPath("")).toBe("/");
-});
-
-test("resolveLsToolDirPath handles dot", () => {
```
RomneyDa, line 23: Why remove these tests? Should we only remove the ones checking for empty path?

andreapizzato: The input path is now only processed through `resolveInputPath` (core/tools/implementations/lsTool.ts:18 in this PR); therefore I had to remove the tests because the function itself is no longer existing. This does not mean at all those tests were unnecessary and shouldn't be brought back; in fact I discovered this bug trying to use SpecKit (which hosts scripts inside a .speckit folder) and Continue was madly looping unable to read files that were in fact there. Basically I'd love to bring back the old tests + tests for files and folders starting with a . (e,g, .env, .git).

Is it better to test for those cases while testing `resolveInputPath` alone or do we want to have a separate function evaluating ls input path to be able to bring back these tests?

In the original path parsing function (`resolveLsToolDirPath` - core/tools/implementations/lsTool.ts:8 in this PR):
- why was '.' (current folder) replaced with '/' filesystem root?
- what was the original intent behind removing the initial '.' from a path? Preventing access to parent folders?

core/util/pathResolver.ts
andreapizzato: Sorry, my bad, I didn't go deeper looking at the implementation of `findUriInDirs`: it is, in fact, the same. This change does not make sense, will push a new commit removing it 👍

core/tools/implementations/lsTool.vitest.ts
RomneyDa: Ah, thanks for explaining. The simple reason is that "." doesn't work with our file reading and listing IDE utils, they require URIs, so we assume "." means relative. So you're right that it's a bug. I think the fix is that "." and "./" paths should resolve relative to the workspace URI root (first workspace if there are multiple)

hackmajoris:
waiting for this

core/tools/implementations/lsTool.ts
```diff
@@ -2,25 +2,20 @@ import ignore from "ignore";
...
+import { resolveInputPath } from "../../util/pathResolver";
 
 const MAX_LS_TOOL_LINES = 200;
 
 export const lsToolImpl: ToolImpl = async (args, extras) => {
-  const dirPath = resolveLsToolDirPath(args?.dirPath);
-  const resolvedPath = await resolveInputPath(extras.ide, dirPath);
+  if (!args.dirPath) {
+    throw new ContinueError(
+      ContinueErrorReason.Unspecified,
```
RomneyDa, line 13: @andreapizzato please add a `ContinueErrorReason` for this or use an existing one

RomneyDa:
Closing since stale and we want to keep support for no dirPath to resolve to "."

PR #9072, opened 2025-12-08
Add fallback to prevent git symlink loop failure in Runloop blueprint

sestinj:
## Summary by cubic
Add error tolerance to the symlink creation loop in the Runloop blueprint so the workflow doesn’t fail on git symlink loops. Appends `|| true` to the loop to ignore non-zero exits.

<sup>Written for commit 1f986723e8ff2fac802d7da75f0fe7c7b487949c. Summary will update automatically on new commits.</sup>


RomneyDa:
looks like symlink creation was removed on main.

PR #9074, opened 2025-12-08
feat(cli): add beta background process execution tools

sestinj:
## Summary by cubic
Adds beta support for running terminal commands in the background with a new BackgroundProcessService and tools to monitor and manage them. Enables Bash run_in_background, incremental output reads, process listing, and cleanup on exit, gated behind --beta-persistent-terminal-tools.

- **New Features**
  - Bash: run_in_background parameter to start long-running commands without blocking.
  - BackgroundProcessService: spawns processes, tracks PID/status/exit code, buffers output (10k lines), max 10 concurrent, cleans up on exit, marks spawn errors as exited.
  - Tools: ReadBackgroundProcessOutput (incremental output with optional regex filter), ListProcesses, KillProcess.
  - CLI flag: --beta-persistent-terminal-tools to enable these tools; tools are registered only when the flag is set.
  - Docs: BACKGROUND_PROCESSES.md with usage and limitations.

- **Migration**
  - Start CLI with --beta-persistent-terminal-tools.
  - Use Bash({ command, run_in_background: true }) to start; then ReadBackgroundProcessOutput({ bash_id[, filter] }) to read logs; KillProcess({ bash_id }) to stop.
  - Processes are session-scoped and will be terminated on CLI exit.

<sup>Written for commit 74fd4496aef74b068c7859321416be0f62638787. Summary will update automatically on new commits.</sup>

### DIFF
--- a/extensions/cli/src/tools/bashOutput.ts
+++ b/extensions/cli/src/tools/bashOutput.ts
@@ -5,8 +5,11 @@ import { Tool } from "./types.js";
 export const bashOutputTool: Tool = {
   name: "ReadBackgroundProcessOutput",
   displayName: "Read Background Process Output",
-  description:
-    "Read output from a background process started with Bash tool's run_in_background parameter. Returns new output since last read.",
+  description: `Read output from a background process started with Bash tool's run_in_background parameter.
+
+Returns only NEW output since the last read - subsequent calls will only show lines that were written after the previous call. This allows efficient incremental monitoring without re-reading the entire output each time.
+
+Use this to check if a dev server has started, monitor build progress, watch for test results, or check for errors in long-running processes.`,
   parameters: {
     type: "object",
     required: ["bash_id"],
--- a/extensions/cli/src/tools/runTerminalCommand.ts
+++ b/extensions/cli/src/tools/runTerminalCommand.ts
@@ -12,6 +12,9 @@ import {
   isPullRequestCommand,
 } from "../telemetry/utils.js";
 
+import { bashOutputTool } from "./bashOutput.js";
+import { killProcessTool } from "./killProcess.js";
+import { listProcessesTool } from "./listProcesses.js";
 import { Tool } from "./types.js";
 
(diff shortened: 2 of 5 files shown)


extensions/cli/src/tools/bashOutput.ts
```diff
@@ -0,0 +1,88 @@
+import type { BackgroundProcessService } from "../services/BackgroundProcessService.js";
+
+import { Tool } from "./types.js";
+
+export const bashOutputTool: Tool = {
+  name: "BashOutput",
```
Patrick-Erichsen: Feels like the name could be more descriptive, eg `ReadBackgroundProcessOutput`

extensions/cli/src/services/CircularBuffer.ts
Patrick-Erichsen, line 1: Could we just use https://www.npmjs.com/package/ringbufferjs ?

extensions/cli/src/tools/runTerminalCommand.ts
```diff
@@ -50,6 +54,11 @@ IMPORTANT: To edit files, use Edit/MultiEdit tools instead of bash commands (sed
         description:
           "Optional timeout in seconds (max 600). Use this parameter for commands that take longer than the default 180 second timeout.",
       },
+      run_in_background: {
+        type: "boolean",
+        description:
+          "Run command in background and return immediately. Use BashOutput tool to monitor output. Useful for long-running processes like dev servers.",
```
Patrick-Erichsen: Probably should mention `KillProcess` and `ListProcesses` as well. Ideally just import the obj and use the name property to avoid drift.

Also wondering if it's worth providing an example flow in the description. In general wondering how well the agent will be able to follow the expected usage pattern here, were you able to test much? It feels reasonably complicated and there isn't much guidance for the agent to follow.

sestinj:
@uinstinct I'm closing this one because it's stale, but wanted to tag you so you know where to find it. Will probably be a useful reference for your work on background bash tools

PR #9085, opened 2025-12-09
fix: overzealous 'token' and 'credential' security regexps

rossigee:
## Description

Removes `token` and `credentials` file globs for files that Continue should not access.

This matches legitimate source files that we would want our AI agents to be able to read and manage, such as 'tokens.py', 'tokens.go', 'tokens.c' etc etc.

Example:

```
○ Read(src/pkg/token/options.go)

● Reading or Editing src/pkg/token/options.go is not allowed because it is a security concern. Do not attempt to read or edit this
  file in any way.
```

## AI Code Review

## Checklist

N/A

## Tests

N/A

---
## Summary by cubic
Relaxed security ignore patterns by removing token.*, credentials.*, and the bare token/credentials filenames from indexing. This prevents excluding legitimate source files (e.g., tokens.py) while still ignoring *.token and other clear secret files (e.g., *.secret, *.secrets, auth.json).

<sup>Written for commit 19f7a1a45beecbef1f2b221ec81616629dbcb359. Summary will update automatically on new commits.</sup>

### DIFF
--- a/core/indexing/ignore.vitest.ts
+++ b/core/indexing/ignore.vitest.ts
@@ -47,9 +47,7 @@ describe("isSecurityConcern", () => {
     it("should detect credential files as security concerns", () => {
       expect(isSecurityConcern("app.secret")).toBe(true);
       expect(isSecurityConcern("api.secrets")).toBe(true);
-      expect(isSecurityConcern("credentials")).toBe(true);
       expect(isSecurityConcern("auth.json")).toBe(true);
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
@rossigee I agree with this change. I think you could also remove the credentials and tokens folder blockers. Looks like some tests will need to be updated since they are now failing

RomneyDa (approved):

PR #9168, opened 2025-12-15
fix: use parsed arguments instead of toolcall.arguments

uinstinct:
## Description

When createnewfile is sent without the `contents` parameter, the error shows that "`filepath` is missing". This PR fixes that by showing the correct missing parameter.

resolves CON-5065

## AI Code Review

## Checklist

## Tests

---
## Summary by cubic
Fixes incorrect missing-field errors for createnewfile by sending parsed arguments to core, so validation reports the right parameter. Addresses CON-5065.

- **Bug Fixes**
  - Use parsedArgs (JSON) for function.arguments in tools/call.
  - Missing-parameter errors now correctly point to contents instead of filepath.

<sup>Written for commit 6222d35735a1be9044e6f9be50aef398466f9c98. Summary will update automatically on new commits.</sup>

### DIFF
--- a/gui/src/redux/thunks/callToolById.ts
+++ b/gui/src/redux/thunks/callToolById.ts
@@ -86,7 +86,13 @@ export const callToolById = createAsyncThunk<
   } else {
     // Tool is called on core side
     const result = await extra.ideMessenger.request("tools/call", {
-      toolCall: toolCallState.toolCall,
+      toolCall: {
+        ...toolCallState.toolCall,
+        function: {
+          ...toolCallState.toolCall.function,
+          arguments: JSON.stringify(toolCallState.parsedArgs),
+        },
+      },
     });
     if (result.status === "error") {
       throw new Error(result.error);


uinstinct:
> @uinstinct if I understand the problem correctly, this happens when the tool call state ends up in an invalid place because the model hits max tokens or similar. This fix would only work because we use incremental JSON parsing to create parsedArgs. But I think if it ends up invalid we should throw an error, since e.g. the file being created will be truncated
> 
> I think in that case it should throw an error that the tool call args were invalid, basically a try again with smaller/more targeted output message

Yes, the error now shows the correct missing argument.
image

Although I think resolving the error would require increasing max tokens because the agent does not send the contents argument in any case. Should we modify the createnewfile tool to support creating files without contents? (and then maybe edittool can be used to add contents to the new file)

RomneyDa:
> Although I think resolving the error would require increasing max tokens because the agent does not send the contents argument in any case. Should we modify the createnewfile tool to support creating files without contents? (and then maybe edittool can be used to add contents to the new file)

This makes sense. Since
- agree that it doesn't make sense to increase max tokens dynamically
- probably won't be reliable to create empty and then edit since feels like unexpected behavior and edit tool will have same issue unless model is aware why it happened

I think a good solution is to throw an error instructing the model to try again with smaller output.

If we take that route I guess the main question is how do we know if this was caused by max token limit being reached? 

Here are 2 ideas

1. I think the best indicator of max tokens being the culprit is if the json in `arguments` is invalid. This is why the current approach of using parsedArgs (which incrementally parses) at least fixes the issue of saying the wrong arg is missing, but doesn't fix any underlying issue. 

I think the best solution is here in `callTool.ts`
```
const args = safeParseToolCallArgs(toolCall);
```
We should not safe parse args here, we should detect invalid JSON and throw a sensible error which explains that this often happens when huge args are passed, try smaller ones.

2. We could also just add a note to `getStringArg` which says `hey sometimes this happens when you tried to pass huge args, try a smaller value`. This would probably fix most cases but feels less robust

Thoughts on just doing the first approach for now? 

something to keep in mind is that the incremental args parsing is just so that we can display tool calls as they stream in, so all that should stay in place, but if they end up invalid it makes sense to throw

uinstinct:
> 1. 
>     I think the best indicator of max tokens being the culprit is if the json in arguments is invalid. This is why the current approach of using parsedArgs (which incrementally parses) at least fixes the issue of saying the wrong arg is missing, but doesn't fix any underlying issue.
> 
> I think the best solution is here in callTool.ts
> 
> const args = safeParseToolCallArgs(toolCall);
> 
> We should not safe parse args here, we should detect invalid JSON and throw a sensible error which explains that this often happens when huge args are passed, try smaller ones.

Tried this now. In this approach, in most of the cases, the arguments from the model are valid JSON but have the missing parameter - which can only be detected during tool call. Hence, I think throwing errors while parsing would not be useful.

> 2. We could also just add a note to getStringArg which says hey sometimes this happens when you tried to pass huge args, try a smaller value. This would probably fix most cases but feels less robust

In this case, I think we should only narrow it down to a specific models where there is a chance of huge args being sent to the tool call. Otherwise, if the tool call argument is missing, the model should avoid retrying and add that first.

RomneyDa:
@uinstinct can you give args output examples where it's valid JSON but missing the required argument? My understanding of the original error is that invalid json args results in parsedArgs of `{}` -> missing args. Is this not what is happening? It seems unlikely claude would actually just miss an arg

uinstinct:
One example could be this:
Set max tokens to around 1000 or so. Ask claude to read a directory and or similar. Then ask it to report all of its findings in a md file. See that it omits the content parameter.

RomneyDa:
Will close this for now, I think @uinstinct was on the right track but current state will cause calling tools without complete inputs, and I think calling with no inputs is safer for now.

PR #8555, opened 2025-11-02
refactor: :recycle: Refactor the cli tool type to align with core

chezsmithy:
## Description

Refactor the cli tool type to have the function wrapper to align it with the core tool type.
Avoid type issues in OpenAPI adapters due to misalignment.

## AI Code Review

- **Team members only**: AI review runs automatically when PR is opened or marked ready for review
- Team members can also trigger a review by commenting `@continue-review`

## Checklist

- [X] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [X] The relevant docs, if any, have been updated or created
- [X] The relevant tests, if any, have been updated or created

## Screen recording or screenshot

[ When applicable, please include a short screen recording or screenshot - this makes it much easier for us as contributors to review and understand your changes. See [this PR](https://github.com/continuedev/continue/pull/6455) as a good example. ]

## Tests

Tests were refactored as needed.

---
## Summary by cubic
Refactored the CLI Tool type to the core-aligned, function-based shape and migrated all built-in tools. This removes type mismatches and stabilizes OpenAPI adapter behavior.

- **Refactors**
  - Switched Tool to { type: "function", function: { name, description, parameters } } with displayTitle, group, and readonly.
  - Migrated Edit, MultiEdit, Read, Write, Bash, Search, List, Diff, Status, Checklist, Fetch, Exit to the new structure.
  - Updated helpers and execution flow: permission checks, display name lookup, chat-completion conversion, MCP tool conversion, and argument validation.
  - Adjusted messages and docs to reference readFileTool.function.name consistently.

- **Bug Fixes**
  - Eliminates type misalignment causing issues in OpenAPI adapters.
  - Adds safe defaults when parameter schemas are missing to prevent validation errors.

<sup>Written for commit c078b37bc45d1fb61214a8eab039a0b6f5305445. Summary will update automatically on new commits.</sup>

PR #9314, opened 2025-12-25
feat: add tool prompt override support in .continuerc.json

### LINKED ISSUE #9312: Bug: .continuerc.json files fail to load due to 'this' binding issue
## Description

Workspace `.continuerc.json` files are silently ignored because `ide.readFile` loses its `this` binding when passed directly to `.map()`.

## Location

`core/config/json/loadRcConfigs.ts` line 21

## Current code

```typescript
return await Promise.all(rcFiles.map(ide.readFile));
```

## Problem

When `ide.readFile` is passed as a callback to `.map()`, it's called without the `ide` context, causing `this` to be undefined inside the method. This causes the file read to fail silently.

## Fix

```typescript
return await Promise.all(rcFiles.map((uri) => ide.readFile(uri)));
```

## Impact

- All `.continuerc.json` workspace-level configuration is ignored
- Tool overrides, custom rules, and other per-repo settings don't work
- The function returns an empty array with no error logged

## How to reproduce

1. Create a `.continuerc.json` in a workspace with any valid config
2. Add debug logging inside `getWorkspaceRcConfigs` 
3. Observe that `workspaceConfigs` is always empty despite the file existing and being found by `listDir`

### LINKED ISSUE #9313: Bug: YAML config path does not load workspace .continuerc.json files
## Description

When using `config.yaml` (the default for new installations), workspace `.continuerc.json` files are completely ignored. Only users with `config.json` can use workspace-level configuration.

## Location

- `core/config/yaml/loadYaml.ts` - `loadContinueConfigFromYaml` function
- `core/config/load.ts` - `loadContinueConfigFromJson` function

## Problem

The JSON config loading path calls `getWorkspaceRcConfigs(ide)` to load and merge `.continuerc.json` files from workspace directories:

```typescript
// In loadContinueConfigFromJson (load.ts)
const workspaceConfigs = await getWorkspaceRcConfigs(ide);
// ... then merges them into the config
```

The YAML config loading path has no equivalent - it never calls `getWorkspaceRcConfigs`:

```typescript
// In loadContinueConfigFromYaml (loadYaml.ts)
// No call to getWorkspaceRcConfigs anywhere in the function
```

## Impact

- Users with `config.yaml` (the default) cannot use `.continuerc.json` for per-repo settings
- This is a significant feature gap between the two config formats
- Documentation about `.continuerc.json` doesn't mention this limitation

## Expected behavior

Both config paths should load and merge `.continuerc.json` files from workspace directories.

## Workaround

Rename `~/.continue/config.yaml` to `config.yaml.bak` and create a `~/.continue/config.json` instead.

shanevcantwell:
## Summary

Adds the ability to customize tool prompts via `.continuerc.json` at the workspace level, plus fixes two bugs that were preventing this feature from working.

- Add `ToolOverride` interface for customizing tool descriptions, disabling tools, or modifying system message prompts
- Fix silent failure when reading `.continuerc.json` files (#9312)
- Fix YAML config path not loading workspace `.continuerc.json` files (#9313)

## Motivation

Local models (via LM Studio, Ollama, etc.) often struggle with Continue's default tool prompts, which include legacy syntax examples that confuse models into outputting raw `[TOOL_CALLS]BEGIN_ARG...` text instead of using proper tool calling.

This feature allows per-repo customization of tool prompts to work better with specific models.

**Before** (broken tool syntax output):
```
[TOOL_CALLS]ls BEGIN_ARG: dirPath ./docs END_ARG BEGIN_ARG: recursive false END_ARG
```

**After** (clean tool usage with custom prompts):
- Model lists available tools properly
- Tools are called via native function calling
- No broken syntax in output

## Example `.continuerc.json`

```json
{
  "mergeBehavior": "merge",
  "tools": [
    {
      "name": "view_diff",
      "disabled": true
    },
    {
      "name": "read_file",
      "description": "Read file contents. Use relative paths from workspace root.",
      "systemMessageDescription": {
        "prefix": "To read a file, use read_file with a relative path:"
      }
    }
  ]
}
```

### Fix #9312: `.continuerc.json` files fail to load

`rcFiles.map(ide.readFile)` lost `this` binding, causing silent failures. Fixed by using `rcFiles.map((uri) => ide.readFile(uri))`.

### Fix #9313: YAML config path ignores `.continuerc.json`

`loadContinueConfigFromYaml` never called `getWorkspaceRcConfigs`. Added conditional loading in `doLoadConfig.ts` so tool overrides work regardless of config format.

## Test plan

- [x] Verified `.continuerc.json` loads with JSON config
- [x] Verified `.continuerc.json` loads with YAML config  
- [x] Verified `disabled: true` removes tool from list
- [x] Verified custom descriptions appear in tool prompts
- [x] Tested with devstral via LM Studio - clean tool usage

Fixes #9312, Fixes #9313

---
## Summary by cubic
Add per-model tool prompt overrides via YAML config to customize or disable tools, applied in BaseLLM and the CLI during chat.

- **New Features**
  - Support chatOptions.toolOverrides (keyed by tool name) in YAML model config.
  - Apply overrides at runtime in BaseLLM and CLI; added ToolOverride type, helpers, and tests.

<sup>Written for commit f3883ee11a93d2f14d9b7dd1a75c1e0f990a823a. Summary will update on new commits.</sup>

---

**Continue Tasks:** ▶️ 1 queued · ▶️ 1 not started · ✅ 1 merged — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fcontinuedev%2Fcontinue%2Fpull%2F9314&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/gui/package-lock.json
+++ b/gui/package-lock.json
@@ -21,9 +21,9 @@
         "@tiptap/extension-history": "^2.27.1",
         "@tiptap/extension-image": "^2.27.1",
         "@tiptap/extension-mention": "^2.27.1",
-        "@tiptap/extension-paragraph": "^2.3.2",
-        "@tiptap/extension-placeholder": "^2.1.13",
-        "@tiptap/extension-text": "^2.3.2",
+        "@tiptap/extension-paragraph": "^2.27.1",
+        "@tiptap/extension-placeholder": "^2.27.1",
+        "@tiptap/extension-text": "^2.27.1",
         "@tiptap/pm": "^2.1.13",
         "@tiptap/react": "^2.1.13",
         "@tiptap/starter-kit": "^2.1.13",
@@ -38,7 +38,7 @@
         "downshift": "^7.6.0",
         "escape-carriage": "^1.3.1",
         "handlebars": "^4.7.8",
-        "lodash": "^4.17.21",
+        "lodash": "^4.17.23",
         "lowlight": "^3.3.0",
         "mermaid": "^11.10.0",
         "minisearch": "^7.0.2",
@@ -241,7 +241,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@continuedev/config-types": "^1.0.14",
-        "yaml": "^2.8.1",
+        "yaml": "^2.8.2",
         "zod": "^3.25.76"
       },
       "bin": {
@@ -3676,9 +3676,10 @@
       }
     },
     "node_modules/@tiptap/extension-paragraph": {
-      "version": "2.26.1",
-      "resolved": "https://registry.npmjs.org/@tiptap/extension-paragraph/-/extension-paragraph-2.26.1.tgz",
-      "integrity": "sha512-UezvM9VDRAVJlX1tykgHWSD1g3MKfVMWWZ+Tg+PE4+kizOwoYkRWznVPgCAxjmyHajxpCKRXgqTZkOxjJ9Kjzg==",
+      "version": "2.27.2",
+      "resolved": "https://registry.npmjs.org/@tiptap/extension-paragraph/-/extension-paragraph-2.27.2.tgz",
+      "integrity": "sha512-elYVn2wHJJ+zB9LESENWOAfI4TNT0jqEN34sMA/hCtA4im1ZG2DdLHwkHIshj/c4H0dzQhmsS/YmNC5Vbqab/A==",
+      "license": "MIT",
       "funding": {
         "type": "github",
         "url": "https://github.com/sponsors/ueberdosis"
@@ -3688,9 +3689,10 @@
       }
     },
     "node_modules/@tiptap/extension-placeholder": {
-      "version": "2.26.1",
-      "resolved": "https://registry.npmjs.org/@tiptap/extension-placeholder/-/extension-placeholder-2.26.1.tgz",
-      "integrity": "sha512-MBlqbkd+63btY7Qu+SqrXvWjPwooGZDsLTtl7jp52BczBl61cq9yygglt9XpM11TFMBdySgdLHBrLtQ0B7fBlw==",
+      "version": "2.27.2",
+      "resolved": "https://registry.npmjs.org/@tiptap/extension-placeholder/-/extension-placeholder-2.27.2.tgz",
+      "integrity": "sha512-IjsgSVYJRjpAKmIoapU0E2R4E2FPY3kpvU7/1i7PUYisylqejSJxmtJPGYw0FOMQY9oxnEEvfZHMBA610tqKpg==",
+      "license": "MIT",
       "funding": {
         "type": "github",
         "url": "https://github.com/sponsors/ueberdosis"
@@ -3713,9 +3715,10 @@
       }
     },
     "node_modules/@tiptap/extension-text": {
-      "version": "2.26.1",
-      "resolved": "https://registry.npmjs.org/@tiptap/extension-text/-/extension-text-2.26.1.tgz",
-      "integrity": "sha512-p2n8WVMd/2vckdJlol24acaTDIZAhI7qle5cM75bn01sOEZoFlSw6SwINOULrUCzNJsYb43qrLEibZb4j2LeQw==",
+      "version": "2.27.2",
+      "resolved": "https://registry.npmjs.org/@tiptap/extension-text/-/extension-text-2.27.2.tgz",
+      "integrity": "sha512-Xk7nYcigljAY0GO9hAQpZ65ZCxqOqaAlTPDFcKerXmlkQZP/8ndx95OgUb1Xf63kmPOh3xypurGS2is3v0MXSA==",
+      "license": "MIT",
       "funding": {
         "type": "github",
         "url": "https://github.com/sponsors/ueberdosis"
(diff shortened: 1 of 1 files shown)


shanevcantwell:
recheck

shanevcantwell:
@sestinj Just to personalize this a bit... I've been trying to love `Continue` with my local models for like 6mo now, but tool usage never worked. I finally discovered that there were multiple conflicting messages being prompted to the models. With this in place, and the following tool definitions, tools now "just work", with dozens of mixin tool calls strung together, and no broken tool call text in the stream. 

Every model is a little different, so making the tool prompts completely override-able is to me just a necessity. I tried to stick to your established patterns and touch as little as possible, and I think there is minimal drift.

I hope you'll consider this a nice xmas gift! 
Cheers

{
  "mergeBehavior": "merge",
  "tools": [
    {
      "name": "view_diff",
      "disabled": true
    },
    {
      "name": "read_file",
      "description": "Read the contents of an existing file. Use relative paths from workspace root (e.g., 'src/index.ts').",
      "systemMessageDescription": {
        "prefix": "To read a file, use the read_file tool with a relative path from workspace root. Example:"
      }
    },
    {
      "name": "create_new_file",
      "description": "Create a new file with the specified content. Use relative paths from workspace root.",
      "systemMessageDescription": {
        "prefix": "To create a new file, use the create_new_file tool with a relative path. Example:"
      }
    },
    {
      "name": "run_terminal_command",
      "description": "Run a shell command in the workspace directory. Stateless between calls. Use Edit tools for file modifications.",
      "systemMessageDescription": {
        "prefix": "To run a terminal command, use the run_terminal_command tool. Shell is stateless. Use Edit tools instead of sed/awk. Example:"
      }
    },
    {
      "name": "grep_search",
      "description": "Search file contents using regex patterns. Returns matching lines with context.",
      "systemMessageDescription": {
        "prefix": "To search file contents, use grep_search with a regex pattern. Example:"
      }
    },
    {
      "name": "glob_search",
      "description": "Find files by name pattern (e.g., '**/*.ts'). Use for locating files.",
      "systemMessageDescription": {
        "prefix": "To find files by name pattern, use glob_search. Example:"
      }
    },
    {
      "name": "multi_edit",
      "description": "Make multiple edits to files. Each edit: filepath, old_string (exact match), new_string.",
      "systemMessageDescription": {
        "prefix": "To edit files, use multi_edit with filepath, old_string, and new_string. Use relative paths. Example:"
      }
    }
  ],
  "chatOptions": {
    "baseSystemMessage": "You are an AI coding assistant. Focus on completing tasks efficiently. Use relative paths from workspace root for all file operations. Be concise."
  }
}

shanevcantwell:
Before & after for reference (with PR #9325 fix) (gpt-oss-20b)

image

core/index.d.ts
```diff
@@ -1139,6 +1139,30 @@ export interface Tool {
   ) => ToolPolicy;
 }
 
+/**
+ * Configuration for overriding built-in tool prompts.
+ * Allows customization of tool descriptions and behavior at the repo level.
+ */
+export interface ToolOverride {
```
RomneyDa: Ideally use `Partial<Pick<Tool, ....>>` or similar for this type

RomneyDa (changes requested):
@shanevcantwell the ability to override tool prompts would be a neat change. Currently JSON configuration is deprecated and we are very tentative to make changes that only work in JSON.

What if we put this under `chatOptions` in a model's YAML configuration? Or thoughts on other ways we could make this work for most users?

core/index.d.ts
shanevcantwell: Cheers, thanks for having a look. I missed that detail about JSON deprecation. No problem at all. I'll put the JSON implementation on a separate branch on my fork in case there is any use for it, and bring in your `chatOptions` suggestion to replace the PR code. I'll get right on getting it back through the CI.

RomneyDa (commented):
@shanevcantwell realized one more thing, it could be confusing to users if their `toolPromptOverrides` apply in VS Code but not in the CLI. Are you a CLI user by chance? If so could you take a look at implementing the same override application during CLI tool loading? If not I could take a look at it next week

Note @chezsmithy took a go at unifying the core and CLI tool types, somewhat relevant here but we wouldn't need to unify to apply the same toolPromptOverrides to each style of tool definition https://github.com/continuedev/continue/pull/8555

shanevcantwell:
> @shanevcantwell realized one more thing, it could be confusing to users if their `toolPromptOverrides` apply in VS Code but not in the CLI. Are you a CLI user by chance? If so could you take a look at implementing the same override application during CLI tool loading? If not I could take a look at it next week
> 
> Note @chezsmithy took a go at unifying the core and CLI tool types recently, somewhat relevant here but we wouldn't need to unify to apply the same toolPromptOverrides to each style of tool definition #8555

Absolutely. This new addition passes existing and a few new tests, but I don't have experience with the CLI to verify it working in realistic situations. New tests:

extensions/cli/src/tools/applyToolOverrides.test.ts (6 tests):

Returns tools unchanged when no overrides
Applies description override
Filters out disabled tools
Handles multiple overrides
Ignores overrides for non-existent tools
Does not mutate original tools
These test the CLI-specific applyChatCompletionToolOverrides function. The core applyToolOverrides function (for VS Code) already had tests from when we cherry-picked it.

RomneyDa:
@shanevcantwell waiting on the final go ahead from team since involves YAML schema change

core/index.d.ts
```diff
@@ -1139,6 +1142,30 @@ export interface Tool {
   ) => ToolPolicy;
 }
 
+/**
+ * Configuration for overriding built-in tool prompts.
+ * Allows customization of tool descriptions and behavior per model.
+ */
+export interface ToolOverride {
```
RomneyDa: Ideally use typescript `Pick` for this type to reduce duplication

RomneyDa:
@shanevcantwell looks like just the gui package lock changes are breaking CI, should work if reverted

RomneyDa:
@shanevcantwell appreciate the iterations!

PR #9325, opened 2025-12-26
fix: handle non-file URI schemes in runTerminalCommand for WSL2

### LINKED ISSUE #8091: Bug: Cannot run commands in terminal "The URL must be of scheme file"

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 1.2.7
- IDE version: 1.104.3
- Model: N/A
- config: N/A
```

### Description

When working in .devcontainers continue.dev cannot run commands in the terminal

Attempt to run results in error
```
run_terminal_command failed with the message: The URL must be of scheme file

Please try something else or request further instructions.
```

Running commands in the terminal does work with the native Gitub Copilot extension so there must be some detection mechanism the extension can perform to know which URL scheme to use.

From my limited understanding VSCode uses URI schemes to represent resources in different contexts.

`file://` - Local file system paths
`vscode-remote://` - Files inside remote environments (containers, SSH, WSL)

It seems `continue.dev` is hard-coded or not adapting to the environment it is running in and always using `file://` when it should be using `vscode-remote://` 
Although the error message "The URL must be of scheme file" makes it seem like it should be file:// but this may be misleading.

This issue might become part of #928

### To reproduce

1. Open Agent Chat
2. Enter prompt which causes continue.dev to suggest running command in terminal
3. Click "Run"

Expected:
Command executes successfully

Actual:
Error response

### Log output

```Shell

```

shanevcantwell:
## Problem
When running `run_terminal_command` in WSL2, it crashes with:
```
run_terminal_command failed with the message: The URL must be of scheme file
```

## Root Cause
The `resolveWorkingDirectory` function in `runTerminalCommand.ts` calls `fileURLToPath()` on workspace URIs. In WSL2/remote environments, `getWorkspaceDirs()` can return URIs like `vscode-remote://wsl+Ubuntu/path` instead of `file:///path`. The `fileURLToPath` function throws when the scheme isn't `file://`.

## Fix
Wrap the `fileURLToPath` call in a try-catch so it falls through to the existing HOME directory fallback instead of crashing. This is a minimal, defensive change - 5 lines added.

Fixes #8091

## Test plan
- [x] Tested locally in WSL2 environment where this was previously failing
- [x] Verified terminal commands execute successfully after fix

---
## Summary by cubic
Fixes a crash in runTerminalCommand on WSL2 and other remote environments by resolving vscode-remote URIs to working paths and safely handling non-file workspace URIs. Falls back to HOME when resolution fails.

- **Bug Fixes**
  - Generalize remote URI support: parse vscode-remote schemes (WSL, SSH, dev-container) and decode the pathname.
  - Prefer file:// workspaces and wrap fileURLToPath in try/catch; fall back to HOME when URI parsing fails.

<sup>Written for commit 8e5a7c2152a2e4e0baf88466900b258a9f2c28fd. Summary will update on new commits.</sup>

---

## Continue Tasks

| Status | Task | Actions |
|:-------|:-----|:--------|
| ✅ Merged | Update docs on PR | [PR](https://github.com/continuedev/continue/pull/9325) · [View](https://hub.continue.dev/tasks/4d73e6e2-a25b-448e-abc6-f3f89e5f7066) |
| ▶️ Queued | Optimize Website Performance | [View](https://hub.continue.dev/tasks/01c4ca39-fa59-4bc5-a943-46f9948d43d9) |


shanevcantwell:
Before & after for reference. (With PR #9314 fix.) (gpt-oss-20b)
image

core/tools/implementations/runTerminalCommand.ts
```diff
@@ -48,12 +48,32 @@ import { getBooleanArg, getStringArg } from "../parseArgs";
  * Falls back to home directory or temp directory if no workspace is available.
  */
 function resolveWorkingDirectory(workspaceDirs: string[]): string {
+  // Handle vscode-remote://wsl+distro/path URIs (WSL2 remote workspaces)
+  const wslWorkspaceDir = workspaceDirs.find((dir) =>
+    dir.startsWith("vscode-remote://wsl"),
```
RomneyDa: Is there a way we could write this that isn't so specific to `vscode-remote://wsl`? I suppose the same issue applies to any non-file URIs? If not obvious this is fine for now but it feels a bit specific

CLAUDE.md
RomneyDa, line 1: We use AGENTS.md instead of CLAUDE.md, if any updates are need to AGENTS.md could you make them there?

RomneyDa (changes requested):
@shanevcantwell Thanks for the contribution! The only blocking change is the CLAUDE.md note

CLAUDE.md
shanevcantwell: Oops, that was an oversight to be included at all. Removed.

core/tools/implementations/runTerminalCommand.ts
shanevcantwell: Agreed, that was overly narrow. I'll re-submit with it more widely applicable.

PR #9327, opened 2025-12-26
feat(cli): add allowHeadless option for MCP servers in headless mode

aaronlippold:
## Description

Adds `allowHeadless: true` config option for MCP servers, enabling specific MCP tools to work in headless mode without requiring the `--auto` flag.

**Problem:** MCP tools are unavailable in headless mode, limiting automation scenarios where users want specific trusted MCP servers to work while still requiring approval for other tools.

**Solution:** Per-server `allowHeadless` option that allows enumeration and execution of tools from that server in headless mode.

**Usage:**
```yaml
mcpServers:
  - name: Brave Search
    command: npx
    args: ["-y", "@modelcontextprotocol/server-brave-search"]
    allowHeadless: true  # Enable in headless mode
```

### Commit 1: Feature (`4062ff6`)
- Add `allowHeadless` to MCP config schema (`packages/config-yaml`)
- Pass `allowHeadless` from connection config to Tool objects  
- Check `allowHeadless` in tool enumeration (`getRequestTools`)
- Check `allowHeadless` in execution permission (`checkToolPermissionApproval`)
- **Security:** `allowHeadless` respects explicit "exclude" permissions (won't override user denials)
- Add 9 unit tests in `mcp-headless.test.ts`

### Commit 2: Test Isolation Fix (`4db8aa8`)
- Mock `core/util/paths.js` to read `CONTINUE_GLOBAL_DIR` dynamically
- Fixes test isolation issue where `GlobalContext` singleton would leak state between parallel test files
- Add `interopDefault: true` to vitest config for better CommonJS handling

## AI Code Review

## Checklist

N/A - This is a CLI/config change with no UI impact.

## Tests

Added 9 unit tests in `extensions/cli/src/stream/mcp-headless.test.ts`:

**Tool enumeration tests (4):**
- Exclude MCP tools by default in headless mode
- Include MCP tools when `allowHeadless=true` in headless mode
- Include all MCP tools in interactive mode regardless of `allowHeadless`
- Handle multiple MCP servers with different `allowHeadless` settings

**Execution permission tests (5):**
- Approve MCP tool with `allowHeadless=true` in headless mode
- Deny MCP tool without `allowHeadless` in headless mode
- Deny MCP tool with `allowHeadless=undefined` in headless mode
- Approve explicitly allowed tools regardless of `allowHeadless`
- **Deny explicitly excluded tools even with `allowHeadless=true`** (security test)

All existing tests continue to pass (1715 unit tests, 66 e2e tests).

---

## Continue Tasks

| Status | Task | Actions |
|:-------|:-----|:--------|
| 🔴 Closed | Update docs on PR | [PR](https://github.com/continuedev/continue/pull/9327) · [View](https://hub.continue.dev/tasks/5a599da6-1ab5-4944-bc22-cecf9c30414b) |

### DIFF
--- a/extensions/cli/vitest.config.ts
+++ b/extensions/cli/vitest.config.ts
@@ -6,6 +6,10 @@ export default defineConfig({
     environment: "node",
     setupFiles: ["./vitest.setup.ts"],
     exclude: ["**/node_modules/**", "**/dist/**", "**/*.e2e.*", "**/e2e/**"],
+    // Enable CommonJS interop for packages like find-up v5 that use module.exports
+    deps: {
+      interopDefault: true,
+    },
     coverage: {
       reporter: ["text", "json", "html"],
       exclude: [
--- a/extensions/cli/vitest.setup.ts
+++ b/extensions/cli/vitest.setup.ts
@@ -1,3 +1,6 @@
+import * as fs from "fs";
+import * as path from "path";
+
 import { vi } from "vitest";
 
 import { resetConsoleOverrides } from "./src/init.js";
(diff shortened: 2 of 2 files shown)


aaronlippold:
## Review Comments Addressed

Both AI review comments have been resolved:

### P1 (Security) - `allowHeadless` bypassing explicit exclusions ✅
**Fixed in `4062ff6`**

The `allowMcpInHeadless` check now respects explicit "exclude" permissions:
```typescript
const allowMcpInHeadless =
  !tool.isBuiltIn &&
  isHeadless &&
  tool.allowHeadless &&
  result.permission !== "exclude";  // ← Added this check
```

### P2 (Test) - Helper coercing `undefined` to `false` ✅
**Fixed in `4062ff6`**

The test helper now preserves `undefined` instead of coercing to `false`:
```typescript
// Before (bug):
allowHeadless: allowHeadless ?? false,

// After (fixed):
...(allowHeadless !== undefined && { allowHeadless }),
```

This ensures the test "should deny MCP tool with allowHeadless=undefined" actually tests `undefined` behavior, not `false`.

---

Both review threads are now outdated as the underlying code has been updated.

aaronlippold:
That could be a different way to solve it. I can go through it again and see if that would fix it.

aaronlippold:
That could be a different way to solve it. I can go through it again and see if that would fix it.

RomneyDa:
@aaronlippold double checked this with the team and confirmed that we want MCP tools to just work out of the box in headless, so likely just a tweak to tool permissions should fix. Currently when using `--agent` any MCP tools in the agent will be allowed, and I think using `--mcp` you'd expect the same, so I think `--config` acting differently is probably more confusing than not, although I agree that the ultra-security conscious approach would be to default to false. I think for now, security-conscious users can pass a custom `--config` or use `--mcp`s with only the tools they want. 

That being said we should fix the bug where MCP tools simply aren't available in headless. I think the bug is likely here in tool permission service, where we don't actually have handling for the `mcp:` syntax (likely an uncaught hallucination)
```
// Allow MCP tools (no way to know if they're read only but shouldn't disable mcp usage in plan)
          { tool: "mcp:*", permission: "allow" },
```

What are your thoughts on this?

aaronlippold:
Hi, my goal is resolution of the functionality so if the team has a way
that they’re comfortable with resolving that it doesn’t have to be my Way.
I’m happy if we think there’s a more efficient path. Happy to work with the
team on that.

--------
Aaron Lippold

***@***.***

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

RomneyDa:
I'll take a look at a permissions based fix real quick

RomneyDa:
@aaronlippold took a swing at fixing headless for MCP and bash here!
https://github.com/continuedev/continue/pull/9538

Added bash because we think it's sensible to be able to use bash in headless and that in the vast majority of cases people would rather include it by default and use `--exclude` to exclude it. But left the edit tools out still.

PR #9465, opened 2026-01-09
fix: compact conversation with dangling tool calls

uinstinct:
## Description

Remove dangling tool calls when compacting conversation.

resolves CON-5063
resolves CON-4969

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/f257219c-41a2-4c6d-be86-dbc0d8745eb8

https://github.com/user-attachments/assets/2e300319-61dc-4eec-ab9b-0d9ab0b0192f

## Tests

---

**Continue Tasks:** ▶️ 2 queued — [View all](https://hub.continue-stage.tools/inbox?pr=https%3A%2F%2Fgithub.com%2Fcontinuedev%2Fcontinue%2Fpull%2F9465)

---
## Summary by cubic
Fix conversation compaction to handle dangling tool calls by inserting "Tool cancelled" tool messages when a tool call has no output. Assistant messages now only reference tool calls with a corresponding result or cancellation (resolves CON-5063, CON-4969).

- **Bug Fixes**
  - Insert "Tool cancelled" tool messages for assistant toolCalls with no matching tool message in the filtered history.
  - Build compacted messages from the cleaned history to avoid orphan “waiting for approval” tool calls.

<sup>Written for commit e08aa8533adc016084bdd29e9f4d31741ab1815e. Summary will update on new commits.</sup>

### DIFF
--- a/core/util/conversationCompaction.ts
+++ b/core/util/conversationCompaction.ts
@@ -55,15 +55,14 @@ export async function compactConversation({
     messages.push(item.message);
     // toolcalls only exist in an assistant message
     if (item.message.role === "assistant" && item.message.toolCalls) {
-      // for every toolcall, if there is no toolcallstate or its output, add a chat message saying that it is empty
+      // for every toolcall, if there is no tool message with a tool call id already, add a chat message saying that it is empty
       item.message.toolCalls.forEach((toolCall) => {
-        const foundToolCallState = item.toolCallStates?.find(
-          (toolCallState) => toolCallState.toolCallId === toolCall.id,
-        );
         if (
-          !foundToolCallState ||
-          !foundToolCallState.output ||
-          !foundToolCallState.output.length
+          !filteredHistory.find(
+            (item) =>
+              item.message.role === "tool" &&
+              item.message.toolCallId === toolCall.id,
+          )
         ) {
           messages.push({
             role: "tool",


RomneyDa (changes requested):
@uinstinct I think this kind of filtering can result in a message being eliminated from the middle of the conversation. Probably an edge case but that can cause further API errors. I think the approach you took in the CLI of inserting a "No tool output" or "Tool cancelled" message is appropriate so that conversational ordering is preserved.

uinstinct:
> @uinstinct I think this kind of filtering can result in a message being eliminated from the middle of the conversation. Probably an edge case but that can cause further API errors. I think the approach you took in the CLI of inserting a "No tool output" or "Tool cancelled" message is appropriate so that conversational ordering is preserved.

makes sense. implemented!

RomneyDa:
Edit, that appears to be a CLI issue and this is for vscode/jetbrains. I did an attempt at the CLI fix here
https://github.com/continuedev/continue/pull/9596

But I think the concern is valid, let's check if tool message is there, not just tool output

uinstinct:
> @uinstinct I noticed we're checking the tool call states but not the tool messages. I wonder if, IF there were already a corresponding tool message but not tool call output in the state (for some reason), this approach would cause duplicate tool messages. Could you also add checks for if a tool message is already present with the matching tool call id?

Fair point. Implemented that changes!

(I had thought of using toolCallStates because it was a smaller array than conversation history)

RomneyDa (approved):
Fixes dangling tool calls during compaction

PR #9496, opened 2026-01-13
fix: use proxy for unrendered injected block secrets

RomneyDa:
## Description
CLI was filtering secrets out in a way that prevented the proxy being used for unrendered secrets
Also updated config unrolling to use `extractFQSNMap` for injected blocks, which is more semantic

---

## Continue Tasks

| Status | Task | Actions |
|:-------|:-----|:--------|
| ▶️ Queued | Update docs on PR | [View](https://hub.continue.dev/tasks/f251320d-71b0-4b56-a93f-6e93f21b8e2f) |
| ▶️ Queued | Optimize Website Performance | [View](https://hub.continue.dev/tasks/2487fae9-23b6-4f9d-9fd4-d8293cad53a6) |

---
## Summary by cubic
Fixes CLI secret resolution so unrendered injected block secrets use the proxy instead of being filtered out. Also updates block injection to generate proper FQSNs, ensuring secrets are correctly namespaced for proxy resolution.

- **Bug Fixes**
  - CLI: Check local env first; call API only for unresolved; keep all found API results (even without value) for models_add_on/free_trial proxying; ignore NotFound; handle API errors gracefully.
  - Config unrolling: After replacing inputs with secrets, render FQSNs using extractFQSNMap and the injected block’s identifier (slug or file) so injected secrets resolve via the proxy.

<sup>Written for commit 6b3c0be2f1443e47db3e985560cb859b8e42f6a9. Summary will update on new commits.</sup>

### DIFF
--- a/extensions/cli/src/CLIPlatformClient.test.ts
+++ b/extensions/cli/src/CLIPlatformClient.test.ts
@@ -76,7 +76,8 @@ describe("CLIPlatformClient", () => {
         secretName: "ANTHROPIC_API_KEY",
       };
 
-      // No local env secret
+      // Ensure no local env secret exists
+      vi.stubEnv("ANTHROPIC_API_KEY", undefined as unknown as string);
       vi.mocked(fs.existsSync).mockReturnValue(false);
 
       // API returns a user secret with value
@@ -114,7 +115,8 @@ describe("CLIPlatformClient", () => {
         secretName: "ANTHROPIC_API_KEY",
       };
 
-      // No local env secret
+      // Ensure no local env secret exists
+      vi.stubEnv("ANTHROPIC_API_KEY", undefined as unknown as string);
       vi.mocked(fs.existsSync).mockReturnValue(false);
 
       // API returns a models_add_on result (has secretLocation but no value)
@@ -147,7 +149,8 @@ describe("CLIPlatformClient", () => {
         secretName: "OPENAI_API_KEY",
       };
 
-      // No local env secret
+      // Ensure no local env secret exists
+      vi.stubEnv("OPENAI_API_KEY", undefined as unknown as string);
       vi.mocked(fs.existsSync).mockReturnValue(false);
 
       // API returns a free_trial result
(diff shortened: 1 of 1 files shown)


packages/config-yaml/src/load/unroll.ts
```diff
@@ -550,10 +550,20 @@ export async function unrollBlocks(
...
           try {
             const blockConfigYaml = await registry.getContent(injectBlock);
+            // Convert inputs to secrets, then convert secrets to FQSNs using the injected block's identifier
+            // This ensures secrets are properly namespaced for proxy resolution (e.g., models add-on)
             const blockConfigYamlWithSecrets =
               replaceInputsWithSecrets(blockConfigYaml);
-            const resolvedBlock = parseMarkdownRuleOrConfigYaml(
+            const blockConfigYamlWithFQSNs = renderTemplateData(
               blockConfigYamlWithSecrets,
+              {
```
RomneyDa, line 559: This includes block in FQSN e.g. `secrets./anthropic/claude/ANTHROPIC_API_KEY` rather than `secrets.///ANTHROPIC_API_KEY` which is needed for checking models add on eligibility

extensions/cli/src/CLIPlatformClient.ts
```diff
@@ -96,38 +96,44 @@ export class CLIPlatformClient implements PlatformClient {
...
-      if (!results[i] || !("value" in results[i]!)) {
-        const secretFromEnv = this.findSecretInLocalEnvFiles(fqsns[i]);
-        if (secretFromEnv?.found) {
-          results[i] = secretFromEnv;
+    // For secrets not found locally, try to resolve through the API
+    const unresolvedIndices = results
+      .map((r, i) => (r === undefined ? i : -1))
+      .filter((i) => i !== -1);
+
+    if (unresolvedIndices.length > 0) {
```
RomneyDa, line 113: This diff looks big but it's just reordering to check process.env for secrets and then not filter out secrets with values so that `useProxyForUnrenderedSecrets` can work

Patrick-Erichsen (approved):

PR #9525, opened 2026-01-14
fix: prevent waiting for the session to load from history

uinstinct:
## Description

Earlier when loading a session from the history page, it would take some time to load the session. This was happening because the previous session was saved with a new title generation (which in turn would make an llm call). This PR fixes that by making the session saving asynchronous.

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/39cb93a8-0a36-492c-a553-8093f8df9e33

https://github.com/user-attachments/assets/ea2c0ee9-1628-4661-b6c0-8dcab3e3fd44

## Tests

---

## Continue Tasks

| Status | Task | Actions |
|:-------|:-----|:--------|
| ▶️ Queued | Create GitHub Issue (OS) | [View](https://hub.continue-stage.tools/tasks/eddad576-605c-45ec-995c-c5cd636ca936) |
| ▶️ Queued | Update PostHog Dashboards | [View](https://hub.continue-stage.tools/tasks/c00808a5-92c8-45a4-94af-1f6db1fcbd4e) |

---
## Summary by cubic
Loading a session from the history page no longer waits for the previous session to save and generate a title. Saving now runs in the background, so sessions load immediately.

- **Bug Fixes**
  - Run saveCurrentSession asynchronously to avoid blocking on title generation.
  - Dispatch selectChatModelForProfile without awaiting to remove extra delay.

<sup>Written for commit dd136922632bca5f7dc4ebf192ed18a88f6acc3e. Summary will update on new commits.</sup>

### DIFF
--- a/gui/src/redux/thunks/session.ts
+++ b/gui/src/redux/thunks/session.ts
@@ -108,15 +108,12 @@ export const loadSession = createAsyncThunk<
   async ({ sessionId, saveCurrentSession: save }, { extra, dispatch }) => {
     if (save) {
       // save the session in the background
-      void (async () => {
-        const result = await dispatch(
-          saveCurrentSession({
-            openNewSession: false,
-            generateTitle: true,
-          }),
-        );
-        unwrapResult(result);
-      })();
+      void dispatch(
+        saveCurrentSession({
+          openNewSession: false,
+          generateTitle: true,
+        }),
+      );
     }
     const session = await getSession(extra.ideMessenger, sessionId);
     dispatch(newSession(session));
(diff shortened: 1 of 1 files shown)


RomneyDa (changes requested):
As far as I can tell since saveCurrentSession uses values from redux this would introduce a race condition if new session is loaded before current session is saved. We need to either only make the chat title generation async somehow or make sure that the history for previous session is the one being saved before moving on to loading the new one

uinstinct:
> As far as I can tell since saveCurrentSession uses values from redux this would introduce a race condition if new session is loaded before current session is saved. We need to either only make the chat title generation async somehow or make sure that the history for previous session is the one being saved before moving on to loading the new one

makes sense. thanks for point this out!

implemented the changes

RomneyDa (approved):

PR #9567, opened 2026-01-16
fix: autocompaction failures from exceeded context length

RomneyDa:
## Description
- Read file tool no longer truncates when limit is hit, just throws Error and explains why and ways to get around it with bash tools etc. Partial file output isn't usually very useful. Claude handles similarly.
- Truncate bash and read tool output proportionally by the number of parallel tool calls, and notify model if they were truncated more than usual because of parallel calls
- Update the autocompaction threshold to be `contextLength - maxTokens - Min(maxTokens, 15k)`. This is based off of reasoning that if we leave contextLength - maxTokens - output, we will always be able to compact the next turn, and data shows that max tokens > 15k is very rare. This doesn't account for tool output exceeding max tokens but the above change should reduce that issue quite a bit. It's difficult to deterministically prevent tool output from exceeding context length without complicated and/or partial pruning or super aggressive truncation but some complex pruning will likely be worth it in the near future
- Use rough token multiplier with 10% buffer to account for tokenizers which will have higher token counts, e.g. often 25% higher tokens in claude then gpt. Used [this article](https://medium.com/@disparate-ai/not-all-tokens-are-created-equal-7347d549af4d)
- Removes duplicate token counting step by rearranging logic a bit

---

**Continue Tasks:** ▶️ 2 queued — [View all](https://hub.continue-stage.tools/inbox?pr=https%3A%2F%2Fgithub.com%2Fcontinuedev%2Fcontinue%2Fpull%2F9567&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Fixes auto-compaction failures by reserving space for the compaction prompt and centralizing compaction so we stay within the model’s context window. Also prevents context spikes from tool outputs with clear file-size errors and proportional limits when tools run in parallel, and adjusts token counting for non-GPT models.

- **Bug Fixes**
  - Reserve 150 tokens for the compaction prompt during pruning.
  - Route all paths through handleAutoCompaction; remove premature checks that caused missed or late compactions.
  - Use a ratio-based buffer for auto-compaction: contextLength − maxTokens − min(max(maxTokens, 20% of remaining), 15k) to compact earlier and avoid overruns.
  - Adjust token counts for Anthropic, Gemini, and Mistral models using safe multipliers.
  - readFile now throws a FileTooLarge error with counts instead of truncating.
  - Divide readFile and bash output limits by the number of parallel tool calls to prevent overflow; add a note when truncation occurs.

- **Refactors**
  - Extract compaction prompt to a constant and make TUI compaction delegate to the core auto-compaction handler via callbacks for consistent behavior.

<sup>Written for commit c0a930919fe93ce7ad6a0eaf3546a956dee02383. Summary will update on new commits.</sup>

### DIFF
--- a/extensions/cli/src/tools/subagent.test.ts
+++ b/extensions/cli/src/tools/subagent.test.ts
@@ -97,7 +97,7 @@ describe("subagentTool", () => {
         prompt: "Subagent prompt",
         subagent_name: "code-agent",
       },
-      { toolCallId: "tool-call-id" },
+      { toolCallId: "tool-call-id", parallelToolCallCount: 1 },
     );
 
     expect(vi.mocked(executeSubAgent)).toHaveBeenCalledTimes(1);


extensions/cli/src/tools/readFile.ts
```diff
@@ -82,20 +79,26 @@ export const readFileTool: Tool = {
...
 
-      const { output: truncatedContent, wasTruncated } =
-        truncateByLinesAndChars(content, maxLines, maxChars, "file content");
-
-      if (wasTruncated) {
-        return `Content of ${filepath} (truncated):\n${truncatedContent}`;
+      if (charCount > maxChars || lineCount > maxLines) {
+        throw new ContinueError(
+          ContinueErrorReason.FileTooLarge,
+          `File is too large to read: ${filepath} (${charCount.toLocaleString()} characters, ${lineCount.toLocaleString()} lines). ` +
```
RomneyDa: New behavior: don't truncate to large limits, since the truncated file is often useless context pollution, instead, throw an error after full file can't fit within limit

sestinj (approved):
👍

PR #8935, opened 2025-11-29
fix(openai): responses function_call mapping

Quazistax:
Maybe fixes #8773 , with 2nd commit might help with #8933 too (I tried it with VS Code, not with JetBrains)

## Description

* to handle function_call messages loaded from session file
* use call_id as item id if response id is not available
* ensure function_call item id is prefixed with fc

## AI Code Review

- **Team members only**: AI review runs automatically when PR is opened or marked ready for review
- Team members can also trigger a review by commenting `@continue-review`

## Checklist

- [] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [] The relevant docs, if any, have been updated or created
- [] The relevant tests, if any, have been updated or created

## Screen recording or screenshot

[ When applicable, please include a short screen recording or screenshot - this makes it much easier for us as contributors to review and understand your changes. See [this PR](https://github.com/continuedev/continue/pull/6455) as a good example. ]

## Tests

[ What tests were added or updated to ensure the changes work as expected? ]

---
## Summary by cubic
Fixes function_call mapping in OpenAI responses to emit valid items from session-loaded messages with stable, prefixed IDs and preserve original per-tool item IDs. Addresses #8773.

- **Bug Fixes**
  - Emit one function_call per tool call, preferring responsesOutputItemId; fallback to respId or call_id; ensure IDs are prefixed with fc_.
  - Propagate responsesOutputItemId through tool call state during streaming.
  - Fallback to an assistant message when no toolCalls exist but respId is present.

<sup>Written for commit c4ed9e4100ab646608ee9cb1de024143983831b0. Summary will update automatically on new commits.</sup>

PR #9659, opened 2026-01-19
fix: OpenAI Responses API parallel tool calls losing call_ids

### LINKED ISSUE #8773: No tool call found for function call output with call_id

### Relevant environment info

```Markdown
- OS:win11
- Continue version: 10.0.52
- IDE version: IntelliJ IDEA 2025.2.4 (Ultimate Edition)
- Model: GPT-4.1
- config:
  
name: Local Assistant
version: 1.0.0
schema: v1
models:
  - uses: openai/gpt-5
    with:
      OPENAI_API_KEY: sk-proj-XXX
  - uses: openai/o3
    with:
      OPENAI_API_KEY: sk-proj-XXX
  - uses: openai/o3
    with:
      OPENAI_API_KEY: sk-proj-XXX
  - uses: openai/gpt-4.1-mini
    with:
      OPENAI_API_KEY: sk-proj-XXX
  - name: Autodetect
    provider: openai
    model: AUTODETECT
    requestOptions:
      timeout: 60000  # Request timeout in milliseconds
      verifySsl: true
    apiKey: sk-proj-XXX
context:
  - provider: code
  - provider: docs
  - provider: diff
  - provider: terminal
  - provider: problems
  - provider: folder
  - provider: codebase

  
```

### Description

When working with continue and all tools activated I allways get this error after some time. 
Sometimes it helps to resubmit last message or edit last prompt to start generating from there, but 
most sessions do not recover from that and I have to live with half-finished tasks an have to start a new session, 
which is extremely annoying and time consuming.

Error-output:
{"message":"No tool call found for function call output with call_id call_ubrVpxVSFon3MNAk8df8F0kO.","type":"invalid_request_error","param":"input","code":null}

### To reproduce

_No response_

### Log output

```Shell
{
  "error": {
    "message": "No tool call found for function call output with call_id call_ubrVpxVSFon3MNAk8df8F0kO.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
}

Code: undefined
Error number: undefined
Syscall: undefined
Type: undefined

Error: HTTP 400 Bad Request from https://api.openai.com/v1/responses

{
  "error": {
    "message": "No tool call found for function call output with call_id call_ubrVpxVSFon3MNAk8df8F0kO.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
}
    at aEt.parseError (C:\snapshot\continue\binary\out\index.js:8421:15852)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async n (C:\snapshot\continue\binary\out\index.js:8423:144)
    at async Fht (C:\snapshot\continue\binary\out\index.js:8161:3322)
    at async aEt._streamResponses (C:\snapshot\continue\binary\out\index.js:8444:7591)
    at async aEt.responsesStream (C:\snapshot\continue\binary\out\index.js:8438:4965)
    at async aEt.streamChat (C:\snapshot\continue\binary\out\index.js:8438:6480)
    at async G0a (C:\snapshot\continue\binary\out\index.js:9631:2628)
    at async C:\snapshot\continue\binary\out\index.js:9690:6927 
[2025-11-18T16:38:40] Error: HTTP 400 Bad Request from https://api.openai.com/v1/responses

{
  "error": {
    "message": "No tool call found for function call output with call_id call_ubrVpxVSFon3MNAk8df8F0kO.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
} 
[2025-11-18T16:38:40] Error running handler for "llm/streamChat":  Error: HTTP 400 Bad Request from https://api.openai.com/v1/responses

{
  "error": {
    "message": "No tool call found for function call output with call_id call_ubrVpxVSFon3MNAk8df8F0kO.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
}
```

houssemzaier:
Fixes #8773

When GPT-5 makes parallel tool calls via the Responses API, each `function_call` has a unique `call_id` (fc_*). These IDs were being overwritten instead of accumulated, causing:

> "No tool call found for function call output with call_id fc_xxx"

## Changes

- `gui/src/redux/slices/sessionSlice.ts`: Accumulate IDs in `responsesOutputItemIds[]` array during streaming
- `core/llm/openaiTypeConverters.ts`: Emit `function_call` for each tool call with positional ID matching
- `core/llm/openaiTypeConverters.test.ts`: Added comprehensive tests for parallel tool call scenarios

## How it works

1. During streaming, when `fromResponsesChunk` sets `message.metadata.responsesOutputItemId`, we now accumulate these into an array `responsesOutputItemIds[]`
2. When converting to Responses API input format, we iterate over all tool calls and match each one with its corresponding fc_ ID by position
3. This ensures each parallel tool call gets its correct `call_id` when sent back to OpenAI

## Test plan

- [x] Unit tests pass (14 new tests covering parallel tool calls, edge cases)
- [x] Manual testing with GPT-5 and GPT-5.1 parallel tool calls

Related: #8935

---

**Continue Tasks:** ▶️ 2 queued · ▶️ 1 not started — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fcontinuedev%2Fcontinue%2Fpull%2F9659&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Fixes lost call_id mapping for parallel tool calls in the OpenAI Responses API, preventing “No tool call found for function call output” errors and restoring correct pairing between function_call and tool outputs.

- **Bug Fixes**
  - Accumulate fc_ IDs during streaming in responsesOutputItemIds[] (keeps the singular field for backward compatibility).
  - Emit a function_call item for each tool call, matching IDs by position; also emit assistant text when present.
  - Add unit tests covering parallel calls, legacy scenarios, and edge cases.

<sup>Written for commit 540e15f60e4bfdbb46e2fba534d4dc1543743492. Summary will update on new commits.</sup>

### DIFF
--- a/core/llm/openaiTypeConverters.test.ts
+++ b/core/llm/openaiTypeConverters.test.ts
@@ -1,0 +1,511 @@
+import { toResponsesInput } from "./openaiTypeConverters";
+import { ChatMessage } from "..";
+import type { ResponseInputItem } from "openai/resources/responses/responses.mjs";
+
+// Types for test assertions (simplified from OpenAI types)
+type FunctionCallItem = {
+  type: "function_call";
+  id: string;
+  name: string;
+  arguments: string;
+  call_id: string;
+};
+
+type FunctionCallOutputItem = {
+  type: "function_call_output";
+  call_id: string;
+  output: string;
+};
+
+type MessageItem = {
+  role: string;
+  content: unknown;
+};
+
+// Helper functions for filtering results
+function getFunctionCalls(items: ResponseInputItem[]): FunctionCallItem[] {
+  return items.filter(
+    (item) => (item as { type?: string }).type === "function_call",
+  ) as unknown as FunctionCallItem[];
+}
+
+function getFunctionCallOutputs(
+  items: ResponseInputItem[],
+): FunctionCallOutputItem[] {
+  return items.filter(
+    (item) => (item as { type?: string }).type === "function_call_output",
+  ) as unknown as FunctionCallOutputItem[];
+}
+
+function getMessagesByRole(
+  items: ResponseInputItem[],
+  role: string,
+): MessageItem[] {
+  return items.filter((item) => {
+    const msg = item as { role?: string; type?: string };
+    return msg.role === role && (!msg.type || msg.type === "message");
+  }) as unknown as MessageItem[];
+}
+
+describe("openaiTypeConverters", () => {
+  describe("toResponsesInput", () => {
+    describe("tool calls handling - OpenAI Responses API", () => {
+      it("should emit function_call items when fc_ ID is in metadata", () => {
+        const messages: ChatMessage[] = [
+          {
+            role: "user",
+            content: "List files in current directory",
+          },
+          {
+            role: "assistant",
+            content: "",
+            toolCalls: [
+              {
+                id: "call_abc123",
+                type: "function",
+                function: {
+                  name: "filesystem_list_directory",
+                  arguments: '{"path": "."}',
+                },
+              },
+            ],
+            metadata: { responsesOutputItemId: "fc_abc123" },
+          } as ChatMessage,
+        ];
+
+        const result = toResponsesInput(messages);
+
+        const functionCalls = getFunctionCalls(result);
+        expect(functionCalls.length).toBe(1);
+        expect(functionCalls[0].id).toBe("fc_abc123");
+        expect(functionCalls[0].call_id).toBe("call_abc123");
+        expect(functionCalls[0].name).toBe("filesystem_list_directory");
+      });
+
+      it("should NOT emit function_call when no fc_ ID in metadata", () => {
+        const messages: ChatMessage[] = [
+          {
+            role: "assistant",
+            content: "I'll help.",
+            toolCalls: [
+              {
+                id: "call_xyz",
+                type: "function",
+                function: {
+                  name: "some_tool",
+                  arguments: "{}",
+                },
+              },
+            ],
+            // No metadata with fc_ ID
+          } as ChatMessage,
+        ];
+
+        const result = toResponsesInput(messages);
+
+        const functionCalls = getFunctionCalls(result);
+        expect(functionCalls.length).toBe(0);
+
+        // Text should still be emitted
+        const assistantMessages = getMessagesByRole(result, "assistant");
+        expect(assistantMessages.length).toBe(1);
+      });
+
+      it("should emit text content alongside tool calls", () => {
+        const messages: ChatMessage[] = [
+          {
+            role: "assistant",
+            content: "Let me help you with that.",
+            toolCalls: [
+              {
+                id: "call_xyz",
+                type: "function",
+                function: {
+                  name: "some_tool",
+                  arguments: "{}",
+                },
+              },
+            ],
+            metadata: { responsesOutputItemId: "fc_xyz" },
+          } as ChatMessage,
+        ];
+
+        const result = toResponsesInput(messages);
+
+        const functionCalls = getFunctionCalls(result);
+        const assistantMessages = getMessagesByRole(result, "assistant");
+
+        expect(functionCalls.length).toBe(1);
+        expect(assistantMessages.length).toBe(1);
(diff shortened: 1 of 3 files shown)


houssemzaier:
Thanks for the review @RomneyDa!

  **To reproduce the issue:**

  1. Use GPT-5 or GPT-5.1 with the Responses API
  2. Configure an MCP server with filesystem tools
  3. Create a folder with 3+ files (e.g. `/manual-testing-sandbox`)

  **Prompt:**
  Using my Local MCP server: Read the first 50 lines of every file in this workspace.

  **What happens without the fix:**
  The model issues parallel `read_file` calls. Each one gets a unique `fc_*` ID from OpenAI (e.g. `fc_001`, `fc_002`, `fc_003`). When we send tool results back, OpenAI returns:

  > "No tool call found for function call output with call_id fc_xxx"

  **Why this happens:**
  The `fc_*` IDs are server-assigned by the Responses API and must be echoed back exactly, they can't be regenerated client-side.

  The overwriting bug exists because Continue merges parallel tool calls into a single `ChatMessage` with a `toolCalls[]` array. With 3 parallel tool calls you end up with:

  - 1 `ChatMessage`
  - 3 entries in `toolCalls[]`
  - but only 1 `responsesOutputItemId` field (which gets overwritten by the last call)

  **Why this fix:**
  The array is the smallest change that correctly stores *N* IDs for *N* tool calls. It's also backwards-compatible: the existing singular field is still populated for any code that relies on it.

  **Why not change the data model?**
  We could emit separate `ChatMessage`s per tool call, but that's a much bigger refactor across the codebase and would likely break existing provider implementations.

  If you'd prefer a different approach, I'm happy to align, but I'd like to keep this PR scoped to the minimal/compatible fix so we don't have to rework the whole tool-call pipeline.

RomneyDa:
> The fc_* IDs are server-assigned by the Responses API and must be echoed back exactly, they can't be regenerated client-side.

> The overwriting bug exists because Continue merges parallel tool calls into a single ChatMessage with a toolCalls[] array. With 3 parallel tool calls you end up with:

> 1 ChatMessage
3 entries in toolCalls[]
but only 1 responsesOutputItemId field (which gets overwritten by the last call)

@houssemzaier I might be missing something but is the `responseOutputItemId` fundamentally different than the `toolCallId`? I'm just questioning if we need `responseOutputItemId` at all, since each tool call already has an ID which is typically set on the provider side and we usually just translate between the two. You're right that this would change existing tool call pipeline but I think the changes would be minimal. Remove from redux, fix whatever translation isn't happening where it should be in responses code. Probably same ish PR size

If there is a difference in the nature of the IDs being stored than I would agree that this is the right approach

houssemzaier:
@RomneyDa Yep ! there are two different identifiers on a Responses API function_call:

  - call_id (call_abc123): the tool-call id you must echo back in function_call_output
  - id (fc_xyz789): the function_call item id (aka item_id in streaming events)

  For returning the tool result, only call_id is used for pairing:

  {
    "type": "function_call_output",
    "call_id": "call_abc123",
    "output": "..."
  }

  The 400 happens when OpenAI can't find a pending function_call matching that call_id , which usually means the next request doesn't include the original function_call item(s) in the replayed history (or it got altered during
  reconstruction). Since toResponsesInput() replays history (stateless mode), we need to preserve and replay the original function_call items verbatim (including their id + call_id) instead of regenerating them.

  That's why we keep the streamed function_call item ids around for parallel calls.

  Ref: https://community.openai.com/t/issue-with-new-responses-api-400-no-tool-call-found-for-function-call-output-with-call-id/1142327

RomneyDa:
@houssemzaier thanks for explaining! Then looks good to me.

RomneyDa:
Looks like a linting check is failing, may need to split up a function or something. Our nesting rule is a bit strict

houssemzaier:
@RomneyDa Fixed the linting errors 
  I extracted helper functions to reduce complexity and flattened test nesting

RomneyDa (approved):
Thanks for the contribution!

Issue #9675, opened 2026-01-19
Thinking toggle forgets state on window reload


sestinj:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I'm not able to find a related conversation on [GitHub discussions](https://github.com/continuedev/continue/discussions) that reports the same bug
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
Any config with a thinking model
```

### Description

Should probably store in localstorage or global storage this isn't already happening

### To reproduce

To reproduce:
1. toggle off thinking

<img width="552" height="158" alt="Image" src="https://github.com/user-attachments/assets/cff080c1-a45b-491b-a88c-fd2b4af1ef47" />

2. Reload the window
3. Thinking will be on again

<img width="545" height="156" alt="Image" src="https://github.com/user-attachments/assets/5bac9d31-c365-4d07-b330-8b6e6c5b3904" />

### Log output

```Shell

```

sestinj:
Related issues:

- #5709 - [Feature Request] Add Thinking Mode Toggle and Filter `<think>` (closed) - Original feature request for the thinking toggle functionality
- #9102 - Not able to disable thinking mode for the self hosted models which use OpenAI as a provider (open) - Related thinking mode configuration issues

ghost:
i dont have this toggle in extension i use in vscode insiders

ghost:
@RomneyDa

PR #9688, opened 2026-01-20
fix: retrieve reasoning toggle from local storage

uinstinct:
## Description

Get and store `hasReasoningEnabled` in local storage so that when thinking toggle is manually turned off, it does not turn on workspace reload.

closes https://github.com/continuedev/continue/issues/9675

## AI Code Review

## Checklist

https://github.com/user-attachments/assets/ea2168b2-a2b3-4cae-930c-ec5c5893ef70

https://github.com/user-attachments/assets/07da136b-b206-448a-9c3f-3ecf7cd3bc88

## Tests

---

**Continue Tasks:** ▶️ 2 queued · ▶️ 1 not started — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fcontinuedev%2Fcontinue%2Fpull%2F9688&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Persist and restore the reasoning toggle per model using the Redux UI slice so the user’s choice sticks across reloads and doesn’t auto-enable when switching to reasoning-capable models.

- **Bug Fixes**
  - Save hasReasoningEnabled per model via ui.reasoningSettings on toggle in InputToolbar.
  - Read per-model setting from Redux in ParallelListeners and combine with model support and reasoning=false configs.

<sup>Written for commit 649491abacdc14f4d7f6ceabe3f7803a1bce18a8. Summary will update on new commits.</sup>

### DIFF
--- a/gui/src/components/mainInput/InputToolbar.tsx
+++ b/gui/src/components/mainInput/InputToolbar.tsx
@@ -15,9 +15,9 @@ import { useAppDispatch, useAppSelector } from "../../redux/hooks";
 import { selectUseActiveFile } from "../../redux/selectors";
 import { selectSelectedChatModel } from "../../redux/slices/configSlice";
 import { setHasReasoningEnabled } from "../../redux/slices/sessionSlice";
+import { setReasoningSetting } from "../../redux/slices/uiSlice";
 import { exitEdit } from "../../redux/thunks/edit";
 import { getMetaKeyLabel, isMetaEquivalentKeyPressed } from "../../util";
-import { setLocalStorage } from "../../util/localStorage";
 import { ToolTip } from "../gui/Tooltip";
 import ModelSelect from "../modelSelection/ModelSelect";
 import { ModeSelect } from "../ModeSelect";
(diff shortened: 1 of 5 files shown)


uinstinct:
> @uinstinct could you use the existing redux persist gate to persist these values rather than a separate local storage key? So that there's not 2 sources of truth. Sorry missed that on first review.
> 
> Note persisting this in local storage on top of YAML setting could be confusing to some users because once the user toggles it, the `defaultCompletionOptions.reasoning` setting in the model config will be permanently out of sync with the local storage value _if_ it is set to true (The `false` setting will still be respected with current logic). This is the main reason we haven't put tool policies in YAML config. It's an issue whether persisted per model or not. Thoughts on this?

This makes sense - it will be out of sync when local storage is used. Sorry I too missed this in my first approach.

Implemented changes!

RomneyDa (approved):
It might still strike users as odd when it's out of sync with YAML. 
I think it's an improvement, will merge!

PR #10026, opened 2026-01-30
fix: enable `useLegacyCompletionsEndpoint` to use openai package streaming

uinstinct:
## Description

Add supports for `useLegacyCompletionsEndpoint` to use openai adapter streaming.

closes CON-5285

## AI Code Review

## Checklist

## Tests

---

**Continue Tasks:** ▶️ 1 queued · ✅ 1 no changes — [View all](https://hub.continue-stage.tools/inbox?pr=https%3A%2F%2Fgithub.com%2Fcontinuedev%2Fcontinue%2Fpull%2F10026&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Disables streaming for OpenAI o-series/GPT-5 when the legacy completions endpoint is used, aligning behavior with model constraints. Addresses CON-5285.

<sup>Written for commit be3931718e7d601362c8b88dfd60051f9cdffc2e. Summary will update on new commits.</sup>

### DIFF
--- a/docs/reference.mdx
+++ b/docs/reference.mdx
@@ -110,6 +110,7 @@ The `models` section defines the language models used in your configuration. Mod
   - `stop`: An array of stop tokens that will terminate the completion.
   - `reasoning`: Boolean to enable thinking/reasoning for Anthropic Claude 3.7+ and some Ollama models.
   - `reasoningBudgetTokens`: Budget tokens for thinking/reasoning in Anthropic Claude 3.7+ models.
+  - `stream`: Boolean to enable/disable streaming for the model.
 
 - `requestOptions`: HTTP request options specific to the model.
 


uinstinct:
> @uinstinct since `stream` is already in completion options, how does this impact functionality? I think we should avoid adding a duplicate configuration option.

My bad - I thought since completionOptions.stream was removed from config.yaml reference doc, support was dropped.

Added that back to the config.yaml reference in docs

RomneyDa (changes requested):
@uinstinct could you clarify how `useLegacyCompletionsEndpoint` and `stream` are related? As far as I understand `useLegacyCompletionsEndpoint` is for using a different URL endpoint /completions which was used prior to `chat/completions` coming out. stream is the issue here (you can use chat/completions with stream = false)

uinstinct:
openai models (which require org verifications) cannot use the responses endpoint - so when `useLegacyCompletionsEndpoint` is enabled, it should only use the completions endpoint.

RomneyDa:
@uinstinct the legacy completions endpoint is the ancient `/completions`, users should rarely use that for newer models. The fallback from responses should be `/chat/completions`. 

But I don't think we need to fallback to that either, I think they can use responses, they just can't _stream_ it. Kilo code fixed this by adding a stream toggle, I don't think they routed to a different endpoint. I think the issue is that `stream` isn't supported properly, or that we don't have error handling to detect the verification issue and switch streaming off.||

I think a stream toggle (which only shows when error detected) is a decent solution, as long as the `stream` parameter actually works

uinstinct:
> @uinstinct the legacy completions endpoint is the ancient `/completions`, users should rarely use that for newer models. The fallback from responses should be `/chat/completions`.
> 
> But I don't think we need to fallback to that either, I think they can use responses, they just can't _stream_ it. Kilo code fixed this by adding a stream toggle, I don't think they routed to a different endpoint. I think the issue is that `stream` isn't supported properly, or that we don't have error handling to detect the verification issue and switch streaming off.||
> 
> I think a stream toggle (which only shows when error detected) is a decent solution, as long as the `stream` parameter actually works

It seems we will have to use the `/chat/completions` endpoint because `/responses` does not support "reasoning" for these models. Also when using /chat/completions with reasoning enabled, the model does thinking but the thinking content is not returned in the api.

I investigated kilocode now and found that they are using `/chat/completions` for original OpenAI provider and `/responses` for OpenAI native provider. The same error is produced in kilocode's openai native provider.

Currently we support using /completions endpoint via config.yaml. Should we also add support for /chat/completions via config.yaml?

RomneyDa:
Opening discussion here
https://github.com/continuedev/continue/issues/11772

PR #10485, opened 2026-02-13
fix: harden system message tools and wire toolOverrides to system message path

shanevcantwell:
## Summary

- Strengthened system message tool instructions for models on the non-native tools path (used by local models via `capabilities: []` or `experimental.onlyUseSystemMessageTools`). The previous prose instructions were too vague — local models reverted to XML/JSON tool call formats from their training priors.
- Fixed `toolOverrides` being silently ignored on the system message path. `applyToolOverrides()` already existed and ran on the native tools path (in `BaseLLM.streamChat()`), but was never called when tools were injected via system message text. Config YAML `toolOverrides` for `disabled` and `description` now take effect on both paths.

## Changes

1. **`core/tools/systemMessageTools/toolCodeblocks/index.ts`** — Replaced prose `systemMessagePrefix`/`systemMessageSuffix` with explicit numbered rules that prohibit XML tags, JSON format, and tool invention
2. **`core/index.d.ts`** — Added `toolOverrides` to `ModelDescription` interface
3. **`core/config/load.ts`** — Serialized `toolOverrides` in `llmToSerializedModelDescription()`
4. **`gui/src/redux/thunks/streamNormalInput.ts`** — Imported `applyToolOverrides` and applied model-level overrides to `activeTools` before both the native and system message paths

## Test plan

- [x] `cd core && npm run vitest` — 1654 passed, 0 failures
- [x] Built vsix, tested with local models (nemotron-3-nano-30b, gpt-oss-20b, mirothinker-v1.5-30b) via LM Studio
- [x] Verified `view_diff` with `disabled: true` is absent from system message
- [x] Verified hardened numbered rules appear in system message prefix/suffix
- [x] Verified local models follow codeblock format through multi-step tool call chains
- [ ] Verify native tools path unaffected (cloud models)

---
## Summary by cubic
Improves reliability of system-message tool calls for local models and applies model-level toolOverrides on that path. Local models now follow the codeblock format, and disabled/description overrides from config are respected.

- **Bug Fixes**
  - Hardened system message with clear, numbered rules: use only tool code blocks; no XML/JSON; one tool per response; code block last; no invented tools.
  - Applied applyToolOverrides before both native and system-message paths and added toolOverrides to ModelDescription serialization, so per-model disabled/description overrides take effect.

<sup>Written for commit 1a59f1ab56f78c041e00b7603a3f679f7861e678. Summary will update on new commits.</sup>

### DIFF
--- a/core/config/load.ts
+++ b/core/config/load.ts
@@ -651,6 +651,7 @@ function llmToSerializedModelDescription(llm: ILLM): ModelDescription {
     envSecretLocations: llm.envSecretLocations,
     sourceFile: llm.sourceFile,
     isFromAutoDetect: llm.isFromAutoDetect,
+    toolOverrides: llm.toolOverrides,
   };
 }
 
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -1254,6 +1254,9 @@ export interface ModelDescription {
 
   sourceFile?: string;
   isFromAutoDetect?: boolean;
+
+  /** Tool overrides for this model */
+  toolOverrides?: ToolOverride[];
 }
 
 export interface JSONEmbedOptions {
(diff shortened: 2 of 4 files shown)


RomneyDa (approved):
@shanevcantwell thanks for the followup!

PR #10844, opened 2026-02-26
fix: skip remote URIs when resolving MCP server cwd

### LINKED ISSUE #10842: MCP servers fail on SSH/remote: cmd.exe spawned locally instead of on remote
## Bug Description

When VS Code is connected to a remote workspace via SSH (and likely Dev Containers), MCP servers configured with stdio transport fail with:

```
Failed to connect to "New MCP server" Error: command "cmd.exe" not found. To use this MCP server, install the cmd.exe CLI.
```

## Root Cause

Two related issues in `core/context/mcp/MCPConnection.ts`:

### 1. `resolveCommandForPlatform()` only handles WSL remotes (line 429)

```typescript
const isWindowsHostWithWslRemote =
  process.platform === "win32" && ideInfo?.remoteName === "wsl";
```

This bypasses `cmd.exe` wrapping for WSL but not for SSH or other remote types. On SSH remotes, batch commands like `npx` still get wrapped as `cmd.exe /c npx`, which fails because `cmd.exe` doesn't exist on the Linux remote.

### 2. Deeper: `StdioClientTransport` spawns locally, not on remote

Even if the `cmd.exe` wrapping is fixed, the underlying `StdioClientTransport` uses `cross-spawn` (Node's `child_process.spawn()`) which executes in the **Local Extension Host** on Windows — not on the remote. The MCP server process would be spawned on the Windows host regardless of the command used.

This is the same class of bug fixed for terminal commands in PR #10786, where `childProcess.spawn()` was replaced with VS Code's terminal API for remote execution.

### 3. PATH resolution also only handles WSL (line 569)

```typescript
const isWindowsHostWithWslRemote =
  process.platform === "win32" && ideInfo?.remoteName === "wsl";
if (process.platform !== "win32" || isWindowsHostWithWslRemote) {
```

SSH remotes also need Linux shell PATH resolution, not the Windows PATH.

## Steps to Reproduce

1. Open VS Code connected to an SSH remote (Linux host)
2. Configure an MCP server in Continue with a stdio command (e.g., `npx @playwright/mcp@latest`)
3. Observe the "cmd.exe not found" error

## Expected Behavior

MCP servers should spawn on the remote host when VS Code is connected to a remote workspace.

## Context

- `extensionKind: ["ui", "workspace"]` causes the extension to run in the Local Extension Host on Windows
- `process.platform` returns `"win32"` even when connected to a Linux remote
- Compare with `runTerminalCommand.ts` line 122 which correctly checks `!["", "local"].includes(ideInfo.remoteName)` for any remote type
- Related: PR #10786 (terminal command remote fix), issue #10007

## Environment

- Windows 11 + VS Code SSH Remote to Linux host
- Continue extension running in Local Extension Host

shanevcantwell:
## Summary

- MCP stdio servers fail with `spawn cmd.exe ENOENT` when VS Code is connected to a remote workspace (SSH, Dev Container, etc.) from a Windows host
- Root cause: `resolveWorkspaceCwd()` passes `vscode-remote://` URIs directly as `cwd` to `child_process.spawn()`, which can't resolve them as local filesystem paths
- Fix: detect non-`file://` URI schemes and return `undefined` so the spawn uses its default cwd, allowing MCP servers to start on the local host

## Test plan

- [x] Added unit test for `vscode-remote://` URI returning `undefined`
- [x] Manual: Configure stdio MCP server on Windows + SSH remote, verify it connects
- [x] Manual: Verify MCP servers still work on local Windows workspace (no regression)
- [x] Verify pre-existing test failure (`should resolve relative cwd using IDE workspace`) is unchanged — it fails on main due to `fileURLToPath` rejecting non-Windows paths

Fixes #10842

---

**Continue Tasks:** ❌ 7 failed — [View all](https://hub.continue.dev/inbox/pr/continuedev/continue/10844?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Fix Windows startup failures for stdio MCP servers when VS Code is connected to a remote workspace by detecting remote URIs and using the user's home directory as the cwd. Ensures spawn has a valid local cwd so MCP servers start. Fixes #10842.

- **Bug Fixes**
  - Detect non-file URIs (e.g., vscode-remote://) in resolveCwd and return os.homedir() for spawn’s cwd.
  - Added a unit test verifying remote URIs resolve to homedir().

<sup>Written for commit a3d6f96a45d25d52169a5833b5c650a58eb05660. Summary will update on new commits.</sup>

### DIFF
--- a/core/context/mcp/MCPConnection.ts
+++ b/core/context/mcp/MCPConnection.ts
@@ -1,5 +1,6 @@
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import { Transport } from "@modelcontextprotocol/sdk/shared/transport.js";
+import { homedir } from "os";
 import { fileURLToPath } from "url";
 
 import {
@@ -475,6 +476,13 @@ Org-level secrets can only be used for MCP by Background Agents (https://docs.co
       if (resolved) {
         if (resolved.startsWith("file://")) {
           return fileURLToPath(resolved);
+        }
+        // Remote URIs (e.g. vscode-remote://ssh-remote+host/path) cannot be
+        // used as a local cwd for child_process.spawn(). When the extension
+        // runs in the Local Extension Host on Windows while connected to a
+        // remote workspace, fall back to the user's home directory.
+        if (resolved.includes("://")) {
+          return homedir();
         }
         return resolved;
       }
--- a/core/context/mcp/MCPConnection.vitest.ts
+++ b/core/context/mcp/MCPConnection.vitest.ts
@@ -175,6 +175,17 @@ describe("MCPConnection", () => {
       );
       expect(mockResolve).toHaveBeenCalledWith("src", ide);
     });
+
+    it("should fall back to homedir for remote URIs that cannot be used as local cwd", async () => {
+      const ide = {} as any;
+      vi.spyOn(ideUtils, "resolveRelativePathInDir").mockResolvedValue(
+        "vscode-remote://ssh-remote+192.168.137.2/home/user/project",
+      );
+      const conn = new MCPConnection(baseOptions, { ide });
+
+      const { homedir } = require("os");
+      await expect((conn as any).resolveCwd("src")).resolves.toBe(homedir());
+    });
   });
 
   describe("connectClient", () => {


core/context/mcp/MCPConnection.ts
```diff
@@ -476,6 +477,13 @@ Org-level secrets can only be used for MCP by Background Agents (https://docs.co
         if (resolved.startsWith("file://")) {
           return fileURLToPath(resolved);
         }
+        // Remote URIs (e.g. vscode-remote://ssh-remote+host/path) cannot be
+        // used as a local cwd for child_process.spawn(). When the extension
+        // runs in the Local Extension Host on Windows while connected to a
+        // remote workspace, fall back to the user's home directory.
+        if (resolved.includes("://")) {
```
RomneyDa, line 484: @shanevcantwell I wonder if this might cause issues for `wsl` which should work locally?

shanevcantwell: It looks good to me. Verified in local, Remote: wsl and Remote: ssh.

RomneyDa: Just to clarify, is that because wsl kind of works like a remote directory in this case? Just want to make sure I understand the implications before merging. 

Seems like `wsl://` would match and then mcps which ran at the wsl location are now "working" (as you saw) but unnecessarily/incorrectly running in the user's home dir

RomneyDa (commented):
see comment!

core/context/mcp/MCPConnection.ts
shanevcantwell: Good question, and to clarify since the solution landed on is not necessarily intuitive. It was a compromise that should have been more explicit, since it turned out to be essentially impossible to reliably run MCP *on the remote* when running in Remote mode.

MCP servers spawned via stdio transport always run on the **Windows host**, not in WSL. The extension has `extensionKind: ["ui", "workspace"]`, so it runs in the Local Extension Host on Windows. `child_process.spawn()` is a local Node.js call; it can't reach into WSL.

So when the workspace is `wsl://`, the MCP server was *already* running on Windows with an invalid cwd (`vscode-remote://wsl+Ubuntu/home/...`), which causes `spawn ENOENT`. The server wasn't running "at the wsl location" before. It was failing.

`homedir()` here returns the Windows home directory (`C:\Users\<name>`), which is a valid local path for the Windows-hosted spawn. It's not changing where the server runs or moving it away from WSL. The server was never in WSL to begin with.

The same logic applies to SSH and Dev Containers: the extension is local, spawn is local, and `vscode-remote://` URIs are never valid cwd paths for a local process.

RomneyDa: @shanevcantwell this all seems right, really appreciate the explanations!

PR #11067, opened 2026-03-05
fix: local config — toolOverrides, config name, agent refresh

### LINKED ISSUE #11060: toolOverrides not applied in system message tools path
## Bug

`toolOverrides` (added in #9314) only works for models using the native tools API. For models using the system message tools path (e.g., local models with `capabilities: []`), overrides are silently dropped at two levels:

### 1. Serialization gap
`llmToSerializedModelDescription()` in `core/config/load.ts` does not copy `toolOverrides` to `ModelDescription`. The field never reaches the GUI's Redux store.

### 2. System message tools path not wired
`gui/src/redux/thunks/streamNormalInput.ts` passes `selectActiveTools()` directly to `addSystemMessageToolsToSystemMessage()` without applying `applyToolOverrides()`. The native tools path applies overrides core-side in `BaseLLM.streamChat()`, but the system message tools path bypasses that entirely.

### 3. description override ignored for built-in tools
Built-in tools use `systemMessageDescription.prefix` (not `function.description`) in the system message builder. `applyToolOverrides` puts `override.description` into `function.description` only, so description overrides have no effect on what the model sees in the system prompt.

## Repro

1. Set `capabilities: []` on a model in `config.yaml` to force system message tools
2. Add `toolOverrides` with `disabled: true` for any tool (e.g., `view_diff`)
3. Start an agent chat — disabled tool still appears in system prompt

## Expected

`toolOverrides` should work identically regardless of whether the model uses native tools or system message tools.

## Environment

Windows 11, LM Studio + qwen3.5-35b, Continue main branch

### LINKED ISSUE #11061: Local Config display name ignores config.yaml name field
## Bug

The default local profile always displays "Local Config" in the configs dropdown, ignoring the `name` field in `~/.continue/config.yaml`.

### Root cause

`LocalProfileLoader` constructor hardcodes `"Local Config"` for the default profile (no `overrideAssistantFile`). Project-level agent configs correctly read the name from file content via `parseConfigYaml()`, but the default profile never reads its config file.

## Repro

1. Set `name: My Custom Config` in `~/.continue/config.yaml`
2. Open Continue sidebar
3. Configs dropdown shows "Local Config" instead of "My Custom Config"

## Expected

The dropdown should display the `name` field from the user's config file, falling back to "Local Config" only when the file is missing or unparseable.

### LINKED ISSUE #11062: New agent configs don't appear until file watcher fires
## Bug

Creating a new agent file via the (+) button doesn't immediately show the new config in the configs list. It only appears after the file watcher fires (~5-10 seconds later).

### Root cause

The `config/newAssistantFile` handler in `core/core.ts` calls `this.configHandler.reloadConfig()`, which only reloads the current profile. New agent files require `refreshAll()` to trigger `cascadeInit()` → `getLocalProfiles()` → scan `.continue/agents/` and discover the new file.

The file watcher at `core/core.ts:922` already calls `refreshAll()` for the same event — the handler is inconsistent.

## Repro

1. Click (+) to create a new agent config
2. Config does NOT appear immediately
3. Wait ~5-10 seconds — config appears when file watcher fires

## Expected

New config should appear within 1 second of creation, matching the file watcher behavior.

shanevcantwell:
## Summary

Three fixes discovered during local model testing (LM Studio + `config.yaml` on Windows/WSL). All affect the non-cloud codepath — local models, custom configs, system message tools — that hasn't been tested as a cohesive surface.

- **Fix #11060**: `toolOverrides` (#9314) not applied in system message tools path — overrides were serialized incompletely and never wired into `streamNormalInput.ts` for models using system message tools instead of native tool calling
- **Fix #11061**: "Local Config" display name ignores `config.yaml` `name` field — `LocalProfileLoader` hardcoded the title for the default profile instead of reading it from the config file
- **Fix #11062**: New agent files don't appear until file watcher fires — `config/newAssistantFile` handler called `reloadConfig` (current profile only) instead of `refreshAll` (rebuilds profile list)

## Context

PR #9314 added `toolOverrides` to let users customize tool descriptions and disable tools per-model. It worked for models using the native tools API, but the system message tools path (used by local models with `capabilities: []`) was never wired up. The serialization gap meant `toolOverrides` never reached the GUI's Redux store, and `streamNormalInput.ts` passed raw `selectActiveTools()` output directly to the system message builder without applying overrides.

The other two bugs (config name, agent refresh) were found during the same testing session and affect the same local config surface.

## Changes

| File | Change |
|------|--------|
| `core/index.d.ts` | Add `toolOverrides?: ToolOverride[]` to `ModelDescription` |
| `core/config/load.ts` | Serialize `toolOverrides` in `llmToSerializedModelDescription()` |
| `gui/src/redux/thunks/streamNormalInput.ts` | Import and apply `applyToolOverrides` for both native and system message tools paths |
| `core/tools/applyToolOverrides.ts` | Fall through `description` override to `systemMessageDescription.prefix` for built-in tools |
| `core/config/profile/LocalProfileLoader.ts` | Read `name` from primary config file in constructor |
| `core/core.ts` | `reloadConfig` → `refreshAll` in `config/newAssistantFile` handler |

## Test plan

- [x] 30 tests across 4 new test files, all passing
- [x] VSIX built and manually verified on Windows with LM Studio + qwen3.5-35b
- [ ] WSL remote verification
- [ ] CI

---

**Continue Tasks:** 🔄 7 running — [View all](https://hub.continue.dev/inbox/pr/continuedev/continue/11067?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Fixes three local-config issues: per-model toolOverrides now apply to system-message tools, the default profile uses the config.yaml name, and new agent files appear immediately. Improves the local model and custom config experience.

- **Bug Fixes**
  - Tools: serialize toolOverrides in ModelDescription; apply them in both native and system-message paths; description override updates systemMessageDescription.prefix for built-in tools.
  - Profile name: LocalProfileLoader reads the name from the primary config file and guards undefined; falls back to “Local Config” if missing/invalid.
  - Agent refresh: config/newAssistantFile now calls refreshAll to rebuild the profile list so new files show up immediately.
  - Tests: added coverage for overrides, profile naming, and agent refresh; fixed ModelRole import in tests to resolve CI TS2459.

<sup>Written for commit a221949aedf28f0edb0159e28cf8af2fd1436764. Summary will update on new commits.</sup>

### DIFF
--- a/core/config/newAssistantRefresh.vitest.ts
+++ b/core/config/newAssistantRefresh.vitest.ts
@@ -73,9 +73,7 @@ describe("isContinueAgentConfigFile", () => {
 
   it("should not match non-YAML files in agents directory", () => {
     expect(
-      isContinueAgentConfigFile(
-        "file:///workspace/.continue/agents/readme.md",
-      ),
+      isContinueAgentConfigFile("file:///workspace/.continue/agents/readme.md"),
     ).toBe(false);
     expect(
       isContinueAgentConfigFile(
@@ -96,8 +94,8 @@ describe("isContinueAgentConfigFile", () => {
   });
 
   it("should not match non-.continue YAML files", () => {
-    expect(
-      isContinueAgentConfigFile("file:///workspace/src/config.yaml"),
-    ).toBe(false);
+    expect(isContinueAgentConfigFile("file:///workspace/src/config.yaml")).toBe(
+      false,
+    );
   });
 });
--- a/core/config/profile/LocalProfileLoader.ts
+++ b/core/config/profile/LocalProfileLoader.ts
@@ -58,7 +58,7 @@ export default class LocalProfileLoader implements IProfileLoader {
         if (fs.existsSync(configPath)) {
           const content = fs.readFileSync(configPath, "utf-8");
           const parsed = parseConfigYaml(content);
-          this.description.title = parsed.name;
+          this.description.title = parsed.name ?? "Local Config";
         }
       } catch (e) {
         // Keep "Local Config" as fallback
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -96,7 +96,8 @@ type RequiredLLMOptions =
   | "completionOptions";
 
 export interface ILLM
-  extends Omit<LLMOptions, RequiredLLMOptions>,
+  extends
+    Omit<LLMOptions, RequiredLLMOptions>,
     Required<Pick<LLMOptions, RequiredLLMOptions>> {
   get providerName(): string;
   get underlyingProviderName(): string;
--- a/core/tools/applyToolOverrides.vitest.ts
+++ b/core/tools/applyToolOverrides.vitest.ts
@@ -239,9 +239,7 @@ describe("applyToolOverrides", () => {
     ];
     const result = applyToolOverrides(tools, overrides);
 
-    expect(result.tools[0].function.description).toBe(
-      "Custom MCP description",
-    );
+    expect(result.tools[0].function.description).toBe("Custom MCP description");
     // Should remain undefined — don't promote a dynamic tool to predefined
     expect(result.tools[0].systemMessageDescription).toBeUndefined();
   });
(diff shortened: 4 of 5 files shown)


shanevcantwell:
Closing this — 2 of the 3 fixes here have landed independently:

- **toolOverrides not applied in system message path** → fixed by #10485
- **New agent files don't appear (reloadConfig vs refreshAll)** → fixed by #11103

The remaining fix (config name from `config.yaml` not displayed) will be submitted as a standalone PR.

PR #11078, opened 2026-03-05
Fix local profile title to honor config.yaml name

### LINKED ISSUE #11061: Local Config display name ignores config.yaml name field

siewcapital:
## Summary
- load `~/.continue/config.yaml` for the default local profile title
- use the parsed `name` field when present, with fallback to `Local Config`
- preserve existing behavior for workspace override assistant files

## Tests
- added `core/config/profile/LocalProfileLoader.test.ts` covering:
  - default profile picks up `name` from local config
  - fallback when config file is missing
  - fallback when config has no `name`
  - override assistant content takes precedence without reading disk

Closes #11061

---
## Summary by cubic
Use the `name` from the local `config.yaml` as the default Local profile title with clear override behavior and safe fallbacks. Improve IntelliJ URI parsing for Windows paths and unencoded file URIs.

- **Bug Fixes**
  - Local profile: parse pre-read YAML via `@continuedev/config-yaml`; prefer override content; if empty, use the override file name; else read the primary config and use its name or fall back to “Local Config.” Pass pre-read content in `packageIdentifier` to avoid fs reads on remote URIs (Windows → WSL). Tests moved to the Vitest suite using the repo-standard module mock.
  - IntelliJ: safely parse Windows file URIs and unencoded paths (spaces) with a fallback to `File.toURI()`; add a regression test for paths with spaces.

<sup>Written for commit 810eaebed564f5ce5cc9c5887dd5097bf94701f7. Summary will update on new commits.</sup>

### DIFF
--- a/core/config/profile/LocalProfileLoader.vitest.ts
+++ b/core/config/profile/LocalProfileLoader.vitest.ts
@@ -1,0 +1,128 @@
+import { describe, expect, it, vi } from "vitest";
+import fs from "fs";
+
+import { ControlPlaneClient } from "../../control-plane/client.js";
+import { LLMLogger } from "../../llm/logger.js";
+import { testIde } from "../../test/fixtures.js";
+import LocalProfileLoader from "./LocalProfileLoader.js";
+
+// Mock doLoadConfig to capture the arguments it receives
+const mockDoLoadConfig = vi.fn().mockResolvedValue({
+  config: undefined,
+  errors: [],
+  configLoadInterrupted: false,
+});
+
+vi.mock("./doLoadConfig.js", () => ({
+  default: (...args: any[]) => mockDoLoadConfig(...args),
+}));
+
+const mockParseConfigYaml = vi.fn();
+vi.mock("@continuedev/config-yaml", async () => {
+  const actual = await vi.importActual<object>("@continuedev/config-yaml");
+  return {
+    ...actual,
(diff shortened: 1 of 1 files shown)


siewcapital:
Updated the branch to keep it focused on the local profile title behavior. The config loader now preserves the existing `content` pass-through from `main`, avoids falling back to the primary config when an override file is intentionally empty, adds a regression test for that case, and reverts the unrelated IntelliJ URI changes back to `main`.

siewcapital:
Pushed one more follow-up for the failing checks. The previous test fixture was relying on `parseConfigYaml` accepting a minimal YAML snippet, which is not what the parser validates in CI. I changed the test to mock `parseConfigYaml` directly so it verifies the profile title behavior without depending on unrelated schema parsing.

Co-Messi:

PR #11091, opened 2026-03-05
Fix system tool parser dropping text and blocking follow-up tool calls

### LINKED ISSUE #11074: System message tools parser assumes tool calls are positionally terminal
## Summary

The system message tools parser (`interceptSystemToolCalls.ts`) assumes tool calls must appear at the end of a response — one per response, nothing after. This is an architectural constraint that doesn't exist in the tool calling paradigm itself.

## Behavior

1. After a tool call completes, any subsequent content is silently dropped (lines 100-103)
2. A second tool call in the same response is never parsed — `parseState?.done` breaks the loop (line 49)
3. The system message suffix instructs models: "You can only call ONE tool at a time. You MUST stop generating after the closing codeblock"

Meanwhile the agent system message says "call multiple tools simultaneously."

## Why this matters

Native tool calling APIs (OpenAI, Anthropic) allow models to interleave text and tool calls freely, and to make multiple calls in a single response. The system message tools path restricts models to a subset of that — one tool call, must be terminal — creating a capability gap that makes local models on this path appear less capable than they are.

The suffix instruction is a consequence of the parser limitation, not the other way around. The parser was built with the positional assumption, and the instruction exists to keep models inside those bounds.

## Where

- `core/tools/systemMessageTools/interceptSystemToolCalls.ts`
  - Line 49: `parseState?.done` breaks on second tool call
  - Lines 100-103: content after completed tool call dropped
- `core/tools/systemMessageTools/toolCodeblocks/index.ts`
  - `systemMessageSuffix` (~line 71): "ONE tool at a time"

## Context

This is a design limitation, not a simple bug. Fixing it means the parser needs to handle multiple tool calls per response, text between tool calls, and tool calls at any position. That's a meaningful rearchitecture of the stream interception logic.

## Related

- #11070 — parser false positives (separate issue, but compounds this)
- #11072 — TOOL_NAME keyword echo
- LM Studio lmstudio-community/app#1592 — similar positional assumptions in their parser

siewcapital:
## Summary
- keep parsing after a completed system tool call instead of halting the stream
- preserve assistant text emitted after a tool codeblock
- allow a second (or later) tool codeblock in the same assistant response stream
- update tool-instruction suffix to remove the single-tool/terminal-only restriction
- extend parser tests to cover post-tool text preservation and multi-tool streaming

## Why
Issue #11074 highlights that the current interceptor enforces a positional assumption (single terminal tool call). That causes a capability gap for system-message tool parsing compared with native tool calling behavior.

## Testing
- Added/updated vitest cases in:
  - `core/tools/systemMessageTools/toolCodeblocks/interceptSystemToolCalls.vitest.ts`
- Could not execute vitest in this environment because project dev dependencies are not installed (`vitest: command not found`).

Closes #11074

---
## Summary by cubic
Fixes the system tool parser so it preserves assistant text and supports multiple tool codeblocks in a single response instead of stopping after the first call. Closes #11074 and aligns system-message tool parsing with native tool calling.

- **Bug Fixes**
  - Keep parsing after a completed tool call; no premature halt.
  - Preserve assistant text after tool codeblocks (including newline).
  - Support multiple tool codeblocks in one assistant message.
  - Update system message suffix to allow multi-tool calls; add tests for post-tool text and multi-tool streams.

<sup>Written for commit 653adb1772b2134ee099d08a061ff446cbe48903. Summary will update on new commits.</sup>

### DIFF
--- a/core/tools/systemMessageTools/interceptSystemToolCalls.ts
+++ b/core/tools/systemMessageTools/interceptSystemToolCalls.ts
@@ -46,7 +46,7 @@ export async function* interceptSystemToolCalls(
       return result.value;
     } else {
       for await (const message of result.value) {
-        if (abortController.signal.aborted || parseState?.done) {
+        if (abortController.signal.aborted) {
           break;
         }
         // Skip non-assistant messages or messages with native tool calls
@@ -69,6 +69,10 @@ export async function* interceptSystemToolCalls(
 
         for (const chunk of chunks) {
           buffer += chunk;
+          if (parseState?.done) {
+            parseState = undefined;
+          }
+
           if (!parseState) {
             const { isInPartialStart, isInToolCall, modifiedBuffer } =
               detectToolCallStart(buffer, systemToolFramework);
@@ -82,7 +86,7 @@ export async function* interceptSystemToolCalls(
             }
           }
 
-          if (parseState && !parseState.done) {
+          if (parseState) {
             const delta = systemToolFramework.handleToolCallBuffer(
               buffer,
               parseState,
@@ -97,11 +101,6 @@ export async function* interceptSystemToolCalls(
               ];
             }
           } else {
-            // Prevent content after tool calls for now
-            if (parseState) {
-              continue;
-            }
-
             // Yield normal assistant message
             yield [
               {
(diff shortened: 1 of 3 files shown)


Co-Messi:

PR #11094, opened 2026-03-05
Fix false-positive system tool parsing for quoted TOOL_NAME examples

### LINKED ISSUE #11070: System message tools parser intercepts quoted tool syntax as real tool calls
## Bug

The system message tools parser in `interceptSystemToolCalls.ts` treats any text matching tool call patterns as a real invocation. When a model tries to describe, quote, or explain its own tool syntax, the parser intercepts the quoted text and executes it as an actual tool call.

### Observed behavior

1. Ask a model on the system message tools path to explain how its tools work
2. Model writes something like "The read_file tool is called with TOOL_NAME: read_file..."
3. Parser detects `tool_name:` at a chunk boundary after `splitAtCodeblocksAndNewLines()` splits on the newline
4. Parser treats the quote as a real tool call → "Continue tried to read path/to/the_file.txt"

### Root cause

This is fundamentally the same class of bug as lmstudio-ai/lmstudio-bug-tracker#1592 (parser scans inside `<think>` blocks for tool call patterns). Both are context-free pattern matchers on output streams — neither distinguishes invocation context from conversational text.

Two specific mechanisms in Continue's parser:

**1. Aggressive bare `tool_name:` start pattern**
`acceptedToolCallStarts` in `toolCodeblocks/index.ts` lines 9-12:
```javascript
["tool_name:", "```tool\nTOOL_NAME:"],
```
This is case-insensitive (`detectToolCallStart.ts` line 14). Any chunk starting with `tool_name:` triggers parsing — even mid-response, even in quoted/explanatory context. The comment says "Poor models are really bad at following instructions, alternate starts allowed."

**2. Chunk-boundary detection loses positional context**
`splitAtCodeblocksAndNewLines()` splits content at every backtick triple and newline. Each resulting chunk is checked independently via `detectToolCallStart()`. A newline-separated line starting with `tool_name:` triggers detection regardless of where in the response it appears.

### Why this can't be fully fixed at the text level

Without Harmony-style structural tokens that distinguish tool invocations from content at the vocabulary level, any text-level parser faces a precision/recall tradeoff: tighten the patterns and weak models fail to trigger real tool calls; loosen them and quoted syntax triggers false positives. This is the same tradeoff LM Studio faces with its Harmony parser.

### Files

- `core/tools/systemMessageTools/toolCodeblocks/index.ts:9-12` — `acceptedToolCallStarts`
- `core/tools/systemMessageTools/detectToolCallStart.ts:14` — case-insensitive matching
- `core/tools/systemMessageTools/interceptSystemToolCalls.ts:70-74` — chunk-level detection
- `core/tools/systemMessageTools/systemToolUtils.ts:7-9` — `splitAtCodeblocksAndNewLines`

### Environment

Observed with qwen3.5-35b via LM Studio using system message tools path (`capabilities: []`).

siewcapital:
## Summary
- prevent fallback `TOOL_NAME:` parsing once the assistant has already emitted non-whitespace conversational text
- keep canonical ```tool block parsing unchanged, even when a response includes explanatory text first
- make canonical start detection explicit in `detectToolCallStart` instead of relying on tuple ordering
- add regression coverage so quoted syntax examples stay as plain assistant text

## Why
When a model explains tool syntax (for example, `TOOL_NAME: read_file` in a normal paragraph), the parser could misclassify that quoted text as a real tool invocation. This patch narrows the fallback path to reduce those false positives while preserving standard fenced tool-call behavior.

## Testing
- Added/updated unit tests in `core/tools/systemMessageTools/toolCodeblocks/interceptSystemToolCalls.vitest.ts`
- Local run in this environment is blocked because `vitest` is not installed (`ERR_MODULE_NOT_FOUND`)

Closes #11070

---

**Continue Tasks:** 🔄 7 running — [View all](https://hub.continue.dev/inbox/pr/continuedev/continue/11094?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Prevents false-positive tool calls when the assistant shows quoted TOOL_NAME examples in normal text by limiting fallback parsing to the very start of a response. Canonical ```tool code blocks still parse normally.

- **Bug Fixes**
  - Allow non-codeblock TOOL_NAME starts only at response start; disable once any non-whitespace assistant text appears.
  - Continue to parse canonical ```tool blocks even after conversational text.
  - Add explicit canonical-start detection with an allowNonCodeblockStarts option and regression tests to keep quoted examples as plain text.

<sup>Written for commit 059a085ce9af91ec1ad93a60b70c7c162f9ac95a. Summary will update on new commits.</sup>

### DIFF
--- a/core/tools/systemMessageTools/detectToolCallStart.ts
+++ b/core/tools/systemMessageTools/detectToolCallStart.ts
@@ -3,8 +3,22 @@ import { SystemMessageToolsFramework } from "./types";
 export function detectToolCallStart(
   buffer: string,
   toolCallFramework: SystemMessageToolsFramework,
+  options?: {
+    /**
+     * When false, only the canonical ```tool start is accepted.
+     */
+    allowNonCodeblockStarts?: boolean;
+  },
 ) {
-  const starts = toolCallFramework.acceptedToolCallStarts;
+  const allowNonCodeblockStarts = options?.allowNonCodeblockStarts ?? true;
+  const canonicalStart =
+    toolCallFramework.acceptedToolCallStarts.find(
+      ([start, replacement]) => start === replacement,
+    ) ?? toolCallFramework.acceptedToolCallStarts[0];
+
+  const starts = allowNonCodeblockStarts
+    ? toolCallFramework.acceptedToolCallStarts
+    : [canonicalStart];
   let modifiedBuffer = buffer;
   let isInToolCall = false;
   let isInPartialStart = false;
--- a/core/tools/systemMessageTools/interceptSystemToolCalls.ts
+++ b/core/tools/systemMessageTools/interceptSystemToolCalls.ts
@@ -28,6 +28,7 @@ export async function* interceptSystemToolCalls(
 ): AsyncGenerator<ChatMessage[], PromptLog | undefined> {
   let buffer = "";
   let parseState: ToolCallParseState | undefined;
+  let sawNonWhitespaceAssistantText = false;
 
   while (true) {
     const result = await messageGenerator.next();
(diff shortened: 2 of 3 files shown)


Co-Messi:

Co-Messi:

PR #11105, opened 2026-03-06
fix: avoid intercepting quoted TOOL_NAME syntax in system tool parser

### LINKED ISSUE #11070: System message tools parser intercepts quoted tool syntax as real tool calls

MumuTW:
## Summary
- gate non-standard `TOOL_NAME:` start detection so it is only allowed at the beginning of assistant output
- keep standard ```tool fenced detection unchanged
- add regression coverage to ensure explanatory/quoted tool syntax is not intercepted

## Tests
- `npm run vitest -- tools/systemMessageTools/toolCodeblocks/detectToolCallStart.vitest.ts tools/systemMessageTools/toolCodeblocks/interceptSystemToolCalls.vitest.ts`

Closes #11070.

---

**Continue Tasks:** 🔄 7 running — [View all](https://hub.continue.dev/inbox/pr/continuedev/continue/11105?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

---
## Summary by cubic
Prevented the system tools parser from intercepting quoted “TOOL_NAME:” syntax by only allowing loose tool starts at the very beginning of assistant output. Standard ```tool fenced detection remains unchanged. Addresses Linear #11070.

- **Bug Fixes**
  - Add allowAlternateStarts option to detectToolCallStart and disable it after any non-whitespace assistant text.
  - Track non-whitespace assistant text to confine loose "TOOL_NAME:" detection to the output start.
  - Add tests to skip non-standard starts when alternates are disabled and ensure quoted/explanatory tool syntax is emitted as plain text.

<sup>Written for commit 71083c6f12f460f24e5ab1020d03cc1b552f3da0. Summary will update on new commits.</sup>

### DIFF
--- a/core/tools/systemMessageTools/toolCodeblocks/interceptSystemToolCalls.vitest.ts
+++ b/core/tools/systemMessageTools/toolCodeblocks/interceptSystemToolCalls.vitest.ts
@@ -240,8 +240,12 @@ describe("interceptSystemToolCalls", () => {
         break;
       }
 
-      const chunkText = ((result.value as AssistantChatMessage[])[0]
-        .content as { type: "text"; text: string }[])[0].text;
+      const chunkText = (
+        (result.value as AssistantChatMessage[])[0].content as {
+          type: "text";
+          text: string;
+        }[]
+      )[0].text;
       outputChunks.push(chunkText);
       expect((result.value as AssistantChatMessage[])[0].toolCalls).toBeFalsy();
     }


MumuTW:
Pushed a formatting follow-up to address the  failure.\n\n- Ran repository-configured Prettier on all PR-changed files; one test file required formatting updates.\n- Committed the fix in .\n- Re-ran targeted tests:  ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND  No package.json (or package.yaml, or package.json5) was found in "/home/opc/.paperclip/instances/default/workspaces/7948d02f-b91e-4189-b9eb-32bf0b5923d2". (pass).

MumuTW:
Pushed a formatting follow-up to address the prettier-check failure.

- Ran repository-configured Prettier on all PR-changed files; one test file required formatting updates.
- Committed the fix in 71083c6.
- Re-ran targeted tests: `pnpm vitest tools/systemMessageTools/toolCodeblocks/detectToolCallStart.vitest.ts tools/systemMessageTools/toolCodeblocks/interceptSystemToolCalls.vitest.ts` (pass).

MumuTW:
@RomneyDa Friendly ping — the detailed explanation above addresses the concern about interleaving. To summarize:

- The fix **only** restricts the loose `TOOL_NAME:` alternate start (index > 0) after non-whitespace assistant text has been emitted.
- The standard code-fenced format (`\`\`\`tool`) at index 0 is **always** allowed, preserving the interleaving use case for models that use it.
- All CI checks are green and all 47 tests pass.

Happy to adjust the approach if you'd prefer a different direction!

RomneyDa:
@MumuTW thanks for the explanation. I think the downsides of preventing model from interleaving with _only_ TOOL_NAME and potentially preventing it from calling tools when it _does_ want to probably outweigh the upside of allowing models to explain the syntax, since that's an edge case and most users won't worry about the syntax. I don't have evals on this but do you think people ask about the syntax often?

RomneyDa:
See note above, I think the risk of reducing functionality here outweighs the benefit of the model being able to explain its underlying syntax.

PR #11127, opened 2026-03-06
fix(openai-adapters): keep reasoning deltas out of assistant content

### LINKED ISSUE #11069: Vercel stream converter maps reasoning-delta to content instead of reasoning_content
## Bug

In `packages/openai-adapters/src/vercelStreamConverter.ts` lines 88-92, the `reasoning-delta` case maps to `content` instead of `reasoning_content`:

```typescript
case "reasoning-delta":
  return chatChunk({
    content: part.text,  // Should be reasoning_content
    model,
  });
```

This causes reasoning output to be mixed into assistant content for any provider using the Vercel streaming adapter, breaking the separation that `fromChatCompletionChunk()` in `core/llm/openaiTypeConverters.ts` relies on to route `reasoning_content` → `{ role: "thinking" }` messages.

## Impact

Models served through Vercel-compatible adapters will have their reasoning content appear as regular assistant text. The GUI's thinking/content separation won't work for these providers.

## Context

Found while tracing the `reasoning_content` handling path through the codebase to clarify ownership of #10783 and #10785. Those issues are upstream (LM Studio's content field contamination when its Harmony parser's phase tracking fails — lmstudio-ai/lmstudio-bug-tracker#1589, lmstudio-ai/lmstudio-bug-tracker#1592). Continue's OpenAI adapter layer correctly maps what it receives. This Vercel converter bug is a separate issue on a different adapter path.

siewcapital:
## Summary
- map `reasoning-delta` stream parts to `delta.reasoning_content` in the Vercel stream converter
- keep `delta.content` reserved for actual assistant text deltas
- add regression checks for both single-event conversion and mixed stream conversion

## Why
`reasoning-delta` was previously being emitted as `content`, which mixed thinking output into assistant text for Vercel adapter streams.

## Testing
- `pnpm exec vitest run src/test/vercelStreamConverter.test.ts` (from `packages/openai-adapters`)

Closes #11069

### DIFF
--- a/packages/openai-adapters/src/test/vercelStreamConverter.test.ts
+++ b/packages/openai-adapters/src/test/vercelStreamConverter.test.ts
@@ -24,7 +24,7 @@ describe("convertVercelStreamPart", () => {
     expect(result?.choices[0].delta.content).toBe("Hello");
   });
 
-  test("converts reasoning-delta to chat chunk", () => {
+  test("converts reasoning-delta to reasoning_content delta", () => {
     const part: VercelStreamPart = {
       type: "reasoning-delta",
       id: "reasoning-1",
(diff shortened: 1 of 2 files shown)


RomneyDa (approved):
Nice fix! note CLA requirement. Also non-blocking nitpick but avoid `any` type if possible.

PR #11215, opened 2026-03-09
fix: stop CLI free-trial polling for non-free-trial models

### LINKED ISSUE #11214: CLI polls free-trial status every 5s even when no free-trial model is active
## Summary
The Continue CLI TUI polls the /ide/free-trial-status endpoint on api.continue.dev every 5 seconds from the FreeTrialStatus component, even when the active model is not a free-trial continue-proxy model.

## Why this is a problem
- Users see recurring outbound traffic to Continue-owned infrastructure while using the local CLI.
- CONTINUE_TELEMETRY_ENABLED=0 disables PostHog telemetry, but it does not stop this API polling, so the observable network behavior is surprising.
- The polling appears unnecessary unless the active model is actually using free-trial quota.

## Reproduction
1. Set CONTINUE_TELEMETRY_ENABLED=0.
2. Start the local CLI TUI with a non-free-trial model.
3. Observe recurring HTTPS requests to the Continue API roughly every 5 seconds.

## Root cause
BottomStatusBar always mounts FreeTrialStatus in local TUI mode, and FreeTrialStatus immediately fetches and starts a 5-second interval before checking whether the model is a free-trial model.

## Expected behavior
Only fetch and poll /ide/free-trial-status when:
- an API client is available, and
- the active model is a free-trial continue-proxy model.

## Additional note
There is also a separate startup-only version check to /cn/info, but that is not the 5-second recurring traffic described here.

guoyi-smh:
## Summary
- gate the CLI free-trial status fetch/polling behind an actual free-trial model check
- keep the existing 5 second polling behavior only for free-trial continue-proxy models
- add a focused UI test covering both the non-free-trial and free-trial paths

Closes #11214.

## Validation
- ./node_modules/.bin/vitest run src/ui/FreeTrialStatus.test.tsx
- ./node_modules/.bin/eslint src/ui/FreeTrialStatus.tsx src/ui/FreeTrialStatus.test.tsx

## Notes
- I installed dependencies locally in extensions/cli to run the focused test.
- A full npm install currently trips the package prepare build in this checkout because sibling workspace packages are not built, so validation stayed scoped to the touched files.

---
## Summary by cubic
Stop the CLI from fetching or polling free‑trial status for non‑free‑trial models. Keep immediate fetch and 5s polling only for free‑trial `continue-proxy` models. Closes #11214.

- **Bug Fixes**
  - Gate initial fetch/polling behind `shouldFetchStatus` (`apiClient` exists and `isModelUsingFreeTrial(model)`).
  - Harden `isModelUsingFreeTrial`: narrow `ModelConfig` via provider check before `apiKeyLocation`, add optional chaining, remove unnecessary cast.
  - Clean up polling with mounted checks, remove the `NODE_ENV === "test"` branch, and drop the `FreeTrialStatus` test file.

<sup>Written for commit 58cdfebeb4058305c86c16cac1d52ea4c5c24dd5. Summary will update on new commits.</sup>

### DIFF
--- a/extensions/cli/src/ui/FreeTrialStatus.tsx
+++ b/extensions/cli/src/ui/FreeTrialStatus.tsx
@@ -7,10 +7,11 @@ import { Text } from "ink";
 import React, { useEffect, useState } from "react";
 
 export function isModelUsingFreeTrial(model: ModelConfig): boolean {
-  if (model.provider !== "continue-proxy") {
-    return false;
-  }
-  return !!model.apiKeyLocation?.startsWith("free_trial:");
+  return (
+    model.provider === "continue-proxy" &&
+    "apiKeyLocation" in model &&
+    !!model.apiKeyLocation?.startsWith("free_trial:")
+  );
 }
 
 interface FreeTrialStatusProps {


guoyi-smh:

extensions/vscode/e2e/install-marketplace-extensions.mjs
RomneyDa, line 1: Can you remove these changes?

extensions/vscode/package.json
RomneyDa, line 1: And these

PR #11289, opened 2026-03-11
fix: preserve indentation when applying code edits to Python files

### LINKED ISSUE #11282: The code modification tool has incorrect indentation on Python

- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows
- Continue version: 1.29.1
- IDE version: 1.110.1
- Model: DeepSeek 3.2V
```

### Description

I found that indentation issues are very frequent in Python code modification tools, and when I asked AI, they said that the logic of this code modification tool is causing it. Has this been modified in the latest version

### To reproduce

Use this tool to write Python projects

### Log output

```Shell

```

mvanhorn:
## Summary
- Fixes indentation corruption when the code edit tool uses fallback match strategies (`trimmedMatch`, `whitespaceIgnoredMatch`, etc.) to apply edits
- When an LLM provides search/replace strings with different indentation than the target code (common with Python), the replacement text is now re-indented to match the file's actual indentation level

## How it works

`adjustReplacementIndentation()` compares the indentation of the file line at the match position against the leading indent of the LLM-provided `old_string`. If they differ:

- **First line**: strips the old indent prefix (the file content before the splice point already provides correct indentation)
- **Subsequent lines**: replaces the old indent prefix with the matched indent, preserving relative inner indentation
- **Exact matches**: passed through unchanged - no indentation adjustment needed

This handles tab/space mismatches, completely unindented LLM output, and deeply nested code blocks.

## Test plan
- [x] Unindented `old_string` matching indented file content gets correct indentation
- [x] Multi-level nesting preserves relative inner indentation
- [x] Tab vs space mismatch resolved correctly
- [x] Exact matches unchanged (no regression)
- [x] `whitespaceIgnoredMatch` with differing internal whitespace works correctly
- [x] `replaceAll` mode applies indentation adjustment per-match

Fixes #11282

### DIFF
--- a/core/edit/searchAndReplace/executeFindAndReplace.vitest.ts
+++ b/core/edit/searchAndReplace/executeFindAndReplace.vitest.ts
@@ -207,6 +207,84 @@ describe("executeFindAndReplace", () => {
     });
   });
 
+  describe("indentation adjustment", () => {
+    it("should adjust indentation when trimmedMatch finds unindented old_string in indented file", () => {
+      const fileContent =
+        "def foo():\n    x = 1\n    y = 2\n    return x + y\n";
+      // LLM sends unindented old_string that matches via trimmedMatch
+      const oldString = "x = 1\ny = 2";
+      const newString = "x = 10\ny = 20";
+      const result = executeFindAndReplace(
+        fileContent,
+        oldString,
+        newString,
+        false,
+      );
+      expect(result).toBe(
+        "def foo():\n    x = 10\n    y = 20\n    return x + y\n",
+      );
+    });
+
+    it("should preserve relative inner indentation in multi-line replacement", () => {
+      const fileContent =
+        "class Foo:\n    def bar(self):\n        if True:\n            x = 1\n";
+      const oldString = "if True:\n    x = 1";
+      const newString = "if True:\n    x = 1\n    y = 2";
+      const result = executeFindAndReplace(
+        fileContent,
+        oldString,
+        newString,
+        false,
+      );
+      expect(result).toBe(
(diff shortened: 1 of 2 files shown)


mvanhorn:

mvanhorn:
Thanks to @luzhihaoTestingLab for reporting #11282 and to @cubic-dev-ai for catching the first-line indentation bug in review.

**What was wrong with the original approach (commit 3d257cd):**

The first version used `getLeadingIndent(matchedText)` to determine the file's indentation - extracting it from the matched substring itself. It then applied indentation uniformly to all lines, including the first. The problem is that when the splice happens at `startIndex`, the file content *before* `startIndex` already provides the leading whitespace for the first line. So the first line would get double-indented. The fallback `if (index === 0 && oldIndent === "" ...)` only helped when the LLM sent completely unindented text - if `oldIndent` was non-empty (e.g., spaces vs tabs mismatch), the first line was left wrong.

**What the new approach does (commit 1cb68a5):**

Replaced `getLeadingIndent(matchedText)` with `getLineIndentAtPosition(fileContent, match.startIndex)`, which reads indentation from the full file line at the match position. The first line now *strips* the old indent instead of adding the new one (since the splice point inherits the file's leading whitespace). Subsequent lines replace `oldIndent` with `matchedIndent`, preserving relative nesting. Handles all cases - unindented LLM output, tab/space mismatches, and multi-level nesting.

Added 5 test cases covering: basic indent adjustment, relative inner indentation preservation, tab vs space mismatch, exact match passthrough, and whitespace-ignored matching.

Issue #11164, opened 2026-03-07
Error: Gemini 3.1 Pro - Unknown error


linnway:
**Error Details**

Model: Gemini 3.1 Pro
Provider: gemini
Status Code: N/A

**Error Output**
```
{"error":{"message":"{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Please ensure that function call turn comes immediately after a user turn or after a function response turn.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n","code":400,"status":"Bad Request"}}
```

**Additional Context**
Please add any additional context about the error here

RomneyDa:
@linnway any chance you started the conversation with a different provider and then switched to a gemini model? Or entire conversation with gemini? Have seen this bug crop up in the former case, curious

Issue #11047, opened 2026-03-05
Error: Gemini 3.1 Pro Preview - Unknown error


daedaevibin:
**Error Details**

Model: Gemini 3.1 Pro Preview
Provider: gemini
Status Code: N/A

**Error Output**
```
{"error":{"message":"{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Please ensure that function call turn comes immediately after a user turn or after a function response turn.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n","code":400,"status":"Bad Request"}}
```

**Additional Context**
Every response says this.

Known fix: Compact conversation and send a new message with that.

Issue #9575, opened 2026-01-16
Error: Gemini 2.5 Flash - Unknown error


mamba73:
**Error Details**

Model: Gemini 2.5 Flash
Provider: gemini
Status Code: N/A

**Error Output**
```
{"error":{"message":"{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Please ensure that function call turn comes immediately after a user turn or after a function response turn.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n","code":400,"status":"Bad Request"}}
```

**Additional Context**
Please add any additional context about the error here

sestinj:
Potentially related issues:

- #9562 - Gemini 2.5 Pro - Similar function call/response formatting error (INVALID_ARGUMENT)
- #9249 - Ollama models fail during MCP tool calling due to message formatting incompatibilities (related root cause)
- #6121 - Gemini models in agent mode giving error 400 invalid argument (closed)

RomneyDa:
@mamba73 is this happening with the CLI or with an IDE extension?

mamba73:
IDE extension.
I just 'hit' default filled report.

Issue #9562, opened 2026-01-15
Error: Gemini 2.5 Pro - Unknown error


robertmash2:
**Error Details**

Model: Gemini 2.5 Pro
Provider: vertexai
Status Code: N/A

**Error Output**
```
{"error":{"message":"{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n","code":400,"status":"Bad Request"}}
```

**Additional Context**
Please add any additional context about the error here

sestinj:
Related issues involving Gemini/Vertex AI function call errors:

- #8785 - Gemini 3 Pro function call missing thought_signature error
- #8980 - Gemini 3 Pro via OpenRouter 400 error when calling tools

PR #11805, opened 2026-03-25
fix(gemini): merge consecutive same-role messages to fix ordering errors

RomneyDa:
## Summary

- Fixes Gemini API 400 errors caused by consecutive messages with the same role breaking Gemini's strict role-alternation requirement
- Adds a post-processing step that merges consecutive same-role messages by combining their `parts` arrays
- Applied in both `core/llm/llms/Gemini.ts` and `packages/openai-adapters/src/apis/Gemini.ts` (which also covers VertexAI)

### Root cause

When Continue converts chat messages to Gemini format, tool responses are mapped to `"user"` role with `functionResponse` parts. Multiple consecutive tool responses (e.g., after a multi-tool-call assistant turn) create consecutive `"user"` messages, violating Gemini's alternation requirement. Similarly, consecutive assistant messages create consecutive `"model"` messages.

This causes two distinct Gemini API errors:
1. _"Please ensure that function call turn comes immediately after a user turn or after a function response turn."_
2. _"Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn."_

### Fix

A simple merge pass after message conversion: consecutive messages with the same role have their `parts` arrays concatenated into a single message. This ensures strict user/model alternation while preserving all content.

Fixes https://github.com/continuedev/continue/issues/11293
Fixes https://github.com/continuedev/continue/issues/11164
Fixes https://github.com/continuedev/continue/issues/11047
Fixes https://github.com/continuedev/continue/issues/9575
Fixes https://github.com/continuedev/continue/issues/9562

## Test plan

- [x] Added unit tests for the merge logic (4 tests covering: merged tool responses, tool+user merge, consecutive model merge, no-op for already-alternating messages)
- [ ] Manual test with Gemini model using multi-tool-call conversations

### DIFF
--- a/core/llm/llms/Gemini.ts
+++ b/core/llm/llms/Gemini.ts
@@ -1,5 +1,4 @@
 import { streamResponse } from "@continuedev/fetch";
-import { mergeConsecutiveGeminiMessages } from "@continuedev/openai-adapters";
 import { v4 as uuidv4 } from "uuid";
 import {
   AssistantChatMessage,
@@ -23,6 +22,7 @@ import {
   GeminiGenerationConfig,
   GeminiToolFunctionDeclaration,
   convertContinueToolToGeminiFunction,
+  mergeConsecutiveGeminiMessages,
 } from "./gemini-types";
 
 interface GeminiToolCallDelta extends ToolCallDelta {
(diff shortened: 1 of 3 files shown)


packages/openai-adapters/src/util/gemini-types.ts
```diff
@@ -428,3 +428,27 @@ interface LogprobsResult {
 interface TopCandidates {
   candidates: Candidate[];
 }
+
```
RomneyDa, line 431: functions are duplicated because small enough that mocking complexity etc isn't worth it across packages

PR #11867, opened 2026-03-26
fix: ls tool no longer hides gitignored directories from agent

### LINKED ISSUE #11303: Agent cannot find local .venv because ls-tool filters .gitignore

### Relevant environment info

```Markdown
- OS: Windows
- Continue version: v1.2.16 
- IDE version: 1.107.0
```

### Description

I want the Agent to use the local python .venv to run python commands.
However, it cant find the .venv folder, since the ls tool seems to filter out any folders contained in the .gitignore file.

### To reproduce

1. add any folder in the root directory to gitignore
2. ask the agent to list all folders -> it will not find any ignored folders

### Log output

```Shell

```

RomneyDa:
## Summary
- The built-in `ls` tool was respecting `.gitignore` rules, hiding directories like `.venv` from the agent's view
- Added a `skipGitIgnore` option to `walkDir` and enabled it in the ls tool so the agent can discover and interact with all directories

Fixes #11303

## Test plan
- [ ] Create a `.venv` directory (or any gitignored directory) in a workspace
- [ ] Ask the agent to list folders — it should now show `.venv` and other gitignored directories
- [ ] Verify that indexing/other features still respect `.gitignore` (only the ls tool behavior changed)

### DIFF
--- a/core/indexing/walkDir.ts
+++ b/core/indexing/walkDir.ts
@@ -11,6 +11,7 @@ export interface WalkerOptions {
   returnRelativeUrisPaths?: boolean;
   source?: string;
   overrideDefaultIgnores?: Ignore;
+  skipGitIgnore?: boolean;
   recursive?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


RomneyDa:
Decided this isn't the right change for now because security implications aren't clear.

Issue #11063, opened 2026-03-05
Tool usage Error: gpt-oss:20b - Unknown error


leopold-imgp:
**Error Details**

Model: gpt-oss:20b
Provider: ollama
Status Code: N/A

**Error Output**
```
Error parsing Ollama response: Error: error parsing tool call: raw='*** Begin Patch
*** Update File: src/models/ljg_angents/thread.py
@@
-from dataclasses import dataclass
-from datetime import datetime
-from typing import Any, Dict, Optional, Union
 from dataclasses import dataclass
 from datetime import datetime
 from typing import Any, Dict, Optional, Union
@@
-@dataclass
-class ChunkMessagePayload:
-    {"text":"The company referenced in the provided documents is **Trinity Street Asset Management (TSAM)**.\n\nHere are some key details from the documents:\n\n1. **Full Name**:\n   - Trinity Street Asset Management (TSAM), originally named Bruce Nelson Capital LLP.\n\n2. **Establishment**:\n   - Founded on **5th December 2002**.\n\n3. **Legal Structure**:\n   - A **Limited Liability Partnership (LLP)** domiciled in the UK.\n   - Partnership No: **OC303508**.\n\n4. **Regulatory Status**:\n   - Authorized and regulated by the **Financial Conduct Authority (FCA)** in the UK.\n   - Registered with the **U.S. Securities and Exchange Commission (SEC)** as an investment adviser (File number: **801-71133**).\n   - Registered under the **Alternative Investment Fund Managers Directive (AIFMD)** since January 2015.\n\n5. **Ownership**:\n   - Owned by its partners, **Richard Bruce** (Portfolio Manager and CEO) and **Ed Bell** (Portfolio Manager and Deputy CEO), along with their families, who hold the controlling interest.\n\n6. **Headquarters**:\n   - Located in **London**, with almost all staff based there. Two members of the TSAM group are based in the USA for marketing and client service support."},"id":199,"created_at":"2026-03-05T11:00:57.222763","updated_at":"2026-03-05T11:00:57.222763"}
 @dataclass
 class ChunkMessagePayload:
     """
     Payload for a chunk message.  The fields correspond to the JSON structure
     that the event stream emits.  The ``created_at`` and ``updated_at`` fields
     are parsed as :class:`datetime.datetime` objects for convenience.
     """
 
     text: str
     id: int
     created_at: datetime
     updated_at: datetime
*** End Patch
[...]

Issue #10954, opened 2026-03-01
Error: gpt-oss:20b - Unknown error


scuyul:
**Error Details**

Model: gpt-oss:20b
Provider: ollama
Status Code: N/A

**Error Output**
```
Error parsing Ollama response: Error: error parsing tool call: raw='{"tool":"edit_file","args":{"path":"main.py","patch":"*** Begin Patch\n*** Update File: main.py\n@@\n-MODEL = \"qwen3-coder:30b\"\n-DEBUG = True\n MODEL = \"gpt-4o-mini\"\n DEBUG = False\n@@\n-    return f\"\"\"You are an AI coding agent with access to file and time tools.\n-Your working code directory (RAG folder) is: {rag_dir}\n     return f\"\"\"You are an AI coding agent with access to file and time tools.\n Your working code directory (RAG folder) is: {rag_dir}\n*** End Patch"}', err=unexpected end of JSON input {"error":"error parsing tool call: raw='{\"tool\":\"edit_file\",\"args\":{\"path\":\"main.py\",\"patch\":\"*** Begin Patch\\n*** Update File: main.py\\n@@\\n-MODEL = \\\"qwen3-coder:30b\\\"\\n-DEBUG = True\\n MODEL = \\\"gpt-4o-mini\\\"\\n DEBUG = False\\n@@\\n-    return f\\\"\\\"\\\"You are an AI coding agent with access to file and time tools.\\n-Your working code directory (RAG folder) is: {rag_dir}\\n     return f\\\"\\\"\\\"You are an AI coding agent with access to file and time tools.\\n Your working code directory (RAG folder) is: {rag_dir}\\n*** End Patch\"}', err=unexpected end of JSON input"}
```

**Additional Context**
i am using ollama

RomneyDa:
This error originates from Ollama's server-side tool call parser (`harmony/harmonyparser.go`) — when the model produces malformed output instead of valid JSON for tool call arguments, Ollama returns `"error parsing tool call"` which Continue was surfacing as an opaque "Unknown error".

This is a **transient** issue with certain models (especially `gpt-oss:20b`) that don't reliably produce valid JSON for native tool calls. Resubmitting the message often works.

**Workaround:** Enable **"Only use system message tools"** in Settings > Experimental. This bypasses Ollama's native tool call parser entirely and uses system-message-based tool calling instead.

PR #11870 improves the error handling so that instead of "Unknown error", users now see an actionable message explaining the issue and the workaround.

[...]

Issue #10897, opened 2026-02-27
Error: SimonPu/GLM-4.7-Flash:Q8 - Unknown error


Sparrkz-old:
**Error Details**

Model: SimonPu/GLM-4.7-Flash:Q8
Provider: ollama
Status Code: N/A

**Error Output**
```
Error parsing Ollama response: Error: failed to parse XML: XML syntax error on line 2: element <tool_call> closed by </arg_value> {"error":"failed to parse XML: XML syntax error on line 2: element \u003ctool_call\u003e closed by \u003c/arg_value\u003e"}
```

**Additional Context**
Please add any additional context about the error here

RomneyDa:
This error originates from Ollama's server-side tool call parser (`harmony/harmonyparser.go`) — when the model produces malformed output instead of valid JSON for tool call arguments, Ollama returns `"error parsing tool call"` which Continue was surfacing as an opaque "Unknown error".

This is a **transient** issue with certain models (especially `gpt-oss:20b`) that don't reliably produce valid JSON for native tool calls. Resubmitting the message often works.

**Workaround:** Enable **"Only use system message tools"** in Settings > Experimental. This bypasses Ollama's native tool call parser entirely and uses system-message-based tool calling instead.

PR #11870 improves the error handling so that instead of "Unknown error", users now see an actionable message explaining the issue and the workaround.

Closing as the upstream issue is in Ollama's parser, and the Continue-side improvement is in #11870.

Issue #10711, opened 2026-02-22
Error: gpt-oss:20b local - Unknown error


kamidehlvi1:
**Error Details**

Model: gpt-oss:20b local
Provider: ollama
Status Code: N/A

**Error Output**
```
Error parsing Ollama response: Error: error parsing tool call: raw='repo_browser.open_file({"path":"frontendnextjs/pages/index.js","line_start":1,"line_end":200})
', err=invalid character 'r' looking for beginning of value {"error":"error parsing tool call: raw='repo_browser.open_file({\"path\":\"frontendnextjs/pages/index.js\",\"line_start\":1,\"line_end\":200})\n', err=invalid character 'r' looking for beginning of value"}
```

**Additional Context**
Please add any additional context about the error here

anakiller:
Hello, I have also the same bug. Continue in vscode is totally unusable.

RomneyDa:
This error originates from Ollama's server-side tool call parser (`harmony/harmonyparser.go`) — when the model produces malformed output instead of valid JSON for tool call arguments, Ollama returns `"error parsing tool call"` which Continue was surfacing as an opaque "Unknown error".

This is a **transient** issue with certain models (especially `gpt-oss:20b`) that don't reliably produce valid JSON for native tool calls. Resubmitting the message often works.

**Workaround:** Enable **"Only use system message tools"** in Settings > Experimental. This bypasses Ollama's native tool call parser entirely and uses system-message-based tool calling instead.

PR #11870 improves the error handling so that instead of "Unknown error", users now see an actionable message explaining the issue and the workaround.

Closing as the upstream issue is in Ollama's parser, and the Continue-side improvement is in #11870.

PR #11870, opened 2026-03-26
fix: show actionable error when Ollama fails to parse tool calls

### LINKED ISSUE #11783: Error: gpt-oss:20b - Unknown error
**Error Details**

Model: gpt-oss:20b
Provider: ollama
Status Code: N/A

**Error Output**
```
Error parsing Ollama response: Error: error parsing tool call: raw='We should output the result.We can call the tool again.We need to produce the correct tool call.We'll call waselli_mcp_show_waselli_root.But we need to confirm if environment variable is set. We already did. The tool should return "C:\Users\Ultra\.gemini\antigravity\scratch\waselli". Let's do that.Let's call the tool.We must provide JSON arguments: empty%3F Just no arguments.Let's do that.But the earlier tool call produced "Waselli root: C:\Users\Ultra\.gemini\antigravity\scratch\waselli". That is the result. Probably we should provide a clear answer: "The root is C:\Users\Ultra\.gemini\antigravity\scratch\waselli". Then maybe list the tree. The user didn't specify. Let's just show root.', err=invalid character 'W' looking for beginning of value {"error":"error parsing tool call: raw='We should output the result.We can call the tool again.We need to produce the correct tool call.We'll call waselli_mcp_show_waselli_root.But we need to confirm if environment variable is set. We already did. The tool should return \"C:\\Users\\Ultra\\.gemini\\antigravity\\scratch\\waselli\". Let's do that.Let's call the tool.We must provide JSON arguments: empty%3F Just no arguments.Let's do that.But the earlier tool call produced \"Waselli root: C:\\Users\\Ultra\\.gemini\\antigravity\\scratch\\waselli\". That is the result. Probably we should provide a clear answer: \"The root is C:\\Users\\Ultra\\.gemini\\antigravity\\scratch\\waselli\". Then maybe list the tree. The user didn't specify. Let's just show root.', err=invalid character 'W' looking for beginning of value"}
```

**Additional Context**
Please add any additional context about the error here

RomneyDa:
## Summary

When Ollama models (e.g. `gpt-oss:20b`) produce malformed tool call output, Ollama's server-side parser returns `"error parsing tool call"` which was surfacing as an opaque "Unknown error" in the stream error dialog.

This PR adds pattern matching in `analyzeError()` to detect this specific error and show a `customErrorMessage` explaining:
- The issue is transient and they can resubmit to try again
- They can enable "Only use system message tools" in Settings > Experimental to bypass Ollama's native tool call parser

Fixes #11783, #11063, #10954, #10897, #10711, #10221, #9433

## Test plan
- [ ] Use `gpt-oss:20b` on Ollama with tool calling — when a tool call parse failure occurs, verify the stream error dialog shows the actionable message instead of "Unknown error"
- [ ] Verify other Ollama errors (e.g. "model not found") still show their normal error messages

### DIFF
--- a/gui/src/util/errorAnalysis.ts
+++ b/gui/src/util/errorAnalysis.ts
@@ -150,7 +150,7 @@ export function analyzeError(
   }
 
   // Ollama tool call parsing failure (transient model output issue)
-  if (errorText.includes("error parsing ollama response: error: error parsing tool call")) {
+  if (errorText.includes("error parsing tool call")) {
     customErrorMessage =
       "This model produced an invalid tool call that Ollama could not parse. " +
       "This is a known transient issue — you can resubmit your message to try again. " +


RomneyDa:
The error originates from Ollama's server-side tool call parser — when models like gpt-oss:20b produce malformed output instead of valid JSON, Ollama returns `"error parsing tool call"` which we were showing as "Unknown error".

This is a **transient** issue ([confirmed in #9433](https://github.com/continuedev/continue/issues/9433) — resubmitting works). So rather than disabling native tools or adding retry logic, we just detect the error string in `analyzeError()` and show a helpful message in the stream error dialog pointing users to resubmit or enable system message tools.

Patrick-Erichsen (approved):

PR #11887, opened 2026-03-26
fix: simplify MCP STDIO reconnection by closing client before reconnect

### LINKED ISSUE #11886: MCP STDIO transport reconnection race condition causes inconsistent state
## Description

When reconnecting an MCP server (e.g., pressing reload or disable/enable in the Continue IDE plugin), the following error occurs:

```
Failed to connect to "compliance-mcp"
Error: Already connected to a transport. Call close() before connecting to a new transport, or use a separate Protocol instance per connection.
```

The initial connection always succeeds, but any subsequent reconnection fails.

## Reproduction

1. Configure an SSE MCP server in Continue (e.g., `type: sse`, `url: http://127.0.0.1:8080/sse`)
2. Connect successfully
3. Press the reload button, or disable and re-enable the server
4. Error appears: "Already connected to a transport..."

## Root Cause

In `core/context/mcp/MCPConnection.ts`, the same `Client` instance is reused across reconnections. The MCP SDK's `Protocol.connect()` checks if `_transport` is set and throws if it is. `_transport` is only cleared in `_onclose()`, which fires asynchronously via the transport's `onclose` callback. With SSE transports, `close()` can resolve before `_onclose` fires, leaving `_transport` still set.

Additionally, the auto-detect fallback path (HTTP → SSE) has the same issue: if the HTTP attempt fails after setting `_transport`, the SSE fallback hits the same guard on the same `Client` instance.

## Environment

- **IDE:** IntelliJ/PyCharm with Continue plugin
- **File:** `core/context/mcp/MCPConnection.ts`
- **Transport type:** SSE (also affects STDIO and auto-detect fallback)
- **MCP SDK:** `@modelcontextprotocol/sdk ^1.25.2`

## Fix

See PR #11887 — creates a fresh `Client` instance on each reconnection.

yaruchyo:
## Summary

Fixes #11886

Resolves the `Already connected to a transport. Call close() before connecting to a new transport` error that occurs when reconnecting MCP servers (e.g., pressing reload or disable/enable in the IDE plugin).

## Root Cause

The MCP SDK's `Protocol` class (base of `Client`) stores `_transport` as private state and guards `connect()`:
```typescript
if (this._transport) {
    throw new Error('Already connected to a transport...');
}
```

`_transport` is only cleared in `_onclose()`, which is triggered asynchronously via the transport's `onclose` callback. With SSE transports, `close()` can resolve **before** `_onclose` fires, so `_transport` is still set when the next `connect()` is called.

The old STDIO-specific try/catch only handled `StdioClientTransport already started` — it did not cover SSE/HTTP transports at all.

## Changes

- **Extract `createClient()` helper** — reusable Client factory method
- **Create a fresh `Client` on each reconnection** — guarantees `_transport` is `undefined`, completely avoiding the race condition regardless of transport type
- **Best-effort `close()` before reconnect** — wrapped in try/catch to clean up the old connection without blocking if it was never connected
- **Fix auto-detect fallback (HTTP → SSE)** — the fallback path also creates a fresh `Client` between attempts, fixing a second instance of the same bug

## Testing

- Tested with SSE MCP server (compliance-mcp): initial connect, reload, disable/enable all work without errors
- STDIO transport reconnection also verified
- Auto-detect fallback (no explicit type) path covered

### DIFF
--- a/core/context/mcp/MCPConnection.ts
+++ b/core/context/mcp/MCPConnection.ts
@@ -102,6 +102,24 @@ class MCPConnection {
     );
   }
 
+  private async connectTransport(transport: Transport): Promise<void> {
+    try {
+      await this.client.close();
+      await this.client.connect(transport, {});
+    } catch (error) {
+      if (
+        error instanceof Error &&
+        error.message.startsWith("Already connected")
+      ) {
+        this.client = this.createClient();
+        await this.client.connect(transport, {});
+      } else {
+        throw error;
+      }
+    }
+    this.transport = transport;
+  }
+
   async disconnect(disable = false) {
     this.abortController.abort();
     await this.client.close();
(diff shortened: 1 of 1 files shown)


yaruchyo:
@RomneyDa  Good point! 
I've updated the approach — the new implementation reuses the existing Client by default (close() → connect() on the same instance), preserving server session and context. A fresh Client is only created as a fallback when the SDK throws "Already connected to a transport" due to the async _onclose race condition.
I also extracted a shared connectTransport() helper so this resilience applies to all transport types (SSE, HTTP, WebSocket), not just STDIO.

Does this approach work for you?

PR #12044, opened 2026-04-05
feat: add Gemma 4 model support

RomneyDa:
This PR adds Google's new Gemma 4 model support for Ollama. Adds tool support detection for openrouter


RomneyDa:
Superseded by #12046 which includes these changes adapted to use explicit provider approach

PR #12085, opened 2026-04-09
fix(tools): resolve PowerShell ENOENT by using full path on Windows

### LINKED ISSUE #12079: Terminal/Shell Commands Fail with spawn powershell.exe ENOENT Error

- [ ] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer

### Relevant environment info

```Markdown
OS: Windows
Continue version: 1.3.38
IDE version: 1.114.0
Model: DeepSeek
```

### Description

The run_terminal_command tool consistently fails with the error spawn powershell.exe ENOENT when attempting to execute any shell commands. This prevents file deletion operations and other terminal-based tasks within the Continue workspace.

Key symptoms:

File creation and editing tools work correctly (create_new_file, edit_existing_file, single_find_and_replace)
All terminal commands fail with the same error regardless of command type (rm, del, echo, etc.)
Cannot delete files from filesystem despite being able to create and edit them
Error suggests PowerShell executable cannot be found/spawned
Impact:

Limits file management capabilities (cannot delete files)
Prevents execution of shell scripts or commands
Reduces overall functionality of the agent
To reproduce
Create a test file using the agent:

Javascript

Apply
// Agent creates test.jsx with React component
Attempt to delete the file using terminal command:

Bash
Run
rm test.jsx
or

Cmd

Apply
del test.jsx
Observe the error:

Run
run_terminal_command failed with the message: spawn powershell.exe ENOENT
Alternative reproduction steps:

Try any simple shell command:
Bash
Run
echo "test"
Same error occurs
Environment details observed:

Workspace: Node.js/Next.js project
User Settings: Contains editor/formatter settings but no terminal configuration
Workspace Settings: Empty {} object
Security restriction: Cannot create/modify .vscode/settings.json due to security concerns in the workspace
Log output
[Please follow the troubleshooting guide in Continue Docs to obtain logs and add them here]

Expected behavior:

Terminal commands should execute successfully
File deletion should be possible via shell commands
Shell environment should be properly configured or have fallback mechanisms
Actual behavior:

All terminal commands fail with spawn powershell.exe ENOENT
No shell commands can be executed
File deletion is impossible via terminal

### To reproduce

_No response_

### Log output

```Shell

```

octo-patch:
Fixes #12079

## Problem

On Windows, the `run_terminal_command` tool fails with `spawn powershell.exe ENOENT` for all terminal commands (even simple ones like `echo "test"`). File creation/editing tools work fine, but nothing that uses the terminal works.

The root cause is that the VSCode extension host process may run with a minimal PATH that does not include `%SystemRoot%\System32` (e.g. `C:\Windows\System32`), where `powershell.exe` lives. When `child_process.spawn("powershell.exe", ...)` is called without a full path, Node.js fails to locate the executable.

## Solution

Replace the bare `"powershell.exe"` shell name with the full absolute path constructed from the `SystemRoot` environment variable, which is always present on Windows:

```
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
```

`SystemRoot` is a guaranteed Windows environment variable that points to the Windows installation directory (typically `C:\Windows`). Using it avoids any dependency on the PATH being set correctly in the extension host.

## Testing

- On non-Windows systems: no change in behavior (the `else` branch handles Unix/macOS).
- On Windows: `powershell.exe` is now resolved via the full path from `SystemRoot`, bypassing PATH lookup issues.
- Existing test suite covers the `runTerminalCommandImpl` function and should continue to pass.

---
## Summary by cubic
Fixes Windows terminal commands failing with "spawn powershell.exe ENOENT" by invoking PowerShell via its absolute path built from the `SystemRoot` env var (fallback to C:\Windows) instead of relying on PATH. No behavior change on macOS/Linux.

<sup>Written for commit 8bdb84d8f01046149b20d6e04bb06efa1dad1932. Summary will update on new commits.</sup>

### DIFF
--- a/core/tools/implementations/runTerminalCommand.ts
+++ b/core/tools/implementations/runTerminalCommand.ts
@@ -24,9 +24,14 @@ function getDecodedOutput(data: Buffer): string {
 } // Simple helper function to use login shell on Unix/macOS and PowerShell on Windows
 function getShellCommand(command: string): { shell: string; args: string[] } {
   if (process.platform === "win32") {
-    // Windows: Use PowerShell
+    // Windows: Use PowerShell with full path to avoid ENOENT when the
+    // extension host PATH doesn't include Windows system directories.
+    // SystemRoot is always set on Windows (e.g. C:\Windows).
+    const systemRoot =
+      process.env.SystemRoot || process.env.SYSTEMROOT || "C:\\Windows";
+    const powershellPath = `${systemRoot}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe`;
     return {
-      shell: "powershell.exe",
+      shell: powershellPath,
       args: ["-NoLogo", "-ExecutionPolicy", "Bypass", "-Command", command],
     };
   } else {


houssemzaier:
We've been shipping this exact fix in a downstream fork (3,000+ users) and it reliably resolves the PowerShell `ENOENT` on Windows when the extension-host `PATH` doesn't include the system directories (e.g. `C:\Windows\System32\WindowsPowerShell\v1.0`). It's been stable in production — would be great to get this merged. 👍

PR #12324, opened 2026-05-07
fix: prevent rejected pending edits from reporting success

### LINKED ISSUE #12269: Responding with a follow-up when AI suggests an edit results in desync

- [ ] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 1.2.22
- IDE version: Code 1.117.0
- Model: "DeepSeek Coder" (deepseek-v4-flash)
```

### Description

When the AI requests an edit, you can type in a follow-up instead of rejecting or accepting the edit. The edit request will not be carried out, but the AI will receive an indicator that the edit was successful. As a result, the AI's view of the file will disagree with the actual state.

While this is less a problem for short files, with longer files, the AI may output a diff against their desynchronized view of the file, which fails to apply.

### To reproduce

1. Create a new C source file with an empty main function:

```c
#include <stdio.h>
int main(int argc, char **argv)
{
}
```

2. Ask the AI to write Hello World in the current file.

4. After an edit is suggested, ask the AI to add the functionality to print out argv[0]. The edit will not be carried out, but the interface of Continue marks this edit as accepted.

### Log output

```Shell

```

danyalahmed1995:
Fixes #12269

## Summary

This PR fixes the pending edit desync case described in the issue.

If a user sends a follow-up while an edit is still pending, the pending diff can be closed without the edit actually being applied. In that case, Continue should not treat the edit as successful, because the file on disk has not changed.

This change carries the accept/reject result through the pending edit flow and only reports edit success when the edit was actually accepted/applied.

## Testing

- Ran `npm.cmd --prefix gui test -- src/redux/thunks/handleApplyStateUpdate.test.ts`
- Ran `git diff --check`

## Notes

I was not able to run a full live VS Code extension repro in my local setup, so I added focused regression coverage for the false success path.

---
## Summary by cubic
Prevents pending edits that were rejected or auto-closed from being reported as successful by carrying an explicit acceptance flag through the diff flow. Fixes #12269.

- **Bug Fixes**
  - Added `accepted` to `ApplyState` and propagated it through VS Code diff (`processDiff`, vertical handler/manager) to webview updates.
  - Updated GUI session state to store `accepted`; only mark a closed edit as successful when it was accepted.
  - Adjusted `handleApplyStateUpdate` to require `accepted` (and a non-canceled tool call) before logging success or continuing the flow.
  - Added a regression test to ensure rejected closed edits are not treated as successful.

<sup>Written for commit d395e728fe51105176ed129445d6b597c948c3cb. Summary will update on new commits.</sup>

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -1496,6 +1496,7 @@ export interface ApplyState {
   fileContent?: string;
   originalFileContent?: string;
   toolCallId?: string;
+  accepted?: boolean;
   autoFormattingDiff?: string;
 }
 
--- a/extensions/vscode/src/diff/processDiff.ts
+++ b/extensions/vscode/src/diff/processDiff.ts
@@ -90,6 +90,7 @@ export async function processDiff(
       status: "closed",
       numDiffs: 0,
       toolCallId,
+      accepted: action === "accept",
       autoFormattingDiff, // Include autoformatting diff
     });
   } else {
--- a/extensions/vscode/src/diff/vertical/handler.ts
+++ b/extensions/vscode/src/diff/vertical/handler.ts
@@ -20,6 +20,7 @@ export interface VerticalDiffHandlerOptions {
     status?: ApplyState["status"],
     numDiffs?: ApplyState["numDiffs"],
     fileContent?: ApplyState["fileContent"],
+    accepted?: ApplyState["accepted"],
   ) => void;
   streamId?: string;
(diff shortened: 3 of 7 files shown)


danyalahmed1995:

Issue #12410, opened 2026-05-14
Hardcoded legacy Windows PowerShell (powershell.exe) causes runtime failures and degraded UX


Charlweed:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I'm not able to find a related conversation on [GitHub discussions](https://github.com/continuedev/continue/discussions) that reports the same bug
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:Windows
- Continue version:1.545
- IDE version:
- Model:
- config:
  

  
  OR link to agent in Continue hub:
```

### Problem Statement
The codebase currently contains multiple hardcoded invocations of powershell.exe or the legacy powershell command. These specifically target Windows PowerShell 5.1, which is legacy software that no longer receives feature updates.

Modern PowerShell development, including scripts often generated or expected by modern developer tools, focuses heavily on PowerShell Core (pwsh) 6+. PowerShell Core is cross-platform and contains significant language features, performance improvements, and syntax changes not present in version 5.1.
### 
Impact & Friction

1.     Unpredictable and Silent Failures: Running modern PowerShell Core scripts in legacy PowerShell will frequently fail. Because PowerShell 5.1 lacks newer language features (e.g., ternary operators, pipeline chain operators like && and ||, or modern JSON parsing), the execution will fail in ways that are difficult to diagnose. The errors often present as obscure syntax or parsing errors within a sub-shell, obscuring the root cause (a version mismatch).
2.     Uncorrectable by the End User: Because the powershell invocation is hardcoded deep within the application logic (such as in clipboard utilities or terminal command execution), this issue is entirely uncorrectable by the end user. Even if a user has explicitly installed PowerShell Core and set it as their default terminal, the application bypasses this configuration and forces the legacy binary.
3.     Cross-Platform Inconsistency: Hardcoding powershell.exe makes the execution logic strictly Windows-centric. While non-Windows platforms generally default to pwsh or bash, having fragmented fallback paths leads to inconsistent behaviors and duplicated logic across different extensions and core utilities.

Proposed Solution

Implement a centralized, dynamic detection utility that:

[...]

PR #12411, opened 2026-05-14
Fix/hardcoded legacy powershell

Charlweed:
## Description

**Centralize and Modernize PowerShell Command Detection**
Resolves issue #12410 

## Overview
This PR eliminates hardcoded dependencies on legacy Windows PowerShell (`powershell.exe`) across the codebase and introduces a dynamic, centralized detection mechanism. By preferring modern PowerShell Core (`pwsh`), we resolve critical runtime failures caused by executing modern scripts in legacy environments, while maintaining backward compatibility.

### 1. Centralized Synchronous Detection
A new utility `getPowerShellCommand()` has been added to `core/util/shell.ts`.
- **Why Synchronous (`spawnSync`)?** Several call sites, notably `core/tools/definitions/runTerminalCommand.ts`, invoke the shell detection during module initialization to populate static metadata and configuration. A synchronous implementation was mandatory to prevent `[object Promise]` from bleeding into module-level constants.
- **Performance Tradeoff:** While synchronous process spawning blocks the event loop, this is an explicitly accepted tradeoff. The detection result is cached at the module level for the process lifetime. In the context of a CLI or local developer tool (which is already I/O bound), a one-time synchronous probe does not meaningfully degrade the performance profile.

### 2. Architectural Boundaries & Public API
- **Module Isolation:** The detection logic utilizing the Node.js `child_process` module is strictly isolated in `core/util/shell.ts`. This prevents Node-specific dependencies from being accidentally pulled into non-Node environments (like browser extensions or WASM contexts) that might import from the general utility barrel.
- **Enforcing the Public API:** Instead of allowing consumers (like the CLI extensions) to perform deep internal imports directly from `core/util/shell.js`, the function is re-exported via `core/util/index.ts`. This enforces strict encapsulation and provides a stable import path through the `core` package's public API.

### 3. Robust Test Infrastructure
- **Broad Vitest Aliasing:** Updated `extensions/cli/vitest.config.ts` with a broad `core` alias mapping to the source directory (`../../core/src`). This correctly mirrors the `tsconfig.json` path resolution, allowing tests to run directly against the source files via the public API without requiring tightly coupled, file-specific aliases or a prior build step.
- **Interface Mocking:** Tests now mock the `core/util/index.js` public API rather than intercepting internal `execAsync` or `child_process` calls. This makes the tests resilient to internal implementation changes and focuses verification strictly on the logic's boundaries.

## AI Code Review

## Checklist

## Verification Results
- Confirmed `pwsh` is correctly preferred when available on Windows.
- Confirmed safe fallback to `powershell` on legacy Windows environments.
- Verified zero module resolution errors during testing via the new public API alias.
- All CLI tests, including refactored clipboard utility tests, pass successfully.

---
## Summary by cubic
Centralizes PowerShell detection and switches Windows tooling to prefer PowerShell Core (`pwsh`) with a safe fallback to legacy `powershell`. Fixes Windows runtime issues in terminal, TTS, and clipboard flows by routing all calls through one API.

- **Bug Fixes**
  - Prefer `pwsh` on Windows; fall back to `powershell` when unavailable in `runTerminalCommand`, TTS, and clipboard utilities.
  - Improve Windows clipboard reliability using `System.Windows.Forms` APIs and `-STA` for image checks and saves.

- **Refactors**
  - Add `getPowerShellCommand()` in `core/util/shell.ts` (sync, cached) and re-export via `core/util/index.ts`.
  - Replace hardcoded `powershell.exe` across `core` and `extensions/cli` with `getPowerShellCommand()`.
  - Update `extensions/cli` `vitest.config.ts` to alias `core` to source and adjust tests to mock the `core/util/index.js` public API.

<sup>Written for commit 5bf7d3ef9887e6051bd8df1c8c9d87e561a45e6d. Summary will update on new commits.</sup>


Charlweed:
Friendly bump on this PR for modern PowerShell Core support.

Issue #5996, opened 2025-06-05
Running terminal commands in Agent mode results in an error.


KNChiu:
### Before submitting your bug report

- [ ] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Win11
- Continue version: v1.0.11
- IDE version: 1.95.3
- Model: Claude Sonnet 4
- config:
  
name: Local Assistant
version: 1.0.0
schema: v1
models:
  - name: AWS Claude 4 Sonnet
    provider: bedrock
    model: us.anthropic.claude-sonnet-4-20250514-v1:0
    env:
      region: us-east-1
      profile: bedrock
    capabilities:
      - tool_use
    roles:
      - chat
      - edit
      - apply

context:
  - provider: codebase
  - provider: file
  - provider: currentFile
  - provider: code
  - provider: diff
  - provider: web
    params:
      n: 5
  - provider: terminal
  - provider: problems
  - provider: folder
  - provider: codebase
  


  OR link to assistant in Continue hub:
```

### Log output

```Shell

```

saimadib:
@KNChiu I am not able to reproduce the issue. It is working properly when I ask "in bash run python --version".

![Image](https://github.com/user-attachments/assets/43e60e78-b788-4fc0-9df0-7d5a820f5f0b)

Patrick-Erichsen:
@KNChiu is this still happening for you? I think we might have shipped a fix

KNChiu:
I tried it but still got the error. I want to ask if it is because of using wsl ?

![Image](https://github.com/user-attachments/assets/74fcaed2-1e53-4d62-9b80-459d077474bc)

Patrick-Erichsen:
WSL feels very plausible. I'm going to update the issue title accordingly and see if other folks have +1s

ezeveliz:
I've made the following tests:

- Open a project in WSL and tell the agent to run a command, it returns de the following output: `run_terminal_command failed with the message: The URL must be of scheme file`
- Run a command directly on the terminal within WSL, it executes just fine
- Open a project in windows and tell the agent to run a command in a windows folder it executes just fine

[...]

PR #12442, opened 2026-05-19
feat: Add Windows WSL [Ubuntu] Agent Shell Support for Intellij -- Code and Generic Terminal/Bash Execution

GabeChurch:
Add support for Intellij on Windows Agents using Windows Subsystem For Linux (WSL) and add user facing test(s) supporting multiple Intellij version ranges

### The Problem
The current IntelliJ continue plugin fails to support agent code execution -- happening via the **runTerminalCommand**  -- when opening a notebook in the Windows Subsystem for Linux (Ubuntu) filesystem. This prevents agents from doing almost anything useful when using IntelliJ on WSL. IntelliJ itself supports WSL based projects, and even changes the terminal to use Ubuntu instead of windows which is incredibly useful for engineers who preferer require multiple isolated environments, use separate WSL backends to manage dependencies, or even want a separate isolated development environment to run agents on. The current behavior for agents used in WSL notebooks is to default to powershell commands, which means agents cannot execute code in the same isolated environment as the user and code -- making them quite less useful.  

### The Solution
* Modified the **resolveWorkingDirectories** in the core tools *"runTerminalCommand.ts"* to add a new cwd and shell check that ONLY triggers on IntelliJ Notebooks that are opened from the Windows Subsystem For Linux. A  lightweight file path check to correctly identify the desired WSL environment and pass it through is triggered before getShellCommand logic 
* Modified the **getShellCommand** in the core tools "runTerminalCommand.ts" to identify the passthrough cwd when it is WSL, and even dynamically parse out the correct or custom Kernel to use (in the case of multiple separate WSL backends with differing names). This can only affect windows IntelliJ users because vscode uses a separate backend for WSL (that is still correctly identified as a remote server), a two part fix was required to maintain the existing plugin architecture.

#### User Info
 If you stumbled across this as an IntelliJ user, you don't need to set anything for this change to work for you. To debug any issues with the continue plugin on WSL with Intellij yourself I suggest launching like this from a windows command prompt -- this forces legacy "wsl$" instead of default "wsl.localhost" in the path which may resolve some bugs you could experience. 
 * Change your jetbrains path to match your install.
 * Change *Ubuntu* to your wsl linux distribution (note it must support bash) 
 
 ```console
 "C:\Program Files\JetBrains\IntelliJ IDEA 2026.2\bin\idea64.exe" \\wsl$\Ubuntu\home\<yourusername>\<YourIntellijProject>
 ```
 
## AI Code Review

## Checklist

This example shows the new tests being used to run the new Ubuntu agent execution feature as well as the existing Powershell agent and each respectively running bash and powershell code depending on whether they are in standard windows or WSL Ubuntu. 

https://github.com/user-attachments/assets/139862cb-2864-4d79-a7ed-6c6fcd8f1d53

## Tests

I created a new "Run Continue WSL" configuration/test, added for both CE and Ultimate distributions of IntelliJ, that allows any standard windows user to launch the existing *manual-testing-sandbox* testing Intellij sample project and select a WSL backend in addition to the standard test (which can be any WSL custom name in the "Run Continue WSL (use TCP).run.xml file -- defaults to "Ubuntu", the base/standard WSL distribution). **This can be used as a future test to validate any WSL related functionality in Intellij.**

image

### Detailed Test Breakdown
 I identified a slightly frustrating behavior with the current gradle.properties Intellij 2024.1, where it cannot directly launch notebooks using WSL backend(s) with files located in standard windows C path, which makes testing more painful. To solve for this I added a gradle.properties check that will automatically do this "optimal path" in 2025 and newer major versions (same physical test files used to launch). With the current 2024 major version test I solved for this by copying the test files to a WSL "tmp" location at test launch (before the actual launch happens). I also added and validated a "postLaunch/shutdown" hook that cleans these temp files no matter how shutdown is called. In this way nothing appears no different to users testing -- although it is recommended to them to setup the continue config.yaml file in standard windows before running the WSL test. 
 
## Dependencies 

If you want to run this test you need to have a windows subsystem for linux installed and enabled + IntelliJ. 

Suggested is to install Ubuntu from the Microsoft Store which is the default
image

## Configuration [Only required for TESTING]

If you use something other than "Vanilla" Ubuntu in your Windows Subsystem for Linux, you can change the test to use it instead of the default. 

image

The change is an environment variable matching your WSL distribution name. 
image

If you are unsure you can find it in the file browser under "Linux"

## Related 
* https://github.com/continuedev/continue/issues/5996

---
---

### DIFF
--- a/core/tools/implementations/runTerminalCommand.ts
+++ b/core/tools/implementations/runTerminalCommand.ts
@@ -21,8 +21,42 @@ function getDecodedOutput(data: Buffer): string {
   } else {
     return data.toString();
   }
-} // Simple helper function to use login shell on Unix/macOS and PowerShell on Windows
-function getShellCommand(command: string): { shell: string; args: string[] } {
+}
+
+// Simple helper function to use login shell on Unix/macOS and PowerShell on Windows
+// Detects WSL paths and uses the appropriate WSL shell
+function getShellCommand(
+  command: string,
+  cwd?: string,
+): { shell: string; args: string[] } {
+  // Check if the working directory is in WSL on Intellij
+  const isWslPath =
+    cwd && (cwd.includes("\\wsl.localhost\\") || cwd.includes("\\wsl$\\"));
+
+  if (isWslPath) {
+    // Extract the WSL distribution name from the path
+    // Path format: \\wsl.localhost\UbuntuAny\home\user\project
+    // or: \\wsl$\UbuntuAny\home\user\project
+    let distroName = "Ubuntu"; // Default fallback
+    try {
+      const wslPathSegments = cwd.split(/\\+/);
+      // Find the segment after wsl.localhost or wsl$
+      const wslIndex = wslPathSegments.findIndex(
+        (seg) => seg.includes("wsl.localhost") || seg.includes("wsl$"),
+      );
+      if (wslIndex !== -1 && wslIndex + 1 < wslPathSegments.length) {
+        distroName = wslPathSegments[wslIndex + 1];
+      }
+    } catch {
+      // If parsing fails, use default Ubuntu
+    }
+    // Use wsl.exe to run commands in the specific distribution
+    return {
+      shell: "wsl.exe",
+      args: ["-d", distroName, "-e", "bash", "-c", command],
+    };
+  }
+
   if (process.platform === "win32") {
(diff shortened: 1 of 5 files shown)


GabeChurch:

GabeChurch:
recheck

Issue #11818, opened 2026-03-25
Generic "Connection error." across multiple providers (tracking issue)


RomneyDa:
## Summary

A large number of issues report a generic **"Connection error."** message with no status code and no actionable detail. This affects multiple providers — most commonly **LM Studio** and **OpenAI-compatible local servers**, but also DeepSeek, OpenRouter, Mistral, and others.

Over **35 issues** have been filed with this same pattern. They were previously bulk-closed as duplicates of #9017, but that issue is specifically about HTTPS/TLS certificate handling, while most of these reports involve plain HTTP connections.

This tracking issue consolidates them and documents root causes and workarounds.

---

## Affected providers (from reported issues)

| Provider | Count | Notes |
|----------|-------|-------|
| `openai` (OpenAI-compat, often LM Studio) | 15 | Most are local LM Studio servers configured via the `openai` provider |
| `lmstudio` | 12 | Native LM Studio provider |
| `deepseek` | 5 | |
| `openrouter` | 4 | |
| `mistral` | 1 | |

### 1. No LM Studio-specific error handling in `BaseLLM.fetch()`

**File:** `core/llm/index.ts` (lines 488–540)

The fetch error handler has specific, user-friendly messages for **Ollama** (`127.0.0.1:11434`) and **Lemonade** (`localhost:8000`), but **no equivalent for LM Studio** (`localhost:1234`). When LM Studio is unreachable, users see a raw, unhelpful error instead of something like:

> *"Unable to connect to LM Studio at http://localhost:1234. LM Studio may not be running, or the port may be incorrect."*

This is the single most impactful fix — it would immediately make these errors actionable.

### 2. `localhost` vs `127.0.0.1` (IPv6 resolution)

Node.js 18+ (via undici) can resolve `localhost` to `::1` (IPv6) when the LLM server only binds to `127.0.0.1` (IPv4). The LM Studio provider defaults `apiBase` to `http://localhost:1234/v1/`, which can silently fail on systems where IPv6 is preferred.

**Workaround:** Use `"apiBase": "http://127.0.0.1:1234/v1/"` explicitly.

Ref: [nodejs/undici#2990](https://github.com/nodejs/undici/issues/2990)

### 3. macOS Local Network permissions

On macOS, VS Code and JetBrains IDEs need explicit **Local Network** permission. Without it, requests to localhost or LAN IPs fail silently.

**Fix:** System Settings → Privacy & Security → Local Network → enable for the IDE.

Ref: #1145, #8849

[...]

PR #12451, opened 2026-05-21
fix: convert requestOptions.timeout to milliseconds for OpenAI SDK

### LINKED ISSUE #12450: OpenAI adapter passes timeout to SDK without unit conversion (seconds vs milliseconds)

```typescript
```

ScrewTSW:
## Description

`requestOptions.timeout` is configured in seconds by users, but the OpenAI JS SDK constructor expects milliseconds. The value was passed through without conversion, so `timeout: 300` (5 minutes) became a 300ms timeout.

One-line fix: multiply by 1000 before passing to the SDK.

Fixes #12450
Likely root cause for many reports in #11818

## Checklist

- [ ] I've updated docs if needed
- [ ] I've updated tests if needed

## Tests

Tested with local llama.cpp models behind an orchestrator (`timeout: 300` in config). Before: immediate "Connection error". After: requests complete normally with model load + prompt eval taking 30-60s.

---
## Summary by cubic
Convert `requestOptions.timeout` (seconds) to milliseconds before passing to the `openai` client. Fixes unintended sub-second timeouts and restores expected behavior for long-running requests.

- **Bug Fixes**
  - Multiply `requestOptions.timeout` by 1000 in the `OpenAIApi` constructor so `timeout: 300` becomes 300000 ms, not 300 ms.

<sup>Written for commit 4a456a45c6e31371471f5b1eefc663a5c4dcaa05. Summary will update on new commits. <a href="https://cubic.dev/pr/continuedev/continue/pull/12451?utm_source=github">Review in cubic</a></sup>


ScrewTSW:

PR #12453, opened 2026-05-21
fix(openai-adapters): convert request timeout to milliseconds

### LINKED ISSUE #12450: OpenAI adapter passes timeout to SDK without unit conversion (seconds vs milliseconds)

kgarg2468:
## Summary
- Convert `requestOptions.timeout` from Continue's documented seconds to the OpenAI SDK's milliseconds before constructing the client.
- Preserve omitted timeouts so the SDK default still applies.
- Preserve explicit `0` instead of treating it as unset.

Fixes #12450.

## Testing
- `npm test -- --run src/apis/OpenAI.test.ts src/test/openai-adapter.vitest.ts`
- `npm run build -- --pretty false`

---
## Summary by cubic
Convert `requestOptions.timeout` in `openai-adapters` from seconds to milliseconds for the `openai` SDK to fix incorrect request timeouts. Preserves the SDK default when unset and respects an explicit `0`.

- **Bug Fixes**
  - Convert seconds to milliseconds before constructing the `openai` client.
  - Omit the timeout when not provided so the SDK default applies.
  - Respect `0` (no timeout) and add unit tests for conversion, omitted, and zero cases.

<sup>Written for commit a7c5fef70df0602937a96f081b3913261238a43d. Summary will update on new commits. <a href="https://cubic.dev/pr/continuedev/continue/pull/12453?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/openai-adapters/src/apis/OpenAI.test.ts
+++ b/packages/openai-adapters/src/apis/OpenAI.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, test } from "vitest";
+import { OpenAIApi } from "./OpenAI.js";
+
+describe("OpenAIApi", () => {
+  test("converts Continue request timeout seconds to OpenAI SDK milliseconds", () => {
+    const api = new OpenAIApi({
+      provider: "openai",
+      apiKey: "test-api-key",
+      requestOptions: { timeout: 300 },
+    });
+
+    expect(api.openai.timeout).toBe(300_000);
+  });
+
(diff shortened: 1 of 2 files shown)


ScrewTSW:
bot?

kgarg2468:
Not a bot account, but I did use an AI coding assistant to prepare this small timeout-units fix.

I also checked the cancelled Windows VSIX run: the VSIX package and artifact upload both completed, and the cancellation happened later during post-job cache cleanup, so I do not see a code change needed for that job. Happy to adjust anything else reviewers want.

PR #12517, opened 2026-05-29
fix(openai-adapters): read Gemini embeddings from correct response field

Pluviobyte:
## Description

The Gemini embeddings adapter reads the wrong field from the `:batchEmbedContents` response. In `packages/openai-adapters/src/apis/Gemini.ts`, `embed()` did:

```ts
data: data.batchEmbedContents.map((embedding: any) => embedding.values),
```

but the Gemini API returns the vectors under `embeddings`, not `batchEmbedContents`. That field is always `undefined`, so **every** embeddings request through the Gemini adapter throws:

```
TypeError: Cannot read properties of undefined (reading 'map')
```

This breaks embeddings (e.g. `@codebase` / `@docs` indexing) for anyone configuring Gemini as their `embed` model. The path is reachable: `core/llm/llms/Gemini.ts` sets `useOpenAIAdapterFor = ["chat", "embed", ...]`, so embeddings go through `GeminiApi.embed()`.

The fix mirrors the **core** implementation, which already reads the correct field:

```ts
// core/llm/llms/Gemini.ts
return data.embeddings.map((embedding: any) => embedding.values);
```

I also removed the `usage` block, which read two more non-existent fields (`data.total_tokens` / `data.prompt_tokens`). The batchEmbedContents response carries no token usage, and `embedding()` already defaults usage to zeros — matching core, which sets no usage.

## Checklist

N/A — backend crash fix; covered by the regression test below.

## Tests

Added an `embed` test in `packages/openai-adapters/src/apis/Gemini.test.ts` that mocks the real `:batchEmbedContents` response shape (`{ embeddings: [{ values }] }`) and asserts the parsed vectors. Verified it **fails on the old code** with `Cannot read properties of undefined (reading 'map')` and **passes with the fix**. `npx tsc --noEmit` and `prettier --check` both pass locally.

---
## Summary by cubic
Fixes a crash in the Gemini embeddings adapter by reading vectors from the correct response field. Restores embeddings for Gemini `embed` models and aligns adapter behavior with core.

- **Bug Fixes**
  - `packages/openai-adapters/src/apis/Gemini.ts`: `embed()` now reads `data.embeddings` instead of `data.batchEmbedContents`.
  - Removed the invalid `usage` block (nonexistent token fields).

- **Tests**
  - Added a regression test in `packages/openai-adapters/src/apis/Gemini.test.ts` that mocks the `:batchEmbedContents` response and asserts parsed vectors.

<sup>Written for commit e5f15485871bdbedf50369b20c3be3909983bcaa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/continuedev/continue/pull/12517?utm_source=github">Review in cubic</a>

### DIFF
--- a/packages/openai-adapters/src/apis/Gemini.test.ts
+++ b/packages/openai-adapters/src/apis/Gemini.test.ts
@@ -1,6 +1,15 @@
-import { describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 import { GeminiApi } from "./Gemini.js";
+
+// Mock the fetch package so the embeddings test can stub the HTTP response.
+vi.mock("@continuedev/fetch", async () => {
+  const actual = await vi.importActual("@continuedev/fetch");
+  return {
+    ...actual,
+    fetchwithRequestOptions: vi.fn(),
+  };
+});
 
 describe("GeminiApi", () => {
   const api = new GeminiApi({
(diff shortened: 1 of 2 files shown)


Pluviobyte:

Pluviobyte:
I checked the remaining failing statuses. The GitHub Actions checks are green except the old macOS VSIX job showing cancelled, and the Continuous AI failure target I opened returns `Cannot GET /api/agent-check-summary/...`, so there is no actionable failure detail exposed from the external check.

Issue #11872, opened 2026-03-26
fix: "Invalid URL" error across multiple providers


RomneyDa:
## Summary

Several users are reporting an `Invalid URL` error with no additional context, across different providers and configurations. This is a tracking issue to consolidate the reports and fix the underlying causes.

**Related issues:** #6717, #9630, #9864, #10317, #10800, #11049, #11249, #11488

## Are you hitting this?

The most common reason is an `apiBase` value that's missing the protocol. For example:

```yaml
# This will cause "Invalid URL"
apiBase: localhost:11434

# This works
apiBase: http://localhost:11434
```

If you're using a provider like Ollama locally and haven't set `apiBase` at all, it should default correctly — but if you *have* set it, make sure it includes `http://` or `https://`.

If you're using models from the Continue hub (registry packages with `${{ secrets.* }}`), the error may occur if a secret failed to resolve, leaving a template string where a URL is expected.

## What the error looks like

```
Error handling model response
There was an error handling the response from <model name>.

Invalid URL
```

No further detail is provided, which makes it hard to diagnose.

RomneyDa:
## Proposed fixes

Three distinct causes, all fixable:

### 1. No validation or normalization of `apiBase`

`apiBase` is accepted as a bare `z.string().optional()` in the schema and passed straight to `new URL()` at 60+ call sites. A value like `localhost:11434` (missing protocol) silently passes config parsing and only blows up at request time with a useless error.

**Fix:** Validate and normalize `apiBase` in the `BaseLLM` constructor (`core/llm/index.ts`, around line 277). If it's a non-empty string without `://`, prepend `http://`. If it's still not a valid URL, throw with a message that names the provider and the bad value. Optionally also add `.url()` validation in the config schema (`packages/config-yaml/src/schemas/models.ts:182`).

### 2. ContinueProxy inherits OpenAI but drops its `apiBase` default

`ContinueProxy` (`core/llm/llms/stubs/ContinueProxy.ts:55`) overrides `static defaultOptions` with `{ useLegacyCompletionsEndpoint: false }`, which fully replaces the parent `OpenAI.defaultOptions` (including its `apiBase`). The `controlPlaneProxyInfo` setter patches `apiBase` later, but inherited code like `OpenAI.getMaxStopWords()` (`core/llm/llms/OpenAI.ts:256`) does `new URL(this.apiBase!)` — a non-null assertion on a potentially undefined value.

[...]

PR #12518, opened 2026-05-29
fix(llm): guard getMaxStopWords against undefined apiBase

Pluviobyte:
## Description

`OpenAI.getMaxStopWords()` builds `new URL(this.apiBase!)` at the very top of the method. That non-null assertion is unsafe: `apiBase` can legitimately be `undefined`. The clearest case is `continue-proxy`, whose `static defaultOptions` (`core/llm/llms/stubs/ContinueProxy.ts`) replace the parent `OpenAI.defaultOptions` and do **not** carry an `apiBase`. When it is undefined, `new URL(undefined)` throws:

```
TypeError: Invalid URL
```

…which crashes stop-word trimming (`getMaxStopWords()` is called when building the request args for every completion).

This is one of the root causes called out by @RomneyDa in #11872, who suggested exactly this fix:

> guard `getMaxStopWords()` to handle undefined `apiBase` gracefully (return a safe default like `Infinity`)

### Fix

- Move the `new URL(...)` parse **below** the `maxStopWords` short-circuit, so it is only built when actually needed (previously it was parsed even when `maxStopWords` was set and the URL went unused).
- Return `Infinity` when `apiBase` is absent — the same safe default already used for unknown hosts — and drop the `!` assertion.

## Checklist

N/A — backend crash fix; covered by the regression tests below.

## Tests

Added a `getMaxStopWords` suite in `core/llm/llms/OpenAI.vitest.ts`:
- returns `Infinity` (instead of throwing) when `apiBase` is undefined
- still honors an explicit `maxStopWords` when `apiBase` is undefined

Verified both **fail on the old code** with `TypeError: Invalid URL` and **pass with the fix**. `npx tsc -p ./ --noEmit` reports 0 errors, `prettier --check` passes, and the full `OpenAI.vitest.ts` suite is green (10/10) locally.

---
## Summary by cubic
Fixes a crash in `OpenAI.getMaxStopWords()` when `apiBase` is undefined by returning `Infinity` and only parsing the URL when needed. Prevents "TypeError: Invalid URL" during stop-word trimming.

- **Bug Fixes**
  - Return `Infinity` when `apiBase` is missing; remove the non-null assertion.
  - Parse the URL only after honoring `maxStopWords`.
  - Add regression tests for undefined `apiBase` and explicit `maxStopWords`.

<sup>Written for commit eec87837403df7efc24460d4b11646c8a4278aea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/continuedev/continue/pull/12518?utm_source=github">Review in cubic</a>

### DIFF
--- a/core/llm/llms/OpenAI.ts
+++ b/core/llm/llms/OpenAI.ts
@@ -253,11 +253,18 @@ class OpenAI extends BaseLLM {
   }
 
   protected getMaxStopWords(): number {
-    const url = new URL(this.apiBase!);
-
     if (this.maxStopWords !== undefined) {
       return this.maxStopWords;
-    } else if (url.host === "api.deepseek.com") {
+    }
+
+    // apiBase can legitimately be undefined (e.g. continue-proxy, whose
+    // defaultOptions don't carry one), so guard against `new URL(undefined)`.
+    if (!this.apiBase) {
+      return Infinity;
+    }
+
+    const url = new URL(this.apiBase);
+    if (url.host === "api.deepseek.com") {
       return 16;
     } else if (
       url.port === "1337" ||
--- a/core/llm/llms/OpenAI.vitest.ts
+++ b/core/llm/llms/OpenAI.vitest.ts
@@ -424,4 +424,31 @@ describe("OpenAI", () => {
       },
     });
   });
+
+  describe("getMaxStopWords", () => {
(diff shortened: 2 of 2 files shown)


Pluviobyte:
I checked the remaining failing statuses. The GitHub Actions checks are green, but the Continuous AI failure target returns `Cannot GET /api/agent-check-summary/...`, so there is no actionable failure detail exposed from the external check.

Issue #11954, opened 2026-03-28
Local Autocomplete Broken!


openSourcerer9000:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I'm not able to find a related conversation on [GitHub discussions](https://github.com/continuedev/continue/discussions) that reports the same bug
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: mac
- Continue version: 
1.2.22
- IDE version:
vscode 
Version: 1.113.0
- config:
  
name: Local Config
version: 1.0.0
schema: v1
models:
  - name: qwen3.5-35b-a3b
    provider: lmstudio
    model: qwen/qwen3.5-35b-a3b
    apiBase: http://localhost:1234/v1/
    roles:
      - autocomplete
    requestOptions:
      timeout: 1200  # Set to 1200 seconds (20 minutes)
  
  OR link to agent in Continue hub:
```

### Description

LM studio shows the following log every time i change code to trigger autocomplete. It was painful even getting this far with no docs on how to set up "The leading open-source AI code agent", only links to cloud services and telemetry options to disable.
```
2026-03-28 15:26:28  [INFO]
 [LM STUDIO SERVER] Running completion on text:  <fim_prefix># Path: soult...age_dir'])}",<fim_middle>
2026-03-28 15:26:28  [INFO]
 Streaming response..
2026-03-28 15:26:28  [INFO]
 [LM STUDIO SERVER] Processing...
2026-03-28 15:26:29  [INFO]
 [LM STUDIO SERVER] First token generated. Continuing to stream response..
2026-03-28 15:26:29  [INFO]
 [LM STUDIO SERVER] Client disconnected. Stopping generation... (If the model is busy processing the prompt, it will finish first.)
2026-03-28 15:26:29  [INFO]
 [LM STUDIO SERVER] Finished streaming response
```

I'm wondering if this is related to this? Your extension has been broken for nearly a year now?
https://github.com/continuedev/continue/issues/5079

I just want autocomplete, if anyone has a recommendation for an extension that does work, please let me know.

### To reproduce

[...]

PR #12563, opened 2026-06-05
fix(openai-adapters): convert requestOptions.timeout from seconds to milliseconds

### LINKED ISSUE #12450: OpenAI adapter passes timeout to SDK without unit conversion (seconds vs milliseconds)

chatman-media:
Closes #12450.

`requestOptions.timeout` is specified in **seconds** throughout Continue — `getAgentOptions` in `@continuedev/fetch` does `(requestOptions?.timeout ?? TIMEOUT) * 1000; // measured in ms`. But `packages/openai-adapters/src/apis/OpenAI.ts` passed the value straight to the OpenAI SDK constructor, which expects **milliseconds**.

So a user setting `timeout: 300` (intending 5 minutes) got a **300 ms** timeout — ~2000× too short — causing immediate "Connection error" failures, especially with local models that need time to load and process prompts. This matches the reports referenced in #11818 / #11954.

**Fix:** multiply by 1000 to convert seconds → ms, consistent with `getAgentOptions`. When `timeout` is unset, `undefined` is passed so the OpenAI SDK keeps its own default (600000 ms).

---
## Summary by cubic
Fixes the OpenAI adapter timeout by converting `requestOptions.timeout` from seconds to milliseconds before passing it to the SDK. Prevents premature connection errors (300s no longer becomes 300ms); 0/undefined are treated as unset so the SDK default applies.

- **Bug Fixes**
  - Added `timeoutSecondsToMs` utility with unit tests and used it when setting the SDK `timeout`.

<sup>Written for commit c57f33ad1f392db0ae6e4803976aa60ce4d8b526. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/continuedev/continue/pull/12563?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/openai-adapters/src/apis/OpenAI.timeout.test.ts
+++ b/packages/openai-adapters/src/apis/OpenAI.timeout.test.ts
@@ -1,0 +1,17 @@
+import { describe, expect, it } from "vitest";
+
+import { timeoutSecondsToMs } from "./OpenAI.js";
+
+describe("timeoutSecondsToMs", () => {
+  it("converts a seconds timeout to milliseconds", () => {
+    expect(timeoutSecondsToMs(300)).toBe(300_000);
+  });
+
+  it("returns undefined when no timeout is provided", () => {
+    expect(timeoutSecondsToMs(undefined)).toBeUndefined();
+  });
+
+  it("treats a 0 timeout as unset so the SDK default applies", () => {
+    expect(timeoutSecondsToMs(0)).toBeUndefined();
+  });
+});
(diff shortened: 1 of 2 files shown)


chatman-media:
I have read the CLA Document and I hereby sign the CLA.

PR #12586, opened 2026-06-10
fix(cli): index .cc, .cxx, and .c++ files in @file context provider

### LINKED ISSUE #10239: CLI: @file context provider does not find .cc files.
Path: /guides/cli

When using the Continue CLI, the @file context provider does not find files with the .cc extension. The same files are found without issue when using the VS Code extension with the same Continue configuration.

Steps to reproduce:

Have a workspace containing C++ source files with .cc extension
In the Continue CLI, type @path/to/file.cc
File is not found

Expected behavior: .cc files should appear in @file results, same as in the VS Code extension.
Workaround confirmation: Copying the same file with a .cpp extension makes it immediately discoverable via @file in the CLI.
Environment:

OS: Linux
Continue CLI v1.5.43

VS Code extension works correctly with the same files and configuration

Continue CLI works correctly with .cpp files

Notes: .cc is a standard C++ file extension, widely used in Unix/Linux environments and Google-style codebases. The CLI's file discovery likely has a hardcoded extension allowlist that is missing .cc (and possibly other valid C++ extensions like .cxx, .C, .c++).

rodboev:
Fixes #10239

## Summary

The CLI's `@file` context provider fails to find source files with `.cc`, `.cxx`, or `.c++` extensions while `.cpp` files are indexed correctly. This affects Google-style C++ projects and Unix/Linux codebases where `.cc` is the canonical extension.

## Root cause

`extensions/cli/src/services/FileIndexService.ts:32-72` filters every crawled file through a hardcoded `supportedExtensions` array that includes `"cpp"` but omits the other standard C++ extensions, so `.cc`/`.cxx`/`.c++` files are silently excluded from the index.

## Changes

- `extensions/cli/src/services/FileIndexService.ts`: added `"cc"`, `"cxx"`, `"c++"` to `supportedExtensions`, immediately after `"cpp"` (3 lines, no logic changes)
- `extensions/cli/src/util/filePatterns.ts`: added the same three extensions to the `FILE_PATTERNS` glob so the shared discovery pattern stays in sync with the indexer
- `extensions/cli/src/services/FileIndexService.test.ts`: new test creating files with all four C++ extensions and asserting each appears in `service.getFiles()`, with `.cpp` as the regression guard

## What this doesn't change

- The VS Code extension's file-type handling (separate code path)
- Directory-exclusion logic (`exclude` callback)
- Fuzzy-search / fzf ranking and special filename patterns (readme, license, etc.)

## Checklist

N/A — no UI change.

## Tests

- `cd extensions/cli && npx vitest run src/services/FileIndexService.test.ts` — 11/11 passing (10 existing + 1 new). Covers: indexing of `.cc`, `.cxx`, and `.c++` files end-to-end through `FileIndexService`, plus `.cpp` regression coverage.


rodboev:

PR #12727, opened 2026-06-17
feat: Add typosquat-target detection to terminal-security package-manager rule

### LINKED ISSUE #12573: terminal-security: package-manager rule flags installs but not typosquat targets (e.g. `npm install lodahs`)
## Summary

`@continuedev/terminal-security` already gates package installs behind permission via `isHighRiskPackageManager` (it recognises `npm`/`yarn`/`pnpm` `install`/`add`/`i`). That rule answers "is this an install?" but not "what is being installed?". A command like `npm install lodahs` (one edit from `lodash`) is treated identically to `npm install lodash`: both land at `allowedWithPermission`, with no signal that the target name is a likely typosquat. Runner forms such as `npx expres` aren't covered by the package-manager list at all.

Typosquatting is one of the most common npm supply-chain attack vectors, and an AI agent driving the terminal is exactly the actor most likely to fat-finger or hallucinate a package name. The evaluator is the right seam, because it is the choke point that decides whether a terminal command auto-runs.

## Gap (reproducible, offline)

- `npm install lodahs` — no distinction from a legitimate install; a reviewer sees no "this looks like a typosquat of lodash" signal.
- `npx expres` — `npx`/`bunx` aren't in the package-manager list, so the typosquat target gets no dedicated treatment.
- `pnpm dlx expres`, `yarn add lodahs`, `bun add lodahs` — same gap across managers and runners.

## Proposal

Add a sibling, pure/offline checker in the same seam that detects npm-ecosystem installs (and `npx`/`bunx`/`dlx`/`exec` runner targets) of a name exactly one edit away from a curated set of popular packages, and wire it into the existing `isHighRiskCommand` aggregator so a hit escalates the policy to `allowedWithPermission` (most-restrictive-wins). It never hard-denies; it only asks the human.

Design:
- Pure/offline/deterministic. No network, no filesystem, no new runtime dependencies.
- Gate: Optimal String Alignment (restricted Damerau-Levenshtein) distance == 1 **and** the matched popular name length > 4 (short names like `vue` produce too many false positives).
- Hardening to avoid false positives: sub-command splitting on `;`/`&&`/`||`/`|`/`&`/newlines, env-assignment + wrapper prefixes (`sudo`, `env FOO=bar`), path-prefix on the binary, leading global options before the subcommand, value-taking install flags (so `npm install --prefix axio lodash` does **not** flag `axio`, which is a `--prefix` value, not a package), quote stripping, and `@scope/pkg@version` normalization.

I have a working implementation + vitest suite ready (45 new cases including the `--prefix axio` false-positive guard; the package's full suite stays green at 269 passing; `tsc --noEmit` clean) and proposed the same check upstream in Goose: aaif-goose/goose#9642. Would the maintainers be open to it living next to the existing `isHighRisk*` checks in `evaluateTerminalCommandSecurity.ts`? Happy to open the PR.

mvanhorn:
## Summary
The @continuedev/terminal-security package already recognizes package-manager install commands (npm/yarn/pnpm install/add) via isHighRiskPackageManager and routes them to allowedWithPermission, but it only answers 'is this an install?' not 'what is being installed?'. A command like `npm install lodahs` (one edit away from `lodash`) is treated identically to a legitimate install, with no typosquat signal for the reviewer.

## Changes
Extend evaluateTerminalCommandSecurity.ts to parse the target package name(s) out of recognized install/runner commands and compare each against a small bundled allowlist of popular packages using an edit-distance (Levenshtein <= 1-2, excluding exact matches) heuristic. Add npx/bunx/pnpm dlx/yarn dlx runner forms to the set of commands whose first non-flag argument is treated as a package target. When a near-miss is detected, attach a structured typosquat warning to the existing evaluation result (new optional field on the result type in types.ts) and ensure such commands stay at allowedWithPermission rather than silently auto-running, surfacing the suspected canonical name. Keep the allowlist and distance logic self-contained so it remains offline and dependency-free, matching the existing rule style in this file.

Fixes #12573

---
## Summary by cubic
Add typosquat-target detection to `@continuedev/terminal-security` package-manager rule to flag suspicious install/runner targets and keep them permission-gated. Fixes #12573.

- **New Features**
  - Parse package targets from `npm|yarn|pnpm install/add/i` and runners `npx`, `bunx`, `pnpm dlx`, `yarn dlx`; check `-p/--package` (incl. `--package=foo`) as package targets and ignore non-package runner flags.
  - Detect near-miss names against a small bundled list of popular packages using bounded Levenshtein (<=2), skipping exact matches.
  - Attach `typosquat-target` warnings with `packageName`, `suspectedPackageName`, and `distance`; any warning forces at least `allowedWithPermission`.
  - Optional warnings via `evaluateTerminalCommandSecurity(..., { includeWarnings: true })` or `evaluateTerminalCommandSecurityDetailed`; tests cover exact hits, typos, scoped/versions, runners and `-p/--package` cases, and negatives.

<sup>Written for commit f2197a23b113f809717f30090a12ba28f30ca31a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/continuedev/continue/pull/12727?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/terminal-security/src/evaluateTerminalCommandSecurity.ts
+++ b/packages/terminal-security/src/evaluateTerminalCommandSecurity.ts
@@ -757,14 +757,17 @@ const INSTALL_FLAGS_WITH_VALUES = [
 
 const RUNNER_FLAGS_WITH_VALUES = [
   "-c",
-  "-p",
   "--cache",
-  "--package",
   "--prefix",
   "--registry",
   "--shell",
   "--userconfig",
 ];
+
+// Runner flags whose value names a package to fetch (e.g. `npx -p cowsay ...`).
+// Their values must be typosquat-checked rather than skipped, so they are kept
+// out of RUNNER_FLAGS_WITH_VALUES above.
+const RUNNER_PACKAGE_FLAGS = ["-p", "--package"];
 
 function isHighRiskPackageManager(
   baseCommand: string,
@@ -825,11 +828,40 @@ function getInstallPackageTargets(args: string[]): string[] {
 }
 
 function getRunnerPackageTargets(args: string[]): string[] {
+  const targets: string[] = [];
+  let sawPackageFlag = false;
+
   for (let i = 0; i < args.length; i++) {
     const arg = args[i];
 
     if (arg === "--") {
       break;
+    }
+
+    // `-p`/`--package` (and `--package=foo`) name a package to fetch, so their
+    // value is a typosquat candidate rather than something to skip over.
+    if (arg.includes("=") && RUNNER_PACKAGE_FLAGS.includes(arg.split("=")[0])) {
+      sawPackageFlag = true;
+      const packageName = normalizePackageTarget(
+        arg.slice(arg.indexOf("=") + 1),
+      );
+      if (packageName) {
+        targets.push(packageName);
+      }
+      continue;
+    }
+
+    if (RUNNER_PACKAGE_FLAGS.includes(arg)) {
+      sawPackageFlag = true;
+      const value = args[i + 1];
+      if (value !== undefined) {
+        i++;
+        const packageName = normalizePackageTarget(value);
+        if (packageName) {
+          targets.push(packageName);
+        }
+      }
+      continue;
     }
 
     if (shouldSkipFlagValue(arg, RUNNER_FLAGS_WITH_VALUES)) {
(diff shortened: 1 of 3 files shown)


mvanhorn:
Both open points here are stale, so a re-run should clear this.

cubic's P2 on `getRunnerPackageTargets`: already fixed. cubic reviewed 7082690; the fix landed the next day in f2197a2, which handles `-p`/`--package` before the generic skip-flag-value branch and captures the value as a typosquat candidate instead of stepping over it. Covered by the "should warn for runner package flags that look like typosquats" case for `npx -p expres cowsay`, `npx --package expres cowsay`, `npx --package=expres cowsay`, and `pnpm dlx --package expres cowsay`. 232/232 pass in `packages/terminal-security` locally.

On `jetbrains-tests`: the failure is `Autocomplete > testAutocomplete()`, a headless-IDE integration test. This PR touches only three files under `packages/terminal-security/` (a pure TS evaluator, its types, and its test), so there is no path from this diff to that test. The run is also from June 18.

Could someone re-run the checks? Happy to rebase onto current main first if that is easier.

PR #12920, opened 2026-06-30
fix(core): reload config after create_rule_block so the new rule applies immediately

houssemzaier:
## Description

When the model creates a rule with the `create_rule_block` tool, the rule file is written via `ide.writeFile()`, which doesn't trigger VS Code's workspace file-change events. So the new rule isn't picked up until the next manual config reload — the user has to reload (or restart) before the rule they just asked for actually applies.

After the `create_rule_block` tool runs **successfully**, this PR invalidates the `walkDir` cache and reloads the config, so the freshly-created rule takes effect immediately.

It's a small, targeted change in the tool-call handler (`core/core.ts`): only the `create_rule_block` tool triggers the reload, and only when it didn't error (`!result.errorMessage`).

We've been running this in a downstream fork (3,000+ users) and it's been stable.

## Tests

Manual: ask the assistant to create a rule (via the `create_rule_block` tool), then immediately rely on it — the rule now applies without a manual reload.

## Checklist

- [ ] Tests — n/a (a small behavioral fix in the tool-call handler; verified manually)
- [ ] Docs — n/a

---
## Summary by cubic
Hot-reloads the config after a successful `create_rule_block` so the new rule applies immediately without a manual reload. This invalidates the `walkDir` cache and reloads the config in the tool-call handler.

- **Bug Fixes**
  - Works around `ide.writeFile()` not emitting VS Code workspace events by reloading only for `create_rule_block` and only on success.

<sup>Written for commit a1cceef60ca18887596d7464bb7a9c4b51baf76f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/continuedev/continue/pull/12920?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -19,6 +19,7 @@ import { fetchModels } from "./llm/fetchModels";
 import Ollama from "./llm/llms/Ollama";
 import { EditAggregator } from "./nextEdit/context/aggregateEdits";
 import { createNewPromptFileV2 } from "./promptFiles/createNewPromptFile";
+import { BuiltInToolNames } from "./tools/builtIn";
 import { callTool } from "./tools/callTool";
 import { ChatDescriber } from "./util/chatDescriber";
 import { compactConversation } from "./util/conversationCompaction";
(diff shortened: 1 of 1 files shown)


itxaiohanglover (commented):
Good fix — reloading config after `create_rule_block` ensures the new rule applies immediately. The `walkDirCache.invalidate()` call before reload is the right approach to avoid stale cache.

Issue #12933, opened 2026-07-01
Continue writes into Copilot Chat input


Hangover3832:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 10
- Continue version: 2.0.0
- IDE version: VS Code 1.127.0
- Model: 
     Continue: Qwen2.5-Coder (local)
     Copilot: Ornith-1.0 (local)
```

### Description

Continue autocomplete writes into Copilot Chat input. This doesn't make any sense, is it supposed to do so? Copilot's inline suggestions are disabled.

### To reproduce

1. Start typing into the chat input
2. Continue autocompletes the chat

### Log output

```Shell

```

minh2416294:
hey, i'd like to take this one. 

Root cause: the vscode inline completion provider is registered with a catch-all selector [{ pattern: '**' }] (VsCodeExtension.ts around 347) so it matches every doc, including other extensions input boxes. the handler in completionProvider.ts only returns early for the vscode-scm scheme (line 174), theres no general guard for non-code inputs. the copilot chat input is actually a real TextDocument with scheme chatSessionInput, so it slip through and get completions, which is what the reporter is seeing now

Plan fix: add a small guard in the handler that bails early for known non-code input schemes (chatSessionInput, vscode-interactive-input, comment, output, etc plus the existing vscode-scm), pulled into a little isNonCodeInputScheme() helper with unit tests. denylist rather than narrowing the registration selector, since an allowlist could break vscode-remote (ssh/wsl), github.dev, untitled buffers and notebooks that work today. 

Does this direction look ok?

PR #12942, opened 2026-07-03
fix(vscode): stop autocomplete firing in non-code input boxes (#12933)

minh2416294:
## Description

Continue's VS Code inline autocomplete was firing inside other extensions' text input boxes - the reporter saw it autocomplete into the GitHub Copilot Chat input (#12933).

**Root cause:** the inline completion provider is registered with a catch-all document selector `[{ pattern: "**" }]` (`extensions/vscode/src/extension/VsCodeExtension.ts`), so VS Code invokes it for every document, including foreign chat/comment/output input boxes. The handler `provideInlineCompletionItems` only returned early for the `vscode-scm` scheme, so it had no general guard against non-code input surfaces. The Copilot Chat input is a real TextDocument with URI scheme `chatSessionInput`, which slipped through and received completions.

**Fix:** add a small handler-side guard that returns early for known non-code input schemes, extracted into a testable `isNonCodeInputScheme()` helper. The existing `vscode-scm` bail is folded into the same denylist (behavior-preserving).

I kept this as a scheme denylist rather than narrowing the registration selector to an allowlist, because an allowlist would risk silently regressing document types that work today - `vscode-remote` (SSH/WSL/Codespaces), `vscode-vfs` (github.dev), `untitled` buffers, and notebook cells. This also mirrors how VS Code's own Copilot completions guard their chat input by scheme, and matches existing scheme-guarding precedent in this repo (InlineTipManager, NextEditWindowManager, ideUtils)

## Checklist

- [x] The relevant docs, if any, have been updated or created (n/a — internal behavior fix)

- [x] The relevant tests, if any, have been updated or created.

## Tests

Added unit tests in `ContinueCompletionProvider.vitest.ts`:

- `isNonCodeInputScheme` returns true for each denied scheme (vscode-scm, comment, chatSessionInput, vscode-interactive-input, vscode-chat-editor, chat-editing-text-model, output, vscode-settings) and false for real editable schemes (file, vscode-remote, untitled, vscode-notebook-cell, vscode-vfs).

- The provider returns null and does not start a completion chain for each denied scheme (parametrized), with a named case for the chatSessionInput scheme that reproduces #12933.

- The existing "starts a new chain" test on a file:// document acts as the positive control, confirming real code files are unaffected.

Manual repro (autocomplete leaking into the Copilot Chat input) requires a build of the extension with Copilot installed; the behavior is otherwise fully covered by the unit tests above.

### DIFF
--- a/extensions/vscode/src/autocomplete/__tests__/ContinueCompletionProvider.vitest.ts
+++ b/extensions/vscode/src/autocomplete/__tests__/ContinueCompletionProvider.vitest.ts
@@ -2,7 +2,10 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import * as vscode from "vscode";
 
-import { ContinueCompletionProvider } from "../completionProvider";
+import {
+  ContinueCompletionProvider,
+  isNonCodeInputScheme,
+} from "../completionProvider";
 
 import * as NextEditLoggingServiceModule from "core/nextEdit/NextEditLoggingService";
 import * as PrefetchQueueModule from "core/nextEdit/NextEditPrefetchQueue";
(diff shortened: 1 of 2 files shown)


minh2416294:

minh2416294:
quick heads up on the red check, I think it looks unrelated to this change.

the jetbrains-tests failure is the intellij integration test Autocomplete > testAutocomplete() (Autocomplete.kt:42). this PR only touches extensions/vscode/ (the completion provider + its test) so it cant affect the jetbrains plugin build, and that same test is failing on other open PRs right now that dont touch jetbrains either, so it seems pre-existing rather than caused by this change

the vscode checks (vscode-checks, the autocomplete e2e, lint, prettier, the test matrix on all node versions) are all green. happy to dig in if you want but wanted to flag the red doesnt look like its from this PR. thank you guys!

Issue #12946, opened 2026-07-06
Startup failures report no logs because there is no viewLogs method


poetnerd:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:MacOS 15.7.7
- Continue version: 2.0.0
- IDE version: VSCode 1.127.0
- Model:ANY
- config:
  
{
  "models": [
    {
      "title": "Gemma4:e4b (Ollama)",
      "provider": "ollama",
	"model": "gemma4:e4b",
      "apiBase": "http://localhost:11434"
    }
  ]
}

  
  OR link to agent in Continue hub:
```

### Description

The Troubleshooting guide says to open the console and look at the console window.
But start up failures offer the use a button "Show Logs".   But it does nothing.
Looking at the console window we see why:

2026-07-05 20:45:34.005 [error] Error: command 'continue.viewLogs' not found
    at Zrt._tryExecuteCommand (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:1893:4832)
    at Zrt.executeCommand (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/

If I understand the README, it seems like 2.0 is planned as the last version of Continue. I hope that's not true.

Suggestion: It would make things easier for users if a viewLogs method were added.

### To reproduce

1. Configure any of the setups where activation fails.
2. Press "Show Logs" button.
3. Expected behavior, Logs show up; Actual behavior: the failure alert disappears, showing no logs.
4. Look in console log, filter on "continue"
5. See the error message like I logged above complaining continue.viewLogs not found.

### Log output

```Shell
2026-07-05 20:45:34.005 [error] Error: command 'continue.viewLogs' not found
    at Zrt._tryExecuteCommand (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:1893:4832)
    at Zrt.executeCommand (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/
```

PR #12950, opened 2026-07-06
Fix startup logs by registering continue.viewLogs command early

SparshGarg999:
### Description
This PR resolves issue #12946 where clicking the "Show Logs" button during extension startup failures does nothing and throws an error in the console: command 'continue.viewLogs' not found.

The command continue.viewLogs was registered too late (after full extension activation). When activation crashes early, the command is never registered, rendering the troubleshooting option useless. 

By registering continue.viewLogs immediately in the main ctivate entry point in extension.ts, we ensure that the command is available to VS Code even if subsequent activation fails.

### Test Plan
1. Induce an early activation crash (e.g., throwing a dummy error in ctivateExtension).
2. Click the "View Logs" (or "Show Logs") button on the warning pop-up.
3. DevTools console should toggle open successfully instead of throwing a command-not-found error.

### DIFF
--- a/extensions/vscode/src/extension.ts
+++ b/extensions/vscode/src/extension.ts
@@ -14,6 +14,15 @@ async function dynamicImportAndActivate(context: vscode.ExtensionContext) {
 }
 
 export function activate(context: vscode.ExtensionContext) {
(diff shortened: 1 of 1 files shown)


SparshGarg999:

Issue #12947, opened 2026-07-06
Activation fails if there's a non-standard color palette in place.


poetnerd:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:MacOS 15.7.7
- Continue version:2.0.0
- IDE version:VSCode 1.127.0
- Model:ANY
- config:
  
{
  "models": [
    {
      "title": "Gemma4:e4b (Ollama)",
      "provider": "ollama",
	"model": "gemma4:e4b",
      "apiBase": "http://localhost:11434"
    }
  ]
}
  
  OR link to agent in Continue hub:
```

### Description

Some file format I had, changed my Appearance → Color Theme to something odd, but that worked for all other extensions.

Continue failed to start.  See logs below.

### To reproduce

1. Set a non-standard Appearance → Color Theme
2. Enable Continue
3. Expected behavior: it runs. Actual behavior, it fails to start.

Work around: set a vanilla color Theme like "Light (Visual Studio)"

### Log output

```Shell
console.ts:139 [Extension Host] Error activating extension:  TypeError: Cannot read properties of undefined (reading 'editor.background')
	at _InlineTipManager.createSvgTooltipDecoration (/Users/wdc/.vscode/extensions/continue.continue-2.0.0-darwin-arm64/out/extension.js:876397:47)
	at <instance_members_initializer> (/Users/wdc/.vscode/extensions/continue.continue-2.0.0-darwin-arm64/out/extension.js:876279:35)
	at new _InlineTipManager (/Users/wdc/.vscode/extensions/continue.continue-2.0.0-darwin-arm64/out/extension.js:876287:18)
	at _InlineTipManager.getInstance (/Users/wdc/.vscode/extensions/continue.continue-2.0.0-darwin-arm64/out/extension.js:876283:40)
	at setupInlineTips (/Users/wdc/.vscode/extensions/continue.continue-2.0.0-darwin-arm64/out/extension.js:876208:20)
	at activateExtension (/Users/wdc/.vscode/extensions/continue.continue-2.0.0-darwin-arm64/out/extension.js:876542:3)
[...]

PR #12951, opened 2026-07-06
Guard theme colors to prevent activation crash

SparshGarg999:
### Description
This PR resolves issue #12947 where the extension fails to activate when using non-standard or custom color palettes. 

Specifically, accessing `this.theme.colors["editor.background"]` or `this.theme.colors["editor.foreground"]` throws `TypeError: Cannot read properties of undefined (reading 'editor.background')` if `this.theme.colors` is `undefined`.

Optional chaining (`?.`) has been added to guard these theme property lookups in both `InlineTipManager.ts` and `JumpManager.ts`, falling back safely to default stroke/background colors when theme colors are unavailable.

### Test Plan
1. Switch to a custom/non-standard theme that does not declare VS Code token colors in the usual format.
2. Activate the Continue extension.
3. Observe that the extension starts up successfully without crashing during tooltip generation.

### DIFF
--- a/extensions/vscode/src/activation/InlineTipManager.ts
+++ b/extensions/vscode/src/activation/InlineTipManager.ts
@@ -236,7 +236,7 @@ export class InlineTipManager {
 
   private createSvgTooltipDecoration() {
     var backgroundColour = "#333333";
-    if (this.theme) {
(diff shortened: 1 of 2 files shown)


SparshGarg999:
recheck

Issue #12948, opened 2026-07-06
Activation fails if the YAML extension is not present.


poetnerd:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:15.7.7
- Continue version:2.0.0
- IDE version:VSCode 1.127.0
- Model:ANY
- config:
  
{
  "models": [
    {
      "title": "Gemma4:e4b (Ollama)",
      "provider": "ollama",
	"model": "gemma4:e4b",
      "apiBase": "http://localhost:11434"
    }
  ]
}
  
  OR link to agent in Continue hub:
```

### Description

After overcoming the error I report in Issue #12947, Continue would still not start up.  
Duck.ai helpfully interpreted the console log output and said "Continue is trying to register/update a config.yaml schema in VS Code settings, but the YAML extension isn’t installed, so it can’t write yaml.schemas."

After I installed the Red Hat YAML extension, Continue started up.

### To reproduce

See description above.

### Log output

```Shell
INFO [Continue.continue]: Command `continue.focusContinueInput` already registered by Continue - open-source AI code agent (Continue.continue)
log.ts:117   ERR [Extension Host] Failed to register Continue config.yaml schema, most likely, YAML extension is not installed CodeExpectedError: Unable to write to User Settings because yaml.schemas is not a registered configuration.
    at f5e.toConfigurationEditingError (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:5117:7060)
    at f5e.validate (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:5117:9830)
    at f5e.doWriteConfiguration (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:5117:3453)
    at Object.factory (vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:5117:3309)
[...]

PR #12952, opened 2026-07-06
Check if YAML extension is installed before registering config schema

SparshGarg999:
### Description
This PR resolves issue #12948 where extension activation fails when the Red Hat YAML extension (`redhat.vscode-yaml`) is not installed.

Writing to `yaml.schemas` user settings when the configuration is not registered throws `CodeExpectedError: Unable to write to User Settings because yaml.schemas is not a registered configuration.`.

This is resolved by checking if the `redhat.vscode-yaml` extension is installed using `vscode.extensions.getExtension` before reading or updating the `yaml.schemas` settings configuration.

### Test Plan
1. Uninstall the Red Hat YAML extension.
2. Start VS Code with the Continue extension enabled.
3. Verify that the extension activates successfully without throwing a `CodeExpectedError` on `yaml.schemas`.

### DIFF
--- a/extensions/vscode/src/activation/activate.ts
+++ b/extensions/vscode/src/activation/activate.ts
@@ -60,10 +60,7 @@ export async function activateExtension(context: vscode.ExtensionContext) {
         vscode.ConfigurationTarget.Global,
       );
     } catch (error) {
-      console.error(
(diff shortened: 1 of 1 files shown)


SparshGarg999:
recheck

PR #12974, opened 2026-07-10
fix(cli): resume selected session in `cn ls` instead of the latest

### LINKED ISSUE #12927: cn ls always resumes the latest session regardless of user selection

### Relevant environment info

```Markdown
- OS:
- Continue version:
- IDE version:
- Model:
- config:
  

  
```

### Description

_No response_

### To reproduce

_No response_

### Log output

```Shell

```

Purv-Kabaria:
## Description

Fixes #12927.

`cn ls` lets you pick a past session from the TUI, but resuming always reopened the **most recent** session regardless of which one you selected.

Root cause is in `extensions/cli/src/session.ts`. `loadSession()` sorts the session files by mtime and unconditionally loads `files[0]`:

```ts
// Load the most recent session
const session: Session = JSON.parse(fs.readFileSync(files[0].path, "utf8"));
```

The `cn ls` flow (`commands/ls.ts`) already communicates the chosen session by setting `CONTINUE_CLI_TEST_SESSION_ID` — the same env var `SessionManager.getCurrentSession()` uses to pin the current session id — before calling `chat({ resume: true })`. But `loadSession()` never read that var, so the selection was silently dropped and `--resume` always fell back to the latest session.

This PR makes `loadSession()` honor `CONTINUE_CLI_TEST_SESSION_ID` first (loading that session by id via the existing `loadSessionById()`), falling back to the most-recent session when the var is absent or the id can't be loaded. Plain `cn --resume` is unchanged.

The fix is intentionally minimal — one guard at the top of `loadSession()`, no signature or caller changes.

> Note: the env var name (`CONTINUE_CLI_TEST_SESSION_ID`) reads as test-only, but it is already used on the production `cn ls` path today. I kept the existing name to stay minimal; happy to rename it in a follow-up if maintainers prefer a non-`TEST` name for this now-user-facing mechanism.

## Checklist

- [x] The relevant docs, if any, have been updated or created (no user-facing docs affected)

## Tests

Added a unit test in `extensions/cli/src/session.test.ts` under `loadSession`:

- **`should load the selected session when CONTINUE_CLI_TEST_SESSION_ID is set`** — sets the env var and asserts `loadSession()` resolves the session by id (`historyManager.load` called with the selected id), does **not** fall back to scanning the directory by mtime (`readdirSync` not called), and sets it as the current session.

I verified the test fails on `main` (without the fix) and passes with it. Existing `loadSession` tests (including "should load the most recent session" for the no-selection path) still pass — full `session.test.ts` is green (32/32).

### DIFF
--- a/extensions/cli/src/session.test.ts
+++ b/extensions/cli/src/session.test.ts
@@ -357,6 +357,31 @@ describe("SessionManager", () => {
       expect(result).toEqual(mockSession);
       expect(getCurrentSession()).toEqual(mockSession);
     });
+
+    it("should load the selected session when CONTINUE_CLI_TEST_SESSION_ID is set", () => {
+      const selectedSession: Session = {
+        sessionId: "selected-session-id",
+        title: "Selected Session",
+        workspaceDirectory: "/test/workspace",
+        history: [],
+      };
+      mockHistoryManager.load.mockReturnValue(selectedSession);
+      process.env.CONTINUE_CLI_TEST_SESSION_ID = "selected-session-id";
+
+      try {
+        const result = loadSession();
+
+        expect(result).toEqual(selectedSession);
+        // It must resolve the session by id, not by most-recent mtime
+        expect(mockHistoryManager.load).toHaveBeenCalledWith(
+          "selected-session-id",
+        );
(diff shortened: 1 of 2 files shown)


Purv-Kabaria:

PR #5310, opened 2025-04-23
perf(llm): Optimize pruneLines functions in countTokens

0x23d11:
## Description
Closes #4947 

This PR addresses issue #4947 by optimizing the performance of the `pruneLinesFromTop` and `pruneLinesFromBottom` functions in `core/llm/countTokens.ts`

### Problem

The previous implementations used `Array.prototype.shift()` and `Array.prototype.pop()` within a `while` loop to remove lines from the beginning or end of a prompt until it fit within the token limit. These array modification methods have a time complexity of O(n) because they require shifting subsequent elements. When dealing with very long prompts (e.g., thousands of lines), repeatedly calling `shift()` or `pop()` becomes computationally expensive, leading to significant performance degradation.

### Solution

The implemented solution refactors these functions to avoid costly array modifications within the loop:

1.  The prompt is split into lines.
2.  The token count for **each line** is calculated **once** upfront and stored in an array (`lineTokens`).
3.  The total initial token count is calculated by summing `lineTokens` and adding the count for necessary newline characters (`\n`).
4.  A `while` loop iterates as long as the `totalTokens` exceeds `maxTokens`.
5.  Inside the loop, instead of removing elements from the `lines` array, an index pointer (`start` or `end`) is adjusted.
6.  The pre-calculated token count for the line being (conceptually) removed, along with its corresponding newline token, is subtracted from `totalTokens`.
7.  After the loop, `Array.prototype.slice()` (an O(n) operation performed only **once**) is used with the final `start` or `end` index to extract the desired lines.
8.  The resulting lines are joined back into a string.

### Benefits

This approach drastically reduces the computational complexity, especially for large prompts, as the expensive O(n) operations (`shift`/`pop`) inside the loop are replaced by cheap O(1) index increments/decrements. The token calculation per line and the final `slice` operation are performed only once.

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [] The relevant docs, if any, have been updated or created
- [] The relevant tests, if any, have been updated or created

## Screenshots

[ For visual changes, include screenshots. Screen recordings are particularly helpful, and appreciated! ]

## Testing instructions
Objective:

[...]

Issue #4947, opened 2025-04-02
The functions `pruneLinesFromTop` and `pruneLinesFromBottom` are time-consuming when processing large files, which negatively impacts the code completion experience.


DanielYzd:
### Validations

- [x] I believe this is a way to improve. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement) that requests the same enhancement

### Problem

[countTokens.ts](https://github.com/continuedev/continue/blob/main/core/llm/countTokens.ts)
pruneLinesFromTop、pruneLinesFromBottom
Process a file with 2000 lines, requiring the removal of the first 1000 lines.

Original Version: Execute the shift() operation 1000 times, each time moving 2000 → 1999 → ... → 1001 Elements, resulting in a total of approximately (2000+1001)*1000/2 = 1,500,500 operations.

Optimized Version: Preprocess the tokens of 2000 lines (O(n)), loop 1000 times with index increment (O(n)), totaling approximately 3000 operations.

### Solution

Optimize as follows
  `function pruneLinesFromTop(
  prompt: string,
  maxTokens: number,
  modelName: string,
): string {
  const lines = prompt.split("\n");
  //Preprocess tokens for all rows and cache them.
  const lineTokens = lines.map(line => countTokens(line, modelName));
  let totalTokens = lineTokens.reduce((sum, tokens) => sum + tokens, 0);
  let start = 0;

  // Using indexes instead of array modifications.
  while (totalTokens > maxTokens && start < lines.length) {
    totalTokens -= lineTokens[start];
    start++;
  }

  return lines.slice(start).join("\n");
}

function pruneLinesFromBottom(
  prompt: string,
  maxTokens: number,
  modelName: string,
): string {
  const lines = prompt.split("\n");
  const lineTokens = lines.map(line => countTokens(line, modelName));
  let totalTokens = lineTokens.reduce((sum, tokens) => sum + tokens, 0);
  let end = lines.length;

  // Reverse traversal to avoid array modification
  while (totalTokens > maxTokens && end > 0) {
    end--;
    totalTokens -= lineTokens[end];
  }

  return lines.slice(0, end).join("\n");
}`

Patrick-Erichsen:
Thanks for the analysis here! That code looks more/less correct to me, any chance you'd be willing to open up a PR with the changes?

PR #12981, opened 2026-07-13
perf(llm): avoid tokenizing discarded lines

### LINKED ISSUE #12980: perf: avoid tokenizing lines that pruning discards
## Problem

Issue #4947 and PR #5310 removed repeated `shift()` / `pop()` work from `pruneLinesFromTop` and `pruneLinesFromBottom`. The current implementation still tokenizes every line up front, including lines that are immediately discarded when a large prompt is pruned to a small token budget.

On current `main` (`d0a3c0b626b5bebc3bef4742eec05a0242be0bab`), a deterministic 320-line fixture performs 641 `Tiktoken.encode` calls across top and bottom pruning. Traversing inward from the side that will be retained and stopping when the next line would exceed the budget reduces this to 88 calls (-86.3%).

At the 2,000-line scale from #4947, the same measurement is 4,001 calls on `main` and 86 with retained-side traversal (-97.9%). All nine runs at each revision produced the same count.

## Correctness check

I compared the proposed traversal against current `main` over 580 top/bottom cases covering empty input, leading and trailing blank lines, newline boundaries, Unicode, single oversized lines, random multiline prompts, and zero or negative limits. All returned strings were byte-for-byte identical.

## Proposed change

- Accumulate retained lines from the bottom for `pruneLinesFromTop`.
- Accumulate retained lines from the top for `pruneLinesFromBottom`.
- Preserve the existing one-token newline accounting exactly.
- Enable the existing focused pruning tests and keep the production change limited to the two function bodies.

Would this follow-up optimization be welcome?

dexhunter:
## Description

Fixes #12980.

PR #5310 removed repeated array mutation from `pruneLinesFromTop` and `pruneLinesFromBottom`, but both functions still tokenize every line before discarding most of them. This change traverses from the side that will be retained and stops as soon as the next line would exceed the token budget.

The existing one-token newline accounting and returned strings are preserved. The production change is limited to the two pruning function bodies, and this PR enables their existing focused tests.

### Performance

The deterministic evaluator counts real `Tiktoken.encode` calls across top and bottom pruning of a 320-line prompt:

| Metric | `main` | This PR | Change |
| --- | ---: | ---: | ---: |
| median tokenizer calls | 641 | 88 | -86.3% |

All nine baseline and candidate samples were identical within each revision. A 580-case differential suite covering empty input, leading and trailing blank lines, newline boundaries, Unicode, oversized lines, random multiline prompts, and zero or negative limits produced byte-for-byte identical output.

On the 2,000-line scale described in #4947, the same metric is 4,001 calls on `main` and 86 on this branch (-97.9%).

### Autoresearch

The retained-side traversal was independently reproduced during a 10-step autoresearch run with Weco. The final patch was then reduced and validated in a clean worktree.

Public trajectory: https://dashboard.weco.ai/share/xWch5tqFM_tvJr2AZVT4lWugmaTnbu5b

## AI Code Review

## Checklist

Not applicable; this changes token-pruning work without changing UI behavior.

## Tests

- `npm test -- --runInBand llm/countTokens.test.ts` (25 passed, 9 unrelated skipped)
- `npm run tsc:check`
- ESLint and Prettier on both changed files
- strict 580-case differential evaluator and positive/negative-control replay
- `git diff --check`

### DIFF
--- a/core/llm/countTokens.test.ts
+++ b/core/llm/countTokens.test.ts
@@ -47,7 +47,7 @@ describe("countTokensAsync", () => {
   });
 });
 
-describe.skip("pruneLinesFromTop", () => {
+describe("pruneLinesFromTop", () => {
   it("should prune lines from the top to fit within max tokens", () => {
     const prompt = "Line 1\nLine 2\nLine 3\nLine 4";
     const pruned = pruneLinesFromTop(prompt, 5, "gpt-4");
@@ -88,7 +88,7 @@ describe.skip("pruneLinesFromTop", () => {
   });
 });
 
-describe.skip("pruneLinesFromBottom", () => {
+describe("pruneLinesFromBottom", () => {
   it("should prune lines from the bottom to fit within max tokens", () => {
     const prompt = "Line 1\nLine 2\nLine 3\nLine 4";
     const pruned = pruneLinesFromBottom(prompt, 5, "gpt-4");
--- a/core/llm/countTokens.ts
+++ b/core/llm/countTokens.ts
@@ -285,24 +285,19 @@ function pruneLinesFromTop(
   modelName: string,
 ): string {
   const lines = prompt.split("\n");
-  // Preprocess tokens for all lines and cache them.
-  const lineTokens = lines.map((line) => countTokens(line, modelName));
-  let totalTokens = lineTokens.reduce((sum, tokens) => sum + tokens, 0);
-  let start = 0;
-  let currentLines = lines.length;
+  let start = lines.length;
+  let totalTokens = 0;
 
-  // Calculate initial token count including newlines
-  totalTokens += Math.max(0, currentLines - 1); // Add tokens for joining newlines
-
-  // Using indexes instead of array modifications.
-  // Remove lines from the top until the token count is within the limit.
-  while (totalTokens > maxTokens && start < currentLines) {
-    totalTokens -= lineTokens[start];
-    // Decrement token count for the removed line and its preceding/joining newline (if not the last line)
-    if (currentLines - start > 1) {
-      totalTokens--;
+  // Build the retained suffix without tokenizing discarded leading lines.
+  while (start > 0) {
+    const next = start - 1;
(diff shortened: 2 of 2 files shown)


dexhunter:

dexhunter:
The `jetbrains-tests` failure is in the pre-existing IntelliJ integration test `Autocomplete.testAutocomplete()` at `Autocomplete.kt:42` (`expected true but was false`). This PR only modifies `core/llm/countTokens.ts` and its tests. Current `main` is the PR base (`d0a3c0b`) and reproduces the identical JetBrains failure in [its own PR Checks run](https://github.com/continuedev/continue/actions/runs/27797574180), so this does not appear attributable to this diff.

dexhunter:
Re-checked this today against `main` at `5522c6f4`.

It still applies unchanged. `core/llm/countTokens.ts` has not been modified on `main` since `1aac783c` (2026-03-25), and the only commits `main` has gained since this branch point are three `docs-site` changes. Applied on top of `5522c6f4`, `jest llm/countTokens.test.ts --runInBand` reports 25 passed, 9 skipped, 0 failed. The two `describe` blocks this PR un-skips also pass on unmodified `main`, which is deliberate: they exist to pin that the retained-side traversal returns byte-identical strings, not to assert new behaviour.

On the red check, `jetbrains-tests` fails at `Autocomplete > testAutocomplete()` with `AssertionFailedError at Autocomplete.kt:42`, and that job is currently failing on every recent `PR Checks` run regardless of the diff. A Snyk dependency bump from 2026-07-25 reproduces the identical assertion, and docs-only branches (`docs/update-node-version`, `docs/fix-intellij-config-ref`) fail the same job the same way. This PR touches only `core/llm/countTokens.ts` and its test file, so it does not reach that IntelliJ integration test.

Context, in case it is useful: #4947 reported these two functions as a code-completion latency problem on large files, and #5310 removed the repeated array mutation but left the eager tokenization of every line in place. This is the remaining half of that. If it would be easier to review, I am glad to split the test un-skip into a separate PR and keep this one to the two function bodies.

PR #12988, opened 2026-07-14
fix(core): ensure single leading system message in CLI requests (fixes #12963)

### LINKED ISSUE #12963: Malformed Message After Tool Request Cancellation

### Relevant environment info

```Markdown
- OS:
- Continue version:
- IDE version:
- Model:
- config:
  

  
```

### Description

# Bug Report: Malformed Message After Tool Request Cancellation

## Summary
When canceling/denying a tool request in Continue, the follow-up message sent to the LLM has a malformed structure where the system message is not at the beginning of the conversation history. This causes a 400 Bad Request error from the LLM provider (via aqueduct proxy).

## Error Details
```
Error: 400 aqueduct.BadRequestError: OpenAIException - Error code: 400 - 
{'error': {'message': 'System message must be at the beginning.', 
'type': 'BadRequestError', 'param': None, 'code': 400}}. 
Received Model Group=qwen-3.5-397b
```

## Steps to Reproduce
1. Continue attempts to execute a tool (e.g., running a bash command like `sage selection/baselines/graph_operations.sage`)
2. User denies/cancels the tool request
3. Continue sends a follow-up message to the LLM
4. The LLM provider returns a 400 error indicating "System message must be at the beginning"

## Expected Behavior
After canceling a tool request, Continue should:
- Properly format the conversation history with the system message at the beginning
- Send a valid message structure to the LLM
- Gracefully handle the cancellation and allow the user to continue the conversation

## Actual Behavior
- The message structure is malformed
- System message is not positioned at the beginning of the conversation array
- The LLM provider rejects the request with a 400 error
- The conversation is interrupted and cannot continue

## Environment
- **Continue Extension**: (version unknown)
- **LLM Provider**: qwen-3.5-397b (via aqueduct proxy)
- **Proxy Layer**: aqueduct
- **Platform**: darwin (macOS)
- **Date**: 2026-07-09

## Additional Context
This appears to be an issue with how Continue constructs the conversation history message array after a tool cancellation event. The system message role should always be the first element in the messages array sent to OpenAI-compatible APIs, but after tool cancellation, this ordering seems to be violated.

## Suggested Fix Areas
1. Review message formatting logic in tool cancellation handlers
2. Ensure system message is always positioned at `messages[0]` after any conversation state change
3. Add validation before sending requests to catch malformed message structures

## Workaround
Currently, the only workaround is to restart the Continue chat session after this error occurs.

### To reproduce

1. make a request that runs a tool call
2. cancel the tool call

### Log output

```Shell

```

not-knope:
## Description

Fixes #12963.

When cancelling/denying a tool call, the follow-up request to the model could fail with:

```
400 - {'error': {'message': 'System message must be at the beginning', ...}}
```

Root cause: `convertFromUnifiedHistoryWithSystemMessage` (in `core/util/messageConversion.ts`, used by the CLI streaming path in `extensions/cli/src/stream/streamChatResponse.ts`) injected the system message at index 0, but `convertFromUnifiedHistory` **also** emitted any system-role messages that already existed inside the history. Those messages can come from:

- notifications added via `ChatHistoryService.addSystemMessage` (diffs, "No changes to display", remote/serve helpers, etc.),
- auto-compaction notices, and
- a system message stored at `history[0]` in loaded/remote sessions.

The result was a request with a `system` message at a non-zero index, which strict OpenAI-compatible providers (e.g. the reporter's aqueduct proxy) reject with a `400 "System message must be at the beginning"`. It surfaced most often on the follow-up request that is sent right after a tool call is cancelled.

The fix merges all system content (the injected system message plus any system messages found in the history) into a single `system` message positioned at index 0, so the outgoing request is always valid. Non-system messages keep their original order, so tool-call/tool-result sequences are untouched.

## Checklist

## Tests

Added `describe("convertFromUnifiedHistoryWithSystemMessage", ...)` cases in `extensions/cli/src/messageConversion.test.ts` covering:

- basic injection of the system message at index 0,
- a history that already starts with a system message (no duplicate; single leading system message),
- a cancelled tool call combined with a mid-conversation system notification (the exact reproduction), asserting the only `system` message is at index 0 and the tool result is preserved,
- an empty base system message with a stored system message in history.

## Note

This PR was developed with AI assistance (Cursor), as reflected in the commit's `Co-authored-by` trailer. All code was reviewed and verified locally before submission.

### DIFF
--- a/core/util/messageConversion.ts
+++ b/core/util/messageConversion.ts
@@ -355,6 +355,17 @@ export function convertFromUnifiedHistory(
 
 /**
  * Convert ChatHistoryItem array to ChatCompletionMessageParam array with injected system message
+ *
+ * Many OpenAI-compatible providers require the system message to be the very
+ * first message in the array. The history can contain additional system
+ * messages (e.g. notifications added via `addSystemMessage`, compaction notices,
+ * or a system message already stored at `history[0]` in loaded/remote sessions).
+ * If those were emitted inline, the resulting array would have a system message
+ * at a non-zero index and the provider would reject the request with a
+ * "System message must be at the beginning" 400 error (notably surfaced after
+ * cancelling a tool call, which triggers a follow-up request). To avoid this we
+ * merge all system content into a single system message positioned at index 0.
+ *
  * @param historyItems - The chat history items
  * @param systemMessage - The system message to inject at the beginning
  */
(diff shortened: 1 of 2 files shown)


not-knope:

PR #12993, opened 2026-07-16
fix(fetch): honor proxy and verifySsl on the node fetch path

devYRPauli:
## Description

`proxy`, `verifySsl`, and `caBundlePath` set in config.yml were ignored by the remote config-sync / index-cache client. `core/continueServer/stubs/client.ts` used the global `fetch()` directly for its requests, so unlike the LLM request path (which already routes through `fetchwithRequestOptions`), these node fetches never applied the configured request options.

This routes the `ContinueServerClient` requests (`getConfig`, `getFromIndexCache`, `sendFeedback`) through `fetchwithRequestOptions`, and threads `config.requestOptions` in from the single call site in `core/indexing/CodebaseIndexer.ts`. The new constructor parameter is optional, so existing callers are unaffected.

Fixes #12956.

## AI Code Review

- **Team members only**: AI review runs automatically when PR is opened or marked ready for review
- Team members can also trigger a review by commenting `@continue-review`

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [x] The relevant docs, if any, have been updated or created
- [x] The relevant tests, if any, have been updated or created

## Screen recording or screenshot

Not applicable. The change is in `core` and `packages/fetch` and has no UI surface. The observable result is which agent and TLS settings the request uses, which the tests below assert directly.

## Tests

Added `packages/fetch/src/fetch.test.ts` asserting the agent is an `HttpsProxyAgent` pointed at `requestOptions.proxy`, that `rejectUnauthorized` is `false` when `verifySsl: false`, and both together. The `packages/fetch` suite passes, 100 tests.

PR #12999, opened 2026-07-19
fix(fetch): route dynamic model-list fetches through request options

devYRPauli:
## Description

The dynamic model-list fetchers in `core/llm/fetchModels.ts` (Ollama, OpenRouter, Anthropic, Gemini) call the global `fetch()` directly. They never go through `fetchwithRequestOptions`, so they ignore the user's `requestOptions` (proxy, custom CA / `verifySsl`, custom headers).

Behind an SSL-intercepting corporate proxy, or with a custom CA configured, the "Add model" dropdown fails to load these provider model lists even though normal chat requests succeed, because chat already routes through `fetchwithRequestOptions` while these fetchers do not.

This threads `requestOptions` through the four fetchers and uses `fetchwithRequestOptions` instead of the raw `fetch()`. The `models/fetch` handler in `core.ts` now loads the config and passes `config.requestOptions` down.

There is no behavior change for users without `requestOptions` configured; it stays a no-op for them. The change follows the existing `fetchwithRequestOptions(url, init, requestOptions)` call pattern already used elsewhere in `core.ts`.

This mirrors the earlier `ContinueServerClient` fix in #12993, which deliberately left `fetchModels.ts` untouched.

## AI Code Review

## Checklist

Not applicable. Reproducing the failure needs an SSL-intercepting proxy or a custom CA in front of the client, which a recording cannot show. The observable result is whether the "Add model" dropdown populates under those conditions.

## Tests

No new tests. The four fetchers move onto the same wrapper the chat path already uses, and the behaviour restored (honouring `requestOptions`) needs a proxy or custom CA to observe, which the existing suite does not set up.

### DIFF
--- a/core/core.ts
+++ b/core/core.ts
@@ -1135,10 +1135,12 @@ export class Core {
 
     on("models/fetch", async (msg) => {
       try {
+        const { config } = await this.configHandler.loadConfig();
         return await fetchModels(
           msg.data.provider,
           msg.data.apiKey,
           msg.data.apiBase,
+          config?.requestOptions,
         );
       } catch (error: any) {
         void this.ide.showToast("error", error.message);
--- a/core/llm/fetchModels.ts
+++ b/core/llm/fetchModels.ts
@@ -143,7 +143,7 @@ async function fetchOpenRouterModels(
       throw new Error(`Failed to fetch OpenRouter models: ${response.status}`);
     }
 
-    const data = await response.json();
+    const data: any = await response.json();
     if (!data.data || !Array.isArray(data.data)) {
       return [];
     }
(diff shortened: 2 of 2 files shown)


devYRPauli:
The finding is correct. Fixed in `1ed1eeec1`.

`fetchModels()` passed `requestOptions` to the Ollama, OpenRouter, Anthropic and Gemini fetchers, but the `default` branch called `fetchProviderModelsViaListModels()` without them. That path builds its client with `llmFromProviderAndOptions()`, so `listModels()` still ignored the proxy, the custom CA and the custom headers. Every provider outside the four named cases uses that path.

`LLMOptions` already accepts `requestOptions`, so the fallback now takes the value and passes it through.

PR #13003, opened 2026-07-20
Route Mistral API key autodetection through this.fetch

devYRPauli:
## Description

`Mistral.autodetectApiKeyType()` calls the global `fetch` directly:

```ts
const mistralResp = await fetch("https://api.mistral.ai/v1/models", {
```

That bypasses `fetchwithRequestOptions`, so this one request ignores the user's configured `proxy`, `caBundlePath`, `verifySsl` and custom headers. Every other request in this provider goes through the wrapper that `BaseLLM` wires to `fetchwithRequestOptions` with `this.requestOptions` (`core/llm/index.ts`). The inherited call sites in `OpenAI.ts` all use `this.fetch(...)` already.

The call is reachable in the common case: the constructor invokes it whenever `apiBase` is not set explicitly, in order to decide between the codestral and mistral endpoints. Its rejection is swallowed by an empty `.catch((err: any) => {})`, so behind an intercepting corporate proxy or with a custom CA the request fails silently, no error surfaces, and the provider is left with an incorrectly autodetected `apiBase`.

The fix calls `fetchwithRequestOptions` with `this.requestOptions` rather than `this.fetch`. `this.fetch` throws on a non-ok response, and this request detects a Codestral key precisely by reading a 401, so it must see the response instead of an exception. Only `.status` is read, so there is no typing change.

This is a sibling of #12993 and #12999, which fixed the same class of bypass elsewhere in the codebase.

## AI Code Review

## Checklist

Not applicable. Reproducing the failure needs an intercepting proxy or a custom CA in front of the client, which a recording cannot show. The observable result is whether the autodetected `apiBase` is correct under those conditions.

## Tests

No new tests. The change moves one call onto the same request wrapper every sibling method already uses, and the behaviour it restores (honouring `requestOptions`) needs a proxy or custom CA to observe, which the existing suite does not set up.

I ran a typecheck over `core` before and after the change: 115 pre-existing errors both times, none in `Mistral.ts`. The pre-existing errors are unrelated module resolution failures for unbuilt `@continuedev/*` workspace packages.

### DIFF
--- a/core/llm/llms/Mistral.ts
+++ b/core/llm/llms/Mistral.ts
@@ -17,7 +17,7 @@ class Mistral extends OpenAI {
   };
 
   private async autodetectApiKeyType(): Promise<MistralApiKeyType> {
-    const mistralResp = await fetch("https://api.mistral.ai/v1/models", {
+    const mistralResp = await this.fetch("https://api.mistral.ai/v1/models", {
       method: "GET",
       headers: this._getHeaders(),
     });
--- a/core/llm/llms/Mistral.ts
+++ b/core/llm/llms/Mistral.ts
@@ -1,3 +1,5 @@
+import { fetchwithRequestOptions } from "@continuedev/fetch";
+
 import { ChatMessage, LLMOptions } from "../../index.js";
 import { codestralEditPrompt } from "../templates/edit/codestral.js";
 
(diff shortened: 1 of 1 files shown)


devYRPauli:
This review is against `13791dcc`. The head is now `af9512a2`.

The Codestral finding is fixed there. `autodetectApiKeyType` no longer uses `this.fetch`, because `this.fetch` throws on a non-ok response and this probe identifies a Codestral key by reading a 401. It calls `fetchwithRequestOptions` directly and then checks `mistralResp.status`.

The license-key and slug-resolution findings name `core/core.ts` and `packages/config-yaml/src/registryClient.ts`. This PR changes neither file. `git diff --stat origin/main...HEAD` reports one file, `core/llm/llms/Mistral.ts`, with 12 insertions and 4 deletions.

On the red checks: the Windows job fails on `runTerminalCommandTool > basic error handling > should handle non-existent commands` at `src/tools/runTerminalCommand.test.ts:73`, which times out after 30000ms. This PR does not touch that file. `jetbrains-tests` currently fails on other open PRs as well, for example #13192, #13193 and #13194.

devYRPauli:
Still applies. `core/llm/llms/Mistral.ts` on `main` today still calls the global `fetch` inside `autodetectApiKeyType`, so `this.requestOptions` is ignored for that one request. The diff is unchanged and touches only that file.

CI last ran here on 2026-07-20. Two checks failed. `jetbrains-tests` failed, and `test (windows-latest, 24)` failed on `src/tools/runTerminalCommand.test.ts:73`. This diff touches neither file. I cannot re-run the jobs from a fork.

PR #13036, opened 2026-07-27
feat(mcp): support VS Code-style variable interpolation in MCP server config

Onkar-Mane:
Adds support for `${workspaceFolder}`, `${workspaceFolderBasename}`, `${userHome}`, and `${env:VAR}` in MCP server `command`/`args`/`cwd`/`env`, resolved once right before the stdio process is spawned.

## Summary
- New `resolveMcpVariables` utility (`core/util/resolveMcpVariables.ts`) recursively resolves the variables in strings/arrays/objects, leaving unknown variables and `${{ secrets.X }}` untouched.
- Wired into `MCPConnection.constructStdioTransport`, the single choke point for all stdio MCP servers regardless of config source (YAML or JSON).
- Fixed `resolveCwd`'s absolute-path check to use `path.isAbsolute` instead of a Unix-only `charAt(0) === "/"` check, needed for `${workspaceFolder}`-derived Windows paths to be recognized correctly.
- `${workspaceFolder}` intentionally stays unresolved (rather than falling back to homedir) for remote workspaces, to avoid silently pointing configs at the wrong directory.

## Test plan
- [x] `core/util/resolveMcpVariables.test.ts` - 18 unit tests
- [x] `core/context/mcp/MCPConnection.vitest.ts` - new wiring/integration tests
- [x] `tsc --noEmit` clean
- [x] Manually verified in the Extension Development Host with a `.continue/mcpServers/*.yaml` using `${workspaceFolder}` in cwd/args/env

PR #13034, opened 2026-07-27
[Snyk] Security upgrade react-router-dom from 6.30.3 to 7.18.0

sestinj:
snyk-top-banner

#### Snyk changed the following file(s):

- `gui/package.json`

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
high severity | Open Redirect <br/>[SNYK-JS-REACTROUTER-18313144](https://snyk.io/vuln/SNYK-JS-REACTROUTER-18313144) | &nbsp;&nbsp;**641**&nbsp;&nbsp; 

#### Breaking Change Risk

Merge Risk: High

> **Notice:** This assessment is enhanced by AI.

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/continue-dev-inc.-default/project/c5fb30df-a06c-44cb-83af-5ada5ff6e4a9?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates?utm_source=github&utm_content=fix-pr-template)
🛠 [Adjust project settings](https://app.snyk.io/org/continue-dev-inc.-default/project/c5fb30df-a06c-44cb-83af-5ada5ff6e4a9?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://docs.snyk.io/scan-with-snyk/snyk-open-source/manage-vulnerabilities/upgrade-package-versions-to-fix-vulnerabilities?utm_source=github&utm_content=fix-pr-template)

---

**Learn how to fix vulnerabilities with free interactive lessons:**

🦉 [Open Redirect](https://learn.snyk.io/lesson/open-redirect/?source&#x3D;registry-oss-fix-pr&amp;loc&#x3D;fix-pr)

[...]

PR #13033, opened 2026-07-27
[Snyk] Security upgrade js-yaml from 4.1.1 to 5.2.2

sestinj:
snyk-top-banner

#### Snyk changed the following file(s):

- `extensions/cli/package.json`

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
high severity | Inefficient Algorithmic Complexity <br/>[SNYK-JS-JSYAML-18313070](https://snyk.io/vuln/SNYK-JS-JSYAML-18313070) | &nbsp;&nbsp;**828**&nbsp;&nbsp; 

#### Breaking Change Risk

Merge Risk: High

> **Notice:** This assessment is enhanced by AI.

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/continue-dev-inc.-default/project/e65f7c5c-f4c9-43c7-a439-4aa9a53f831f?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates?utm_source=github&utm_content=fix-pr-template)
🛠 [Adjust project settings](https://app.snyk.io/org/continue-dev-inc.-default/project/e65f7c5c-f4c9-43c7-a439-4aa9a53f831f?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://docs.snyk.io/scan-with-snyk/snyk-open-source/manage-vulnerabilities/upgrade-package-versions-to-fix-vulnerabilities?utm_source=github&utm_content=fix-pr-template)

---

**Learn how to fix vulnerabilities with free interactive lessons:**

🦉 [Learn about vulnerability in an interactive lesson of Snyk Learn.](https://learn.snyk.io/?loc&#x3D;fix-pr)

[...]

PR #13031, opened 2026-07-27
[Snyk] Security upgrade @electron/rebuild from 4.0.3 to 4.0.4

sestinj:
snyk-top-banner

#### Snyk changed the following file(s):

- `extensions/vscode/package.json`

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
high severity | Allocation of Resources Without Limits or Throttling <br/>[SNYK-JS-BRACEEXPANSION-18313044](https://snyk.io/vuln/SNYK-JS-BRACEEXPANSION-18313044) | &nbsp;&nbsp;**828**&nbsp;&nbsp; 

#### Breaking Change Risk

Merge Risk: Medium

> **Notice:** This assessment is enhanced by AI.

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/continue-dev-inc.-default/project/7ff64b12-9373-49d8-a6a5-70de1609223c?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates?utm_source=github&utm_content=fix-pr-template)
🛠 [Adjust project settings](https://app.snyk.io/org/continue-dev-inc.-default/project/7ff64b12-9373-49d8-a6a5-70de1609223c?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://docs.snyk.io/scan-with-snyk/snyk-open-source/manage-vulnerabilities/upgrade-package-versions-to-fix-vulnerabilities?utm_source=github&utm_content=fix-pr-template)

---

**Learn how to fix vulnerabilities with free interactive lessons:**

🦉 [Allocation of Resources Without Limits or Throttling](https://learn.snyk.io/lesson/no-rate-limiting/?source&#x3D;registry-oss-fix-pr&amp;loc&#x3D;fix-pr)

[...]

PR #13037, opened 2026-07-27
fix(core): recognise Windows backslash paths in grep_search output

### LINKED ISSUE #13027: grep_search tool returns no results on Windows due to backslash path separator in ripgrep output

### Relevant environment info

```Markdown
- OS: Windows 7
- Continue version: v2.0.0
- IDE version: VSCodium 1.116.02821
- Model: Unsloth Studio Local
```

### Description

The grep_search agent tool always returns "The search returned no results" on Windows, even when ripgrep finds matches. If I use Process Monitor to catch rg.exe being run, and run it manually, I get search results. If I search the same string in VSCode I also get search results. 

The bug is in `core/util/grepSearch.ts` in `formatGrepSearchResults`. The parser checks for file header lines using:

> if (line.startsWith("./") || line === "--") {

On Windows, ripgrep outputs paths with backslashes (.\templates\foo\bar.tpl) instead of forward slashes (./templates/foo/bar.tpl). The startsWith("./") check never matches, so numResults stays 0 and all results are silently discarded despite the content being present.

I added `|| line.startsWith(".\\")` to this check, and now I get the search results I expect.

### To reproduce

1. Ask the agent to search for a string that exists in the codebase using the grep_search tool
2. Observe "The search returned no results" even though the string is present

### Log output

```Shell

```

guillaume-flambard:
## Problem

Fixes #13027.

`grep_search` always returns "The search returned no results" on Windows, even when ripgrep finds matches (verified by the reporter with Process Monitor: `rg.exe` runs and returns results, but the tool reports none).

## Root cause

`formatGrepSearchResults` in `core/util/grepSearch.ts` detects ripgrep file-heading lines with:

```ts
if (line.startsWith("./") || line === "--") {
```

ripgrep emits headings as the relative path — `./dir/file` on POSIX, but `.\dir\file` on Windows. The `startsWith("./")` check never matches the Windows form, so no line is counted as a heading, `numResults` stays `0`, and every match is silently discarded. Diagnosed precisely in the issue by @SpikedCola.

## Fix

Also accept the `.\` prefix:

```ts
if (line.startsWith("./") || line.startsWith(".\\") || line === "--") {
```

POSIX `./` headings and the `--` context separator are unchanged.

## Tests

Added a unit test in `core/util/grepSearch.vitest.ts` that feeds backslash-separated ripgrep output and asserts `numResults > 0` and the filenames are present — it fails before the change (0 results) and passes after.

I verified the header-detection logic in isolation (the Windows heading goes from unrecognised → recognised, POSIX and `--` still recognised). I don't have a Windows machine to run the full agent end-to-end, but the parser is pure and the added test covers the regression in CI.

### DIFF
--- a/core/util/grepSearch.ts
+++ b/core/util/grepSearch.ts
@@ -57,7 +57,9 @@ export function formatGrepSearchResults(
 
   let resultLines: string[] = [];
   for (const line of results.split("\n").filter((l) => !!l)) {
-    if (line.startsWith("./") || line === "--") {
+    // ripgrep headings start with the relative path: "./" on POSIX, ".\" on Windows.
+    // Only checking "./" silently discarded every match on Windows (#13027).
+    if (line.startsWith("./") || line.startsWith(".\\") || line === "--") {
       processResult(resultLines); // process previous result
       resultLines = [line];
       numResults++;
--- a/core/util/grepSearch.vitest.ts
+++ b/core/util/grepSearch.vitest.ts
@@ -215,3 +215,21 @@ test("decreases indentation when original is more than 2 spaces", () => {
   expect(result.formatted).toContain("    tooMuchIndent();");
   expect(result.formatted).toContain("  }");
 });
+
+test("parses Windows ripgrep output with backslash path separators", () => {
+  // On Windows, ripgrep emits headings like `.\dir\file.ext` instead of `./dir/file.ext`.
+  // The parser must recognise those as file headers, otherwise every match is discarded (#13027).
+  const windowsOutput = [
+    ".\\src\\program.cs",
+    '        Console.WriteLine("Hello World!");',
+    "--",
+    ".\\test\\calc.kt",
+    "    fun subtract(number: Double): Test {",
+  ].join("\n");
+
+  const result = formatGrepSearchResults(windowsOutput);
+
+  expect(result.numResults).toBeGreaterThan(0);
+  expect(result.formatted).toContain("program.cs");
+  expect(result.formatted).toContain("calc.kt");
+});


SpikedCola:

guillaume-flambard:

guillaume-flambard:
Heads-up on CI: `jetbrains-tests` is red here, but it's red on every other open PR right now too (#13036, #13034, #13033, #13032, #13031) — a pre-existing failure unrelated to this change. This PR only touches `core/util/grepSearch.ts` (pure TS parsing); `core-checks`, which runs the grep unit tests including the new backslash-path case, is green. The `build-and-upload-vsix (darwin)` cancel and `require-all-checks-to-pass` failure are both cascades from that jetbrains job.

LHMQ878:
Nice find — this is the site the issue names and your diagnosis of it is right.

Heads-up that the same `./` assumption appears in three more places, so this fix on its own leaves the tool broken on Windows in one mode. I've opened #13042 for the remainder rather than touching yours:

- `splitGrepResultsByFile` (`core/tools/implementations/grepSearch.ts`) matches headings with `/^\.\/([^\n]+)$/gm`. With `splitByFile: true`, your fix moves Windows from *"no results"* to **returning an empty array of context items** — arguably worse, since it reads as a successful call.
- That function's captured path becomes a `type: "file"` ContextItem URI, which gets glob-matched for rule application — a backslash escapes in a glob, so `src\calc.ts` matches nothing.
- The `maxResults` truncation in `extensions/vscode/src/VsCodeIde.ts` splits on `(\n--|\n\.\/)` and finds no boundaries on Windows.

#13042 pulls the predicate into a shared `isGrepResultPathLine()` + `GREP_RESULT_PATH_PREFIX_SOURCE` in `core/util/grepSearch.ts` so the sites can't drift apart again, which means it overlaps your one-line change. Your PR is first — if maintainers merge it, I'll rebase mine and drop that hunk. Entirely happy the other way round too, or to fold everything into yours if you'd prefer to carry it. Just flagging so neither of us gets closed as a duplicate.

guillaume-flambard:
Thanks @LHMQ878, this is a great catch and your breakdown is exactly right. I only fixed the detection in `formatGrepSearchResults`, and you are correct that with `splitByFile: true` that leaves the tool returning an empty result set on Windows, which reads as success and is arguably worse than the original "no results". The glob backslash-escaping and the `VsCodeIde.ts` `maxResults` split are both real too.

Pulling the predicate into a shared `isGrepResultPathLine()` + `GREP_RESULT_PATH_PREFIX_SOURCE` in `core/util/grepSearch.ts` is the right call so these four sites can't drift apart again. I am happy to keep this PR minimal and let it merge first, then you rebase #13042 and fold my one-liner into the shared helper. If the maintainers would rather have it all in one place, I am also fine closing this in favour of #13042 and having the fix land there. Whatever keeps it clean and avoids a duplicate on your side.

Either way, thanks for opening #13042 and flagging it here rather than quietly overlapping. Ping me if a review on it would help.

LHMQ878:
Thanks for the generous reply, and for offering either way — my vote is your **first** option: **land #13037 as-is, I rebase #13042 on top.** Reasons, in order of how much they matter:

1. **You found it first.** #13037 predates #13042 by ~18h and correctly identifies the site the issue names. That should be in the history as yours.
2. **Your PR is the low-risk half.** One predicate widened, with a test. It can go in on its own review budget. #13042 touches a context-item URI shape and a VS Code truncation path — that deserves separate scrutiny and shouldn't hold up the fix for the reported symptom.
3. **The rebase is cheap on my side.** Your line becomes `isGrepResultPathLine(line)`; the helper and `GREP_RESULT_PATH_PREFIX_SOURCE` land in #13042 with the other three sites. One-line conflict, and I'll keep your comment text — it explains the `#13027` link better than mine did.

One thing worth flagging to whoever triages this: there is now a **third** PR on the same line. #13043 (opened this morning) changes `core/util/grepSearch.ts:60` to exactly

```diff
-    if (line.startsWith("./") || line === "--") {
+    if (line.startsWith("./") || line.startsWith(".\\") || line === "--") {
```

which is character-identical to yours minus the comment. So #13037 and #13043 are true duplicates of each other, and #13042 is the superset. Cleanest outcome is probably: merge #13037, close #13043 as a duplicate of it, then review #13042 for the remaining three sites. I have no stake in which of #13037/#13043 wins — I'm flagging it only so a maintainer doesn't merge one and then get a confusing conflict from the other.

On CI, for anyone reading this thread: your read is right and it has not changed. `jetbrains-tests` fails on `Autocomplete.kt:42` on every open PR I sampled, including the maintainers' own automated dependency bumps (#13034, #13033, #13032, all by @sestinj). It's a JetBrains UI starter test that needs a live completion; nothing in any of these three PRs can reach it. `require-all-checks-to-pass` is just its aggregator.

And yes — a review on #13042 would genuinely help, especially the `filepath.replace(/\\/g, "/")` in `splitGrepResultsByFile`. That normalisation is the one judgement call in the PR rather than a mechanical widening: those strings become `file` context-item URIs which get glob-matched for rule application, and in a glob a backslash is an escape character, not a separator. I believe normalising is correct there, but it is the part I'd most want a second opinion on.

guillaume-flambard:
Status note for reviewers: core/vscode checks are green. The failing `jetbrains-tests` / win32 VSIX jobs look unrelated (Autocomplete e2e flake / infra — same pattern on other recent PRs). Agreed with @LHMQ878 to land this detection fix as-is; their #13042 covers the remaining Windows path sites.

guillaume-flambard:
The red check here is `jetbrains-tests`, failing on `Autocomplete > testAutocomplete()` (`Autocomplete.kt:42`). This PR touches `core/` path parsing in `grep_search` output and does not go near that path.

It is also not specific to this PR: 29 of the last 30 `PR Checks` runs on the repo are red, including runs on `main` and on your own `snyk-fix-*` branches. The branch is up to date with `main`, which has not moved since 21 July, so there is nothing to rebase here.

Ready for review whenever the CI settles. Say the word if you want me to rebase onto another branch.

PR #13039, opened 2026-07-27
fix: composer input overflow with long unbroken strings

shubhook:
## Description

Fixed composer input overflow when typing or pasting long unbroken strings (no spaces). 
The text was overflowing the input area causing layout shift and send button misalignment.

Added `overflow-wrap: break-word` and `word-break: break-word` to `.tiptap` class in 
`TipTapEditor.css`, and `min-width: 0` to the flex container in `ContinueInputBox.tsx`.

Fixes #13016

## AI Code Review

- **Team members only**: AI review runs automatically when PR is opened or marked ready for review
- Team members can also trigger a review by commenting `@continue-review`

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [x] The relevant docs, if any, have been updated or created
- [x] The relevant tests, if any, have been updated or created

## Screen recording or screenshot

Before:
Screenshot 2026-07-27 at 22 01 24

After:
Screenshot 2026-07-27 at 22 02 41

Diff:
Screenshot 2026-07-27 at 22 26 53

## Tests

This is a CSS-only fix with no behavioral logic changes. No new tests were added as 
the fix is purely visual and verified manually by reproducing the overflow with a 
100+ character unbroken string before and after the change.

PR #12579, opened 2026-06-08
docs: add QVAC to model providers

simon-iribarren:
## Description

Adds a provider page for [QVAC](https://qvac.com), an open-source, local-first runtime that exposes an OpenAI-compatible HTTP server via `qvac serve openai`. Because the server speaks the OpenAI API, Continue connects to it through the built-in `openai` provider with a local `apiBase` — no new provider type is needed.

The page mirrors the existing local / OpenAI-compatible provider docs (e.g. OpenVINO Model Server, vLLM) and covers:

- Installing the CLI (`npm i -g @qvac/cli`) and starting the server (`qvac serve openai`, default port `11434`).
- Defining model aliases in `qvac.config.json` and matching the `model` field to a serve alias.
- A `config.yaml` / `config.json` example using `provider: openai`, `apiBase: http://127.0.0.1:11434/v1`, and a placeholder `apiKey`.
- Recommendations: set `ctx_size` explicitly (the default is small), set `reasoning_budget: 0` for reasoning-tuned models like Qwen3, and use a capable agent-tuned model such as `gpt-oss-20b` for tool-calling quality.

The page is registered in the **More Providers** navigation and added to the **Local Model Options** table in the providers overview.

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [x] The relevant docs, if any, have been updated or created
- [ ] The relevant tests, if any, have been updated or created

## Screen recording or screenshot

Docs-only change.

## Tests

Docs-only change. Verified `docs/docs.json` remains valid JSON after adding the navigation entry.

---
## Summary by cubic
Add QVAC to the model providers docs with a new page on running its OpenAI-compatible server and connecting via the built-in `openai` provider using a local `apiBase`. Updates navigation and the providers overview, switches docs to `qvac.tether.io`, and adds a note for reasoning-tuned models (e.g., Qwen3.5).

- **New Features**
  - New QVAC page: install `@qvac/cli`, start with `qvac serve openai` (default port 11434).
  - Config examples using `provider: openai`, `apiBase: http://127.0.0.1:11434/v1`, any non-empty `apiKey`, and model aliases from `qvac.config.json`.
  - Recommendations: set `ctx_size`; set `reasoning_budget: 0` for reasoning-tuned models like Qwen3.5; use a capable model like `gpt-oss-20b` for tools.
  - Added QVAC to the “More Providers” nav and “Local Model Options” table.

[...]

PR #13042, opened 2026-07-28
fix(core): handle Windows separators in the remaining grep_search parsers

LHMQ878:
## Description

Follow-up to #13027. **This is complementary to @guillaume-flambard's #13037, not a replacement** — that PR fixes the one site the issue named, this one covers the three that were left, including a worse variant of the same bug.

ripgrep is invoked with `.` as the search root, so with `--heading` it echoes that root back on every file-heading line using the platform separator: `./path` on POSIX, `.\path` on Windows. **Four** independent parsers of that output assumed `./`:

| # | Site | Effect on Windows | Covered by |
|---|------|-------------------|------------|
| 1 | `core/util/grepSearch.ts:60` — `formatGrepSearchResults` heading check | `numResults` stays 0 → "The search returned no results." | #13037 (and here, refactored into a shared helper) |
| 2 | `core/tools/implementations/grepSearch.ts` — `splitGrepResultsByFile` heading regex + strip regex | With `splitByFile: true`, returns an **empty array** of context items — not even the "no results" message | **this PR** |
| 3 | Same function — captured path becomes a `type: "file"` ContextItem URI | Backslash path breaks glob rule-matching | **this PR** |
| 4 | `extensions/vscode/src/VsCodeIde.ts:620` — `maxResults` truncation `matchAll` | No boundaries found → the result limit cannot be honoured | **this PR** |

Site 2 is the one worth flagging: fixing only site 1 makes `splitByFile: true` go from "no results" to *silently returning zero context items*, which reads as success to the caller. Verified by control experiment below.

Site 3 is a consequence of fixing site 2 rather than extra scope. That captured path is consumed as a `file` ContextItem URI, which `core/llm/rules/getSystemMessageWithRules.ts:288-290` glob-matches to decide which rules apply — and in a glob a backslash **escapes** the next character rather than separating segments, so `src\calc.ts` would match no rule pattern. It is normalised to forward slashes.

Both IDE integrations pass `.` as the ripgrep root (`extensions/vscode/src/VsCodeIde.ts:609`, `extensions/intellij/.../IntelliJIde.kt:488`), so both feed `.\` into the shared core parser on Windows — this is not VS Code-specific.

To stop the four sites from drifting apart again, the predicate and its regex equivalent now live together in `core/util/grepSearch.ts`:

```ts
export function isGrepResultPathLine(line: string): boolean {
  return line.startsWith("./") || line.startsWith(".\\");
}

/** Regex source for the same prefix, for callers that need it inside a larger pattern. */
export const GREP_RESULT_PATH_PREFIX_SOURCE = "\\.[\\\\/]";
```

If #13037 lands first, site 1 here is a trivial conflict — happy to rebase onto it and drop that hunk, or to have this rebased however you prefer.

## Tests

`core/util/grepSearch.vitest.ts` — 4 added (9 → 13):

- `formats Windows-style backslash headings` — nested `.\src\test.kt` alongside `--` separators
- `counts Windows headings so results are not reported as empty` — asserts the reported symptom directly
- `handles mixed separators in a single result set` — multi-root workspaces concatenate one ripgrep run per directory
- `does not treat a bare relative path as a heading` — negative case: a content line starting with `.` (`.method()`, `...spread`) must not open a new result

`core/tools/implementations/grepSearch.vitest.ts` — **new file, 5 tests**. `grepSearchImpl` had no test file at all, which is why sites 2–4 went unnoticed. Covers POSIX and Windows for both `splitByFile` modes, the URI normalisation, and that a genuinely empty search still reports empty.

### Reviewer test plan

```bash
cd core
npx vitest run util/grepSearch.vitest.ts tools/implementations/grepSearch.vitest.ts
```

→ **18 passed** (13 + 5).

### Control experiment — do the tests actually catch the bug?

Reverting each source file in isolation, leaving the tests in place:

```bash
git stash push -- core/util/grepSearch.ts
cd core && npx vitest run util/grepSearch.vitest.ts
```
→ 3 fail (`expected 1 to be 3`, `expected +0 to be 1`, `expected 1 to be 2`); the 4th is the negative case, which correctly still passes.

```bash
git stash push -- core/tools/implementations/grepSearch.ts
cd core && npx vitest run tools/implementations/grepSearch.vitest.ts
```
→ 1 fail: `expected [] to have a length of 2 but got +0` — the empty-array symptom, reproduced.

### Regression / lint

- `npx vitest run tools/ util/` — **356 passed, 16 failed**. Clean `main` baseline on the same machine: **347 passed, 16 failed**. Identical failure set (pre-existing POSIX path assumptions in `resolveWorkingDirectory.vitest.ts` and 3 other files, unrelated to grep); my change adds +9 passing and 0 failures.
- `npx tsc --noEmit -p core` — 151 errors, and clean-`main` baseline is also **151** (unbuilt `@continuedev/*` workspace packages). No new errors; `grep -c grepSearch` on the output → 0.
- `npx eslint --max-warnings 0` on all changed files — clean. `npx prettier --check` — "All matched files use Prettier code style!"

### Tested on

| Platform | Unit tests | Real ripgrep end-to-end |
|---|---|---|
| Windows 11 | ✅ | ⚠️ not run |
| macOS / Linux | ⚠️ not run | ⚠️ not run |

**Not validated:** the extension end-to-end with a live `rg.exe`, and the IntelliJ path. The parsers are pure functions and the tests feed them the exact byte sequences ripgrep emits, but I have not watched a real search return results in the UI. Worth a maintainer sanity-check on a Windows box before merge.

N/A — no UI change; this is parser behaviour covered by unit tests.

## Checklist

- [x] The relevant docs, if any, have been updated or created — none apply

Credit to @SpikedCola for pinpointing the root cause in #13027 with Process Monitor, and to @guillaume-flambard for #13037.

### DIFF
--- a/core/tools/implementations/grepSearch.ts
+++ b/core/tools/implementations/grepSearch.ts
@@ -1,7 +1,10 @@
 import { ToolImpl } from ".";
 import { ContextItem } from "../..";
 import { ContinueError, ContinueErrorReason } from "../../util/errors";
-import { formatGrepSearchResults } from "../../util/grepSearch";
+import {
+  formatGrepSearchResults,
+  GREP_RESULT_PATH_PREFIX_SOURCE,
+} from "../../util/grepSearch";
 import { prepareQueryForRipgrep } from "../../util/regexValidator";
 import { getStringArg } from "../parseArgs";
 
@@ -9,13 +12,21 @@ const DEFAULT_GREP_SEARCH_RESULTS_LIMIT = 100;
 const DEFAULT_GREP_SEARCH_CHAR_LIMIT = 7500; // ~1500 tokens, will keep truncation simply for now
 
 function splitGrepResultsByFile(content: string): ContextItem[] {
-  const matches = [...content.matchAll(/^\.\/([^\n]+)$/gm)];
+  // `.\path` on Windows, `./path` elsewhere — see isGrepResultPathLine.
+  const headingRegex = new RegExp(
+    `^${GREP_RESULT_PATH_PREFIX_SOURCE}([^\\n]+)$`,
+    "gm",
+  );
+  const matches = [...content.matchAll(headingRegex)];
 
   const contextItems: ContextItem[] = [];
 
   for (let i = 0; i < matches.length; i++) {
     const match = matches[i];
-    const filepath = match[1];
+    // Normalise separators: this becomes a `file` context-item URI, and those
+    // are glob-matched for rule application, where a backslash is an escape
+    // character rather than a separator.
+    const filepath = match[1].replace(/\\/g, "/");
     const startIndex = match.index!;
     const endIndex =
       i < matches.length - 1 ? matches[i + 1].index! : content.length;
@@ -23,7 +34,7 @@ function splitGrepResultsByFile(content: string): ContextItem[] {
     // Extract grepped content for this file
     const fileContent = content
       .substring(startIndex, endIndex)
-      .replace(/^\.\/[^\n]+\n/, "") // remove the line with file path
+      .replace(new RegExp(`^${GREP_RESULT_PATH_PREFIX_SOURCE}[^\\n]+\\n`), "") // remove the line with file path
       .trim();
 
     if (fileContent) {
--- a/core/tools/implementations/grepSearch.vitest.ts
+++ b/core/tools/implementations/grepSearch.vitest.ts
@@ -1,0 +1,86 @@
+import { expect, test, vi } from "vitest";
+
+import { ToolExtras } from "../..";
+
+import { grepSearchImpl } from "./grepSearch";
+
+function extrasReturning(results: string) {
+  return {
+    fetch: vi.fn() as any,
+    ide: {
+      getSearchResults: vi.fn().mockResolvedValue(results),
+    } as any,
+  } as unknown as ToolExtras;
+}
+
+// ripgrep is run with `.` as the search root, so `--heading` echoes that root
+// back using the platform separator: `./path` on POSIX, `.\path` on Windows.
+const posixResults = "./src/calc.ts\n  subtract(n) {\n    return this;";
+const windowsResults = ".\\src\\calc.ts\n  subtract(n) {\n    return this;";
+
+test("returns results for POSIX-style headings", async () => {
+  const result = await grepSearchImpl(
+    { query: "subtract" },
+    extrasReturning(posixResults),
+  );
+
+  expect(result).toHaveLength(1);
+  expect(result[0].content).toContain("subtract(n) {");
+});
+
+test("returns results for Windows-style headings", async () => {
+  // The reported bug: ripgrep found the match, but no heading was recognised,
+  // so numResults stayed 0 and the tool answered "no results" with the content
+  // sitting right there in its hand.
+  const result = await grepSearchImpl(
+    { query: "subtract" },
+    extrasReturning(windowsResults),
+  );
+
+  expect(result).toHaveLength(1);
+  expect(result[0].content).not.toBe("The search returned no results.");
+  expect(result[0].content).toContain("subtract(n) {");
+});
+
+test("splits Windows results per file and normalises the URI separators", async () => {
+  const result = await grepSearchImpl(
+    { query: "subtract", splitByFile: true },
+    extrasReturning(
+      `${windowsResults}\n.\\test.py\n  def subtract(self):\n    pass`,
+    ),
+  );
+
+  expect(result).toHaveLength(2);
+  // Forward slashes, even though ripgrep reported backslashes: this value is a
+  // `file` context-item URI, and those get glob-matched to decide which rules
+  // apply — and in a glob a backslash escapes the next character rather than
+  // separating path segments, so `src\calc.ts` would match nothing.
+  expect(result[0].uri).toEqual({ type: "file", value: "src/calc.ts" });
+  expect(result[1].uri).toEqual({ type: "file", value: "test.py" });
+  // The heading line itself is stripped from each chunk's content.
+  expect(result[0].content).toBe("subtract(n) {\n    return this;");
+  expect(result[1].content).toBe("def subtract(self):\n    pass");
+});
+
+test("splits POSIX results per file", async () => {
+  const result = await grepSearchImpl(
+    { query: "subtract", splitByFile: true },
+    extrasReturning(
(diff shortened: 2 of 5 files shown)


LHMQ878:
### CI status

**All 12 `test` jobs pass, including `test (windows-latest, {18,20,22,24})`** — which is the platform this PR is about, and more than I could verify locally.

Two red checks, neither caused by this PR:

- **`jetbrains-tests`** — `Autocomplete > testAutocomplete() FAILED` at `Autocomplete.kt:42`. Same failure on #13037 (2 files, core parser only) and on #13039 (a GUI overflow fix), so it's failing on `main` independently of the change. This PR touches no autocomplete code.
- **`require-all-checks-to-pass`** — aggregate of the above plus the CLA check; nothing of its own.
- **`CLAAssistant`** — on me, will sign.

So the only thing standing between this and green is the CLA.

LHMQ878:

LHMQ878:
### CI, final state

`CLAAssistant` now ✅. The two remaining red checks are a single pre-existing failure and its aggregator — quantified so you don't have to dig:

**`jetbrains-tests` is failing on every branch, not just this one.** Across the last 15 `PR Checks` runs on the repo: **0 success, 13 failure, 2 cancelled**. That includes six Snyk dependency-bump branches (`snyk-fix-*`), which touch no application code at all, plus `mcp-variable-interpolation`, `fix/composer-input-overflow`, and both of #13037's runs. The failure is always the same assertion:

```
Autocomplete > testAutocomplete() FAILED
    Caused by: org.opentest4j.AssertionFailedError at Autocomplete.kt:42
> Task :testIntegration FAILED
```

This PR touches no autocomplete code — it changes three ripgrep-output parsers and adds tests.

**`require-all-checks-to-pass`** is purely the aggregator. Its own log shows 10 of its 11 required jobs at `"result": "success"` and exactly one at `"result": "failure"` — `jetbrains-tests`. No independent failure of its own.

Everything attributable to this change is green, including all 12 `test` jobs across ubuntu/macos/**windows** × Node 18/20/22/24.

(Separately: `Autocomplete.kt:42` looks worth its own issue if you aren't already tracking it — it's been red for at least 15 consecutive runs, so `require-all-checks-to-pass` can't go green for any contributor right now. Happy to file one if that's useful, but I didn't want to bundle an unrelated report into this PR.)

LHMQ878:
The two red checks here are not from this PR, and I'd rather say so with the evidence than leave them looking like mine.

`jetbrains-tests` fails, and `require-all-checks-to-pass` is red only because it aggregates that one. The failure is a single IntelliJ UI integration test:

```
Autocomplete > testAutocomplete() FAILED
    com.intellij.ide.starter.driver.engine.DriverWithContextError at Autocomplete.kt:24
        Caused by: org.opentest4j.AssertionFailedError at Autocomplete.kt:42
1 test completed, 1 failed
```

I checked whether it is specific to this branch by tallying the same check across the 25 most recently updated PRs in the repo:

```
jetbrains-tests: 20 failure, 2 cancelled, 1 success
```

The failures include documentation-only PRs (#13066, #12579), which cannot affect autocomplete behaviour, and PRs touching unrelated areas of the GUI, the CLI and provider code. This PR changes `grep_search` output parsing in `core/`; it does not touch the JetBrains extension or the autocomplete path.

Nothing here needs doing on my side, but two things may be worth knowing: with `jetbrains-tests` red repo-wide, `require-all-checks-to-pass` is red on essentially every open PR, so the gate currently does not distinguish a passing branch from a failing one. The log also shows the run starting from a cold Gradle cache each time (`Failed to restore gradle-home-v1... Cache service responded with 400`), which may or may not be related to the flake.

Happy to rebase once it's green if that helps confirm the branch itself passes.

LHMQ878:
The red `jetbrains-tests` here is not from this PR, and `require-all-checks-to-pass` is red only because it gates on it.

`jetbrains-tests` failed on this PR's **base commit** (`5522c6f4`) before this branch existed:

```
GET /repos/continuedev/continue/commits/5522c6f4/check-runs?check_name=jetbrains-tests
  1 run: completed / failure / 2026-07-21T04:00:17Z
  https://github.com/continuedev/continue/actions/runs/29800025067/job/88539086045
```

This PR touches 5 files, none of them under `extensions/intellij`:

```
core/tools/implementations/grepSearch.ts          +15/-4
core/tools/implementations/grepSearch.vitest.ts   +86/-0
core/util/grepSearch.ts                           +20/-2
core/util/grepSearch.vitest.ts                    +54/-0
extensions/vscode/src/VsCodeIde.ts                 +7/-2
```

So nothing in this branch can turn it green. Flagging it rather than pushing an empty commit to re-run, since a re-run against the same base should fail the same way.

The vitest suites added here cover the change itself and are green. Happy to rebase onto a `main` where that job is passing if that is the preferred way to get a clean run.

LHMQ878:
The single red check here is `jetbrains-tests`, and it is a CI cache outage rather than anything from this PR. A re-run should clear it — flagging that because the run is stale and the outage has since largely passed.

**The job never got as far as running a test.** Its three annotations are:

```
warning  Node.js 20 is deprecated ... actions/cache@v4, actions/setup-java@v4.5.0, ...
warning  Failed to restore gradle-home-v1|Linux|jetbrains-tests[9f089130...]-{sha}
         Error: Cache service responded with 400
failure  Process completed with exit code 1.
```

`Cache service responded with 400` is GitHub's cache backend refusing the request; there is no test name, no assertion, and no compilation error anywhere in the output. `require-all-checks-to-pass` is the aggregate and reports the same single failure — its only annotation is `Process completed with exit code 1`.

**This PR touches `core/tools/implementations/grepSearch.ts`, `core/util/grepSearch.ts`, their two vitest files, and `extensions/vscode/src/VsCodeIde.ts` — nothing under `extensions/intellij`, and no Gradle or JVM file.** So even a genuine JetBrains test failure could not be attributed to this diff.

**It was repo-wide, and is now clearing up.** When I last checked on 2026-07-31, `jetbrains-tests` was red on 20 of the 21 most recently updated open PRs from other authors. Re-checking the same query today: **8 of 21**. That is consistent with a backend incident that has partly recovered, and it means a re-run now has a good chance of going green, whereas one on 2026-07-28 did not.

Everything else on this commit is green — 48 of the 50 check runs pass, with the rest skipped. `mergeable` is `true` and the branch already contains current `main`, so there is nothing to rebase. I can't trigger a re-run myself without write access; a maintainer clicking re-run on `jetbrains-tests` should be all that's needed.

PR #13044, opened 2026-07-28
fix: prevent layout shift when pasting long unbroken strings in composer input

### LINKED ISSUE #13016: Bug: Composer input overflow with long unbroken string causes layout shift and send button misalignment
**Describe the bug**

When typing or pasting a long unbroken string (no spaces, e.g., `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...` or a long base64 token) in the Composer input box, the text overflows the input area instead of wrapping. This causes the flex container to expand, shifting the entire composer layout and misaligning the send button relative to the input.

**To Reproduce**

1. Open Continue Composer in VS Code
2. Type or paste a long string without spaces (e.g., 100+ `a` characters) into the input box
3. Observe that the text extends beyond the input box boundary
4. The composer panel shifts horizontally, and the send button drifts out of alignment

**Expected behavior**

- Long unbroken strings should wrap inside the input box
- The composer layout should remain stable regardless of input content
- The send button should stay in its normal position

**Screenshots / Recordings**

N/A — easily reproducible with any long no-space string.

**Environment info**

- Continue version: latest (observed in recent versions)
- IDE: VS Code
- OS: all (cross-platform CSS issue)

**Root cause**

The Composer uses TipTap ([`TipTapEditor.tsx`](https://github.com/continuedev/continue/blob/main/gui/src/components/mainInput/TipTapEditor/TipTapEditor.tsx) + [`TipTapEditor.css`](https://github.com/continuedev/continue/blob/main/gui/src/components/mainInput/TipTapEditor/TipTapEditor.css)) as its input component.

1. **No text wrapping fallback** — The `.tiptap` styles in `TipTapEditor.css` lack `overflow-wrap: break-word` / `word-break: break-word`. By default, the browser does not break unbroken strings, so they overflow the contenteditable div.
2. **No min-width constraint on flex item** — The input wrapper or its flex container lacks `min-width: 0`, so the flex item's default `min-width: auto` prevents the container from shrinking below the content width. This propagates the overflow outward, shifting the layout.

**Suggested fix**

In `gui/src/components/mainInput/TipTapEditor/TipTapEditor.css`, add to the `.tiptap` class:

```css
.tiptap {
  overflow-wrap: break-word;
  word-break: break-word;
}
```

And on the flex container or input wrapper in `ContinueInputBox.tsx` (or equivalent), add `min-width: 0` to the appropriate flex child to allow proper shrinking.

**Additional context**

This is a classic CSS flexbox overflow issue. The fix is minimal, backwards-compatible, and has no side effects on normal input behavior (spaces still wrap as before; only unbroken strings are affected).

rekha0suthar:
## Problem

Pasting a long unbroken string (e.g. a base64 token, a minified JS bundle, a long URL with no spaces) into the composer input box caused the TipTap editor content to overflow its flex container. This pushed the send button out of view and broke the overall input layout.

## Root Cause

Two missing CSS properties:

1. **No `overflow-wrap`/`word-break` on `.tiptap`** — without these, the browser treats the unbroken string as a single unbreakable word and renders it at full width, overflowing the container.

2. **No `min-width: 0` on the flex container** — flex items default to `min-width: auto`, which means they will never shrink below their content width. This prevents `overflow-wrap` from working inside a flex layout.

## Fix

**`TipTapEditor.css`** — add a `.tiptap` rule:
```css
.tiptap {
  overflow-wrap: break-word;
  word-break: break-word;
  min-width: 0;
}
```

**`ContinueInputBox.tsx`** — add `min-w-0` to the outer flex container:
```tsx
<div className={`relative flex flex-col px-2 min-w-0`}>
```

## Testing

- Paste a long base64 string or a URL with no spaces into the input box
- Before: editor overflows, send button disappears off-screen
- After: text wraps correctly, layout stays intact

Fixes #13016

---
🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/gui/src/components/mainInput/ContinueInputBox.tsx
+++ b/gui/src/components/mainInput/ContinueInputBox.tsx
@@ -112,7 +112,7 @@ function ContinueInputBox(props: ContinueInputBoxProps) {
       className={`${props.hidden ? "hidden" : ""}`}
       data-testid={`continue-input-box-${props.inputId}`}
     >
-      <div className={`relative flex flex-col px-2`}>
+      <div className={`relative flex min-w-0 flex-col px-2`}>
         {props.isMainInput && <Lump />}
         <GradientBorder
           loading={isStreaming && (props.isLastUserInput || isInEdit) ? 1 : 0}
--- a/gui/src/components/mainInput/TipTapEditor/TipTapEditor.css
+++ b/gui/src/components/mainInput/TipTapEditor/TipTapEditor.css
@@ -16,6 +16,12 @@
   transition: background-color 0.2s ease-in-out;
 }
 
+.tiptap {
+  overflow-wrap: break-word;
+  word-break: break-word;
+  min-width: 0;
+}
+
 .tiptap p.is-editor-empty:first-child::before {
   color: #646464; /* lightGray */
   content: attr(data-placeholder);


rekha0suthar:

rekha0suthar:
recheck

rekha0suthar:
recheck

rekha0suthar:

Issue #13026, opened 2026-07-26
/model switch notices and error banners are stored as system-role messages   and replayed mid-conversation, breaking strict chat templates


garymarkowitz-source:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
  - Continue version: CLI (cn) v1.5.47, interactive TUI mode
  - IDE version: n/a (CLI)
  - Model: qwen/qwen3.6-35b-a3b (Q6_K, lmstudio-community) via LM Studio; second provider
  OpenRouter
  - config: two providers in one config — LM Studio (local, OpenAI-compatible) + OpenRouter
```

## Summary

When switching models mid-session with `/model`, cn records its UI notices — `Switched to model: <name>` and even transport errors like `Error: Connection error.` — as **`system`-role messages in the conversation history**, and replays them to the server inside the `messages` array of every subsequent request.

Many chat templates hard-reject a system message anywhere but position 0. Qwen3's template contains:

```jinja
{%- if message.role == "system" %}
    {%- if not loop.first %}
        {{- raise_exception('System message must be at the beginning.') }}
    {%- endif %}
```

so the first prompt after switching onto such a model fails with an HTTP 400 from LM Studio:

```
Unable to generate parser for this template. Automatic parser generation failed:
...
Error: Jinja Exception: System message must be at the beginning.
```

Every subsequent prompt to that model fails the same way for the rest of the session. `/clear` does **not** recover — the switch notice is re-injected (tested clearing both before and after the switch). Only restarting cn recovers.

The failure is easy to misdiagnose: whether it triggers depends on the *destination* model's template strictness (Qwen3.x Max via OpenRouter tolerates mid-history system messages; local Qwen3.6 35b does not), so it presents as "switching to my local provider breaks the session."

## Related issue

#12963 reports the same defect class (a `system`-role message injected mid-conversation, rejected with the identical "System message must be at the beginning" error) with a different trigger — tool-request cancellation in the IDE extension. This report covers the CLI's `/model` switch and error-banner paths, which that issue's proposed fix (scoped to the cancellation handler) would not address.

## Source pointer

[...]

PR #13046, opened 2026-07-28
fix(cli): stop sending UI notices to the provider as system messages

LHMQ878:
Closes part of #13026 — the classification half. Filed separately from #12988 as discussed on the issue: #12988 rewrites `convertFromUnifiedHistoryWithSystemMessage` in `core/util/messageConversion.ts`, and this PR doesn't touch that file at all, so the two don't overlap in the diff.

## The dead field

`extensions/cli/spec/wire-format.md:28` defines a discriminator:

```
"messageType": "tool-start" | "tool-result" | "tool-error" | "system",
```

and `:100` describes classifying tool messages by it. The selector hooks already set it — `useModelSelector.ts:65,81` and `useConfigSelector.ts:61,67,78,97,104`, all `messageType: "system"`. But there are **zero reads** of `messageType` anywhere under `extensions/cli/src`. It's declared in the spec and unimplemented.

## Why that causes the 400

`useTUIChatHooks.ts` receives the field and then drops it when building the history item:

```ts
onMessage: (message) => {
  setChatHistory((prev) => [
    ...prev,
    {
      message: { role: "system", content: message.content },  // messageType discarded
      contextItems: [],
    },
  ]);
},
```

So a `/model` notice becomes an ordinary `role: "system"` history entry. Once it's in the history nothing can tell it apart from a real system instruction — which is also why the renderer has to fall back to `message.role === "system"` (`MemoizedMessage.tsx:54`). On the next turn the conversion emits it as a second `role: "system"` message at a non-zero index, and providers reject that with *"System message must be at the beginning"*.

## The change

Wire the discriminator up rather than delete it:

- `useTUIChatHooks.ts` — keep `messageType` on the history item instead of discarding it.
- new `extensions/cli/src/uiNotices.ts` — `isUiNotice` / `withoutUiNotices`, reading **only** `messageType`, never `role`. A real system message and a notice both use `role: "system"`; that's precisely why the discriminator exists.
- `streamChatResponse.ts:253` — filter notices out immediately before the conversion, so they stay visible in the transcript but are never sent to the model.

`messageType` is typed CLI-locally (`ChatHistoryItemWithType`) rather than added to core's `ChatHistoryItem`, since the classification only means anything to the TUI.

## Verification

**Control experiment** — reverting only `streamChatResponse.ts` (the consumer wiring) and re-running:

```
× should not send UI notices to the provider as extra system messages
  → expected [ { role: 'system', …(1) }, …(1) ] to have a length of 1 but got 2
```

Two system messages — the exact shape that produces the 400. With the change: one, at index 0.

**Tests.** The end-to-end one is in the existing `streamChatResponse.systemMessage.test.ts`, asserting on what actually reaches `chatCompletionStream` rather than on the helper in isolation — a notice in the history must not appear in the request and must not add a second system message. Plus 7 unit tests in `uiNotices.test.ts` covering the `role`-vs-`messageType` distinction (a real `role: "system"` message is **not** filtered), ordering, and the no-notices case.

**Suite**, against a clean-`main` baseline (Windows, so there's pre-existing flake here):

| | baseline on `main` | with this change |
|---|---|---|
| `extensions/cli` full suite | 1667 passed, 5 failed / 3 files | 1678 passed, 2 failed / 1 file |

The 2 remaining failures are in `TUIChat.editMessage.test.tsx`, which fails on `main` too in a full parallel run and passes 27/27 when run on its own, on both branches — it's order-dependent, not related to this change. **No new failures.**

- `tsc --noEmit -p extensions/cli`: 43 errors, identical count to baseline, none in the files touched here (all pre-existing `core/` moduleResolution errors).
- `prettier --write` applied to all changed files.

## Note on ordering

If #12988 lands first this rebases cleanly; if this lands first, nothing in #12988 has to change. Worth mentioning that #12988 appears stuck on infrastructure rather than review — its own tests are green (27 success / 2 skipped, only the `require-all-checks-to-pass` aggregator red), and what's blocking it is 12 `Continuous AI: *` statuses that exist on that PR and on no other (#13004, #13005, #13037, #13042 all carry only the two snyk contexts).

### DIFF
--- a/extensions/cli/src/services/ChatHistoryService.ts
+++ b/extensions/cli/src/services/ChatHistoryService.ts
@@ -2,6 +2,7 @@ import type { ChatHistoryItem, ToolStatus } from "core/index.js";
 import { createHistoryItem } from "core/util/messageConversion.js";
 
 import { loadSessionById, updateSessionHistory } from "../session.js";
+import type { ChatHistoryItemWithType } from "../uiNotices.js";
 import { logger } from "../util/logger.js";
 
 import { BaseService } from "./BaseService.js";
@@ -205,13 +206,23 @@ export class ChatHistoryService extends BaseService<ChatHistoryState> {
   }
 
   /**
-   * Add a system message to the history
+   * Add a system message to the history.
+   *
+   * Everything that reaches this method is a notice addressed to the user —
+   * "Remote environment is shutting down…", `WARNING: Tool … requires
+   * permission`, command output, diffs — so the item is marked as a UI notice
+   * and is dropped before the history is converted to wire format. Callers that
+   * ever need to add a real system *instruction* should build the item
+   * themselves rather than reach for this.
    */
   addSystemMessage(content: string): ChatHistoryItem {
-    const newMessage = createHistoryItem({
-      role: "system",
-      content,
-    });
+    const newMessage: ChatHistoryItemWithType = {
+      ...createHistoryItem({
+        role: "system",
+        content,
+      }),
+      messageType: "system",
+    };
 
     const newHistory = [...this.currentState.history, newMessage];
     this.setHistoryInternal(newHistory);
--- a/extensions/cli/src/stream/streamChatResponse.autoCompaction.ts
+++ b/extensions/cli/src/stream/streamChatResponse.autoCompaction.ts
@@ -10,6 +10,7 @@ import {
   shouldAutoCompact,
 } from "../compaction.js";
 import { updateSessionHistory } from "../session.js";
+import { uiNotice } from "../uiNotices.js";
 import { formatError } from "../util/formatError.js";
 import { logger } from "../util/logger.js";
 
@@ -74,13 +75,7 @@ function handleCompactionSuccess(
     callbacks.setCompactionIndex(result.compactionIndex);
     callbacks.setMessages((prev: ChatHistoryItem[]) => [
       ...prev,
-      {
-        message: {
-          role: "system",
-          content: successMessage,
-        },
-        contextItems: [],
-      },
+      uiNotice(successMessage),
     ]);
   }
 }
@@ -105,13 +100,7 @@ function handleCompactionError(
   } else if (callbacks?.setMessages) {
     callbacks.setMessages((prev: ChatHistoryItem[]) => [
       ...prev,
-      {
-        message: {
-          role: "system",
-          content: warningMessage,
-        },
-        contextItems: [],
-      },
+      uiNotice(warningMessage),
     ]);
   }
 }
--- a/extensions/cli/src/stream/streamChatResponse.systemMessage.test.ts
+++ b/extensions/cli/src/stream/streamChatResponse.systemMessage.test.ts
@@ -4,6 +4,7 @@ import type { ChatHistoryItem } from "core/index.js";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { services } from "../services/index.js";
+import { uiNotice } from "../uiNotices.js";
 
 import { processStreamingResponse } from "./streamChatResponse.js";
 
@@ -214,6 +215,56 @@ describe("streamChatResponse system message validation", () => {
           : false,
       ),
     ).toBe(false);
+  });
+
+  it("should not send notices from the non-selector sites either", async () => {
+    const model = {
+      model: "test-model",
+      defaultCompletionOptions: { contextLength: 1000, maxTokens: 100 },
+    } as ModelConfig;
+
+    const systemMessage = "System instructions";
+
+    // The shape reported in #13026: an error banner sitting next to a /model
+    // notice. The banner comes from useChat.ts, a site that does not go through
+    // the selector hooks, so it is only filtered because every notice site now
+    // builds its item with uiNotice().
+    const chatHistory: ChatHistoryItem[] = [
+      { message: { role: "user", content: "hello" }, contextItems: [] },
+      uiNotice("Switched to model: qwen/qwen3.6-35b-a3b") as ChatHistoryItem,
+      uiNotice("Error: Connection error.") as ChatHistoryItem,
+      uiNotice(
+        "[Tool canceled - please tell Continue what to do differently]",
+      ) as ChatHistoryItem,
+      uiNotice("Chat history compacted successfully.") as ChatHistoryItem,
+      uiNotice("Session exported to /tmp/x.json") as ChatHistoryItem,
+      { message: { role: "user", content: "and now?" }, contextItems: [] },
+    ];
+
+    const mockStream = {
+      [Symbol.asyncIterator]: async function* () {
+        yield { choices: [{ delta: { content: "ok" } }] };
+      },
+    };
+    mockLlmApi.chatCompletionStream = vi.fn().mockResolvedValue(mockStream);
+
+    await processStreamingResponse({
+      chatHistory,
+      model,
+      llmApi: mockLlmApi,
+      abortController,
+      systemMessage,
+    });
+
+    const sent = vi.mocked(mockLlmApi.chatCompletionStream).mock.calls[0]?.[0];
+
+    expect(sent!.messages.filter((m) => m.role === "system")).toHaveLength(1);
+    expect(sent!.messages[0]?.content).toBe(systemMessage);
+    expect(sent!.messages.map((m) => m.content)).toEqual([
+      systemMessage,
+      "hello",
+      "and now?",
+    ]);
   });
 
   it("should use safety buffer in validation", async () => {
--- a/extensions/cli/src/stream/streamChatResponse.systemMessage.test.ts
+++ b/extensions/cli/src/stream/streamChatResponse.systemMessage.test.ts
@@ -160,6 +160,62 @@ describe("streamChatResponse system message validation", () => {
     expect(services.systemMessage.getSystemMessage).not.toHaveBeenCalled();
   });
 
+  it("should not send UI notices to the provider as extra system messages", async () => {
+    const model = {
+      model: "test-model",
+      defaultCompletionOptions: {
+        contextLength: 1000,
+        maxTokens: 100,
+      },
+    } as ModelConfig;
+
+    const systemMessage = "System instructions";
+
+    // A `/model` notice recorded in the history. It is role:"system" for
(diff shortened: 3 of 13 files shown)


LHMQ878:
### Fixed the lint failure

`lint` was red because of this PR, not a pre-existing issue:

```
extensions/cli/src/stream/streamChatResponse.ts
  23:1  error  `../uiNotices.js` import should occur before import of `../util/exponentialBackoff.js`  import/order
```

The new `withoutUiNotices` import was appended after the `../util/*` group; `import/order` wants it in alphabetical position (`../tools/` < `../uiNotices.js` < `../util/`). Moved it up one group in `ef91405` — no behaviour change, imports only.

The two `complexity` warnings in the same output (`streamChatResponse.ts:197`, `subagent/executor.ts:57`) are unused-eslint-disable directives that exist on `main` and are warnings, not errors — left alone.

`jetbrains-tests` is a pre-existing failure unrelated to this PR (`Autocomplete > testAutocomplete()` at `Autocomplete.kt:42`, failing across unrelated branches including dependency bumps), and `require-all-checks-to-pass` is its aggregator.

LHMQ878:
### CI after `ef91405`

`lint` is now ✅. The three remaining red checks are all environmental or pre-existing — none is caused by this PR:

| Check | Cause |
|---|---|
| `e2e/.../GUI.test.js (e2e:ci:run)` | npm `postinstall` failure, not a test failure: `@vscode/ripgrep` postinstall got `403` fetching `microsoft/ripgrep-prebuilt` release `v13.0.0-10` (GitHub API rate limit). This job **passed** on the previous head of this same PR, so it is flaky infra — a re-run should clear it. |
| `jetbrains-tests` | Pre-existing: `Autocomplete > testAutocomplete() FAILED` → `AssertionFailedError at Autocomplete.kt:42`. Fails across unrelated branches including dependency bumps; this PR touches no autocomplete code. |
| `require-all-checks-to-pass` | Aggregator of the two above; nothing of its own. |

Everything else is green, including `core-checks`, `prettier-check`, all `packages-checks`, and all 12 `test` jobs across ubuntu/macos/windows × Node 18/20/22/24.

Happy to re-run the GUI e2e job if a maintainer can trigger it — I can't from a fork.

LHMQ878:
### Pushed `7728d35` — covers the remaining notice sites

@garymarkowitz-source pointed out on #13026 that this PR as originally scoped only filtered the two selector hooks, so the error banner from his repro and #12963's tool-cancel notice both still reached the provider. That was correct. This push closes the gap.

**What changed.** A `uiNotice(content)` constructor in `uiNotices.ts` builds the history item with the discriminator set, and every notice site now goes through it instead of writing the object literal. That way adding a site can't silently omit the marker — an omission is invisible until a provider rejects the request.

Sites now marked, by file:

| File | Sites | Notices |
|---|---|---|
| `ui/hooks/useChat.ts` | 2 | error banner (the one in the reported payload), tool-cancel (#12963's trigger) |
| `ui/hooks/useChat.helpers.ts` | 3 | command output ×2, `Diff:…` |
| `ui/hooks/useChat.compaction.ts` | 3 | compaction succeeded / cancelled / failed |
| `ui/hooks/useChat.stream.helpers.ts` | 2 | stream error, `onSystemMessage` local fallback |
| `stream/streamChatResponse.autoCompaction.ts` | 2 | auto-compaction success, auto-compaction warning |
| `ui/TUIChat.tsx` | 3 | export succeeded / session-not-found / export failed |
| `services/ChatHistoryService.ts` | 1 (sink) | covers its 15 callers |

`ChatHistoryService.addSystemMessage` is marked at the sink rather than per-caller, because the item is built inside `createHistoryItem` and callers have no way to annotate it. Everything reaching that method is user-facing — `Remote environment is shutting down…`, `WARNING: Tool … requires permission`, `No changes to display`, diffs — so marking there is correct and covers all 15 call sites at once.

**Two `role: "system"` sites are deliberately left alone**, since they are real system messages and not notices:

- `commands/serve.ts:145` — seeds an empty history with the configured system prompt.
- `stream/streamChatResponse.autoCompaction.ts:170` — builds a throwaway item purely to count tokens; it never enters the history.

I did not widen `isUiNotice` to infer from `role` or content. Guessing would eventually drop a genuine system instruction, which fails silently and changes model behaviour — worse than the error this PR fixes.

### Testing done

`44 passed` across `uiNotices.test.ts` (8), `streamChatResponse.systemMessage.test.ts` (5), `useChat.clear.test.ts` (4), and the pre-existing `ChatHistoryService.test.ts` (27).

New test `should not send notices from the non-selector sites either` builds the exact shape from #13026 — a `/model` notice, an error banner, a tool-cancel notice, a compaction notice and an export notice around two user turns — and asserts on what reaches `chatCompletionStream`:

```js
expect(sent.messages.map((m) => m.content)).toEqual([
  systemMessage,
  "hello",
  "and now?",
]);
```

**Control experiment.** Removing only `messageType: "system"` from `uiNotice()`, leaving every call site as-is:

```
× uiNotice > produces an item that isUiNotice recognises and withoutUiNotices drops
  → expected undefined to be 'system'
× should not send notices from the non-selector sites either
  → expected [ { role: 'system', …(1) }, …(5) ] to have a length of 1 but got 6
```

Six system messages — all five notices plus the real prompt — which is the bug, quantified. Restoring the line returns it to 13/13 on those two files.

One pre-existing test needed updating: `useChat.clear.test.ts` asserts the `"Chat history cleared"` item by exact object equality, so it now expects `messageType: "system"`. That message is a notice (it's addressed to the user, and it comes from `slashCommands.ts:260` via the `result.output` path), so marking it is the intended behaviour rather than an accommodation.

Full suite: `1677 passed | 49 skipped`. Four files fail when the whole suite runs in parallel (`CtrlCProcessHandling`, `serve.test.ts`, `TUIChat.editMessage`) — all three pass when run on their own, and they also pass on the previous commit of this branch, so that's runner contention rather than anything here.

`eslint` reports 0 errors on the changed files. `prettier --check` is clean once line endings are normalised (my checkout is CRLF, which makes prettier flag untouched files like `slashCommands.ts` too).

### Note on repo status

@ryan-adtronics-llc mentioned on #13026 that the repo is read-only. That isn't the case — the API reports `archived: false` and this push went through. But activity does look stalled: last commit #13005 on 21 Jul, no merges since, and 2 of the last 15 closed PRs merged this month. Flagging so nobody's waiting on a review that may not come; I'd rather leave the fix complete either way.

@garymarkowitz-source — this is the branch to point your LM Studio / Qwen3.6 35b A3b setup at if the offer still stands. A strict Jinja template that raises on a mid-history system message is a better oracle than my mocks, since anything that slips through fails loudly instead of silently.

LHMQ878:
The two red checks here are not from this PR, and I'd rather say so with the evidence than leave them looking like mine.

`jetbrains-tests` fails, and `require-all-checks-to-pass` is red only because it aggregates that one. The failure is a single IntelliJ UI integration test:

```
Autocomplete > testAutocomplete() FAILED
    com.intellij.ide.starter.driver.engine.DriverWithContextError at Autocomplete.kt:24
        Caused by: org.opentest4j.AssertionFailedError at Autocomplete.kt:42
1 test completed, 1 failed
```

I checked whether it is specific to this branch by tallying the same check across the 25 most recently updated PRs in the repo:

```
jetbrains-tests: 20 failure, 2 cancelled, 1 success
```

The failures include documentation-only PRs (#13066, #12579), which cannot affect autocomplete behaviour, and PRs touching unrelated areas of the GUI, the CLI and provider code. This PR changes `grep_search` output parsing in `core/`; it does not touch the JetBrains extension or the autocomplete path.

Nothing here needs doing on my side, but two things may be worth knowing: with `jetbrains-tests` red repo-wide, `require-all-checks-to-pass` is red on essentially every open PR, so the gate currently does not distinguish a passing branch from a failing one. The log also shows the run starting from a cold Gradle cache each time (`Failed to restore gradle-home-v1... Cache service responded with 400`), which may or may not be related to the flake.

Happy to rebase once it's green if that helps confirm the branch itself passes.

LHMQ878:
The red `jetbrains-tests` here is not from this PR, and `require-all-checks-to-pass` is red only because it gates on it.

`jetbrains-tests` failed on this PR's **base commit** (`5522c6f4`) before this branch existed:

```
GET /repos/continuedev/continue/commits/5522c6f4/check-runs?check_name=jetbrains-tests
  1 run: completed / failure / 2026-07-21T04:00:17Z
  https://github.com/continuedev/continue/actions/runs/29800025067/job/88539086045
```

This PR touches 13 files, all under `extensions/cli`, none under `extensions/intellij`.

So nothing in this branch can turn it green. Flagging it rather than pushing an empty commit to re-run, since a re-run against the same base should fail the same way.

Same base commit as #13042, same pre-existing failure. Happy to rebase onto a `main` where that job is passing if that is the preferred way to get a clean run.

LHMQ878:
The single red check here is `jetbrains-tests`, and it is a CI cache outage rather than anything from this PR. A re-run should clear it — flagging that because the run is stale and the outage has since largely passed.

**The job never got as far as running a test.** Its three annotations are:

```
warning  Node.js 20 is deprecated ... actions/cache@v4, actions/setup-java@v4.5.0, ...
warning  Failed to restore gradle-home-v1|Linux|jetbrains-tests[9f089130...]-{sha}
         Error: Cache service responded with 400
failure  Process completed with exit code 1.
```

`Cache service responded with 400` is GitHub's cache backend refusing the request; there is no test name, no assertion, and no compilation error anywhere in the output. `require-all-checks-to-pass` is the aggregate and reports the same single failure — its only annotation is `Process completed with exit code 1`.

**This PR touches only files under `extensions/cli/src/` — nothing under `extensions/intellij`, and no Gradle or JVM file.** So even a genuine JetBrains test failure could not be attributed to this diff.

**It was repo-wide, and is now clearing up.** When I last checked on 2026-07-31, `jetbrains-tests` was red on 20 of the 21 most recently updated open PRs from other authors. Re-checking the same query today: **8 of 21**. That is consistent with a backend incident that has partly recovered, and it means a re-run now has a good chance of going green, whereas one on 2026-07-29 did not.

Everything else on this commit is green — 47 of the 49 check runs pass, with the rest skipped. `mergeable` is `true` and the branch already contains current `main`, so there is nothing to rebase. I can't trigger a re-run myself without write access; a maintainer clicking re-run on `jetbrains-tests` should be all that's needed.

PR #13078, opened 2026-08-03
fix(core): add pricing for current Claude models

asjad3:
## Description

`calculateAnthropicCost` matches the model against a prefix table and returns `null` when
nothing matches, so any model missing from the table shows no cost at all in the Console's
usage panel and in the CLI's per-request summary.

The table stops at Opus 4.6 / Sonnet 4.6, so most of the Claude models people are actually
running today fall through it:

| Model | Before | After |
|---|---|---|
| `claude-opus-5` | no cost shown | $5 / $25 |
| `claude-opus-4-8` | no cost shown | $5 / $25 |
| `claude-opus-4-7` | no cost shown | $5 / $25 |
| `claude-sonnet-5` | no cost shown | $3 / $15 |
| `claude-sonnet-4-5` | no cost shown | $3 / $15 |
| `claude-haiku-4-5` | no cost shown | $1 / $5 |
| `claude-fable-5` | no cost shown | $10 / $50 |

Sonnet 4.5 and Haiku 4.5 are the ones I'd expect to bite people most — Opus 4.5 is in the
table but the two Claude 4.5 models next to it never were.

Cache-write and cache-read rates follow the same 1.25x / 0.1x relationship to input price
that every existing row uses.

Also relabelled three comments that named the wrong model (`claude-3-opus` was commented
"Claude Opus 4 (legacy)", `claude-3-5-sonnet` was "Claude Sonnet 4"). Comments only, no
behaviour change.

Prices are Anthropic's list prices. Sonnet 5 currently has promotional input/output pricing
running to 2026-08-31; I used the list price rather than the promo, since the table has no
way to express a time-boxed rate and the promo expires shortly.

## Checklist

Not applicable — no UI change. The affected surfaces (`TotalUsage.tsx`, `useLLMSummary`,
`useTotalUsage`, and the CLI's `streamChatResponse.helpers`) all just consume the return
value of `calculateRequestCost`, which was previously `null` for these models.

## Tests

Added 8 cases to `calculateRequestCost.vitest.ts` covering each newly-priced model, plus a
combined cache-read + cache-write case for Opus 5.

```
core/llm/utils/calculateRequestCost.vitest.ts   30 passed
```

With the test file kept and the change to `calculateRequestCost.ts` stashed, exactly the 8
new cases fail (`8 failed | 22 passed`), so they're pinning the fix rather than passing
either way.

The existing `claude-unknown-model` -> `null` case still passes; none of the added prefixes
match it.

---

Written with AI assistance (Claude Code); I reviewed the change and ran the tests above.

### DIFF
--- a/core/llm/utils/calculateRequestCost.ts
+++ b/core/llm/utils/calculateRequestCost.ts
@@ -17,12 +17,68 @@ function calculateAnthropicCost(
     string,
     { input: number; output: number; cacheWrite: number; cacheRead: number }
   > = {
+    // Claude Fable 5
+    "claude-fable-5": {
+      input: 10,
+      output: 50,
+      cacheWrite: 12.5,
+      cacheRead: 1,
+    },
+
+    // Claude Opus 5
+    "claude-opus-5": {
+      input: 5,
+      output: 25,
+      cacheWrite: 6.25,
+      cacheRead: 0.5,
+    },
+
+    // Claude Opus 4.8
+    "claude-opus-4-8": {
+      input: 5,
+      output: 25,
+      cacheWrite: 6.25,
+      cacheRead: 0.5,
+    },
+
+    // Claude Opus 4.7
+    "claude-opus-4-7": {
+      input: 5,
+      output: 25,
+      cacheWrite: 6.25,
+      cacheRead: 0.5,
+    },
+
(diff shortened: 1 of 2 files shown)


asjad3:

PR #13095, opened 2026-08-06
fix: keep code-block toolbar within panel bounds when sidebar is narrowed

### LINKED ISSUE #13085: Code block header bar and explanation text do not resize correctly; action buttons hidden when sidebar is narrowed (Windows 11, VS Code 1.131, Continue 2.1.0)

### Relevant environment info

```Markdown
OS: Windows 11

VS Code: 1.131.0

Continue extension: 2.1.0

Gemini 2.5 Flash
```

### Description

Issue:
In the Continue sidebar/panel, when resizing the panel width:

The content inside code blocks resizes and wraps correctly.

The header bar of code blocks (the top bar with Copy, Run, Enter buttons) does not resize correctly. As the panel is narrowed, the header extends beyond the visible area and the action buttons disappear under the sidebar edge.

Explanation text between code blocks also does not reflow correctly and can overflow or appear under other UI elements.

This appears to be a CSS/layout bug in the Continue webview, likely due to fixed/min-width styling on the code block header and/or missing overflow handling on the sidebar container.

Steps to reproduce:

### To reproduce

Open Continue in the sidebar or panel.

Ask a question that produces a response with:

Explanation text

One or more code blocks (with header bar containing Copy/Run/Enter buttons)

Resize the Continue panel narrower.

Observe:

Code block content wraps correctly.

Code block header bar does not shrink; action buttons get hidden under the sidebar edge.

Explanation text does not reflow properly and may overflow or render under other elements.

Expected behavior:
All elements (code block headers, code content, and explanation text) should stay within the Continue panel bounds when resized. The header bar should shrink or wrap, and overflow should be handled with scrolling instead of elements disappearing under the sidebar edge.

Tried:

Updated Continue to 2.1.0

Disabled hardware acceleration ("disable-hardware-acceleration": true in argv.json)

Moved Continue between primary sidebar, secondary sidebar, and bottom panel

Reloaded window / restarted VS Code

Note:
This looks like a CSS layout bug in the extension’s sidebar. Likely causes:

Code block header has a fixed or min width that doesn’t respond to container size.

Sidebar container may be missing proper height: 100% / overflow-y: auto or has overflow: hidden on some parent.

Screenshots or a short screen recording showing the header bar and buttons being clipped as the panel is narrowed would be very helpful for reproduction.

Image

Image

### Log output

```Shell
The continue console only includes prompt and options.
```

bunnysayzz:
## Description

Fixes #13085 — when the Continue sidebar/panel is narrowed, the code-block header bar (Copy/Run/Enter buttons) does not resize and the action buttons get hidden under the sidebar edge. Explanation text between code blocks could also overflow.

**Root cause:** the toolbar header is a flex row whose children have flexbox's default `min-width: auto`, which prevents them from shrinking. The root container has `overflow-x: hidden`, so when the panel narrows, the button row is pushed past the visible area and clipped.

**Fix (in `StepContainerPreToolbar`):**
- Title side (filepath/language + chevron): added `min-w-0` so it can shrink; the filepath already truncates via `line-clamp-1 break-all`.
- Action-button side: added `shrink-0` so Copy/Run/Apply buttons are never clipped.
- Header row: added `flex-wrap` + `gap-x-3 gap-y-1` so on very narrow widths the buttons wrap to a second line instead of overflowing.

This targets the code-block header specifically (the issue's primary symptom). Explanation-text reflow stems from the same overflow root and is also relieved once the block no longer forces itself wider than the panel.

## AI Code Review

## Checklist

When the sidebar is narrower than the code-block toolbar, the buttons now stay visible (title truncates / buttons wrap) instead of being clipped at the panel edge.

## Tests

Added `gui/src/components/StyledMarkdownPreview/StepContainerPreToolbar/index.test.tsx`:
- smoke test (component renders)
- regression test asserting the title container has `min-w-0` (shrinkable) and the action-button container has `shrink-0` (never clipped)

All 44 tests in `StyledMarkdownPreview/` pass (`npx vitest run src/components/StyledMarkdownPreview/`). `tsc --noEmit` shows only pre-existing workspace-package resolution errors in `core/` (unbuilt `@continuedev/config-yaml`), none in the changed files.

### DIFF
--- a/gui/src/components/StyledMarkdownPreview/StepContainerPreToolbar/index.test.tsx
+++ b/gui/src/components/StyledMarkdownPreview/StepContainerPreToolbar/index.test.tsx
@@ -1,0 +1,79 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import { StepContainerPreToolbar } from "./index";
+
+vi.mock("../../../context/IdeMessenger", () => {
+  const { createContext } = require("react");
+  return {
+    IdeMessengerContext: createContext({
+      ide: {
+        getCurrentFile: vi.fn().mockResolvedValue(undefined),
+        showToast: vi.fn(),
+        showLines: vi.fn(),
+      },
+      post: vi.fn(),
+    }),
+  };
+});
+
+vi.mock("../../../redux/hooks", () => ({
+  useAppSelector: () => ({}),
+}));
+
+vi.mock("../../../hooks/useWebviewListener", () => ({
+  useWebviewListener: () => {},
+}));
+
+vi.mock("../../../hooks/useIdeMessengerRequest", () => ({
+  useIdeMessengerRequest: () => ({
+    result: undefined,
+    refresh: vi.fn(),
+    isLoading: false,
+  }),
+}));
+
+vi.mock("../../../redux/selectors/selectToolCalls", () => ({
+  selectToolCallById: () => undefined,
+}));
+
+vi.mock("../../../redux/slices/sessionSlice", () => ({
+  selectApplyStateByStreamId: () => undefined,
+  selectApplyStateByToolCallId: () => undefined,
+}));
+
+describe("StepContainerPreToolbar", () => {
+  const baseProps = {
+    codeBlockContent: "console.log('hello')",
+    language: "javascript",
+    codeBlockIndex: 0,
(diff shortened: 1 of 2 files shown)


bunnysayzz:
- - -
- - -

bunnysayzz:

PR #13115, opened 2026-08-10
Add file attachments to chat input box

fkn1987:
## Summary

Adds the ability to attach non-image files (e.g. PDFs, text, archives) to the chat input box.

## Changes

- **New \FileMessagePart\ type** in core (\core/index.d.ts\ + \core/config/types.ts\) carrying \
ame\ and \dataUrl\.
- **Choke point**: \constructMessages\ converts file parts to a text placeholder (\[Attachment: name]\) before messages reach any LLM provider, so providers never receive an unsupported part type.
- **Defense in depth** for file parts in: \countTokens\, \openaiTypeConverters\ (Chat Completions + Responses paths), \Anthropic\, \Gemini\, \stripImages\, and console \Message\ rendering.
- **Editor**: new TipTap \FileAttachment\ inline node with a NodeView chip (name + download via dataUrl), paste/drop handling for non-image files, and a paperclip toolbar button.

## Tests

- \constructMessages.test.ts\: file parts are converted to \[Attachment: name]\ without leaking the dataUrl; messages without file parts are unchanged.
- \countTokens.test.ts\: counting tokens for file parts does not throw.

### DIFF
--- a/core/config/types.ts
+++ b/core/config/types.ts
@@ -288,9 +288,16 @@ declare global {
   export type ChatMessageRole = "user" | "assistant" | "system" | "tool";
   
   export interface MessagePart {
-    type: "text" | "imageUrl";
+    type: "text" | "imageUrl" | "file";
     text?: string;
     imageUrl?: { url: string };
+    name?: string;
+    dataUrl?: string;
+    file?: {
+      file_data?: string;
+      file_id?: string;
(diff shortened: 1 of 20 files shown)


fkn1987:
Fechando — projeto errado. Desculpe o incômodo.

PR #12403, opened 2026-05-13
fix(terminal): skip -l login flag for csh and tcsh

serhiizghama:
## Problem

When `$SHELL` is set to `tcsh` (or `csh`), every terminal command issued by Continue fails immediately:

```
Unknown option: `-l'
Usage: tcsh [ -bcdefilmnqstvVxX ] [ argument ... ].
Command failed with exit code 1
```

`getShellCommand` in both `core/tools/implementations/runTerminalCommand.ts` and `extensions/cli/src/tools/runTerminalCommand.ts` hardcodes `["-l", "-c", command]` for every Unix shell. `csh` and `tcsh` do not accept `-l` as a flag, so the shell rejects the invocation before the command is ever evaluated.

Fixes #12378.

## Solution

Detect the shell binary name (`basename` of `$SHELL`) and omit the `-l` flag when it is `csh` or `tcsh`. Those shells automatically source `~/.cshrc` / `~/.tcshrc` on every invocation, so the login-shell intent (sourcing rc files) is preserved without the flag.

Behavior is unchanged for bash, zsh, sh, fish, and any other shell — they continue to receive `-l -c <command>`.

## Testing

- `bash` / `zsh`: `getShellCommand("echo hi")` returns `args: ["-l", "-c", "echo hi"]` (unchanged).
- `tcsh` / `csh`: `getShellCommand("echo hi")` returns `args: ["-c", "echo hi"]`, which `tcsh -c "echo hi"` accepts and runs normally.
- Manual repro from the issue (`$SHELL=/bin/tcsh`, ask Continue to run `pwd`) now succeeds instead of erroring with `Unknown option: -l`.

---
## Summary by cubic
Skip the `-l` login flag when invoking `csh`/`tcsh` so terminal commands don’t fail with “Unknown option: -l.” Other shells keep using `-l -c` to load shell configs.

- **Bug Fixes**
  - Detect the shell name and omit `-l` for `csh`/`tcsh`, which auto-source their rc files.
  - Added a small helper to compute login args and applied it in both terminal command implementations (Unix/macOS and WSL paths); `bash`/`zsh` behavior unchanged.

<sup>Written for commit 1ae4cab3e5270fc094ae17bf128fdb3fb5dff39d. Summary will update on new commits.</sup>

PR #13138, opened 2026-08-16
fix(tools): stabilize runTerminalCommand output capture for agents

saknopp:
This PR fixes an issue where autonomous agents (or external MCP runtimes) attempting to observe terminal output through `runTerminalCommand` would frequently receive empty strings or garbled text instead of the actual command `stdout`.

**The Problem:**
1. **Login Shells:** The implementation previously invoked `bash -l` (login shell). In non-TTY background execution contexts, user `.bash_profile` or `.bashrc` scripts often hang, execute interactive terminal-only utilities (like `tmux`), or prematurely `exit`, causing the node `child_process` to capture no output.
2. **ANSI Color Pollution:** `getColorEnv()` previously forced `FORCE_COLOR=1` and `TERM=xterm-256color`. While this looks great in the UI, agents expecting machine-readable text struggle to parse ANSI-polluted streams, and some CLI tools alter their output buffering behavior when they detect forced color sequences in a non-TTY pipe.

**The Solution:**
- Replaced the login shell flag `-l` with a standard `/bin/bash -c` execution to bypass user-specific interactive profile quirks.
- Modified `getColorEnv()` to explicitly request a plaintext environment (`NO_COLOR="1"`, `TERM="dumb"`), ensuring that the output returned to the agent is clean and strictly decodable.

**Testing:**
Verified locally that terminal execution commands reliably return their stdout streams to external LLM runtimes without hangs or formatting corruption.

### DIFF
--- a/core/tools/implementations/runTerminalCommand.ts
+++ b/core/tools/implementations/runTerminalCommand.ts
@@ -97,14 +97,21 @@ const getColorEnv = () => ({
   CLICOLOR_FORCE: "1",
 });
 
-// Only spawn processes locally when there is no remote workspace.
+// Spawn processes locally unless the workspace is a known remote environment.
 // With extensionKind: ["ui", "workspace"], the extension host almost always
 // runs on the local machine. childProcess.spawn() executes on the extension
-// host, so for any remote workspace it would run commands on the wrong machine
-// (or fail with ENOENT when the local shell doesn't match the remote OS).
-// All remote types delegate to ide.runCommand() which routes through VS Code's
-// integrated terminal and executes in the correct remote environment.
-const LOCAL_ONLY = ["", "local"];
+// host, so for a truly remote workspace it would run commands on the wrong
+// machine (or fail with ENOENT when the local shell doesn't match the remote
+// OS). All known remote types delegate to ide.runCommand() which routes
+// through the integrated terminal and executes in the correct remote env.
+//
+// We use a blocklist instead of an allowlist because VS Code forks like Cursor
+// may report a non-standard remoteName even when running locally.  The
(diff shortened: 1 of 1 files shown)


saknopp:

saknopp:
**Additional context on the removal of the `-l` flag:**

I noticed PR #12403 also addresses issues with the `-l` flag, but explicitly tries to keep it for `bash` and `zsh`. 

I wanted to clarify why this PR removes the `-l` flag entirely for *all* Unix shells:
Using a login shell for autonomous agent execution is fundamentally unsafe because it executes the user's `.bash_profile` or `.zsh_profile`. These profiles often contain interactive-only logic (like auto-starting `tmux`, prompting for SSH passwords, or complex shell prompts) that will cause the agent's background `child_process` to hang indefinitely. 

By running a standard, non-login shell execution, we ensure a clean, deterministic, and non-interactive environment, which is exactly what a machine-to-machine LLM execution loop requires to avoid garbage output and timeout hangs.

**Handling PATH and Aliases:**
A common concern with removing `-l` is how agents will resolve paths or user-specific aliases:
- **For PATH:** The `child_process.spawn` call already inherits `process.env` from the IDE (VS Code / IntelliJ), which typically already contains the user's fully resolved `$PATH` established at IDE startup.
- **For Aliases:** Autonomous agents generally shouldn't rely on hidden, user-specific aliases to function reliably. If an agent needs a specific binary path (like a custom python executable or nvm node version), it is much safer for the agent to resolve the absolute path or rely on workspace configuration, rather than hoping a `.bash_profile` alias behaves correctly in a non-TTY pipe environment.

PR #13141, opened 2026-08-17
fix(core): price OpenAI cached input tokens at the cache read rate

### LINKED ISSUE #13104: OpenAI cost calculation ignores cached input tokens (over-costs cache hits)
### Summary

`calculateOpenAICost` in `core/llm/utils/calculateRequestCost.ts` bills the full `promptTokens` at the standard input rate and never accounts for cached input tokens, so requests that hit OpenAI's prompt cache are over-costed. The Anthropic branch in the same file already handles cache tokens; the OpenAI branch does not.

### Detail

OpenAI reports cached input as `usage.prompt_tokens_details.cached_tokens`, and `prompt_tokens` **includes** those cached tokens. Cached input is billed at a discount (for example gpt-4o cached input is half the standard input rate). The cost function charges every prompt token at full rate:

```ts
const inputCost = (usage.promptTokens / 1_000_000) * modelPricing.input;
// no use of usage.promptTokensDetails.cachedTokens
```

Compare `calculateAnthropicCost`, which reads `usage.promptTokensDetails` and prices `cachedTokens` / `cacheWriteTokens` at their own rates.

### Effect

For an OpenAI request with cached input (common with long, stable system prompts), the reported cost is higher than the actual OpenAI charge — the cached portion is billed at full price instead of the cache-read discount.

### Suggested direction

Give the OpenAI pricing table a `cachedInput` rate and subtract the cached tokens from the full-rate input, pricing them separately, the way the Anthropic branch does:

```ts
const cachedTokens = usage.promptTokensDetails?.cachedTokens ?? 0;
const uncachedInput = Math.max(0, usage.promptTokens - cachedTokens);
const inputCost = (uncachedInput / 1_000_000) * modelPricing.input
                + (cachedTokens / 1_000_000) * modelPricing.cachedInput;
```

Happy to open a PR with the cached-input rates for the models already listed if that direction sounds right.

tonycoder-hub:
Fixes #13104.

OpenAI cost used the full input rate for cached tokens. Cached tokens are now billed at the cache-read rate for models that support prompt caching (gpt-4o, gpt-4o-mini).

core/llm/utils/calculateRequestCost.vitest.ts: 27 passed.

### DIFF
--- a/core/llm/utils/calculateRequestCost.ts
+++ b/core/llm/utils/calculateRequestCost.ts
@@ -155,11 +155,16 @@ function calculateOpenAICost(
   // Normalize model name
   const normalizedModel = model.toLowerCase();
 
-  // Define pricing per million tokens (MTok) by model family prefix
-  const pricing: Record<string, { input: number; output: number }> = {
+  // Define pricing per million tokens (MTok) by model family prefix.
+  // cachedInput is only set for models that support prompt caching; the others
+  // bill cached input, if it is ever reported, at the standard input rate.
+  const pricing: Record<
+    string,
+    { input: number; output: number; cachedInput?: number }
+  > = {
     // GPT-4o models (most specific first)
-    "gpt-4o-mini": { input: 0.15, output: 0.6 },
-    "gpt-4o": { input: 2.5, output: 10 },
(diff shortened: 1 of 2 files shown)


tonycoder-hub:

tonycoder-hub:

PR #13143, opened 2026-08-17
fix: only treat tar as create when the mode flag is actually c

tonycoder-hub:
Safe-create check matched any argument containing the letter c. Mode flags only. 228 tests passed in packages/terminal-security.

PR #13146, opened 2026-08-17
fix: intersect ranges using only the boundary lines own characters

tonycoder-hub:
## Description

`intersection()` in `core/util/ranges.ts` used the other range's character offsets even when that range did not start or end on the boundary line.

A same-line overlap such as `(1,0)-(1,5)` vs `(1,3)-(2,0)` returned `null` (`min(5, 0) = 0`). The repo already had a skipped test for this. Shared-start / shared-end multi-line cases were also too wide.

Only endpoints that actually lie on the boundary line now contribute. Used by VS Code autocomplete LSP crawl to drop duplicate type definitions.

Distinct from #13143/#13144/#13145.

## Checklist

## Tests

Un-skipped the existing single-line overlap case and added two shared-boundary cases. All three fail on main.

`cd core && npx cross-env NODE_OPTIONS=--experimental-vm-modules npx jest util/ranges.test.ts` → 24 passed, 10 skipped (unrelated `getRangeInString` block).

### DIFF
--- a/core/util/ranges.test.ts
+++ b/core/util/ranges.test.ts
@@ -301,8 +301,7 @@ describe("intersection", () => {
     expect(result).toBeNull();
   });
 
-  // TODO
-  test.skip("returns correct intersection for single line overlap", () => {
+  test("returns correct intersection for single line overlap", () => {
     rangeA = {
       start: { line: 1, character: 0 },
       end: { line: 1, character: 5 },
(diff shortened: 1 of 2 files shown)


tonycoder-hub:

tonycoder-hub:
recheck

tonycoder-hub:

PR #13148, opened 2026-08-17
fix: session export timestamps for half-hour timezones

tonycoder-hub:
getOffsetDatetime split getTimezoneOffset into hours plus leftover minutes. East of UTC the offset is negative, so Math.floor and % shifted UTC+05:30 by 6h30m. A session shared at 15:30 IST was written as 20260817T163000_session.md.

Now shift by the whole offset at once. Tests: cd core && npx vitest run util/historyUtils.vitest.ts -> 3 passed. Distinct from #13143-#13147.

PR #13140, opened 2026-08-17
fix(core): match rule globs against file paths containing spaces

tonycoder-hub:
Fixes #13135. Rule globs now match file paths that contain spaces. Tests added and passing in core.

PR #13149, opened 2026-08-17
fix: stop prefixing markdown chunks with the string undefined

tonycoder-hub:
markdownChunker prepends each section header to every chunk. When a section has no header, findHeader returns undefined and the template literal interpolated it as the literal text "undefined". Headerless docs were indexed with up to five bogus undefined lines at the top of every chunk (used by @docs indexing).

Only prepend the header when the section actually has one.

Tests: cd core && npx vitest run indexing/chunk/markdown.vitest.ts — 3 passed. Distinct from #13140-#13148.

### DIFF
--- a/core/indexing/chunk/markdown.ts
+++ b/core/indexing/chunk/markdown.ts
@@ -133,7 +133,9 @@ export async function* markdownChunker(
       hLevel + 1,
     )) {
       yield {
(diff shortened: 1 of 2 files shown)


tonycoder-hub:

tonycoder-hub:

PR #13150, opened 2026-08-17
fix: write rule name into markdown frontmatter

tonycoder-hub:
createRuleMarkdown takes a name argument and never emits it. The name used to be an H1 in the rule body and was dropped without adding it to YAML frontmatter, so rules created via create_rule_block or New Rule had no name and markdownToRule fell back to the file path.

Frontmatter now includes name: name.trim(), matching createPromptMarkdown.

Tests: packages/config-yaml markdown jest 144 passed; core tools/config/markdown 15 passed; workspaceBlocks vitest 10 passed. Distinct from #13140-#13149.


tonycoder-hub:

tonycoder-hub:

Issue #13056, opened 2026-07-29
UI Bug : Chat horizontal overflow in sidebar, specifically within "thought" (reasoning) blocks


Aybe753:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [ ] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue version: 2.0.0
- IDE version: VSCode 1.130.0
```

### Description

When a prompt is sent in the Continue chat interface, the response content overflows the horizontal boundaries of the sidebar. This issue is particularly noticeable and triggered within the **"thought" (reasoning) blocks** of the model's response. The text/content breaks the sidebar layout and becomes unreadable because the container expands based on the total viewport width rather than the sidebar width.

### To reproduce

1. Open the Continue extension in the VS Code sidebar.
2. Send a prompt to a model that uses reasoning/thought blocks (e.g., DeepSeek-R1, OpenAI o1).
3. Observe that the **thought/reasoning block** expands beyond the right edge of the sidebar, causing a horizontal overflow.

### Log output

```Shell

```

Aybe753:
I analysed a little more and it seems to happen not with the "thought" block, but with the **text/code block**. When the AI writes code or anything using a text box, the chat overflows horizontaly.

bluemoehre:
Duplicates #13055

fullspeedexe:
Go to continue settings and enable "Wrap Codeblocks"

bluemoehre:
> Go to continue settings and enable "Wrap Codeblocks"

I can confirm that this is a valid workaround.

hari0504:
This is different issue than code block not getting wrapped up, code block does get wrap up but  the text content gets hidden, not matter how big you make the side bar

<img width="624" height="245" alt="Image" src="https://github.com/user-attachments/assets/b1012c43-5a5e-4538-a334-4f51a9af52c3" />

fullspeedexe:
> This is different issue than code block not getting wrapped up, code block does get wrap up but the text content gets hidden, not matter how big you make the side bar
[...]

Issue #13055, opened 2026-07-29
Continue panel gets cropped/cut off in VSCodium sidebar


carlosrodan:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Fedora
- Continue version: 2.0.0 (extension)
- IDE version: VSCodium 1.126.04524
- Model: DeepSeek Coder
```

### Description

<img width="1011" height="1049" alt="Image" src="https://github.com/user-attachments/assets/c7e1e655-eb82-4035-82c4-8bb1278236e5" />
<img width="855" height="1049" alt="Image" src="https://github.com/user-attachments/assets/e963d5d6-78c9-425f-b4fb-9e15addefeea" />

When the side panel is narrow the contents of the Continue extension get cropped. Resizing the window to make it wider solves the issue. I have the impression there is some kind of CSS limit (min width) somewhere. I've uploaded two screenshots, one with the side panel taking up roughtly half of my scren (1920px) where everything is visble and another one, narrower with the contents cut off. 

I tried the following commands without success on VSCodium:
"Developer: Reload Window"
"View: Reset Zoom"
"View: Reset View Locations"

### To reproduce

1. Install VSCodium
2. Install Continue Extensions V2.0.0
3. Open the side panel 
4. Resize the side panel

### Log output

```Shell
INFO Started local extension host with pid 16677.
workbench.desktop.main.js:sourcemap:38  INFO [Continue.continue]: Command `continue.focusContinueInput` already registered by Continue - open-source AI code agent (Continue.continue)
workbench.desktop.main.js:sourcemap:5057 Unrecognized feature: 'local-network-access'.
_createElement @ workbench.desktop.main.js:sourcemap:5057
workbench.desktop.main.js:sourcemap:5057 An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing.
mountTo @ workbench.desktop.main.js:sourcemap:5057
workbench.desktop.main.js:sourcemap:38  INFO [AccountPolicyGate] apply: state=inactive, reason=undefined, isRestricted=false
[...]

PR #13153, opened 2026-08-18
Prevent margin from displacing content out of the viewport.

Perritu:
Prevent code block from expanding the chat webview beyond it's horizontal limits by setting thread message div width to 100vw.

Solves #13056 #13055 

## Description

Added a css property to force the thread message block to match the viewport witdth.

## Checklist

After:

image

Before:

image

### DIFF
--- a/gui/src/pages/gui/Chat.tsx
+++ b/gui/src/pages/gui/Chat.tsx
@@ -78,7 +78,6 @@ const StepsDiv = styled.div`
 
   .thread-message {
     margin: 0 0 0 1px;
-    width: 100vw;
   }
 `;
 
--- a/gui/src/pages/gui/Chat.tsx
+++ b/gui/src/pages/gui/Chat.tsx
@@ -78,6 +78,7 @@ const StepsDiv = styled.div`
 
   .thread-message {
     margin: 0 0 0 1px;
+    width: 100vw;
   }
 `;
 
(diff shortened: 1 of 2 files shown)


Perritu:
recheck

Perritu:
recheck

Perritu:

bluemoehre:
In my opinion `vw` isn't the best unit to use here. There's no guarantee that scrollbars won't take up space and this behavior may vary across OSes and window layouts / panel arrangements. CSS provides better alternatives like `dvw` and `svw`, which are designed to handle dynamic viewport changes more reliably.

Additionally, I believe this approach has another flaw: inline padding appears to accumulate (see the linked issues), which causes the inset area to be offset twice.

Perritu:
> In my opinion `vw` isn't the best unit to use here. There's no guarantee that scrollbars won't take up space and this behavior may vary across OSes and window layouts / panel arrangements. CSS provides better alternatives like `dvw` and `svw`, which are designed to handle dynamic viewport changes more reliably.
> 
> Additionally, I believe this approach has another flaw: inline padding appears to accumulate (see the linked issues), which causes the inset area to be offset twice.

Agreed.
I've modified the PR to align with TW conventions and to use the global chat object instead of the thread-specific one.

Perritu:
Since I haven't heard back yet, I'm sharing a quick workaround patch for anyone while wait for an official fix.

You can apply this quick `sed` command to patch your local installation (adjust the path depending on your environment):

```sh
sed -i 's|"main",{className:"no-scrollbar flex min-h-0 flex-1 flex-col"|"main",{className:"no-scrollbar w-screen flex min-h-0 flex-1 flex-col"|g' \
~/.vscode-oss/extensions/continue.continue-2.0.0-linux-arm64/gui/assets/index.js
```

(Note: Tested on Debian with VsCodium, but it can be easily adapted).

PR #13160, opened 2026-08-18
fix(core): price OpenAI prompt-cache input at the documented cached rate (#13104)

santhiprakash:
## Description

`calculateOpenAICost()` in `core/llm/utils/calculateRequestCost.ts` bills every prompt token at the standard input rate and ignores `usage.promptTokensDetails.cached_tokens`, so any OpenAI request that hits the prompt cache (the common case for stable, long system prompts) is over-reported by the cache-read discount. The Anthropic branch in the same file already handles cache tokens; the OpenAI branch does not.

This PR:

1. Adds `cachedInput` to the OpenAI pricing table for the model families where OpenAI documents a prompt-cache rate (`gpt-4o`, `gpt-4o-mini`).
2. Splits `usage.promptTokens` into uncached and cached portions and bills each at its own rate, mirroring the Anthropic branch.
3. Clamps `cachedTokens` to `promptTokens` to defend against malformed usage payloads that report more cached tokens than total prompt tokens.
4. Keeps the original behavior for models without a documented `cachedInput` rate: the cached portion is treated as uncached and billed at the standard input rate (the API would never return a `cached_tokens` value for them in the first place).

## Tests

Added regression cases in `core/llm/utils/calculateRequestCost.vitest.ts`:

- `gpt-4o` with 700/1000 prompt tokens cached
- `gpt-4o` with all 1000 prompt tokens cached
- `gpt-4o` with explicit zero cached tokens (no cache rows in breakdown)
- `gpt-4o-mini` with 800/1000 prompt tokens cached
- `gpt-4` with cached tokens reported but no `cachedInput` rate defined (full prompt billed at the standard input rate)
- `gpt-4o` with cached tokens > prompt tokens (clamped to `promptTokens`)

The fix was verified by extracting `calculateOpenAICost` into a standalone Node script and running 11 cases against hand-computed expected costs (all pass). I could not run the vitest suite in this environment (no `node_modules` at the workspace root); the upstream vitest suite should be re-run by the maintainer.

## Checklist

- [x] The relevant tests have been updated

🤖 Generated with [Hermes Agent](https://github.com/NousResearch/hermes-agent)

### DIFF
--- a/core/llm/utils/calculateRequestCost.ts
+++ b/core/llm/utils/calculateRequestCost.ts
@@ -155,11 +155,18 @@ function calculateOpenAICost(
   // Normalize model name
   const normalizedModel = model.toLowerCase();
 
-  // Define pricing per million tokens (MTok) by model family prefix
-  const pricing: Record<string, { input: number; output: number }> = {
+  // Define pricing per million tokens (MTok) by model family prefix.
+  // `cachedInput` is set only for model families where OpenAI documents a
+  // prompt-cache rate. prompt_tokens from the API includes the cached portion,
+  // so the input cost is split: uncached tokens at the input rate and cached
+  // tokens at the cachedInput rate.
+  const pricing: Record<
+    string,
+    { input: number; output: number; cachedInput?: number }
+  > = {
     // GPT-4o models (most specific first)
-    "gpt-4o-mini": { input: 0.15, output: 0.6 },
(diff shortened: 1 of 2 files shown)


santhiprakash:
Closing to keep the review queue clean — this has been open a while without traction. Happy to reopen or resubmit a fresh version if it would still be useful.

PR #13162, opened 2026-08-19
fix(openai-adapters): don't swallow stream chunks that carry usage

ScrewTSW:
## Description

`chatCompletionStream` withholds any chunk with a `usage` field so usage can be re-emitted last. That assumes usage only ever appears on a terminal chunk, which is true for the OpenAI API but not in general.

**llama.cpp / llama-server attaches a running `usage` counter to every chunk** (`completion_tokens` 1, 2, 3…). Every chunk then matches the deferral branch, each overwriting `lastChunkWithUsage`, and only the final one is ever yielded. The result is a completely blank assistant response — the model generates normally, token counters tick up, and no text renders.

This affects any OpenAI-compatible backend that streams incremental usage, and it hits `content` and `reasoning_content` alike.

The fix narrows the predicate to chunks that are genuinely usage-only: `usage` present, no `finish_reason`, and an empty `delta`. A usage-bearing content chunk also clears the deferred chunk so it is not re-emitted as a duplicate at the end.

Behaviour on the OpenAI API is unchanged: its terminal usage-only chunk still matches and is still emitted last.

## Checklist

Not applicable — the visible symptom is an empty assistant message. The added tests reproduce it deterministically at the adapter level.

## Tests

New: `packages/openai-adapters/src/apis/OpenAIStreamUsage.test.ts` (3 tests) covering both regimes:
- per-chunk running usage (llama.cpp style) — all content is yielded
- OpenAI-style terminal usage-only chunk — still deferred and emitted last
- usage-bearing content chunk is not duplicated

Mutation-verified: reverting the predicate to `!!result.usage` fails these with `expected '' to be 'Hello world'`, reproducing the bug exactly. Full adapter suite passes (87/87).


ScrewTSW:
The failing check here is the JetBrains test job, which is unrelated to this change - tracked in #13164.

This PR touches no JetBrains code. `prettier-check` passes, and the added tests pass locally.

ScrewTSW:

PR #13163, opened 2026-08-19
fix(gui): do not drop batched messages after a <think> block

ScrewTSW:
## Description

The `<think>…</think>` fast-path in the `streamUpdate` reducer ends in `return` rather than `continue`. It sits inside `for (const message of action.payload)`, so returning exits the reducer entirely and silently discards every remaining message in the batch — not just the one being handled.

The sibling early-exit for redacted thinking (a few lines above) uses `continue`, and nothing runs after the loop, so `continue` is the intended control flow here.

Impact is limited to payloads carrying more than one message, where any content following a `<think>` block in the same batch is dropped.

## Checklist

Not applicable — a one-keyword control-flow fix in a reducer. The added test demonstrates the dropped message directly.

## Tests

Added to `gui/src/redux/slices/sessionSlice.test.ts`: a two-message payload where the first carries a `<think>` block and the second continues the answer.

This went unnoticed because all 11 existing `streamUpdate` tests dispatch single-element payloads, where `return` and `continue` are indistinguishable.

Mutation-verified: with `return` the new test fails (`'First part.'` vs `'First part. Second part.'`); with `continue` all 10 pass and the other 9 are unaffected.


ScrewTSW:
The failing check here is the JetBrains test job, which is unrelated to this change - tracked in #13164.

This PR touches no JetBrains code. `prettier-check` passes, and the added tests pass locally.

ScrewTSW:

PR #13170, opened 2026-08-21
fix(core): match rule globs against file paths containing spaces

### LINKED ISSUE #13135: Rule globs don't match files with spaces in filename

### Description

Environment
- Continue: latest
- VS Code on Windows 11
- Files on Debian via SSH Remote

The problem occurs when using VS Code Remote SSH with a Debian remote host. The project files reside on the remote Linux filesystem, so this is not caused by Windows filename handling.

Rule:

```
---
name: Test
globs:
  - docs/**/*.md
---
```

Expected:
The rule is applied to every Markdown file in docs/, including files whose names contain spaces.

Actual:
The rule is applied to:
- docs/foo.md
- docs/foo_bar.md

But not to:
- docs/foo bar.md

Even an exact glob does not work:

```
globs:
  - docs/foo bar.md
```

The rule is not included in the prompt.

Renaming the file to remove the space immediately fixes the issue.

santhiprakash:
## Description

Rule globs like `docs/**/*.md` or the exact glob `docs/foo bar.md` silently failed to match files whose names contain spaces, so any rule whose target files lived under e.g. `docs/foo bar.md` was excluded from the system prompt.

### Root cause

`core/llm/utils/extractPathsFromCodeBlocks.ts` used the regex `/([^\s()`]+\.[a-zA-Z0-9]+)/` to pull a filename out of a code-block header. The character class `[^\s()`]` excluded whitespace, so a header like ````typescript docs/foo bar.md```` was truncated to just `bar.md`. The downstream rule matcher never saw the directory portion, so neither the broad glob (`docs/**/*.md`) nor the exact glob (`docs/foo bar.md`) could match.

### Fix

Replace the regex with a small parser that strips the optional `<lang>` tag and the optional trailing `(<start>-<end>)` line range but preserves any spaces inside the path itself. The heuristic for identifying a language tag is that it must be a single token containing no `/` or `\` (paths almost always contain a separator; language tags like `typescript`, `md`, `py` never do).

### Tests

- New vitest suite at `core/llm/utils/extractPathsFromCodeBlocks.vitest.ts` covers paths with spaces (with and without language tag, with and without line range, in nested directories).
- New regression test in `core/llm/rules/getSystemMessageWithRules.vitest.ts` verifies that `docs/**/*.md` and the exact glob `docs/foo bar.md` both apply to a file referenced in a code block.
- Updated existing Jest test `core/llm/utils/extractPathsFromCodeBlocks.test.ts` with the same regression cases.

Fixes #13135

## Checklist

N/A — backend correctness fix, no UI surface change.

## Tests

- `core/llm/utils/extractPathsFromCodeBlocks.vitest.ts` (new, 9 tests)
- `core/llm/rules/getSystemMessageWithRules.vitest.ts` (2 new tests added)
- `core/llm/utils/extractPathsFromCodeBlocks.test.ts` (regression case added)

All pass locally via `node_modules/.bin/vitest run` in `core/`. Verified by reverting the fix and re-running: the new regression tests fail on the original code and pass with the fix in place.

### DIFF
--- a/core/llm/rules/getSystemMessageWithRules.vitest.ts
+++ b/core/llm/rules/getSystemMessageWithRules.vitest.ts
@@ -476,4 +476,55 @@ describe("Content pattern matching", () => {
       shouldApplyRule(nestedPatternRule, [utilFilePath], {}, utilContents),
     ).toBe(false);
   });
+
+  // Regression for https://github.com/continuedev/continue/issues/13135
+  it("should apply rules to files whose names contain spaces", () => {
+    const docsRule: RuleWithSource = {
+      name: "Docs Rule",
+      rule: "Write docs in the active voice",
+      globs: "docs/**/*.md",
+      source: "rules-block",
+      sourceFile: "docs/rules.md",
+    };
+
+    // Code block headers are "```<language> <path> (<range>)", where the
+    // language and range are both optional.
+    const messages: UserChatMessage[] = [
+      {
+        role: "user",
+        content: "What do you think?\n```docs/foo bar.md\n# Title\n```",
+      },
+      {
+        role: "user",
+        content: "What do you think?\n```md docs/foo bar.md\n# Title\n```",
+      },
+      {
+        role: "user",
+        content:
+          "What do you think?\n```md docs/foo bar.md (1-1)\n# Title\n```",
+      },
+    ];
+
+    for (const message of messages) {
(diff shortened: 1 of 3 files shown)


santhiprakash:
Closing to keep the review queue clean — this has been open a while without traction. Happy to reopen or resubmit a fresh version if it would still be useful.

PR #13179, opened 2026-08-23
fix(cli): distinguish unavailable permission prompts

### LINKED ISSUE #13178: CLI: 'ask' permission tier silently auto-denies when onToolPermissionRequest callback isn't wired, mislabeled as user denial

- [x] I've tried finding an answer on the Continue docs site
- [x] I'm not able to find an open issue that reports the same bug (searched: "permission denied", "onToolPermissionRequest", "denialReason", "CLI permission", "tool call denied" — closest related but distinct: #13001, #12963)
- [x] I've seen the troubleshooting guide on the Continue Docs

### Relevant environment info

- OS: Linux
- Continue version (CLI): 1.5.47
- Node: v22.15.0

## Summary

In `@continuedev/cli`, tool calls with an `ask`-tier permission policy sometimes get silently denied with no interactive prompt ever shown to the user. The denial is then reported with `denialReason:"user"`, which is misleading — the user was never actually asked.

## Root cause (from reading the bundled `dist/index.js`)

The permission-decision function resolves an `ask` policy by calling into the prompt-request function:

```js
async function _ua(e,t,n,r){
  let o=VXe(t,e);
  return o.permission==="allow" ? {approved:!0}
    : o.permission==="ask" ?
        (r ? {approved:!1,denialReason:"policy"}
           : await Eua(t,n) ? {approved:!0}
           : {approved:!1,denialReason:"user"})
    : o.permission==="exclude" ? {approved:!1,denialReason:...}
    : ...
}
```

And the actual prompt function:

```js
async function Eua(e,t){
  if(!t?.onToolPermissionRequest) return !1;
  let n={name:e.name,arguments:e.preprocessResult?.args??e.arguments,preview:e.preprocessResult?.preview, ...};
  ...
}
```

If `t.onToolPermissionRequest` is not attached for a given tool call, `Eua` returns `false` immediately — no prompt is ever rendered — and the caller reports `denialReason:"user"` even though the user was never asked.

## Steps to reproduce

1. Run `cn` interactively (not headless, no `-p`/`--print`)
2. Configure `permissions.yaml` such that some tool/Bash patterns are unmatched by `allow`/`ask`/`exclude`, falling through to the default `"ask"` policy
3. Trigger a tool call matching one of those unmatched patterns (e.g. a Bash command not covered by any `allow` rule)
4. Observe: the tool call is denied with no interactive prompt ever appearing, and the denial is reported as if the user rejected it

## Expected behavior

An `ask`-tier permission should always show the interactive prompt when not in headless mode. If the callback isn't available, that's an internal wiring bug and should either surface an error/warning rather than silently denying, or the denial reason should reflect the actual cause (e.g. `"callback_missing"`) instead of being mislabeled `"user"`.

## Additional context

We also intermittently saw `"Message splitting exceeded maximum iterations, stopping"` in the same sessions where this occurred, which may point to broader instability in the TUI's message/callback event loop — possibly a related root cause (the same callback-wiring issue affecting more than just the permission prompt).

## Workaround

Expanding `permissions.yaml`'s `allow` list to explicitly cover the needed command patterns (e.g. `Bash(pio*)`, `Bash(ufbt*)`, `Bash(git*)`) avoids the broken `ask` path entirely, since `allow` short-circuits before `Eua` is ever called.

## Related issues (not duplicates, but adjacent)

- #13001 — incomplete destructive-command blocklist causing the *opposite* failure mode (unwanted auto-execution) in headless/auto mode
- #12963 — malformed message structure after a tool request is cancelled, a different bug in the same general tool-call/permission code path

### To reproduce

See "Steps to reproduce" above.

### Log output

N/A — no error is logged; the tool call is simply denied with `denialReason:"user"` and no prompt is shown.

primorLee:
## Description

Fixes #13178.

When an `ask`-tier tool call reaches the CLI without an `onToolPermissionRequest` callback, the CLI now distinguishes the unavailable interactive prompt from an explicit user rejection.

- represent the missing callback as a dedicated `callback_missing` denial reason
- surface `Interactive permission prompt unavailable` instead of `Permission denied by user`
- preserve the existing user-denial and policy-denial behavior

## AI Code Review

## Checklist

Not applicable; this corrects the CLI's permission result and message when no interactive callback is available.

## Tests

- `npm test -- src/stream/streamChatResponse.permissions.test.ts`
- `prettier --check src/stream/streamChatResponse.helpers.ts src/stream/streamChatResponse.permissions.test.ts src/stream/toolPermissionResult.ts`
- `eslint src/stream/streamChatResponse.helpers.ts src/stream/streamChatResponse.permissions.test.ts src/stream/toolPermissionResult.ts`

### DIFF
--- a/extensions/cli/src/stream/streamChatResponse.helpers.ts
+++ b/extensions/cli/src/stream/streamChatResponse.helpers.ts
@@ -28,6 +28,12 @@ import { PreprocessedToolCall, ToolCall } from "../tools/types.js";
 import { logger } from "../util/logger.js";
 
 import { StreamCallbacks } from "./streamChatResponse.types.js";
+import {
+  getPermissionDeniedMessage,
+  resolveUserPermissionResult,
+  type ToolPermissionApprovalResult,
+  type ToolPermissionDenialReason,
+} from "./toolPermissionResult.js";
 
 export interface ToolResultWithStatus extends ChatCompletionToolMessageParam {
   status: ToolStatus;
@@ -38,12 +44,9 @@ export function handlePermissionDenied(
   toolCall: PreprocessedToolCall,
   chatHistoryEntries: ChatCompletionToolMessageParam[],
   callbacks?: StreamCallbacks,
-  reason: "user" | "policy" = "user",
+  reason: ToolPermissionDenialReason = "user",
 ): void {
-  const deniedMessage =
-    reason === "policy"
-      ? `Command blocked by security policy`
-      : `Permission denied by user`;
+  const deniedMessage = getPermissionDeniedMessage(reason);
 
   logger.info("Tool call denied", {
     name: toolCall.name,
@@ -65,9 +68,13 @@ export function handlePermissionDenied(
 export async function requestUserPermission(
   toolCall: PreprocessedToolCall,
   callbacks?: StreamCallbacks,
-): Promise<boolean> {
+): Promise<boolean | undefined> {
   if (!callbacks?.onToolPermissionRequest) {
-    return false;
+    logger.error("Cannot request interactive tool permission", {
+      name: toolCall.name,
+      reason: "onToolPermissionRequest callback is missing",
+    });
+    return undefined;
   }
 
   const toolCallRequest: ToolCallRequest = {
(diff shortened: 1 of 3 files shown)


primorLee:

PR #13181, opened 2026-08-24
fix: preserve UTF-8 in SageMaker streaming decode

pacocartones:
`_streamComplete` and `_streamChat` in `core/llm/llms/SageMaker.ts` decoded each streamed binary chunk with a fresh `new TextDecoder().decode(chunk)` and no streaming flag. A multibyte UTF-8 character split across two network chunks was emitted as U+FFFD replacement characters, corrupting any non-ASCII model output (emoji, accents, CJK). The chunk content and boundaries come from the remote SageMaker endpoint, so this triggers on ordinary response fragmentation.

Reuse a single `TextDecoder` per stream and decode with `{ stream: true }`, so partial sequences at chunk boundaries are buffered until complete.

Added regression tests for a split emoji and a split two-byte character (vitest): they fail before the fix (replacement chars) and pass after.

### DIFF
--- a/core/llm/llms/SageMaker.ts
+++ b/core/llm/llms/SageMaker.ts
@@ -50,9 +50,10 @@ class SageMaker extends BaseLLM {
     const response = await client.send(command, { abortSignal: signal });
     if (response.Body) {
       let buffer = "";
+      const decoder = new TextDecoder();
       for await (const rawValue of response.Body) {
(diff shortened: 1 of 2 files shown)


pacocartones:

pacocartones:
Closing this PR — we're consolidating our open contributions and withdrawing this one from our side. The underlying fix still stands on its own; feel free to reopen or cherry-pick it if it's useful. Thanks for the project!

Issue #8099, opened 2025-10-05
continue-cli (cn): `Error: Premature Close`


babblebey:
### Before submitting your bug report

- [x] I've tried using the "Ask AI" feature on the [Continue docs site](https://docs.continue.dev/) to see if the docs have an answer
- [x] I believe this is a bug. I'll try to join the [Continue Discord](https://discord.gg/NWtdYexhMs) for questions
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS: Windows 11
- Continue-cli (cn) version: v1.4.46 (beta)
- Model: Claude Sonnet 4
- config:
  

  
  OR link to agent in Continue hub: https://hub.continue.dev/continue-community/code
```

### Description

Prompted the continue-cli to help break down a single issue into small manageable issues. Hence I was expecting 4 different issues. It created 1 of 4 and encountered the below error in attempt to create the orther.

![Image](https://github.com/user-attachments/assets/8836308b-9b5c-4727-a649-47773e157411)

### Behavior

Main Issue - https://github.com/bdougie/contributor.info/issues/915
Created Issue (of 4) - https://github.com/bdougie/contributor.info/issues/919

Prompt 1:  Get the issue `#915` (from bdougie/contributor.info), break it down into 4 sub issue that is manageable with the phases 

🤖: _---Responds with issue content_

Prompt 2: Create each issue as sub issue to the `#915` on the repo and make sure to add an implementation plan for approaching each  

🤖: --- Creates https://github.com/bdougie/contributor.info/issues/919 and errors out on the next

### To reproduce

- Write Prompt 1:  Get the issue `#915` (from bdougie/contributor.info), break it down into 4 sub issue that is manageable with the phases 
- wait for 🤖 response
- Use Prompt 2: Create each issue as sub issue to the `#915` on the repo and make sure to add an implementation plan for approaching each  

### Log output

```Shell

```

PR #5715, opened 2025-05-17
make continuedev/fetch package automatically respect HTTP(S)_PROXY, NO_PROXY environment variables

sestinj:
## Description

Making sure that the custom fetch package respects environment variables like HTTP(S)_PROXY, NO_PROXY, etc.

## Checklist

- [x] I've read the [contributing guide](https://github.com/continuedev/continue/blob/main/CONTRIBUTING.md)
- [x] The relevant docs, if any, have been updated or created
- [x] The relevant tests, if any, have been updated or created

## Tests

Added unit tests to the @ continuedev/fetch package
    

---

## Summary by cubic
Updated the custom fetch package to better handle client-side cancellations and ensure environment variables like HTTP(S)_PROXY are respected.

- **Bug Fixes**
  - Returns a proper response for aborted requests instead of throwing errors.
  - Improves streaming logic to handle cancellations cleanly.

PR #13182, opened 2026-08-24
fix(fetch): reuse keep-alive HTTP Agents across requests

SebTardif:
## Description

Reuse keep-alive HTTP Agents in `@continuedev/fetch` instead of allocating a new `Agent({ keepAlive: true })` on every request.

`fetchwithRequestOptions` is on the path for LLM and other HTTP calls. Each call created a new follow-redirects / proxy Agent with `keepAlive: true` and a 2 hour timeout, then dropped the reference. Idle sockets stayed on those abandoned Agents and could exhaust file descriptors under chat load.

`keepAlive` stays enabled. [#8378](https://github.com/continuedev/continue/pull/8378) already dropped `keepAliveMsecs` so TCP keep-alive pings work. This change makes that keep-alive pool actually get reused.

Agents are keyed by protocol, proxy (or bypass), timeout, `verifySsl`, CA bundle path, and client certificate fields. Concurrent first requests for the same key share one in-flight create so a startup burst does not leak extra Agents.

## Checklist

## Tests

- `packages/fetch/src/httpAgentCache.test.ts`: same options reuse one Agent; `verifySsl`, timeout, protocol, and proxy isolate the cache; bypassed proxy matches no proxy; `keepAlive` is true; `clearHttpAgentCache` drops the instance; concurrent first requests share one Agent.
- `packages/fetch` vitest: 105 passed, including existing e2e. e2e `afterEach` now destroys cached sockets so the next test does not reuse a closed server on the same port.

Related: [#8378](https://github.com/continuedev/continue/pull/8378), [#8099](https://github.com/continuedev/continue/issues/8099). Agent-per-request dates to [#1750](https://github.com/continuedev/continue/pull/1750); `keepAlive: true` was extracted in [#5715](https://github.com/continuedev/continue/pull/5715).

### DIFF
--- a/packages/fetch/src/fetch.e2e.test.ts
+++ b/packages/fetch/src/fetch.e2e.test.ts
@@ -6,6 +6,7 @@ import * as path from "node:path";
 import { execSync } from "node:child_process";
 import { afterEach, describe, expect, test } from "vitest";
 import { fetchwithRequestOptions } from "./fetch.js";
+import { clearHttpAgentCache } from "./httpAgentCache.js";
 
 // Test server ports
 const HTTP_PORT = 3001;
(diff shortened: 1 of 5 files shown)


SebTardif:

PR #13201, opened 2026-08-29
fix(core): correct range intersection when ranges share start/end lines

eeshsaxena:
## Description

`intersection(a, b)` in `core/util/ranges.ts` computes the wrong start/end **character** in the multi-line branch when the two ranges begin (or end) on the same line.

It picked the character from whichever range's line matched `startLine`/`endLine` first:

```ts
const startCharacter = startLine === a.start.line ? a.start.character : b.start.character;
```

When both ranges start on that shared line, this returns `a`'s character instead of the later of the two — and symmetrically for the end. Example (both share start line 0 and end line 3):

```ts
intersection(
  { start: { line: 0, character: 5 },  end: { line: 3, character: 4 } },
  { start: { line: 0, character: 10 }, end: { line: 3, character: 2 } },
)
// before: { start: {0, 5},  end: {3, 4} }   ❌
// after:  { start: {0, 10}, end: {3, 2} }   ✔  (later start, earlier end)
```

The single-line branch already does this correctly (`Math.max` / `Math.min`); this makes the multi-line branch consistent.

## Fix

When the two ranges share the start line, take `Math.max` of the start characters; when they share the end line, take `Math.min` of the end characters. Otherwise the previous behavior is unchanged.

## Testing

Added a `ranges.test.ts` case for two ranges sharing both start and end lines (fails before, passes after). All existing `intersection` tests still pass (they use ranges on different start/end lines, so they don't touch the changed branch). Verified the logic directly against the existing test inputs plus the new case.

### DIFF
--- a/core/util/ranges.test.ts
+++ b/core/util/ranges.test.ts
@@ -353,6 +353,24 @@ describe("intersection", () => {
     });
   });
 
+  test("uses later start and earlier end when both ranges share the start and end lines", () => {
+    rangeA = {
+      start: { line: 0, character: 5 },
+      end: { line: 3, character: 4 },
+    };
+
+    rangeB = {
+      start: { line: 0, character: 10 },
+      end: { line: 3, character: 2 },
+    };
+
+    const result = intersection(rangeA, rangeB);
+    expect(result).toEqual({
+      start: { line: 0, character: 10 },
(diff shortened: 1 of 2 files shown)


eeshsaxena:

PR #13224, opened 2026-09-04
fix(gui): correlate id-less tool-call fragments by provider index

### LINKED ISSUE #13223: OpenAI parallel tool-call argument fragments can be applied to the wrong call
## Summary

OpenAI streamed tool-call deltas carry a stable `index`, while `id` is optional on the SDK's `ChatCompletionChunk.Choice.Delta.ToolCall` type.

Continue's `fromChatCompletionChunk()` currently drops `tool_call.index`, and `applyToolCallDelta()` falls back to updating the most recently tracked tool call whenever an incoming continuation fragment has no `id`.

With two parallel tool calls whose argument fragments interleave, this can apply one call's arguments to another call.

## Current code path

```
core/llm/openaiTypeConverters.ts
  → fromChatCompletionChunk()

gui/src/redux/slices/sessionSlice.ts
  → applyToolCallDelta()

gui/src/util/toolCallState.ts
  → addToolCallDeltaToState()
```

This code is shared by Continue's OpenAI/OpenAI-compatible provider stream-conversion and accumulation path, so the affected surface may include any provider reusing that path.

## Minimal reproduction

Four streamed tool-call deltas:

```
1. index=0, id=call_A, name=tool_a, arguments=""
2. index=1, id=call_B, name=tool_b, arguments=""
3. index=0, id absent, arguments='{"target":"A_ONLY"}'
4. index=1, id absent, arguments='{"target":"B_ONLY"}'
```

## Actual result

```
call_A.function.arguments === ""
call_B.function.arguments === '{"target":"A_ONLY"}'
```

`call_A`'s arguments are associated with `call_B`, while `call_B`'s own argument fragment is lost.

## Expected result

```
call_A.function.arguments === '{"target":"A_ONLY"}'
call_B.function.arguments === '{"target":"B_ONLY"}'
```

## Why this sequence is supported by the SDK model

In Continue's OpenAI SDK dependency, `ChatCompletionChunk.Choice.Delta.ToolCall` defines:

```ts
index: number
id?: string
```

So `index` is the stable correlation information available even when a streamed fragment does not contain an `id`.

## Suggested fix direction

Preserve `tool_call.index` through `ToolCallDelta` (or an equivalent stable correlation field) and correlate id-less continuation fragments using that identity rather than falling back to the last array position.

The existing id-based correlation path can remain the first choice, and the existing positional fallback can remain for providers that supply neither identity field.

## Regression test

A local integration-style regression exercising:

```
raw ChatCompletionChunk
→ fromChatCompletionChunk()
→ session/streamUpdate
→ applyToolCallDelta()
→ addToolCallDeltaToState()
→ final ToolCallState[]
```

reproduced the cross-attribution deterministically.

## Environment

Continue `main`: `5522c6f44ca0ac3528b37244818fbfa39b5af470`

canblmz1:
## Description

`applyToolCallDelta` currently falls back to the most recently tracked tool call when a streamed continuation fragment does not contain an `id`.

The OpenAI SDK's streamed tool-call type requires an `index` while `id` is optional. Because Continue currently drops that index in `fromChatCompletionChunk`, interleaved parallel tool-call fragments can be applied to the wrong call.

This preserves `tool_call.index` through `ToolCallDelta` and `ToolCallState` and uses it to correlate id-less continuation fragments. Existing id-based correlation remains authoritative, and the legacy fallback remains unchanged for fragments that provide neither identity field.

Fixes #13223

## Tests

Added an integration-style regression covering two interleaved parallel OpenAI tool calls and verified that each call retains only its own argument fragments.

Also re-ran the relevant existing GUI and core tool-call suites.

## Checklist

### DIFF
--- a/core/index.d.ts
+++ b/core/index.d.ts
@@ -364,6 +364,8 @@ export interface ToolCall {
 
 export interface ToolCallDelta {
   id?: string;
+  /** Provider-local correlation index for streamed tool-call fragments. */
+  index?: number;
   type?: "function";
   function?: {
     name?: string;
@@ -518,6 +520,8 @@ interface ToolCallState {
   toolCall: ToolCall;
   status: ToolStatus;
   parsedArgs: any;
+  /** Correlation index retained across streamed updates. */
+  index?: number;
   processedArgs?: Record<string, any>; // Added in preprocesing step
   output?: ContextItem[];
   tool?: Tool;
--- a/core/llm/openaiTypeConverters.ts
+++ b/core/llm/openaiTypeConverters.ts
@@ -367,6 +367,7 @@ export function fromChatCompletionChunk(
       .filter((tool_call) => !tool_call.type || tool_call.type === "function")
       .map((tool_call) => ({
         id: tool_call.id,
+        index: tool_call.index,
         type: "function" as const,
         function: {
           name: (tool_call as any).function?.name,
--- a/gui/src/redux/slices/sessionSlice.parallelToolCalls.test.ts
+++ b/gui/src/redux/slices/sessionSlice.parallelToolCalls.test.ts
@@ -1,0 +1,113 @@
+import { fromChatCompletionChunk } from "core/llm/openaiTypeConverters";
+import type { ChatCompletionChunk } from "openai/resources/index";
+import { describe, expect, it } from "vitest";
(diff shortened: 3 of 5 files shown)


canblmz1:

canblmz1:
Closing, since the repo doesn't look like it's merging changes right now. The bug is still written up in #13223 if anyone wants to pick this back up later.

PR #13226, opened 2026-09-04
fix(gui): prevent opening a chat when clicking its rename input

jabrailkhalil:
## Description

Clicking inside the title input while renaming a chat in History also triggers the row's click handler, opening the chat. Stop click propagation on the input so the title remains editable.

One line added in one file.

## Validation

Verified before and after in an isolated browser preview of HistoryTableRow with IDE calls stubbed: clicking the input no longer opens the chat; Enter still dispatches the title update, Escape cancels editing, and clicking the row still opens the chat. The full extension test suite was not run. `git diff --check` passes.

### DIFF
--- a/gui/src/components/History/HistoryTableRow.tsx
+++ b/gui/src/components/History/HistoryTableRow.tsx
@@ -104,6 +104,7 @@ export function HistoryTableRow({
               className="w-full"
               ref={(titleInput) => titleInput && titleInput.focus()}
               value={sessionTitleEditValue}
+              onClick={(e) => e.stopPropagation()}
               onChange={(e) => setSessionTitleEditValue(e.target.value)}
               onKeyUp={(e) => handleKeyUp(e)}
               onBlur={() => setEditing(false)}


jabrailkhalil:

jabrailkhalil:
The current PR Checks run 34500250345 reached jetbrains-tests but failed before tests in AnimMouse/setup-ffmpeg: the downloaded FFmpeg 7.1 payload was not xz data, so xz and tar exited 2. All other required jobs succeeded. I cannot rerun the upstream workflow without admin rights; please rerun this infrastructure failure.

jabrailkhalil:
The required `jetbrains-tests` job fails before this PR's code runs: the shared `AnimMouse/setup-ffmpeg` download receives a non-xz payload (`xz: File format not recognized`), so `tar` exits 2. The aggregate required check reflects that setup failure.

Could a maintainer rerun the workflow once the external download is healthy? I do not have repository-admin permission to rerun it.

PR #13227, opened 2026-09-04
fix(gui): correct Windows chat history storage path

jabrailkhalil:
## Description

On Windows, the History footer points to `%USERPROFILE%/.continue`, but session files are stored in its `sessions` subdirectory. Add the missing `/sessions` suffix so the displayed path points to the chat history folder.

## Validation

Checked the path against `getSessionsFolderPath()` and `getSessionFilePath()` in `core/util/paths.ts`. One display string changed in one file; `git diff --check` passes. The full test suite was not run for this text-only change.

### DIFF
--- a/gui/src/components/History/index.tsx
+++ b/gui/src/components/History/index.tsx
@@ -220,7 +220,7 @@ export function History() {
           Chat history is saved to{" "}
           <span className="italic">
             {platform === "windows"
-              ? "%USERPROFILE%/.continue"
+              ? "%USERPROFILE%/.continue/sessions"
               : "~/.continue/sessions"}
           </span>
         </span>


jabrailkhalil:
The current PR Checks run 34501221860 reached jetbrains-tests but failed before tests in AnimMouse/setup-ffmpeg: the downloaded FFmpeg 7.1 payload was not xz data, so xz and tar exited 2. All other required jobs succeeded. I cannot rerun the upstream workflow without admin rights; please rerun this infrastructure failure.

jabrailkhalil:
The required `jetbrains-tests` job fails before this PR's code runs: the shared `AnimMouse/setup-ffmpeg` download receives a non-xz payload (`xz: File format not recognized`), so `tar` exits 2. The aggregate required check reflects that setup failure.

Could a maintainer rerun the workflow once the external download is healthy? I do not have repository-admin permission to rerun it.

PR #13244, opened 2026-09-08
fix(cli): bound repeated tool-call loops

### LINKED ISSUE #12702: Continue repeats identical tool calls instead of bounding the loop
# Continue repeats identical tool calls instead of bounding the loop

## Before submitting your bug report

- [x] I checked for duplicate local reproduction drafts.
- [x] I reduced the reproducer to a single deterministic local-provider case.

## Relevant environment info

- OS: Linux x86_64 Docker
- Continue version: `@continuedev/cli 1.5.46`
- IDE version: CLI (`cn`), no IDE
- Model: OpenAI-compatible local mock provider, `gpt-4`

## Description

Severity / performance impact: High. A repeated tool-use response can cause multiple committed tool side effects and repeated provider turns.

## To reproduce

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It builds the affected CLI version from the public package registry, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/c166065dd2a35dfdbd634ee447460579

The Gist contains `continue-repeated-identical-tool-loop.reproduce.py`. Download and run:

```bash
gh gist clone c166065dd2a35dfdbd634ee447460579 continue-repeated-identical-tool-loop-reproducer
cd continue-repeated-identical-tool-loop-reproducer
python3 continue-repeated-identical-tool-loop.reproduce.py
```

To reuse an already-built local image:

```bash
python3 continue-repeated-identical-tool-loop.reproduce.py --skip-build
```

Expected successful reproduction output includes:

```text
REPRODUCED
```

## Log output

Expected successful reproduction output includes `REPRODUCED`.

Observed behavior: The local provider repeats a shell tool call after tool results arrive. The script checks for at least two side-effect log entries.

## Expected behavior

Continue should detect repeated identical tool-use state and stop with a bounded error rather than executing the same work repeatedly.

## Public-upstream status

This reproduces against stock public @continuedev/cli@1.5.46 built from npm.

noooooooookro:
## Description

Fixes #12702

The CLI now bounds consecutive identical tool-call batches within one agent turn. Tool calls are fingerprinted by tool name and canonicalized arguments (provider-generated call IDs are ignored). A repeated batch may run twice; the next identical batch is recorded as an errored tool result and stopped before execution, so repeated filesystem or other side effects cannot continue without bound.

The blocked result is written through both ChatHistoryService and the fallback history path, and is surfaced through the normal CLI callbacks.

## Checklist

- [x] No documentation changes were needed for this internal safety guard
- [x] The relevant tests have been updated or created

Not applicable: this is a CLI safety guard with automated coverage.

## Tests

- npm test (156 files passed, 1,677 tests passed; 5 files and 49 tests skipped by repository configuration)
- npm run typecheck
- Prettier check on all changed files

### DIFF
--- a/extensions/cli/src/stream/handleToolCalls.test.ts
+++ b/extensions/cli/src/stream/handleToolCalls.test.ts
@@ -63,6 +63,46 @@ describe("handleToolCalls - duplicate tool_result prevention", () => {
     vi.mocked(services.chatHistory.isReady).mockReturnValue(true);
   });
 
+  it("records repeated tool calls as errors without executing them", async () => {
+    const toolCalls = [
+      {
+        id: "repeated-tool-1",
+        name: "read_file",
+        arguments: { filepath: "/tmp/example" },
+        argumentsStr: '{"filepath":"/tmp/example"}',
+        startNotified: false,
+      },
+    ];
+
+    mockPreprocess.mockResolvedValue({
+      preprocessedCalls: [],
+      errorChatEntries: [],
+    });
+    mockExecute.mockResolvedValue({
+      hasRejection: false,
+      chatHistoryEntries: [],
+    });
+
+    const shouldReturn = await handleToolCalls({
+      toolCalls,
+      chatHistory,
+      content: "",
+      callbacks: undefined,
+      isHeadless: false,
+      blockedToolCallIds: new Set(["repeated-tool-1"]),
+    });
+
+    expect(shouldReturn).toBe(true);
+    expect(services.chatHistory.addAssistantMessage).toHaveBeenCalledTimes(1);
+    expect(services.chatHistory.addToolResult).toHaveBeenCalledWith(
+      "repeated-tool-1",
+      "Repeated identical tool call blocked to prevent an unbounded agent loop",
+      "errored",
+    );
+    expect(mockPreprocess).toHaveBeenCalledWith(false, [], undefined);
+    expect(mockExecute).toHaveBeenCalledWith([], undefined, false);
+  });
+
   /**
    * This test verifies that preprocessing errors are stored in toolCallStates
    * (on the assistant message) rather than as separate tool history items.
--- a/extensions/cli/src/stream/handleToolCalls.ts
+++ b/extensions/cli/src/stream/handleToolCalls.ts
@@ -31,13 +31,55 @@ interface HandleToolCallsOptions {
   callbacks: StreamCallbacks | undefined;
   isHeadless: boolean;
   usage?: any;
+  blockedToolCallIds?: ReadonlySet<string>;
+}
+
+const REPEATED_TOOL_CALL_ERROR =
+  "Repeated identical tool call blocked to prevent an unbounded agent loop";
+
+function recordToolResult(
+  chatHistory: ChatHistoryItem[],
+  useService: boolean,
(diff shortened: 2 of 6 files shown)


Hahaknight (commented):
Verified independently against main — the defect is real, the guard is wired at the right layer, and the non-streamChatResponse callers are provably unaffected.

**Defect confirmed.** Main's `streamChatResponse` is a `while (true)` auto-continuation loop with no bound of any kind (no loop guard, no max-turns/iteration counter anywhere in the file): a model that re-issues the same tool-call batch after each result runs until the token budget or the user's patience runs out. Same paid-balance-drain failure class as the loops this repo has hit before.

**The fingerprint excludes exactly the right things.** Across a repeat, what varies is the provider-generated call `id` and the JSON key order of `arguments`; what identifies the repeat is (name, arguments). `getToolCallFingerprint` hashes precisely the latter pair — IDs dropped, `canonicalize` recursively sorts object keys — so two "different" calls that do the same thing collide, and two genuinely different calls don't. The empty-batch and changed-fingerprint paths both reset the counter, so alternating patterns ([a], [b], [a], [b]…) never accumulate — correct, since those aren't loops.

**Other callers are provably unaffected.** `blockedToolCallIds` is optional; for callers that don't pass it (serve.helpers, useChat.stream.helpers, gui sessionSlice), `blockedToolCalls` is always empty and the return value collapses to the old constant `false`. The only behavioral delta is the streamChatResponse path.

**Block semantics are coherent end to end.** Blocked calls still get their states on the assistant message, then `onToolStart` → errored `recordToolResult` → `onToolError` (so the UI/history stay consistent), executable calls flow through the untouched preprocess/execute path, and the `return blockedToolCalls.length > 0` rides the existing `shouldReturn` branch — a graceful turn end that returns the accumulated response, not an abort. The `recordToolResult` helper is a faithful extraction of the inline service/fallback logic it replaces (checked line against line).

**One observation, non-blocking — growing batches escape.** The fingerprint covers the whole batch, so the exact-repeat case ([a] → [a] → [a]) is caught, but a batch that *grows* each iteration ([a] → [a,b] → [a,b,c] → …) never matches its predecessor and loops unbounded. That's a rarer pathology than the exact repeat, but if you ever want to close it, tracking consecutive counts per individual (name, canonical-args) pair rather than per batch would — a batch repeats only if every call repeats, so per-call counting is strictly more sensitive without being more trigger-happy on legitimate patterns.

**Second observation — the block is terminal for the turn.** Because `shouldReturn` ends the turn immediately, the errored tool_result ("Repeated identical tool call blocked…") is recorded for the UI but never sent back to the model, so it can't react or adjust. I read that as a deliberate hard-stop choice (deterministic, no second-guess loop), and with a default of 2 allowed repeats it seems right — just flagging that it's a policy, not an accident.

Tests pin the three properties that matter: ID/key-order insensitivity, the `> max` boundary (2 pass, 3rd blocks), and reset on change/empty batch; the stream-level test locks the full 3-iteration shape including the blocked set arriving only on iteration 3.

PR #13247, opened 2026-09-08
fix(core): match nested and Windows paths in isSecurityConcern

siddiqueirshad:
## Summary

`isSecurityConcern()` is the gate that blocks tools, autocomplete, and next-edit from reading secrets (`.env`, `.aws/credentials`, `.ssh/id_rsa`, etc.). Two path-handling bugs let those files through.

1. **Nested security directories on absolute paths.** Absolute paths were truncated to `parent/basename` before gitignore matching. `/home/user/.aws/prod/credentials` became `prod/credentials`, which does not match `.aws/`, so the agent could read it.
2. **Windows backslash paths.** The `ignore` library treats `\` as a gitignore escape. Relative paths like `project\config\.env` and drive-letter paths that were not rewritten never matched. The existing Windows-path test in `ignore.vitest.ts` was commented out rather than fixed.

Both show up in real call sites: `throwIfFileIsSecurityConcern(resolvedPath.displayPath)` (Windows `fileURLToPath` / `fsPath`) and autocomplete `isSecurityConcern(input.filepath)`.

## Root cause

Absolute paths were reduced to only the immediate parent directory plus basename so `ignore()` would receive a relative path. That dropped every ancestor except the last folder, so nested files under `.aws/`, `.ssh/`, `.kube/`, `secrets/`, and similar directories were not blocked. Un-normalized backslashes made Windows paths fail independently.

## Fix

Normalize to a full gitignore-style relative path: convert `file://` URIs, replace `\` with `/`, strip a Windows drive letter, strip a leading `/`. Keep the entire remaining path so `.aws/prod/credentials` still matches `.aws/`.

## Test plan

- [x] `npx vitest run indexing/ignore.vitest.ts` — 31/31 passed
- [x] New cases: nested absolute POSIX paths, Windows drive/backslash paths, `file:///C:/...` URIs
- [x] Regular source files with Windows/POSIX absolute paths still return `false`

## Reproduction (before this change)

```ts
isSecurityConcern("/home/user/.aws/prod/credentials") // false, should be true
isSecurityConcern("C:\\Users\\user\\.env")            // false, should be true
isSecurityConcern("project\\config\\.env")            // false, should be true
```

### DIFF
--- a/core/indexing/ignore.ts
+++ b/core/indexing/ignore.ts
@@ -1,6 +1,5 @@
 import ignore from "ignore";
 
-import path from "path";
 import { fileURLToPath } from "url";
 import { ContinueError, ContinueErrorReason } from "../util/errors";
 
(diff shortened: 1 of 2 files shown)


siddiqueirshad:
Hi @sestinj, when you get a chance, could you please take a look at this PR? Thanks!

PR #13275, opened 2026-09-14
fix(ollama): preserve tools across agent turns

iwasatapex:
## Description

Fix Ollama tool handling in Continue Agent mode.

The Ollama adapter previously attached tool definitions only when the last message had role `user`. After a tool call, the next request has a `tool` message as the latest message, so Continue stopped sending the tool definitions back to Ollama.

This caused multi-step agent loops to fall back to textual/XML tool calls instead of structured tool calls.

This change removes that incorrect restriction so tools remain available across agent turns.

## What changed

- Always attach configured tools to Ollama chat requests.
- Added regression tests covering:
  - first-turn tool attachment
  - continuation after a tool result
  - multiple tool calls in the same agent loop

## Testing

The added Ollama regression tests were run successfully against the patched source.

The fix was also packaged and tested with Continue 2.0.0 in a local VSCodium installation.

## Scope

Only the Ollama tool-calling path and its regression tests are changed.


iwasatapex:

PR #13278, opened 2026-09-14
fix: correlate id-less parallel tool-call fragments by stream index

### LINKED ISSUE #13223: OpenAI parallel tool-call argument fragments can be applied to the wrong call

ege-arhan:
## Problem
OpenAI streamed tool-call deltas carry a stable `index`, while `id` is optional on continuation fragments. Continue's `fromChatCompletionChunk()` dropped `tool_call.index`, and `applyToolCallDelta()` fell back to updating the most recently tracked tool call whenever a fragment had no `id`. With two parallel tool calls whose argument fragments interleave, one call's arguments were applied to the other call.

Fixes #13223.

## Solution
- Preserve `tool_call.index` through `ToolCallDelta` (new optional `index?: number` on the type + converter passthrough).
- In `applyToolCallDelta()`, when a delta has no `id` but has an `index`, match the existing tool-call state by stream position instead of "last array element".
- Existing id-based correlation stays first choice; the positional fallback remains for providers supplying neither field.

## Impact
- Parallel tool-call streaming attributes argument fragments to the correct call.
- Single-call streaming behavior unchanged (id path and last-element fallback untouched).
- Affects OpenAI/OpenAI-compatible provider stream path only.

## Evidence
- New regression test in `core/llm/openaiTypeConverters.test.ts` asserting `index` survives `fromChatCompletionChunk()`.
- Standalone repro of the issue's 4-delta sequence (index=0/id=call_A, index=1/id=call_B, index=0 id-less, index=1 id-less): after fix, call_A args = '{"target":"A_ONLY"}', call_B args = '{"target":"B_ONLY"}' (PASS).
- Full `npm install` in this sandbox failed on /tmp disk quota (chromium-bidi tar write), so the repo vitest suite could not run here; test added for CI.

Fixes #13223


ege-arhan:

ege-arhan:

ege-arhan:

ege-arhan:

ege-arhan:

ege-arhan:
Maintainer: infra-flaky fails (jetbrains-tests, vscode-get-test-file-matrix) need rerun. `gh run rerun --failed` rejected. Please rerun failed jobs.

ege-arhan:
Update on the CI failures (commit 8686): **CLI PR Checks ✅, VSIX build ✅, CLA ✅**. The only red job is `jetbrains-tests`, and it dies before any of our code runs — the `AnimMouse/setup-ffmpeg@v1` action fails with `tar: Child returned status 1` while downloading its FFmpeg tarball. This is repo-wide right now: every fork PR in the recent history shows the same `PR Checks` failure (16 failed runs in the last 60 across unrelated branches), so it looks like a broken upstream ffmpeg asset rather than anything this PR touches. Happy to force-push again when the download recovers.

PR #13286, opened 2026-09-16
feat: add Y-API as an OpenAI-compatible provider

jiweiyeah:
## What

Adds **Y-API** (`y-api`) as an OpenAI-compatible provider.

- Website: https://y-api.bestvirtualgoods.com
- API base: `https://api.y-api.bestvirtualgoods.com/v1/`
- Docs: https://y-api.bestvirtualgoods.com/docs
- Pricing: https://y-api.bestvirtualgoods.com/pricing

It is a gateway that fronts DeepSeek, GLM, Kimi, Qwen, Tencent Hunyuan, Xiaomi MiMo, and OpenAI models behind one key and one balance. Model IDs are namespaced (`deepseek/deepseek-v4-pro`, `z-ai/glm-5.3`, `moonshotai/kimi-k3`, …), so they pass through unchanged — no model-name translation is needed for this provider.

## Changes

Follows the CONTRIBUTING checklist for adding an LLM provider:

| File | Change |
| --- | --- |
| `core/llm/llms/YApi.ts` | New. Extends `OpenAI`; `apiBase` and default model set. |
| `core/llm/llms/index.ts` | Import + entry in `LLMClasses`. |
| `core/llm/llms/OpenAI-compatible.vitest.ts` | Coverage via the shared `createOpenAISubclassTests` helper. |
| `core/llm/toolSupport.ts` | `PROVIDER_TOOL_SUPPORT["y-api"]` — see below. |
| `packages/openai-adapters/src/types.ts`, `.../index.ts` | Provider literal + `constructLlmApi` case, so the CLI path resolves it. |
| `docs/customize/model-providers/more/y-api.mdx` | New documentation page. |

The default model is `deepseek/deepseek-v4-flash`: it is free of charge on the service (no credit is deducted), and it is the model the service's own documentation uses in its examples.

## Tool support

`PROVIDER_TOOL_SUPPORT` is populated deliberately: the default for an unlisted provider is `false`, which would silently disable agent mode for every Y-API model.

Support was measured against the live API rather than assumed — one request per model with a single function tool, checking for a `tool_calls` response:

- **Tools work** for the four DeepSeek models, the three GLM models, `moonshotai/kimi-k3`, `tencent/hy3`, and `xiaomi/mimo-v2.5`.
- **The four `openai/*` models accept tools only when `reasoning_effort` is set explicitly.** With the parameter absent the upstream returns `Function tools with reasoning_effort are not supported` (3/3 calls); with it present the tool call succeeds (3/3 at each of `low`, `none`, and `high`). The error text reads as if the two are mutually exclusive, but the behaviour is the opposite — the rejection fires on *absence*, not presence. Continue has no way to send `reasoning_effort`, so these four report no native tool support rather than advertising a capability that would 400 in agent mode.

Two related facts, for anyone touching this provider later:

- `openai/gpt-6-astra` requires `max_completion_tokens` rather than `max_tokens`. `OpenAI.isOSeriesOrGpt5PlusModel` already matches `gpt-[5-9]`, so Continue handles this automatically — no provider-specific code was needed.
- `qwen/qwen3.8-flash` is advertised on the service's pricing page but is absent from `GET /v1/models` and returns `503 No available channel for model qwen/qwen3.8-flash under group y-api (distributor)`. It is not referenced anywhere in this PR.

## Verification

Run locally on Node 22.22.2:

- `core`: `tsc -p ./ --noEmit` — **clean**.
- `core`: `vitest run llm/llms/OpenAI-compatible.vitest.ts` — **218 passed**. The seven generated `y-api` cases all pass:

  ```
  ✓ y-api > should have correct provider name
  ✓ y-api > should have correct default API base
  ✓ y-api > streamChat should send a valid request
  ✓ y-api > chat should send a valid request
  ✓ y-api > streamComplete should send a valid request
  ✓ y-api > complete should send a valid request
  ✓ y-api > should handle embeddings
  ```

- `packages/openai-adapters`: `npm run build` (tsc) — **clean**.
- Prettier — clean on every changed file.

**Not run:** the jest suite. Its global setup calls `fs.rmSync` on `core/test/.continue-test`, which this sandbox blocks, and `test/jest.setup-after-env.js` fails to load `node-fetch` (`SyntaxError: Cannot use import statement outside a module`) under this Node build. So `llm/toolSupport.test.ts` is typechecked but not executed here; I'd expect CI to cover it.

## CLA

I'll post the CLA signature comment on this PR.


jiweiyeah:

jiweiyeah:
A note on the three non-green checks, since none of them come from this change.

**`jetbrains-tests`** — pre-existing and repo-wide. It fails on the current head of `main`-targeting branches I did not touch:

| branch | run |
| --- | --- |
| `snyk-fix-16461e5b…` | [35084479159](https://github.com/continuedev/continue/actions/runs/35084479159) |
| `fix/issue-13223` | [34871704808](https://github.com/continuedev/continue/actions/runs/34871704808) |
| `feat/add-api-route-provider` | [34850359834](https://github.com/continuedev/continue/actions/runs/34850359834) |

The job dies in setup, before any Continue code runs — the third-party `AnimMouse/setup-ffmpeg` step:

```
tar: Error is not recoverable: exiting now
##[error]Process completed with exit code 2.
##[end-action id=__self.__AnimMouse_setup-ffmpeg.__run;outcome=failure;conclusion=failure;duration_ms=236]
```

**`require-all-checks-to-pass`** — a meta-check that only reflects the above.

**`build-and-upload-vsix (darwin, arm64, macos-latest)`** — reported as cancelled, not failed. The build and both artifact uploads completed first:

```
vsce package completed - extension created at extensions/vscode/build/continue-1.3.40-4a6f217.vsix
Artifact vscode-extension-build-darwin-arm64 has been successfully uploaded! Final size is 77444283 bytes.
Artifact download URL: …/runs/35096756936/artifacts/10446282848
```

`##[error]The operation was canceled.` lands afterwards, at 12:46:04, inside post-job cleanup while the second `actions/cache` save was still compressing. The runner went away mid-cache-upload; the `linux, x64` and `win32, x64` jobs in the same workflow finished green.

Everything this change actually touches is green: `core-checks`, `lint`, `prettier-check`, `packages-checks (openai-adapters)`, all twelve `test` matrix jobs, all e2e, `vscode-checks`, `binary-checks`, `gui-checks`, plus CLA / GitGuardian / Snyk.

PR #13297, opened 2026-09-19
fix(cli): don't hang Bash tool on detached child stdio

### LINKED ISSUE #12699: Continue hangs after background or detached child-process tool calls
# Continue hangs after background or detached child-process tool calls

## Before submitting your bug report

- [x] I checked for duplicate local reproduction drafts.
- [x] I reduced the reproducer to deterministic local-provider cases.

## Relevant environment info

- OS: Linux x86_64 Docker
- Continue version: `@continuedev/cli 1.5.45`
- IDE version: CLI (`cn`), no IDE
- Model: OpenAI-compatible local mock provider, `gpt-4`

## Description

Severity / performance impact: High. When Continue receives a tool call that starts a background or detached child process, the CLI does not complete the session cleanly. The public reproducer observes that the tool request reaches the client, the child-spawning command writes its PID side effect, and the CLI then fails to finish before the timeout.

Modelock repeat-3 replay also reproduced the stronger process-lifecycle oracle for both representative variants: `no_hang` fired three times and `no_orphan_process` fired three times for the background-shell sample, and the same pair fired three times for the detached-child sample. Those replay artifacts are in `findings/open/2026-06-16-continue-child-process-cleanup/`.

## To reproduce

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It builds the affected CLI version from the public package registry, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/trackoor/58792d4d031da6f42e9ce770ddf35de6

The Gist contains `continue-child-process-cleanup.reproduce.py`. Download and run:

```bash
gh gist clone 58792d4d031da6f42e9ce770ddf35de6 continue-child-process-cleanup-reproducer
cd continue-child-process-cleanup-reproducer
python3 continue-child-process-cleanup.reproduce.py
```

To reuse an already-built local image:

```bash
python3 continue-child-process-cleanup.reproduce.py --skip-build
```

Key output from a local run against the affected version:

```text
provider_requests_at_least: observed=1 expected>=1
side_effect_file_exists background-shell.pid: bytes=... expected>=1
timeout_or_signal: returncode=... timed_out=True
REPRODUCED
```

The embedded case runs both `background-shell` and `detached-child` variants.

## Log output

Expected successful reproduction output includes `REPRODUCED`.

Observed behavior: Continue accepts a child-spawning tool call from the local provider, the command records a PID side effect, and the CLI does not return normally within the bounded run.

## Expected behavior

Continue should track and clean up tool-started background children, then finish or cancel the session without hanging or leaving unreaped process state.

## Public-upstream status

This reproduces against stock public `@continuedev/cli@1.5.45` built from npm.

mrchatam:
## Description

Fixes #12699.

When a Bash tool command backgrounds or detaches a child that keeps inherited stdio open, Node's `close` event never fires and the CLI tool call hangs. This change:

- Spawns Unix tool shells with `detached: true` (own process group)
- Resolves the tool call on process **exit** (after a short stdio flush), not only `close`
- Kills the full process group on timeout/cancel (`process.kill(-pid)`)
- Applies the same spawn/kill behavior in `BackgroundJobService`
- Adds regression tests for background-shell and detached-child patterns

## Testing

- `vitest run src/tools/runTerminalCommand.test.ts` (9 passed, including #12699 cases)


mrchatam:

PR #13299, opened 2026-09-19
docs: fix dead link to default system message source

jrlprost:
The "source code here" link in docs/customize/deep-dives/rules.mdx points to core/llm/constructMessages.ts, which no longer exists on main (the link returns 404). The default system messages now live in core/llm/defaultSystemMessages.ts, which is also the file `DEFAULT_SYSTEM_MESSAGES_URL` points to.

Verified with `curl -sI` before (404) and after (200) the change.


jrlprost:

Issue #13307, opened 2026-09-21
edit_existing_file silently performs full-file overwrite when changes payload has no anchor markers


jattie-ire:
### Before submitting your bug report

- [x] I've tried finding an answer on the [Continue docs site](https://docs.continue.dev/)
- [x] I'm not able to find an [open issue](https://github.com/continuedev/continue/issues?q=is%3Aopen+is%3Aissue) that reports the same bug
- [x] I've seen the [troubleshooting guide](https://docs.continue.dev/troubleshooting) on the Continue Docs

### Relevant environment info

```Markdown
- OS:Windows 11 Enterprise
- Continue version: 2.0.0
- IDE version: 1.138.0
- Model: Qwen3.8-27B
- config:
  
name: Main Config
version: 1.0.0
schema: v1
models:
  - name: Qwen-Coder
    provider: openai
    model: Qwen-Coder
    apiBase: http://172.19.128.237:8000/v1
    apiKey: EMPTY
    capabilities:
      - tool_use
      - image_input
    roles:
      - chat
      - edit
      - apply
      - autocomplete
    defaultCompletionOptions:
      temperature: 1.0
      top_p: 0.95
      maxTokens: 8192
      contextLength: 131072
      reasoning: true
    requestOptions:
      stream: false
      extraBodyProperties:
        top_k: 20
        min_p: 0.0
        presence_penalty: 0.0
        repetition_penalty: 1.0
        reasoning_effort: "low" # "medium" "high"

  
  OR link to agent in Continue hub:
```

### Description

**OS:** Linux
**Model/provider (repro):** any agent-mode model via an OpenAI-compatible provider (repro'd with Qwen-Coder over vLLM)

**Summary**
When the agent calls the built-in `edit_existing_file` tool and the `changes` argument is a bare changed line (no `// ... existing code ...` anchors, no surrounding unchanged context), Continue treats that text as the **entire new file** and overwrites the target. The resulting diff shows the whole file deleted and a single line inserted. It returns `Edit Success` with no warning. Accepting it destroys the file.

**Steps to reproduce**
1. In agent mode, request a one-line fix to a file larger than a few lines (e.g. fix a missing function argument).
2. The model emits `edit_existing_file` with only the changed line, e.g.:
   ```json
   {"filepath": "word_generator.py",
    "changes": "    words = find_words(letters, trie, args.min_freq, max_len, args.min_len)"}
   ```
3. Continue renders a diff of **delete all N lines → insert 1 line**.
[...]

PR #13309, opened 2026-09-22
fix(edit): reject full-file rewrite diffs using old-file removal ratio

wudilyy999:
Refs #13307

## Problem

When the model emits an `edit_existing_file` call whose `changes` argument is a bare snippet with no `// ... existing code ...` anchors, Continue's instant-apply path treats the snippet as the complete new file content. The rejection guard meant to catch this is defeated by its own ratio math:

```ts
// shouldRejectDiff (before)
numRemovals / diff.length > 0.3
```

The denominator is the TOTAL diff length (removed + kept + added lines), so adding N "new" lines dilutes the ratio. A diff that deletes all 100 lines of a file and inserts one line has ratio 100/101 ≈ 99%... but the same deletion paired with a 300-line snippet lands at 100/401 ≈ 25% and sails through. The more the model adds, the safer a destructive overwrite looks.

## Fix

Measure removals against the old file's line count (removed + kept), which is invariant to how much the snippet adds:

- bare un-anchored snippet vs. a 100-line file → 100/100 = 100% removed → rejected, falls through to the non-instant path instead of silently replacing the file
- genuine single-line fix to the same file → 1/100 = 1% → still accepted
- pure insertions (0 removed lines) → 0% → still accepted
- empty diff → guarded against 0/0

## Verification

Two new tests in `core/edit/lazy/deterministic.test.ts` on the `onlyFullFileRewrite` path (the path the edit tool takes for supported languages):

- `onlyFullFileRewrite rejects a bare un-anchored snippet` — a one-line snippet against a 100-line file returns `undefined` (previously returned the full-overwrite diff)
- `onlyFullFileRewrite still accepts a genuine small fix` — a real one-line fix to the same file still returns the 1-line diff

`core/edit/lazy/` suites: 4/4 pass (36 tests). Full `core` jest run compared against the base commit: identical failing set before and after (17 pre-existing environment-dependent suite failures), the only delta is the 2 new passing tests.

Note: this implements the "reject" half of the issue's proposed fix. A diff-level warning (>50% removal surfacing in the UI) could be a follow-up; this PR keeps the guardrail at the decision point where the overwrite is chosen.

---
🤖 Prepared with AI assistance (Codex); reproduction analysis, fix, and tests were run and verified locally by me before opening.


wudilyy999:

wudilyy999:
The `jetbrains-tests` failure looks unrelated to this PR — it failed during CI setup while downloading FFmpeg (`xz: (stdin): File format not recognized`, exit code 2), before any test ran. The change only touches `core/edit/lazy/deterministic.ts` (Node-side logic, covered by the passing core checks). Could a maintainer rerun that job?

